PR #55, opened 2019-02-02
Clean up matrix printing in quilc

notmgsk:
I *think* the new stuff is equivalent to the old. The old stuff was
doing a lot of matrix calculations and rescaling and etc. There are
functions now that do all of this for us (gate-apps-to-log-mat,
scale-out-matrix-phases, etc)

I borrowed the matrix equality stuff from cl-quil-tests. Couple
questions,
  1. Is this all equivalent?
  2. Should matrix equality stuff live elsewhere (so that it's not
  duplicated across quilc/cl-quil/cl-quil-tests)
  3. Should that place be magicl?

Have a good weekend xxx

### DIFF
--- a/app/src/entry-point.lisp
+++ b/app/src/entry-point.lisp
@@ -367,7 +367,7 @@ HTTP server for good.
 (defun process-program (program chip-specification)
   (let* ((original-matrix
            (when (and *protoquil* *compute-matrix-reps*)
-             (quil::gate-applications-to-logical-matrix program)))
+             (quil::parsed-program-to-logical-matrix program)))
          (quil::*compiler-noise-stream* *verbose*)
          (*statistics-dictionary* (make-hash-table :test 'equal))
          (*random-state* (make-random-state t)))
@@ -446,7 +446,7 @@ HTTP server for good.
           (print-2Q-gate-depth lschedule)))
 
       (when (and *protoquil* *compute-matrix-reps*)
-        (let* ((processed-program-matrix (quil::gate-applications-to-logical-matrix processed-program :compress-qubits t))
+        (let* ((processed-program-matrix (quil::parsed-program-to-logical-matrix processed-program :compress-qubits t))
                (same-same-but-different (quil::scale-out-matrix-phases processed-program-matrix
                                                                        original-matrix)))
           (format *human-readable-stream* "~%#Matrix read off from input code~%")
--- a/app/src/entry-point.lisp
+++ b/app/src/entry-point.lisp
@@ -367,7 +367,7 @@ HTTP server for good.
 (defun process-program (program chip-specification)
   (let* ((original-matrix
            (when (and *protoquil* *compute-matrix-reps*)
-             (quil::make-matrix-from-quil (coerce (quil::parsed-program-executable-code program) 'list) program)))
+             (quil::gate-applications-to-logical-matrix program)))
          (quil::*compiler-noise-stream* *verbose*)
          (*statistics-dictionary* (make-hash-table :test 'equal))
          (*random-state* (make-random-state t)))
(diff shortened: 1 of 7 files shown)


stylewarning:
@notmgsk If it's now in `cl-quil`, we should just depend on `cl-quil` from tests and quilc.

app/src/printers.lisp
```diff
@@ -13,50 +13,35 @@
...
+
+(defun operator= (mat1 mat2)
+  (multiple-value-bind (mat1 mat2) (quil::matrix-rescale mat1 mat2)
+    (setf mat1 (quil::scale-out-matrix-phases mat1 mat2))
+    (quil::matrix-equality mat1 mat2)))
+
+(defun matrix-equals-dwim (mat1 mat2)
+  "Returns true if mat1 is equal to mat2, with the specific notion of equality
+depending on whether *ENABLE-STATE-PREP-COMPRESSION* is enabled."
+  (funcall (if quil::*enable-state-prep-compression*
```
stylewarning: is this the right predicate? @ecp-rigetti

stylewarning: I myself prefer `if` and two function calls, as opposed to a `(funcall (if ...) ...)`

app/src/entry-point.lisp
```diff
@@ -362,7 +362,7 @@ HTTP server for good.
 (defun process-program (program chip-specification)
   (let* ((original-matrix
            (when (and *protoquil* *compute-matrix-reps*)
-             (quil::make-matrix-from-quil (coerce (quil::parsed-program-executable-code program) 'list) program)))
+             (quil::gate-applications-to-logical-matrix program)))
```
ecp-rigetti: Not your fault, but this is a weird name for this function on account of it doesn't consume a bunch of `gate-application`s.

app/src/printers.lisp
```diff
@@ -13,50 +13,35 @@
...
+    (print-matrix-with-comment-hashes original-matrix *human-readable-stream*)
     (format *human-readable-stream* "~%#Matrix read off from compiled code~%")
-    (print-matrix-with-comment-hashes new-matrix *human-readable-stream*)
-    (setf (gethash "compiled_matrix" *statistics-dictionary*)
-          (with-output-to-string (s)
-            (print-matrix-with-comment-hashes new-matrix s)))
+    (print-matrix-with-comment-hashes same-same-but-different *human-readable-stream*)
     (format *human-readable-stream* "~%")
-    (finish-output *standard-output*)
+    (format *human-readable-stream* "Matrices are~a equal~%" (if (matrix-equals-dwim original-matrix same-same-but-different) "" " "))
```
ecp-rigetti: Is the second option supposed to be `"n't"`?

app/src/printers.lisp
ecp-rigetti: This is the right predicate for now, but it could stand to be improved to match the other DWIM-type machinery around https://github.com/rigetti/quilc/blob/master/src/compressor/compressor.lisp#L440 .

app/src/entry-point.lisp
```diff
@@ -441,15 +441,6 @@ HTTP server for good.
...
-              (initial-l2p (quil::pragma-rewiring
-                            (aref (quil::parsed-program-executable-code processed-program) 0)))
-              (final-l2p (quil::pragma-rewiring
-                          (aref (quil::parsed-program-executable-code processed-program)
-                                    (1- (length (quil::parsed-program-executable-code processed-program)))))))
-          (print-matrix-representations initial-l2p
-                                        (coerce processed-quil 'list)
-                                        final-l2p
-                                        original-matrix)))
+        (print-matrix-representations processed-program original-matrix))
```
ecp-rigetti: This one also feels like a weird name / weird function signature. Does this get reused? If not, why not inline it? Seems to make strictly more sense in context.

app/src/printers.lisp
notmgsk: yep you bet

app/src/entry-point.lisp
stylewarning: can we just fix this name @notmgsk ? `parsed-program-to-logical-matrix`

src/utilities.lisp
```diff
@@ -41,3 +41,22 @@
...
+
+(defun operator= (mat1 mat2)
+  (multiple-value-bind (mat1 mat2) (quil::matrix-rescale mat1 mat2)
+    (setf mat1 (quil::scale-out-matrix-phases mat1 mat2))
+    (quil::matrix-equality mat1 mat2)))
+
+(defun matrix-equals-dwim (mat1 mat2)
+  "Returns true if mat1 is equal to mat2, with the specific notion of equality
+depending on whether *ENABLE-STATE-PREP-COMPRESSION* is enabled."
+  (funcall (if quil::*enable-state-prep-compression*
```
stylewarning: can we make this inside out?

```
(if quil::*enable...
    (first-column...)
    (operator= ...))
```

src/utilities.lisp
```diff
@@ -41,3 +41,22 @@
...
+    (quil::matrix-first-column-equality mat1 mat2)))
+
+(defun operator= (mat1 mat2)
+  (multiple-value-bind (mat1 mat2) (quil::matrix-rescale mat1 mat2)
+    (setf mat1 (quil::scale-out-matrix-phases mat1 mat2))
+    (quil::matrix-equality mat1 mat2)))
+
+(defun matrix-equals-dwim (mat1 mat2)
+  "Returns true if mat1 is equal to mat2, with the specific notion of equality
+depending on whether *ENABLE-STATE-PREP-COMPRESSION* is enabled."
```
stylewarning: comment what DWIM means in the docstring

stylewarning:
ping @notmgsk

src/utilities.lisp
notmgsk: yes

src/utilities.lisp
notmgsk: ok

notmgsk:
@ecp-rigetti @stylewarning all points have been addressed. Maybe a final once over and lets go to LGTM

PR #58, opened 2019-02-06
Compile PHASE to RZ

### LINKED ISSUE #57: Cannot compile parametric PHASE gate
Originally raised on [pyQuil](https://github.com/rigetti/pyquil/issues/791).

Compiling a `PHASE` gate that has a memory-ref as its parameter fails with the following error
```
* (compiler-hook (parse-quil-string "DECLARE gamma REAL[1]
PHASE(2*gamma[0]) 0") (build-8q-chip))

Condition CL-QUIL::UNKNOWN-GATE-PARAMETER was signalled.
   [Condition of type UNKNOWN-GATE-PARAMETER]
```

notmgsk:
The translation employed here is pretty interesting. When you look at
the matrices of
```
    PHASE(θ) = | 1  0   |
               | 0  eⁱᶿ |
```
and
```
    RZ(2α) = | e⁻ⁱᵅ  0   |
             | 0     eⁱᵅ |
```
you might notice that PHASE is equivalent to RZ × e^(iθ/2). So you
seek how to write the follow gate in terms of Quil's native gates
```
    RZ(β) = | eⁱᵝ  0   |
            | 0    eⁱᵝ |.
```
Then it's problem solved.

Instead, remember that during measurement, any phase factor is
irrelevant! Hence PHASE ≡ RZ × e^(iθ/2) ≡ RZ.

This closes #57.

### DIFF
--- a/src/chip-specification.lisp
+++ b/src/chip-specification.lisp
@@ -313,7 +313,8 @@ MISC-DATA is a hashtable of miscellaneous data associated to this hardware objec
                              (find  ':X/2 lower-precedence))
                          (not (find ':RZ  higher-precedence))
                          (not (find ':Z/2 higher-precedence)))
-                    (list #'RY-to-XZX
+                    (list #'PHASE-to-RZ
+                          #'RY-to-XZX
                           #'RX-to-ZXZXZ
                           #'euler-compiler)))
             :into compilation-methods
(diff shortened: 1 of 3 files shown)


stylewarning:
High quality unicode.

Issue #31, opened 2019-01-22
Define arbitrary chip-spec names that can be loaded with `--isa`


notmgsk:
Instead of extending the following as and when necessary
```
       (setf *isa-descriptor*
             (cond
               ((or (null isa)
                    (string= isa "8Q"))
                (quil::build-8Q-chip))
               ((string= isa "20Q")
                (quil::build-skew-rectangular-chip 0 4 5))
               ((string= isa "16QMUX")
                (quil::build-nQ-trivalent-chip 1 1 8 4))
               ((string= isa "bristlecone")
                (quil::build-bristlecone-chip))
               ((probe-file isa)
                (quil::read-chip-spec-file isa))
               (t
                (error "ISA descriptor does not name a known template or an extant file."))))
```
it might be useful to have chip-specs define their names and have `entry-point.lisp` look them up at run-time like
```
    (setf *isa-descriptor* (lookup-isa-descriptor-for-name *isa-name-from-stdin*))
```

matt-y:
Hello, I'd like to give this one a shot

notmgsk:
@matt-y take a good SWAG at it.

PR #65, opened 2019-02-07
Refactoring isa descriptor selection in entry-point.lisp

matt-y:
[More in line now with suggestions outlined here](https://github.com/rigetti/quilc/issues/31)

Simply defined a dynamic variable holding a hash of known isa values and what they map to. There is still some conditional logic outside of that simple lookup however - just less than the original cond. A little more easily edited now perhaps.

Also looks like some white space was picked up by my config - sorry about the noise.

### DIFF
--- a/app/src/entry-point.lisp
+++ b/app/src/entry-point.lisp
@@ -67,10 +67,10 @@
 
 (defparameter *isa-descriptors*
   (alexandria::alist-hash-table
-   '(("8Q" . (quil::build-8Q-chip))
-     ("20Q" . (quil::build-skew-rectangular-chip 0 4 5))
-     ("16QMUX" . (quil::build-nQ-trivalent-chip 1 1 8 4))
-     ("bristlecone" . (quil::build-bristlecone-chip)))
+   `(("8Q" . ,(quil::build-8Q-chip))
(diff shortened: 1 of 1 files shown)


app/src/entry-point.lisp
```diff
@@ -93,6 +101,18 @@
...
+  (if (null isa)
+      (gethash "8Q" *isa-descriptors*)
+      (let ((isa-hash-value (gethash isa *isa-descriptors*)))
+        (cond
+          ((not (null isa-hash-value))
+           isa-hash-value)
+          ((probe-file isa)
+           (quil::read-chip-spec-file isa))
+          (t
+           (error "ISA descriptor does not name a known template or an extant file."))))))
```
notmgsk: I suggest `(error "ISA descriptor ~S does not name a known chip specification or file." isa)`

app/src/entry-point.lisp
```diff
@@ -93,6 +101,18 @@
              (princ matrix s))
            (coerce #(#\Newline #\#) 'string))))
 
+(defun lookup-isa-descriptor-for-name (isa)
+  (if (null isa)
```
notmgsk: Provide default to hash map to remove this condition

PR #66, opened 2019-02-07
Update rewrite-arithmetic endpoint

lcapelluto:
Update endpoint from rewrite-arithmetic -> resolve-gate-parameter-arithmetic to be compatible with pyquil

### DIFF
--- a/app/src/rpc-server.lisp
+++ b/app/src/rpc-server.lisp
@@ -124,7 +124,7 @@
                                    (mapcar (alexandria:compose #'symbol-name #'quil.clifford::base4-to-sym)
                                            (quil.clifford::base4-list pauli-out))))))
 
-(defun rewrite-arithmetic (request)
+(defun rresolve-gate-parameter-arithmetic (request)
   "Rewrites the request program without arithmetic in gate parameters."
   (check-type request rpcq::|RewriteArithmeticRequest|)
   (let ((program (quil::parse-quil (rpcq::|RewriteArithmeticRequest-quil| request))))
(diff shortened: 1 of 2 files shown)


app/src/rpc-server.lisp
```diff
@@ -124,7 +124,7 @@
                                    (mapcar (alexandria:compose #'symbol-name #'quil.clifford::base4-to-sym)
                                            (quil.clifford::base4-list pauli-out))))))
 
-(defun rewrite-arithmetic (request)
+(defun rresolve-gate-parameter-arithmetic (request)
```
notmgsk: `rreally?`

lcapelluto: haha thank you

lcapelluto:
```suggestion
(defun resolve-gate-parameter-arithmetic (request)
```

app/src/rpc-server.lisp
```diff
@@ -124,7 +124,7 @@
                                    (mapcar (alexandria:compose #'symbol-name #'quil.clifford::base4-to-sym)
                                            (quil.clifford::base4-list pauli-out))))))
 
-(defun rewrite-arithmetic (request)
+(defun resolve-gate-parameter-arithmetic (request)
```
stylewarning: I don't like the term "resolve". What's being resolved? We are rewriting the program into an inequivalent program.

notmgsk: True. But `resolve-gate-parameter-arithmetic` is used in pyQuil and we decided (offline) that making changes to pyquil was bad for BC.

lcapelluto: I agree, it is a worse name

stylewarning: Can we at least make an issue to rename it then? @lcapelluto

lcapelluto: If we want to rename it eventually, we should just do it now. The only reason we weren't doing it that way was to avoid changing the user API (cc @ecp-rigetti)

ecp-rigetti: I don't mind picking "none of the above" by selecting a different name entirely, and I agree that if we're going to do that then we ought to do it now. I don't have a good name to suggest though. `make-parameter-arithmetic-blackbox` — idk.

lcapelluto:
I'm going to go with the `rewrite-arithmetic` name, which I think is fairly appropriate. No changes required in quilc. Will follow up with the 1 line change in pyQuil.

PR #67, opened 2019-02-07
Show human readable time units in PRINT-PROGRAM-RUNTIME

### LINKED ISSUE #43: Output readable duration with units when using --compute-runtime

apaszke:
Also, slightly reword the comment of LSCHEDULER-WALK-GRAPH
to clarify the explanation.

Fixes #43.

---

For example, this command
```
echo "H 0\nH 1\nCNOT 0 1" | ./quilc -P --compute-runtime
```
now prints
```
...
# Compiled program duration:    0s    0ms    0us  168ns   30ps
```

Not sure if picosecond precision is really needed, but that's what it seemed to report 🤷‍♂️

### DIFF
--- a/app/src/printers.lisp
+++ b/app/src/printers.lisp
@@ -104,15 +104,16 @@
 (defun print-program-runtime (lschedule chip-specification)
   (let* ((duration (quil::lscheduler-calculate-duration lschedule
                                                        chip-specification))
+         (duration-picos (* duration 1000))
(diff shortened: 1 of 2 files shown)


apaszke:
@stylewarning updated with total picoseconds. I didn't want to use `~A` for the individual fields, because their alignment would change based on the length of each element, making it hard to compare results of two runs.

stylewarning:
For individual fields I agree!

PR #125, opened 2019-02-27
Respect time-limit configuration

### LINKED ISSUE #124: Fix RPCQ timeout
In the CLI code, there's an option to install a time limit ( https://github.com/rigetti/quilc/blob/master/app/src/entry-point.lisp#L62 ), but the RPCQ server timeout is hard coded ( https://github.com/rigetti/quilc/blob/master/app/src/rpc-server.lisp#L172 ) at 60 seconds.

Cross reference: https://github.com/rigetti/rpcq/issues/39 .

notmgsk:
Previously the time-limit option had no effect on the RPCQ server. Closes #124.

Some redundant/bad stuff has been removed.

Squashed commit msgs below

1. a time-limit of zero should correspond to no timeout, yet this code
leaves the default huchentoot timeout of 20s
2. in entry-point.lisp the conversion from milliseconds to seconds is
already performed, so (/ ... 1000) is unnecessary here

Specify time-limit in seconds rather than ms

And prevent negative values

Use configurable time-limit for rpcq server

Ooops - actually set time-limit

### DIFF
--- a/app/src/entry-point.lisp
+++ b/app/src/entry-point.lisp
@@ -56,10 +56,10 @@
     (("enable-state-prep-reductions") :type boolean :optional t :documentation "assume that the program starts in the ground state")
     (("protoquil" #\P) :type boolean :optional t :documentation "restrict input/output to ProtoQuil")
     (("help" #\h) :type boolean :optional t :documentation "print this help information and exit")
-    (("server-mode-http" #\S) :type boolean :optional t :documentation "run as a web server *and* an RPCQ server. ignores --port and uses 6000 and 5555 respectively.")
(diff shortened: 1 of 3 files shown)


app/src/entry-point.lisp
```diff
@@ -250,8 +250,11 @@
   (when benchmark
     (benchmarks))
 
-  (when (plusp time-limit)
-    (setf *time-limit* (/ time-limit 1000.0d0)))
+  (when (minusp time-limit)
```
stylewarning: just error imho, swat them on the wrist

stylewarning:
Make sure to update the manpage

ecp-rigetti:
Modulo what Robert said, this LGTM.

PR #126, opened 2019-02-28
Remove use of buildapp for building

rigettizach:
Instead, use a build-app.lisp file to load quilc and save the quilc
application binary.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -57,45 +57,32 @@ else
 	echo "Non-Linux-based platforms unsupported"
 endif
 
-install-build-deps: install-test-deps
-	$(QUICKLISP) \
-		--eval '(ql:quickload "buildapp")' \
-		--eval '(buildapp:build-buildapp "/usr/local/bin/buildapp")'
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -57,45 +57,32 @@ else
...
-
 ###############################################################################
 # BUILD
 ###############################################################################
 
 .PHONY: quilc
 quilc: system-index.txt
-	buildapp --output quilc \
-		--manifest-file system-index.txt \
+	$(SBCL) $(FOREST_SDK_FEATURE) \
```
stylewarning: @rigettizach Should we run this without the init file?

rigettizach: @stylewarning Yes, and that's how `$(SBCL)` is configured. If the init file is loaded, it defeats the ability to override the Quicklisp location (and could taint the binary with all kinds of unknown stuff).

stylewarning: oops, missed that, thanks!

Issue #84, opened 2019-02-08
Implement a bespoke translator for CCNOT


ecpeterson:
There are standard low-depth circuits implementing CCZ and CCNOT from CZ, which quilc does not discover through generic compilation. Implement this as a translator. (If appropriate, additionally implement a dependent UCR -> CCZ translator!)

stylewarning:
@ecpeterson This is done, yes?

Issue #18, opened 2019-01-15
quilc is full of all of these educated-guess durations


stylewarning:
The file `chip-specification.lisp` is a pretty big offender. Hard-coded duration constants everywhere. But other files have it to. The approach should maybe be a bit more principled.

ecpeterson:
Made configurable in #210.

PR #130, opened 2019-03-01
Install SWAPs on recently used links

### LINKED ISSUE #80: Encourage scheduling of SWAPs on recently used links
Under certain bounded simplification assumptions (e.g.: all 2Q gates can represented using three CZs) as well as certain gateset assumptions (SWAP takes three CZs to represent), it becomes advantageous to schedule SWAPs onto link lines which already have 2Q gates on them, so that (at least) the 2Q gates preceding the SWAP become "free". This is not reflected in the current "stopwatch depth heuristic" that the addresser uses to select SWAPs. Refine this heuristic to take advantage of this.

I think this issue is easy for someone with surgical familiarity with the addresser and fairly hard otherwise.

ecp-rigetti:
This PR changes the behavior of the `:GREEDY-QUBIT` SWAP-search strategy on CZ- or CNOT-based chips: by knowing that the compressor will eventually emit a two-qubit instruction sequence no longer than `optimal-2q` guarantees, we can elect to install a SWAP onto a recently used link, knowing that it will collapse the preceding instructions that we recently placed there.

Without this improvement, quilc decomposed `CCNOT 0 1 2` on a linear chip into a sequence of multiqubit gate depth of 11; with this it comes down to 9; and with this *and* an all-to-all CCNOT translator it comes down to 7.

Closes #80 and #84 . (#80 could be extended to cover the `:A*` search strategy too, but that looks quite complicated.)

Has a bit of a kludge in it for calculating the expected maximum duration. Improving this kludge sounds related to #18 .

### DIFF
--- a/src/addresser/outgoing-schedule.lisp
+++ b/src/addresser/outgoing-schedule.lisp
@@ -97,6 +97,13 @@ BEFORE-INST to make use of RESOURCE."
     (chip-schedule-end-time schedule inst)
     0))
 
+(defun chip-schedule-resource-carving-point (schedule resource)
+  "Returns the latest time in SCHEDULE after which RESOURCE does not communicate with any other resources (i.e., RESOURCE is only touched by instructions whose resource utilization is a subset of RESOURCE)."
+  (loop :for instr :being :the :hash-keys :of (chip-schedule-times schedule)
+      :when (and (resources-intersect-p (instruction-resources instr) resource)
+                 (not (resource-subsetp (instruction-resources instr) resource)))
+        :maximize (chip-schedule-end-time schedule instr)))
+
 (defun chip-schedule-qubit-times (schedule)
   "Find the first time a qubit is available, for each qubit in the schedule."
   (map 'vector
--- a/src/chip-specification.lisp
+++ b/src/chip-specification.lisp
@@ -439,6 +439,10 @@ MISC-DATA is a hash-table of miscellaneous data associated to this hardware obje
       (vector-push-extend (lambda (instr)
                             (CNOT-to-native-CNOTs chip-spec instr))
                           ret))
+    ;; We make this unconditional. We could later conditionalize it if
+    ;; we happen to have better CCNOT translations for specific target
+    ;; gate sets.
+    (vector-push-extend #'CCNOT-to-CNOT ret)
     (cond
       ((optimal-2q-target-meets-requirements architecture ':cz)
        (vector-push-extend #'ucr-compiler ret))
--- a/src/chip-specification.lisp
+++ b/src/chip-specification.lisp
@@ -322,6 +322,14 @@ MISC-DATA is a hash-table of miscellaneous data associated to this hardware obje
     ;; set up the basic optimal 2Q compiler
     (vector-push-extend (optimal-2q-compiler-for type)
                         (hardware-object-compilation-methods obj))
+    ;; based on the optimal 2Q compiler, tag this hardware object with the
+    ;; longest duration any compiled sequence of instructions could possibly
+    ;; take to run on it.
+    (when (or (member ':cz type)
+              (member ':iswap type))
+      ;; TODO: compute this based on duration data
+      (setf (gethash "time-bound" misc-data)
+            #.(+ (* 3 150) (* 6 9)))) ; this is the maximum amt of time that a 2Q program might take
     ;; return the qubit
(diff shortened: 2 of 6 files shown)


ecp-rigetti:
P.S.: `chip-schedule-resource-carving-point` is written about as naively as possible, and in its average case it will scan over the entire outgoing program to make its decision. By cleverly walking the logical scheduler, I think it's possible to improve this to be of order polynomial in the number of qubits on the chip and independent of the number of instructions in the program. This seems a *lot* harder to get right, though.

src/addresser/outgoing-schedule.lisp
```diff
@@ -97,6 +97,16 @@ BEFORE-INST to make use of RESOURCE."
     (chip-schedule-end-time schedule inst)
     0))
 
+(defun chip-schedule-resource-carving-point (schedule resource)
```
stylewarning: document

src/addresser/outgoing-schedule.lisp
```diff
@@ -97,6 +97,16 @@ BEFORE-INST to make use of RESOURCE."
...
+  (let ((max-start-time 0))
+    (dohash ((instr instr-start-time) (chip-schedule-times schedule))
```
stylewarning: this can all be done with `loop`

```
(loop :for instr :being :the :hash-keys :of (chip-sche...)
        :using (hash-value instr-start-time)
      :when (and ...)
        :maximize instr-start-time)
```

src/chip-specification.lisp
```diff
@@ -322,6 +322,14 @@ MISC-DATA is a hash-table of miscellaneous data associated to this hardware obje
...
     (vector-push-extend (optimal-2q-compiler-for type)
                         (hardware-object-compilation-methods obj))
+    ;; based on the optimal 2Q compiler, tag this hardware object with the
+    ;; longest duration any compiled sequence of instructions could possibly
+    ;; take to run on it.
+    (when (or (member ':cz type)
+              (member ':iswap type))
+      ;; TODO: compute this based on duration data
+      (setf (gethash "time-bound" misc-data)
+            504))                       ; (+ (* 3 150) (* 6 9))
```
stylewarning: you can just write `#.(+ (* 3 150) (* 6 9))` instead of `obtuse-number        ; clarifying calculation`

ecp-rigetti:
I'll add a couple of tests, but tbh I'm not sure what's appropriate.

ecp-rigetti:
A couple other fun statistics: compiling a random 4Q gate to target a linear topology, this lowered the CZ depth of the full compilation from 224 to 211 and the state-aware version from 211 to 193.

stylewarning:
@ecp-rigetti The test is failing. Can you investigate?

src/chip-specification.lisp
```diff
@@ -439,6 +447,9 @@ MISC-DATA is a hash-table of miscellaneous data associated to this hardware obje
       (vector-push-extend (lambda (instr)
                             (CNOT-to-native-CNOTs chip-spec instr))
                           ret))
+    (when (or (find ':cnot (alexandria:ensure-list architecture))
```
stylewarning: We ought to install this unconditionally. It gives better `:iswap` compilations as well.

stylewarning:
@ecp-rigetti If we make the above change, then we can add this new test:
```
(deftest test-ccnot-compilation-on-cphase-iswap ()
  "Test that CCNOT compiles nicely on a line having the (:CPHASE ISWAP) architecture."
  (let* ((chip (quil::build-nq-linear-chip 3 :architecture '(:cphase :iswap)))
         (orig-prog (quil::parse-quil-string "CCNOT 0 1 2"))
         (orig-matrix (quil::parsed-program-to-logical-matrix orig-prog))
         (proc-prog (quil::compiler-hook orig-prog chip))
         (proc-matrix (quil::parsed-program-to-logical-matrix proc-prog))
         (2q-code (program-2q-instructions proc-prog)))
    (is (matrix-equals-dwim orig-matrix proc-matrix))
    (is (every (link-nativep chip) 2q-code))
    ;; NOTE: Decomposing into five 2q gates is more of a regression
    (is (= 5 (length 2q-code)))))
```

src/chip-specification.lisp
stylewarning: I added this in a commit below.

PR #141, opened 2019-03-12
Gate fusion transformation

stylewarning:
This PR is/will be an implementation of gate fusion for faster simulation.


src/analysis/fusion.lisp
```diff
@@ -0,0 +1,148 @@
...
+  (:documentation "A node in a program grid, representing a single gate."))
+
+(defun preceding-node-on-qubit (gn qubit)
+  "What is the incoming node to the GRID-NODE GN on qubit QUBIT?"
+  (let ((wire (position qubit (grid-node-qubits gn))))
+    (if (null wire)
+        nil
+        (aref (grid-node-back gn) wire))))
+
+(defun (setf preceding-node-on-qubit) (new-node gn qubit)
```
notmgsk: defun setf is my favourite thing about common lisp

src/analysis/fusion.lisp
```diff
@@ -0,0 +1,526 @@
...
+;;; slot.) Instead, GRID-NODEs just contain the information necessary
+;;; to determine both if and how it can get fused with other, usually
+;;; neighboring nodes.
+;;;
+;;; The other object of interest is the PROGRAM-GRID. This is nothing
+;;; more than a record of which nodes are "roots" (i.e., the start of
+;;; a program), and which nodes are "trailers" (i.e., the last nodes
+;;; to exist on a qubit).
+;;;
+;;; The main way to extend the fusion routines is to teach it *when*
```
ecp-rigetti: I like the sound of another little engine in CL-QUIL.

src/analysis/fusion.lisp
```diff
@@ -0,0 +1,526 @@
...
+  (let* ((back (grid-node-back node))
+         (x (aref back 0)))
+    (if (null x)
+        (values nil nil)
+        (loop :for i :from 1 :below (length back)
+              :unless (eq x (aref back i))
+                :do (return (values nil nil))
+              :finally (return (values t x))))))
+
+(defun jam-node-in (node)
```
ecp-rigetti: Weird function to stand alone, but I think I can see where you'd want to use it.

src/analysis/fusion.lisp
```diff
@@ -0,0 +1,526 @@
...
+      ;; We don't want to see these nodes and they won't be a
+      ;; part of the final program, since they're now fused.
+      (setf (gethash old1 seen-table) t
+            (gethash old2 seen-table) t)
+      (remhash old1 final-table)
+      (remhash old2 final-table)
+      new-node)))
+
+(defun fuse-trivially (pg)
+  "Perform only trivial fusions (i.e., fuse nodes that are trivially subsumed in some direction) on the program grod PG."
```
ecp-rigetti: `grod`

src/analysis/fusion.lisp
```diff
@@ -0,0 +1,526 @@
...
+  (let* ((nodes (program-grid-nodes pg))
+         (seen (make-hash-table :test 'eq
+                                :size (length nodes)))
+         (final (make-hash-table :test 'eq
+                                 :size (length nodes))))
+    (dolist (old-node nodes)
```
ecp-rigetti: Is there ever a scenario where you have, like, nodes 0 <-> 1 <-> 2 <-> 3, you fuse nodes 0 and 1 then nodes 2 and 3, then you don't fuse the results bc both the results were marked as final? I think the answer in this example is 'no', but I get confused if I try to think about a slightly longer chain, and it seems like it might depend on the sort order of the reply from `program-grid-nodes`.

src/analysis/fusion.lisp
```diff
@@ -0,0 +1,526 @@
...
+;;; THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+;;; (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+;;; OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+(defun sort-program-grid (pg)
+  "Produce a list of nodes of the PROGRAM-GRID PG a topologically sorted order, according to a temporal-<= relation.
+
+The list will actually be a list of lists, where each sublist commutes and can be freely ordered. (Note that even up to permutation of the elements of the sublists, the answer is still not unique.)"
+  (let* ((sorted        nil)            ; Final sorted list.
+         ;; XXX: Note: We could merge NEEDS-SORTING with SORTED-SET if
```
ecp-rigetti: IDK what "XXX" means to you, but to me it means "an error lives here".

src/analysis/fusion.lisp
```diff
@@ -0,0 +1,526 @@
...
+         (needs-sorting (program-grid-nodes pg))
+         (sorted-set    (make-hash-table :test 'eq :size (length needs-sorting)))
+         (sinks         (remove-if-not #'root-node-p needs-sorting)))
+    (flet ((dependencies-sorted-p (node)
+             (every (lambda (pred)
+                      (or (null pred)
+                          (gethash pred sorted-set)))
+                    (grid-node-back node))))
+      (loop :until (null sinks)
+            :do (progn
```
ecp-rigetti: surely u know that `:do` has an implicit `progn`

src/analysis/fusion.lisp
```diff
@@ -0,0 +1,526 @@
...
+                                 ;; sinks. It must be cyclic!
+                                 (error "Cannot sort a cyclic graph. ~
+                                         The cycles are ~S." needs-sorting)))))))
+
+
+;;;;;;;;;;;;;;;;;;;; Fusion on Real Programs (TM) ;;;;;;;;;;;;;;;;;;;;
+
+(defmethod fuse-objects ((a gate-application) (b gate-application))
+  ;; A and B are temporally ordered.
+  (premultiply-gates (list a b)))
```
ecp-rigetti: 😎

src/analysis/fusion.lisp
```diff
@@ -0,0 +1,526 @@
...
+      ;; XXX: Note that with this approach to identifying fuseable
+      ;; gates, certain gate sequences will *NOT* be fused. For
+      ;; instance, consider the following program:
+      ;;
+      ;;     X 0
+      ;;     RZ(theta) 1
+      ;;     X 0
+      ;;
+      ;; The RZ will stop the sequence of X gates from getting fused.
+      (partition-sequence-into-segments #'fuseable-gate-p code)
```
ecp-rigetti: I'm not sure why this variant is implemented if it suffers from the above comment. I'd think that you could, say, split the code vector into segments of pure-quantum instructions and do fusion on those whole blocks, perhaps with some extra logic to prevent you from fusing gates once they achieved some monstrous size. What's the reason you wouldn't want to do that? Or is that kind of decision-making part of the to-be-explored "When?" question from far above?

src/analysis/fusion.lisp
stylewarning: The answer is "no" because the final table is just keeping tabs on what I need to know at the end, and doesn't restrict what gets fused.

src/analysis/fusion.lisp
stylewarning: It's just nice to have at the top-level I guess, not that I'd generically use it very much outside of this file

src/analysis/fusion.lisp
stylewarning: the final table is also updated with each fusion pass

src/analysis/fusion.lisp
stylewarning: I like to be able to slurp the entire loop body into a C-k, and the indentation always looks weird without progn

src/analysis/fusion.lisp
stylewarning: The very last question is right. This is the kind of thing to explore when figuring out the `when?` question. For now, it seems OK to just be a little naive until more complicated fusion logic is introduced. The purpose of this right now is to at least partition _definitely_ non-fuseable items. (It would be cool to "fuse" things with e.g. `RESET` or `MEASURE` at some point.)

Right now, fusion doesn't add any additional size to the program and this is Good Enough (TM).

stylewarning:
I think this is ready for review. It's not 100% extensible and doesn't do any complicated fusing, but it's a good foundation, and I don't want this PR to blow up in size.

tests/analysis-tests.lisp
```diff
@@ -269,3 +270,92 @@ B(b) 0 1
...
+
+(deftest test-grid-node ()
+  "Tests on the grid node data structure and associated functions."
+  (let ((q0 (quil::make-grid-node 'q0 0))
+        (q1 (quil::make-grid-node 'q1 1))
+        (q2 (quil::make-grid-node 'q2 2))
+        (q01 (quil::make-grid-node 'q01 0 1))
+        (q12 (quil::make-grid-node 'q12 1 2))
+        (q02 (quil::make-grid-node 'q02 0 2)))
+    ;; All notes start as root nodes.
```
rigettizach: Typo notes -> nodes

PR #145, opened 2019-03-15
ignore network errors when checking for sdk update

notmgsk:


app/src/entry-point.lisp
```diff
@@ -276,7 +276,7 @@
   #+forest-sdk
   (unless skip-version-check
     (multiple-value-bind (available-p version)
-        (sdk-update-available-p)
+        (ignore-errors (sdk-update-available-p))
       (when available-p
           (format t "An update is available to the SDK. You have version ~A. ~
```
stylewarning:
```suggestion
        (format t "An update is available to the SDK. You have version ~A. ~
```

PR #154, opened 2019-03-21
document make-rewriting-rule, other reorganizing

### LINKED ISSUE #153: Document MAKE-REWRITING-RULE
It's not documented but it's a macro with peculiar syntax.

stylewarning:
Fixes #153

### DIFF
--- a/src/compressor/rewriting-rule-data-type.lisp
+++ b/src/compressor/rewriting-rule-data-type.lisp
@@ -106,7 +106,7 @@ BODY is a list of forms that, by the end, should construct a list of instruction
 
 ;;; dumb little macros for cond guards that we were repeating a million times
 
-(defmacro else-give-up-compilation (test &body body)
+(defmacro give-up-compilation-unless (test &body body)
   "If TEST passes, proceed to execute BODY (and return the value of the final expression). If TEST fails, call GIVE-UP-COMPILATION."
   `(cond
      (,test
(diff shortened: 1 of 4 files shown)


src/compressor/rewriting-rule-data-type.lisp
```diff
@@ -19,14 +19,31 @@ CONSUMER is a function that consumes a COMPRESSOR-CONTEXT followed by a list of
...
+  "Create a new REWRITING-RULE object.
+
+READABLE-NAME is an evaluated parameter that should result in a STRING.
+
+CONTEXT-VAR should be a symbol that will be bound to a context, or the symbol _ should it be ignored.
+
+BINDING-CLAUSES is a list of bindings, as if by OPERATOR-MATCH. Specifically, it has the following grammar:
+
+    <BINDING-CLAUSES> ::= (<MATCH-CLAUSE> <VARIABLE>)
+    <MATCH-CLAUSE>    ::= (<GATE-NAME> (<PARAM>*) <QUBIT> <QUBIT>*)
```
ecp-rigetti: It might also be the case that (<PARAM>*) can be replaced by _, in which case all the parameters are ignored *and* their length is not matched against.

src/compressor/rewriting-rule-data-type.lisp
```diff
@@ -42,3 +59,66 @@ CONSUMER is a function that consumes a COMPRESSOR-CONTEXT followed by a list of
...
+(defun param-+ (arg1 arg2)
+  (param-binary-op '+ arg1 arg2))
+
+(defun param-* (arg1 arg2)
+  (param-binary-op '* arg1 arg2))
+
+
+;;; dumb little macros for cond guards that we were repeating a million times
+
+(defmacro else-give-up-compilation (test &body body)
```
ecp-rigetti: While you're playing with this, consider following up on your own suggestion of renaming `else-give-up-compilation` to `give-up-compilation-unless`.

PR #155, opened 2019-03-21
allow transforms w/missing deps to be restarted

stylewarning:


src/package.lisp
```diff
@@ -37,6 +37,21 @@
...
+   #:unsatisfied-transform-dependency-object
+                                        ; READER
+   #:unsatisfied-transform-dependency-needed-transform
+                                        ; READER
+   #:unsatisfied-transform-dependency-attempted-transform
+                                        ; READER
+   #:transformable                      ; CLASS
+   #:transformedp                       ; GENERIC, METHOD
+   #:transform                          ; GENERIC, METHOD
+   #:define-transform                   ; MACRO
```
notmgsk: how useful vs ugly are all these `; MACRO` etc

stylewarning:
> Very cool. CL-QUIL could be a lot more usable in interactive mode than it currently is.

@ecp-rigetti File issues!

ecp-rigetti:
no u

src/transformable-mixin.lisp
```diff
@@ -88,24 +88,51 @@ If any of XFORM or its predecessors have not been performed, return NIL and the
   ;; Check all of its predecessors.
   (validate-instance-for-transform-predecessors xform instance))
 
+(define-condition unsatisfied-transform-dependency (error)
+  ((attempted-transform :reader unsatisfied-transform-dependency-attempted-transform
+                        :initarg :attempted-transform)
+   (needed-transform :reader unsatisfied-transform-dependency-needed-transform
+                     :initarg :needed-transform)
+   (object :reader unsatisfied-transform-dependency-object
+           :initarg object))
```
rigettizach: This should be `:object` instead of `object`. (I thought this would signal an unknown initarg error, but it just errors when trying to print the unbound slot instead.)

src/transformable-mixin.lisp
```diff
@@ -88,24 +88,51 @@ If any of XFORM or its predecessors have not been performed, return NIL and the
...
+  (:documentation "Perform and record the transform XFORM on the instance INSTANCE. Pass INSTANCE and ARGS to the respective function representing this transform.
+
+Signals UNSATISFIED-TRANSFORM-DEPENDENCY if the transform cannot be applied because a transform is needed."))
 
 (defmethod transform :before ((xform symbol) instance &rest args)
   (declare (ignore args))
   (assert (find-transform xform) (xform)
           "The transform ~S is unknown to me." xform)
-  (tagbody
+  (prog ((attempts-left *max-transform-attempts*))
```
rigettizach: This is cute - I might reach for `dotimes`, e.g. `(dotimes (attempts-left *max-transform-attempts* (error ...))` with a return in the body, which eliminates the count maintenance and go statements.

rigettizach: Of course, with dotimes, the variable name doesn't matter, but `attempts-left` would be particularly unsuitable.

PR #157, opened 2019-03-25
Fix type checking of instructions in clifford stuff

notmgsk:
I assume this code was written before pragmas etc were introduced. Executable
code now contains non-application stuff (pragmas) and so must be careful with
types.

Fixes rigetti/pyquil#858.

### DIFF
--- a/src/clifford/benchmarking-procedures.lisp
+++ b/src/clifford/benchmarking-procedures.lisp
@@ -185,29 +185,23 @@
 
 (defun extract-cliffords (parsed-quil)
   "Given PARSED-QUIL generate the CLIFFORD for each gate"
-  (loop :for gate-application :across (quil:parsed-program-executable-code parsed-quil)
-        ;; Collect cliffords of all applications (and unresolved
-        ;; applications). If any other instruction type is
+  (loop :for instr :across (quil:parsed-program-executable-code parsed-quil)
+        ;; Collect cliffords of all applications (and pragmas,
+        ;; unresolved applications). If any other instruction type is
         ;; encountered, raise error.
-        :unless (or (typep gate-application 'quil:application)
+        :unless (or (typep instr 'quil:application)
+                    (typep instr 'quil:pragma)
                     (and quil::*allow-unresolved-applications*
-                         (typep gate-application 'quil:unresolved-application)))
-                (error "ouf")
-        :collect (matrix-to-clifford (quil:gate-matrix gate-application))))
-
-(defun extract-qubits-used (parsed-quil)
-  "Given PARSED-QUIL return the indices of the qubits used. The result is given as a list of the qubits used per instruction in the PARSED-QUIL."
-  (loop :for parsed-clifford :across (quil:parsed-program-executable-code parsed-quil)
-        :when (or (typep parsed-clifford 'quil:application)
-                  (and quil::*allow-unresolved-applications*
-                       (typep parsed-clifford 'quil:unresolved-application)))
-          :collect (mapcar #'quil:qubit-index (quil:application-arguments parsed-clifford))))
+                         (typep instr 'quil:unresolved-application)))
+                (error "Cannot extract clifford from the instr ~/cl-quil:instruction-fmt/"
+                       instr)
+        :collect (matrix-to-clifford (quil:gate-matrix instr))))
 
 (defun clifford-circuit-p (parsed-quil)
   "If the parsed circuit PARSED-QUIL a clifford circuit, return the CLIFFORD corresponding to it. Otherwise return NIL. This will generate a clifford that acts on the number of qubits in the program, rather than a number of qubits that is the difference between the maximum and minimum index."
   (let* ((cliffords (extract-cliffords parsed-quil))
-         (qubit-targets (extract-qubits-used parsed-quil))
-	 (qubits (sort (remove-duplicates (a:flatten qubit-targets)) #'<))
+         (qubit-targets (cl-quil::qubits-used parsed-quil))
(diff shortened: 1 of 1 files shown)


notmgsk:
Had a look round and don't see anymore places where the same fix needs applying.

ecpeterson:
This might also give unexpected behavior in re: https://github.com/rigetti/pyquil/issues/858 , since it strips the PRAGMAs and then doesn't also reattach them to the output program.

Thinking about it some, I'm not sure that this is possible; doesn't the benchmarking API return a list of gate *indices*, so that it's not even possible to interleave non-gate-set-instruction directives? It might be :cool: to allow RPCQ to emit a special reply that contains a warning to be raised with `warn` + an actual payload to be returned as the call result, so that the RPCQ server could lightly slap Marcus's wrist when he passes in a program like this.

notmgsk:
Good points, @ecpeterson. I can imagine how an identity-preserving pragma would be useful, however.

ecpeterson:
For what it's worth, qubits will never be relabeled by this routine, the program that comes out is guaranteed to agree semantically with the program that goes in (+ conjugation), and it's generally not possible to do make such a program without doing a bit of rewriting. A passed-through PRAGMA would only have an effect when the output was then further routed to one of the more traditional quilc compilation routines, and it feels like a misuse to me here. I like the idea of raising a warning.

stylewarning:
Something about this doesn't feel right, both the old and new.

This *is* fixing something, but it's also letting things go wrong. What if there are classical instructions, what if there are other things?

I feel like there should be a type _check_ that requires the input to have some property. If you run across something that's neither a pragma nor a gate application, perhaps it shall error?

What do you guys think? @ecpeterson @notmgsk

ecpeterson:
I think installing such an error is a good idea.

notmgsk:
@stylewarning do we need a word for that? Something like "semiprotoquil"

ecpeterson:
If this https://github.com/rigetti/rpcq/pull/45 gets in first, I'd also recommend installing a `warn` on the PRAGMA path.

src/clifford/benchmarking-procedures.lisp
```diff
@@ -169,30 +169,32 @@
 (defun extract-cliffords (parsed-quil)
```
stylewarning: we need to do proper error checking

src/clifford/benchmarking-procedures.lisp
```diff
@@ -169,30 +169,32 @@
...
+                                                  (magicl:conjugate-transpose gate))))
+                                 (assert (not (null conj)) () "The given program does not represent a Clifford element.")
+                                 (pauli-from-string (concatenate 'string phase conj)))))))))
 
 
 (defun extract-qubits-used (parsed-quil)
   "Given PARSED-QUIL return the indices of the qubits used. The result is given as a list of the qubits used per instruction in the PARSED-QUIL."
   (loop :for parsed-clifford :across (quil::parsed-program-executable-code parsed-quil)
-        :collect (mapcar #'quil::qubit-index (quil::application-arguments parsed-clifford))))
+        :when (typep parsed-clifford 'quil::application)
```
notmgsk: Should this type check for `gate-application` instead?

src/clifford/benchmarking-procedures.lisp
notmgsk: @stylewarning check each instruction type for pragma or gate application?

src/clifford/benchmarking-procedures.lisp
stylewarning: either `gate-application`, or `unresolved-application` when `*allow-unresolved-applications*` is `t`

src/clifford/benchmarking-procedures.lisp
stylewarning: something like that yes

marcusps:
What the status for this? Just curious.

stylewarning:
@marcusps There are some review comments to be addressed.

src/clifford/benchmarking-procedures.lisp
```diff
@@ -185,13 +185,23 @@
...
+        ;; Collect cliffords of all applications (and unresolved
+        ;; applications). If any other instruction type is
+        ;; encountered, raise error.
+        :unless (or (typep gate-application 'quil:application)
+                    (and quil::*allow-unresolved-applications*
+                         (typep gate-application 'quil:unresolved-application)))
+                (error "ouf")
         :collect (matrix-to-clifford (quil:gate-matrix gate-application))))
 
 (defun extract-qubits-used (parsed-quil)
```
stylewarning: i think this function is wrong, and we should use (or augment) the framework for doing this in the analysis/ folder. Check the qubits-used pass; add to it if we need to get qubits.

src/clifford/benchmarking-procedures.lisp
```diff
@@ -185,13 +185,23 @@
...
   "Given PARSED-QUIL generate the CLIFFORD for each gate"
-  (loop :for gate-application :across (quil::parsed-program-executable-code parsed-quil)
+  (loop :for gate-application :across (quil:parsed-program-executable-code parsed-quil)
+        ;; Collect cliffords of all applications (and unresolved
+        ;; applications). If any other instruction type is
+        ;; encountered, raise error.
+        :unless (or (typep gate-application 'quil:application)
+                    (and quil::*allow-unresolved-applications*
+                         (typep gate-application 'quil:unresolved-application)))
+                (error "ouf")
```
stylewarning: obv change this message

PR #164, opened 2019-04-01
Fixup/various

ecpeterson:
This PR has a handful of fixes that have come up in testing:

* More work on the numerical stability of `collinearp`.
* Fix the ordering of instruction qubits during state-aware compression: the wavefunction is ordered against `wf-qc`, which agrees with `qubits-on-obj` as sets but which might not have the same ordering.
* Improves the runtime of the test suite by generating smaller matrices during careful compression.
* Make it easier to see bad state on assert.

### DIFF
--- a/src/compressor/compressor.lisp
+++ b/src/compressor/compressor.lisp
@@ -490,12 +490,12 @@ other's."
                             (eql final-wf-reduced-prep ':not-simulated))
                        (collinearp final-wf final-wf-reduced-prep))))
              (assert final-wf-reduced-instrs-collinearp
-                     ()
+                     (final-wf final-wf-reduced-instrs)
                      "During careful checking of instruction compression, the produced ~
                    wavefunction by instruction reduction was detected to not be ~
                    collinear with the target wavefunction.")
              (assert final-wf-reduced-prep-collinearp
-                     ()
+                     (final-wf final-wf-reduced-prep)
                      "During careful checking of instruction compression, the produced ~
                    wavefunction by state prep reduction was detected to not be ~
                    collinear with the target wavefunction.")))
--- a/src/compressor/compressor.lisp
+++ b/src/compressor/compressor.lisp
@@ -56,10 +56,12 @@
 
 (defun qubits-in-instr-list (instructions)
   "Produces a list of all of the (unboxed) qubit indices appearing in INSTRUCTIONS, a list of applications."
-  (remove-duplicates (mapcan (lambda (isn)
-                               (when (typep isn 'application)
-                                 (application-qubit-indices isn)))
-                             instructions)))
+  (delete-duplicates
+   (mapcan (lambda (isn)
+             (cond
+               ((typep isn 'application)
+                (application-qubit-indices isn))))
+           instructions)))
 
 
 ;;; helper functions for algebraically-reduce-instructions, including the doubly-linked
@@ -372,6 +374,9 @@ other's."
                        (notany (lambda (instr) (typep instr 'state-prep-application))
                                instrs))
               (let* ((reassignment
+                      ;; the actual reassignment we use is unimportant. this is
+                      ;; more along the lines of COMPRESS-QUBITs, so that our
+                      ;; matrices don't take up quite so much space.
                       (loop :for i :in (union (qubits-in-instr-list instrs)
                                               (qubits-in-instr-list translation-results))
                             :for j :from 0
@@ -465,11 +470,15 @@ other's."
     (labels
         ((check-quil-agrees-as-matrices ()
            (let* ((relabeling
+                   ;; the actual reassignment we use is unimportant. this is
+                   ;; more along the lines of COMPRESS-QUBITs, so that our
+                   ;; matrices don't take up quite so much space.
                    (loop :for i :in (reduce #'union
-                                            (list (qubits-in-instr-list instructions)
-                                                  (qubits-in-instr-list reduced-instructions)
-                                                  (qubits-in-instr-list decompiled-instructions)
-                                                  (qubits-in-instr-list reduced-decompiled-instructions)))
+                                            (list instructions
+                                                  reduced-instructions
+                                                  decompiled-instructions
+                                                  reduced-decompiled-instructions)
+                                            :key #'qubits-in-instr-list)
                          :for j :from 0
                          :collect (cons i j)))
                   (stretched-matrix (or (make-matrix-from-quil instructions :relabeling relabeling)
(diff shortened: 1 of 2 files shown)


src/compressor/compressor.lisp
```diff
@@ -56,7 +56,10 @@
 
 (defun qubits-in-instr-list (instructions)
   "Produces a list of all of the (unboxed) qubit indices appearing in INSTRUCTIONS, a list of applications."
-  (remove-duplicates (mapcan #'application-qubit-indices instructions)))
+  (remove-duplicates (mapcan (lambda (isn)
```
stylewarning: if you're going to `mapcan`, might as well `delete-duplicates`

src/compressor/compressor.lisp
```diff
@@ -56,7 +56,10 @@
 
 (defun qubits-in-instr-list (instructions)
```
stylewarning: can this function have `MEASURE` or `RESET`? Does that need to be taken into account?

src/compressor/compressor.lisp
```diff
@@ -368,13 +371,23 @@ other's."
                        (not *enable-approximate-compilation*)
                        (notany (lambda (instr) (typep instr 'state-prep-application))
                                instrs))
-              (let ((ref-mat (make-matrix-from-quil (list (first instrs))))
-                    (mat (make-matrix-from-quil translation-results)))
-                (setf mat (kron-matrix-up mat (1- (integer-length (magicl:matrix-rows ref-mat)))))
+              (let* ((reassignment
+                      (loop :for i :in (union (qubits-in-instr-list instrs)
```
stylewarning: i assume this order is arbitrary. can you comment that that's OK?

src/compressor/compressor.lisp
```diff
@@ -368,13 +371,23 @@ other's."
...
+                                              (qubits-in-instr-list translation-results))
+                            :for j :from 0
+                            :collect (cons i j)))
+                     (ref-mat (make-matrix-from-quil (list (first instrs))
+                                                     :relabeling reassignment))
+                     (mat (make-matrix-from-quil translation-results
+                                                 :relabeling reassignment))
+                     (kron-size (max (1- (integer-length (magicl:matrix-rows ref-mat)))
```
stylewarning: we ought to just write a utility function `ilog2` or so (following `isqrt`). This pattern is used so often

src/compressor/compressor.lisp
```diff
@@ -390,12 +403,12 @@ other's."
     (labels
         (;; produce a sequence of native instructions that have the effect of
          ;; carrying START-WF to FINAL-WF (= INSTRUCTIONS |START-WF>)
-         (decompile-instructions-into-state-prep (start-wf final-wf)
+         (decompile-instructions-into-state-prep (start-wf final-wf qc)
            (expand-to-native-instructions
             (list (make-instance 'state-prep-application
                                  :source-wf (copy-seq start-wf)
                                  :target-wf final-wf
-                                 :arguments (nreverse (mapcar #'qubit qubits-on-obj))))
```
stylewarning: any reason for this change?

src/compressor/compressor.lisp
```diff
@@ -451,25 +464,37 @@ other's."
...
-             (let* ((decompiled-matrix
-                     (make-matrix-from-quil decompiled-instructions))
-                    (reduced-matrix
-                     (kron-matrix-up (make-matrix-from-quil reduced-instructions)
-                                     (1- (integer-length (magicl:matrix-rows stretched-matrix)))))
-                    (reduced-decompiled-matrix
-                     (kron-matrix-up (make-matrix-from-quil reduced-decompiled-instructions)
-                                     (1- (integer-length (magicl:matrix-rows stretched-matrix))))))
+           (let* ((relabeling
+                   (loop :for i :in (reduce #'union
```
stylewarning: comment about ordering

also
```
(reduce #'union (list instrs reduced-instrs ...) :key #'qubits-in-instr-list)
```
is an alternative

src/matrix-operations.lisp
```diff
@@ -259,21 +259,17 @@ as needed so that they are the same size."
...
-                            (not (double= 0d0 s)))
-                       (and (not (double= 0d0 r))
-                            (double= 0d0 s)))
-                   (return-from collinearp nil))
-                  ((and (not (double= 0.0d0 (abs r)))
-                        (< (abs (- 1 (abs r))) peak-abs))
-                   (setf peak-abs (abs (- 1 (abs r))))
-                   (setf scalar (/ r s)))))
+            :when (and (not (double= 0.0d0 (abs r)))
+                       (not (double= 0d0 (abs s)))
```
stylewarning:
```suggestion
                       (not (double= 0.0d0 (abs s)))
```

src/matrix-operations.lisp
```diff
@@ -259,21 +259,17 @@ as needed so that they are the same size."
...
+                       (< (abs (- 1 (abs r))) peak-abs))
+            :do (setf peak-abs (abs (- 1 (abs r))))
+                (setf scalar (- (phase r) (phase s))))
       ;; rather than computing the full norm, this could be replaced by the
       ;; component-wise calculation (loop ... :always (double= 0d0 (expt ...)))
       ;; which has the potential to terminate early on a negative result
-      (let ((norm (sqrt (dot-product vect1 vect2 scalar))))
-        (double= 1d0 norm)))))
+      (when scalar
```
stylewarning: i would replace `when` with `and` since you care about returning `nil`

src/compressor/compressor.lisp
ecpeterson: It can't, and the docstring says something to this effect, but it might be better to make it robust.

On the other hand, this PR isn't introducing this function, so...

src/compressor/compressor.lisp
ecpeterson: This is the main change of this PR: `qubits-on-obj` and `wf-qc` are guaranteed to be equal as sets, but not as lists, and the order matters for how the entries in `source-wf` and `final-wf` are interpreted. `wf-qc` is the list with the correct ordering.

src/compressor/compressor.lisp
ecpeterson: nice, i like this

src/compressor/compressor.lisp
stylewarning: why `reverse` instead of `nreverse` though?

stylewarning: why change nreverse to reverse?

ecpeterson: oh, no, either is fine

ecpeterson:
Note to self: I'd like to merge https://github.com/rigetti/quilc/pull/166 after this gets in.

PR #169, opened 2019-04-02
Line numbers in parser errors

### LINKED ISSUE #103: Add line/column tags to parsed Quil
It would be helpful if errors in {lexing, parsing, compiling} Quil could be assigned to a particular bad point in the input. Threading at least the line number through the output of the parser would enable this kind of error emission.

In the current arrangement, these would go into the `TOKEN` objects. Line numbers would have to be specially dealt with since the current tokenizer is line-based (cf. #93).

notmgsk:
Closes #103 

```
QUIL> (let ((*definitions-allowed* nil))
        (parse-quil "H 0
DEFCIRCUIT TEST a b:
    CNOT a b

TEST 0 1"))
```

```
At line 2: Found DEFCIRCUIT where it's not allowed.
   [Condition of type QUIL-PARSE-ERROR]
```

### DIFF
--- a/src/parser.lisp
+++ b/src/parser.lisp
@@ -224,13 +224,8 @@
 
 (defun quil-parse-error (format-control &rest format-args)
   "Signal a QUIL-PARSE-ERROR with a descriptive error message described by FORMAT-CONTROL and FORMAT-ARGS."
-  (error 'quil-parse-error :format-control format-control
-                           :format-arguments format-args))
-
-(defun quil-parse-error-extended (token format-control &rest format-args)
-  "Signal a QUIL-PARSE-ERROR with a descriptive error message, including line number, described by FORMAT-CONTROL and FORMAT-ARGS."
   (error 'quil-parse-error :format-control (concatenate 'string "At line ~A: " format-control)
-                           :format-arguments (concatenate 'list (list (token-line token)) format-args)))
+                           :format-arguments (concatenate 'list (list *line-number*) format-args)))
 
 (defvar *definitions-allowed* t
   "Dynamic variable to control whether DEF* forms are allowed in the current parsing context.")
@@ -246,7 +241,8 @@
 "
   (let* ((line (first tok-lines))
          (tok (first line))
-         (tok-type (token-type tok)))
+         (tok-type (token-type tok))
+         (*line-number* (token-line tok)))
     (case tok-type
       ;; Gate/Circuit Applications
       ((:CONTROLLED :DAGGER)
@@ -258,7 +254,7 @@
       ;; Circuit Definition
       ((:DEFCIRCUIT)
        (unless *definitions-allowed*
-         (quil-parse-error-extended tok "Found DEFCIRCUIT where it's not allowed."))
+         (quil-parse-error "Found DEFCIRCUIT where it's not allowed."))
 
        (let ((*definitions-allowed* nil)
              (*formal-arguments-allowed* t))
--- a/src/parser.lisp
+++ b/src/parser.lisp
@@ -22,14 +22,17 @@
      :STRING :INDENTATION :INDENT :DEDENT :COMPLEX :PLUS
      :MINUS :TIMES :DIVIDE :EXPT :INTEGER :NAME :AREF)))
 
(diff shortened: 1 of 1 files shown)


src/parser.lisp
```diff
@@ -238,7 +251,8 @@
 "
   (let* ((line (first tok-lines))
          (tok (first line))
-         (tok-type (token-type tok)))
+         (tok-type (token-type tok))
+         (*line-number* (token-line tok)))
```
ecpeterson: `token-line` was already a thing?? I'd been avoiding looking at this in large part because I thought `token-line` would have to be invented.

ecpeterson:
Looks good to me but this is more Robert's domain.

src/parser.lisp
stylewarning: i gotchu fam

src/parser.lisp
```diff
@@ -22,15 +22,18 @@
      :STRING :INDENTATION :INDENT :DEDENT :COMPLEX :PLUS
      :MINUS :TIMES :DIVIDE :EXPT :INTEGER :NAME :AREF)))
 
-(defvar *line-start-position*)
-(defvar *line-number*)
+(defvar *line-start-position* nil)
+(defvar *line-number* nil)
 
-(defstruct (token (:constructor tok (type &optional payload)))
+(defstruct (token (:constructor %tok (type &optional payload line)))
```
```suggestion
(defstruct (token (:constructor tok (type &optional payload (line *line-number*))))
```

src/parser.lisp
```diff
@@ -22,15 +22,18 @@
...
   "A lexical token."
   (line nil :type (or null (integer 1)))
   (type nil :type token-type)
   (payload nil))
 
+(defun tok (type &optional payload line)
```
stylewarning: you can delete this

src/parser.lisp
```diff
@@ -221,8 +224,18 @@
 
 (defun quil-parse-error (format-control &rest format-args)
   "Signal a QUIL-PARSE-ERROR with a descriptive error message described by FORMAT-CONTROL and FORMAT-ARGS."
-  (error 'quil-parse-error :format-control format-control
-                           :format-arguments format-args))
+  ;; There are callers of this function where *line-number* is not set
+  ;; appropriately (e.g. in expand-circuits.lisp) and where it may be impossible
+  ;; to set due to missing context. Most callers of this function however are
+  ;; right here in parser.lisp where *line-number* is more than likely set.
+  (let ((format-control (if *line-number*
```
stylewarning: make a new condition `quil-parse-error-at-line` which includes a line number slot, and does the formatting appropriately with its `:report`. (Don't concatenate format strings & the like.) Then in this function, dispatch based off of whether `*line-number*` has a value.

src/parser.lisp
```diff
@@ -221,8 +224,18 @@
...
+                            (concatenate 'string "At line ~A: " format-control)
+                            format-control))
+        (format-args (if *line-number*
+                         (concatenate 'list (list *line-number*) format-args)
```
stylewarning: this line will go away if you do the above change, but know that you can more simply write this as
```
(cons *line-number* format-args)
```

src/parser.lisp
```diff
@@ -222,20 +219,26 @@
   ()
   (:documentation "Representation of an error parsing Quil."))
 
+(define-condition quil-parse-error-at-line (quil-parse-error)
+  ((line :initarg :line :reader quil-parse-error-at-line-line))
```
stylewarning: rename the reader to `line-of-quil-parse-error`

src/parser.lisp
```diff
@@ -222,20 +219,26 @@
...
+  (:report (lambda (condition stream)
+             (format stream "At line ~A: "
```
stylewarning: use `~D` if we are guaranteed to always have an integer

src/parser.lisp
```diff
@@ -222,20 +219,26 @@
...
+                     (quil-parse-error-at-line-line condition))
```
stylewarning: make sure to change the corresponding reader name here

PR #170, opened 2019-04-02
Error on malformed parameter lists to DEFGATE and DEFCIRCUIT

### LINKED ISSUE #151: fix commas in defgate & defcircuit parsing
In `parser.lisp` there are two instances of:

```
            ;; Go through the supposed parameters, checking that they
            ;; are, and parsing them out.
            ;;
            ;; FIXME: This allows erroneous cases like (%a,,%a) and (%a %a)
```

notmgsk:
Closes #151

### DIFF
--- a/src/parser.lisp
+++ b/src/parser.lisp
@@ -819,9 +819,8 @@
             ;; Some sanity checks for the parameter list. Must be of odd length,
             ;; and every other token should be a :COMMA
             (unless (and (oddp (length found-params))
-                         (or (= 1 (length found-params))
-                             (loop :for c :in (rest found-params) :by #'cddr
-                                   :always (eql ':COMMA (token-type c)))))
+                         (loop :for c :in (rest found-params) :by #'cddr
+                               :always (eql ':COMMA (token-type c))))
               ;; TODO Some printer for tokens?
               (quil-parse-error "Malformed parameter list in DEFGATE: ~A" found-params))
             ;; Go through the supposed parameters, checking that they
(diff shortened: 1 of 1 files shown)


src/parser.lisp
```diff
@@ -991,13 +997,19 @@
             ;; Remove right paren and stash away params.
             (pop rest-line)
 
+            ;; Some sanity checks for the parameter list. Must be of odd length,
+            ;; and every other token should be a :COMMA
+            (unless (and (oddp (length found-params))
+                         (or (= 1 (length found-params))
```
ecpeterson: I'm surprised this is necessary. Lisp doesn't admit to vacuous truth??

notmgsk: I don't follow

ecpeterson: If `(= 1 (length found-params))` is satisfied, what does the `loop` do? Does it error because there aren't two things to bind, or because one gets bound to `nil`?

What I'd hope would happen is that the loop doesn't get run at all b/c there aren't two things to bind, and then the `:always` clause returns `T` because that's what a conjunction with no clauses ought to do ("vacuous truth").

src/parser.lisp
```diff
@@ -816,13 +816,19 @@
             ;; Remove right paren.
             (pop rest-line)
 
+            ;; Some sanity checks for the parameter list. Must be of odd length,
+            ;; and every other token should be a :COMMA
+            (unless (and (oddp (length found-params))
+                         (or (= 1 (length found-params))
```
stylewarning: you shouldn't nee the `= 1` bit and therefore the `or`

```
CL-USER> (loop :for x :in nil :by #'cddr
               :always x)
T
```

src/parser.lisp
stylewarning: this is correct; we can leave out the `= 1`

notmgsk: i smell ya

PR #173, opened 2019-04-03
initial commit of boondoggle, a random quil tester

stylewarning:
All of this was written by Eric Peterson. Edited modestly for open
source.


notmgsk:
What about @rigettizach's project? That would supersede this I guess? No point having both

stylewarning:
@rigettizach's serves a slightly different purpose: this project (boondoggle) is for producing program objects and evaluation pipelines for internal library use. Zach's project produces text for fuzzing. They could potentially be merged, but I don't think they replace one another.

ecpeterson:
`boondoggle` also generates random Quil, and I’d be happy for that part of it to be replaced and outsourced to Zach’s project. Beyond that, it provides functionality for doing things like running circuits on a QVM and on a QPU and getting statistical data about the differences in output. That sort of thing is fresh (if not super intelligently implemented).

PR #175, opened 2019-04-05
Simplify classical instruction printing

rigettizach:
Use a common superclass of unary-, binary-, and trinary-classical-instruction to share a print-instruction-generic method between them.

### DIFF
--- a/src/ast.lisp
+++ b/src/ast.lisp
@@ -235,6 +235,10 @@ EXPRESSION should be an arithetic (Lisp) form which refers to LAMBDA-PARAMS."
   (:documentation "Abstract class representing an executable instruction.")
   (:metaclass abstract-class))
 
+(defclass classical-instruction (instruction)
+  ()
+  (:metaclass abstract-class))
+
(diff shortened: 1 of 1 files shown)


src/ast.lisp
```diff
@@ -983,6 +987,9 @@ N.B., The fractions of pi will be printed up to a certain precision!")
 (defun print-instruction (instr &optional (stream *standard-output*))
   (print-instruction-generic instr stream))
 
+(defun print-instruction-to-string (instr)
+  (print-instruction instr nil))
```
ecpeterson: I'm not sure we've respected the convention that `nil` means "print to a string" in all these methods. I'd want to double check, for instance, that complex float parameters still end up where they're supposed to be when printing an instruction to `nil`.

Otherwise, looks good!

stylewarning: We do, this is forced with an `:around` method. See https://github.com/rigetti/quilc/blob/master/src/ast.lisp#L989.

PR #177, opened 2019-04-08
Short term fix for build-app.lisp on Windows, and use ':win32 in build-app.lisp

notmgsk:
There is something weird going on with build-app.lisp in Windows. My hunch is that somehow shared libraries are getting lost, but can't say more than that.

Also use `':win32` rather than `':windows` in build-app.lisp.


ecpeterson:
For my own edification: what does `zap-info` do?

stylewarning:
zap info deletes unnecessary info from the binary, like inline expansions, documentation, etc.

PR #182, opened 2019-04-09
Nicer error message when ALEXA lexing fails on unexpected input.

rigettizach:
Before:

    Couldn't find match at position 51 within the lexer CL-QUIL::LINE-LEXER.

After:

    At line 7: unexpected input text "?" in "S? 7"

### DIFF
--- a/src/parser.lisp
+++ b/src/parser.lisp
@@ -133,14 +133,44 @@
   ("[^\\S\\n\\r]+"
    nil))
 
+(defun tokenization-failure-context (input-string condition)
+  "Given an ALEXA:LEXER-MATCH-ERROR in CONDITION, return as multiple
+values the individual line and individual character within
+INPUT-STRING that triggered the condition."
+  (flet ((alexa-failure-position (condition)
+           ;; This parses the error string of ALEXA, which looks like
+           ;; "Couldn't find match at position 42 ...". This is
+           ;; obviously a fragile and silly thing to do, so when ALEXA
+           ;; is updated to have an error condition with direct access
+           ;; to the failing position, update this code to suit.
+           (let* ((text (princ-to-string condition))
+                  (start (position-if #'digit-char-p text)))
+             (parse-integer text :start start :junk-allowed t))))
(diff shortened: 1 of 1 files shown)


src/parser.lisp
```diff
@@ -133,14 +133,44 @@
...
-          :collect tok)))
+    (handler-case
+        (let ((f (funcall lexer string)))
+          (loop :for tok := (funcall f)
+            :until (null tok)
+            :collect tok))
+      (alexa:lexer-match-error (c)
+        (multiple-value-bind (context-line failing-char)
+            (tokenization-failure-context string c)
+          (error "At line ~D: unexpected input text ~S in ~S"
```
notmgsk: Would it make sense here to use `quil-parse-error` for consistency?

src/parser.lisp
```diff
@@ -133,14 +133,44 @@
   ("[^\\S\\n\\r]+"
    nil))
 
+(defun tokenization-failure-context (input-string condition)
+  "Given an ALEXA:LEXER-MATCH-ERROR in CONDITION, return as multiple
+values the individual line and individual character within
+INPUT-STRING that triggered the condition."
+  (flet ((alexa-failure-position (condition)
+           ;; This parses the error string of ALEXA, which looks like
```
stylewarning: could corresponding issues be filed in ALEXA and QUILC?

src/parser.lisp
stylewarning: I agree with this.

src/parser.lisp
rigettizach: I see you implemented the fix in ALEXA. For now, this code will work with both new and old ALEXA. Do you want to put off changing quilc until the new alexa has been released and is available in a Quicklisp update?

ecpeterson: My opinion is that since the change is already written we should merge it, and then we can do the superior thing once the new ALEXA propagates.

stylewarning: i agree with that opinion

Issue #168, opened 2019-04-01
automatically test translators


stylewarning:
Translators seem ripe for automatic testing. Maybe DEFINE-TRANSLATOR can stash away the translator somewhere, and as a part of unit tests, we can just test them all on random qubit arguments.

Thoughts @ecpeterson ?

ecpeterson:
Sounds like a good idea to me. We're already encouraging the use of macros, and the macro already receives as input a mungable form of the "input" gate.

Random qubit assignments is probably somewhat excessive, but also won't be hurting anybody.

ecpeterson:
This was "done" in #210, though Mark pointed out in #368 that there's more to be done. I'm going to close this issue in favor of that one.

PR #184, opened 2019-04-10
Test translators automatically

stylewarning:
This PR is to address #168.

- [x] Store translators so that they are later accessible.
- [ ] Write a routine to generate a matching gate to a translator pattern.
- [ ] Write the test routine.


stylewarning:
This is being usurped by better technology in a different PR.

PR #187, opened 2019-04-11
remove make-gate-matrix-from-gate-string in favour of make-matrix-from-quil

### LINKED ISSUE #172: Possible code duplication: `make-gate-matrix-from-string` vs `make-matrix-from-quil`
We recently extended `make-matrix-from-quil` to support an optional relabeling argument. I think that this new functionality subsumes an existing function: `make-gate-matrix-from-string`. I'd like to verify that this is the case and, if so, replace `m-g-m-f-s` with `m-m-f-q`.

notmgsk:
closes #172

### DIFF
--- a/src/compilers/approx.lisp
+++ b/src/compilers/approx.lisp
@@ -714,7 +714,7 @@ Additionally, if PREDICATE evaluates to false and *ENABLE-APPROXIMATE-COMPILATIO
 Both CENTER-CIRCUIT and the return value are lists of GATE-APPLICATIONs; A, D, and B are matrices; and Q1, Q0 are qubit indices."
   (multiple-value-bind (ua ub fidelity)
       (match-matrix-to-an-e-basis-diagonalization
-       (make-matrix-from-quil center-circuit :relabeling `((,q1 . 1) (,q0 . 0)))
+       (make-matrix-from-quil center-circuit :relabeling (standard-qubit-relabeler `(,q1 ,q0)))
        a d b)
     
     (multiple-value-bind (b1 b0) (convert-su4-to-su2x2 ub)
--- a/src/compilers/optimal-2q.lisp
+++ b/src/compilers/optimal-2q.lisp
@@ -385,8 +385,10 @@ The optional argument INSTR is used to canonicalize the qubit indices of the ins
              ;; we also want to know a matrix representation of the bare circuit, so
              ;; that we can use MAGICL to adorn it with SU(2) x SU(2) conjugators.
              (bare-matrix
-               (let ((temp (make-gate-matrix-from-gate-string (application-arguments instr)
-                                                              bare-circuit)))
+               (let ((temp (make-matrix-from-quil bare-circuit
+                                                  :relabeling (standard-qubit-relabeler
+                                                               (mapcar #'qubit-index
+                                                                       (application-arguments instr))))))
                  (unless (= 4 (magicl:matrix-rows temp))
                    (setf temp (kron-matrix-up temp 2)))
                  (let ((det-factor (expt (magicl:det temp) -1/4)))
--- a/src/compressor/compressor.lisp
+++ b/src/compressor/compressor.lisp
@@ -401,7 +401,7 @@ other's."
 
 (defun decompile-instructions-in-context (instructions chip-specification context)
   "This routine is called by COMPRESS-INSTRUCTIONS-IN-CONTEXT to make a decision about how to prefer 'linear algebraic compression': the list of INSTRUCTIONS can always be rewritten as its associated action matrix, but under certain conditions (governed by CONTEXT) we can sometimes get away with something less."
-  (let* ((qubits-on-obj (qubits-in-instr-list instructions)))
+  (let ((qubits-on-obj (qubits-in-instr-list instructions)))
     (labels
         (;; produce a sequence of native instructions that have the effect of
          ;; carrying START-WF to FINAL-WF (= INSTRUCTIONS |START-WF>)
--- a/src/compressor/compressor.lisp
+++ b/src/compressor/compressor.lisp
@@ -224,7 +224,7 @@ other's."
                                            (peephole-rewriter-node-prev node))
                                           (peephole-rewriter-node-instr node)
                                           :destructive? nil)))
-       
+
        ;; having selected an appropriate sequence of instructions, actually
        ;; apply the available rewriting rules. if we find one that applies,
        ;; splices the results in to replace the nodes and return the location
@@ -262,7 +262,7 @@ other's."
                        (mapc #'delete-node relevant-nodes-for-inspection)
                        (return-from apply-rules new-node))))
                (compiler-does-not-apply () nil)))))
-       
+
        ;; main loop for the peephole rewriter. for any particular node, it
        ;; assembles a list of instructions which might be subject to rewriting
        ;; rules, then passes those to APPLY-RULES. if APPLY-RULES returns
@@ -272,9 +272,9 @@ other's."
          ;; for each instruction...
          (setf node (peephole-rewriter-node-next node))
          (unless node (return-from outer-instruction-loop))
-         
+
          (update-context node)
-         
+
          ;; build the noncommuting instruction list
          (let* ((nodes-for-inspection (find-noncommuting-instructions node))
                 (qubit-complex nil)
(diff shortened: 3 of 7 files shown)


src/compressor/compressor.lisp
```diff
@@ -404,7 +404,9 @@ other's."
 
 (defun decompile-instructions-in-context (instructions chip-specification context)
   "This routine is called by COMPRESS-INSTRUCTIONS-IN-CONTEXT to make a decision about how to prefer 'linear algebraic compression': the list of INSTRUCTIONS can always be rewritten as its associated action matrix, but under certain conditions (governed by CONTEXT) we can sometimes get away with something less."
-  (let ((qubits-on-obj (qubits-in-instr-list instructions)))
+  (let* ((qubits-on-obj (qubits-in-instr-list instructions))
+         (qubits-indices (mapcar (alexandria:compose #'qubit-index #'qubit)
```
ecpeterson: aren't these functions mutual inverses?

src/compilers/optimal-2q.lisp
```diff
@@ -385,8 +385,9 @@ The optional argument INSTR is used to canonicalize the qubit indices of the ins
              ;; we also want to know a matrix representation of the bare circuit, so
              ;; that we can use MAGICL to adorn it with SU(2) x SU(2) conjugators.
              (bare-matrix
-               (let ((temp (make-gate-matrix-from-gate-string (application-arguments instr)
-                                                              bare-circuit)))
+               (let ((temp (make-matrix-from-quil bare-circuit
+                                                  :relabeling (mapcar #'qubit-index
```
ecpeterson: I think you also mean to use your helper function `qubits-relabeling` here; otherwise, this doesn't seem to be an alist.

src/compressor/compressor.lisp
stylewarning: yes; if the intention here is to copy the list, we should call `copy-list`

notmgsk: yep. thank McCarthy for code review 🙏

src/compilers/optimal-2q.lisp
notmgsk: yep. thank McCarthy for code review 🙏

src/gates.lisp
```diff
@@ -569,8 +569,16 @@ Note that this is a controlled version of a R_z gate multiplied by a phase."
                   m)))
       (magicl:multiply-complex-matrices mat m))))
 
+(defun qubits-relabeling (qubits-list)
```
stylewarning: How about we use functions instead to represent these maps? I suggest rewriting this as something like this:
```
(defun standard-qubit-relabeler (qubits-list)
  "Construct a standard qubit relabeler and its inverse.

A *relabeler* is a function that takes as input a logical qubit index and produces a relabeled (sometimes called *physical*) qubit index.

The *standard labeling* of qubits (in a context where qubits are supplied as arguments to a gate application) is in descending order to 0. As such, a given QUBITS-LIST (a b c ... y z) of length N would have a labeling of:

    a -> N - 1
    b -> N - 2
    c -> N - 3
    ...
    y -> 1
    z -> 0.

Qubits not present in QUBITS-LIST are mapped to themselves."
  (let ((map
          (loop :for l_i :in qubits-list
                :for p_i :from (1- (length qubits-list)) :downto 0
                :collect (cons l_i p_i))))
    (values (lambda (logical-qubit)
              (or (cdr (assoc logical-qubit map))
                  logical-qubit))
            (lambda (physical-qubit)
              (or (car (rassoc physical-qubit map))
                  physical-qubit)))))
```

We could do away with constructing `map` and just iterating, bumping indexes and the like each time.

```
(defun standard-qubit-relabeler* (qubits-list)
  "Construct a standard qubit relabeler and its inverse.

A *relabeler* is a function that takes as input a logical qubit index and produces a relabeled (sometimes called *physical*) qubit index.

The *standard labeling* of qubits (in a context where qubits are supplied as arguments to a gate application) is in descending order to 0. As such, a given QUBITS-LIST (a b c ... y z) of length N would have a labeling of:

    a -> N - 1
    b -> N - 2
    c -> N - 3
    ...
    y -> 1
    z -> 0.

Qubits not present in QUBITS-LIST are mapped to themselves."
  (let ((length (length qubits-list)))
    (values (lambda (logical-qubit)
              (let ((p (position logical-qubit qubits-list)))
                (if p
                    (- length p 1)
                    logical-qubit)))
            (lambda (physical-qubit)
              (if (<= 0 physical-qubit (1- length))
                  (nth physical-qubit qubits-list)
                  physical-qubit)))))
```

stylewarning: If we do this, then the `:relabeling` argument can just be `:relabeler` and by default be `#'identity`.

ecpeterson: i like this advice

src/compilers/optimal-2q.lisp
stylewarning: @notmgsk this line needs changing still

notmgsk:
@ecpeterson @stylewarning do you know why it's not picking up the function `standard-qubit-relabeler`? I moved it into `src/relabeling.lisp`, added that to `cl-quil.asd` and `src/package.lisp` ... but it remains undefined?

notmgsk:
> @ecpeterson @stylewarning do you know why it's not picking up the function `standard-qubit-relabeler`? I moved it into `src/relabeling.lisp`, added that to `cl-quil.asd` and `src/package.lisp` ... but it remains undefined?

Missing a `(in-package :cl-quil)`.

src/compressor/compressor.lisp
```diff
@@ -403,7 +401,7 @@ other's."
 
 (defun decompile-instructions-in-context (instructions chip-specification context)
   "This routine is called by COMPRESS-INSTRUCTIONS-IN-CONTEXT to make a decision about how to prefer 'linear algebraic compression': the list of INSTRUCTIONS can always be rewritten as its associated action matrix, but under certain conditions (governed by CONTEXT) we can sometimes get away with something less."
-  (let ((qubits-on-obj (qubits-in-instr-list instructions)))
+  (let* ((qubits-on-obj (qubits-in-instr-list instructions)))
```
stylewarning:
```suggestion
  (let ((qubits-on-obj (qubits-in-instr-list instructions)))
```

src/compilers/approx.lisp
```diff
@@ -714,7 +714,7 @@ Additionally, if PREDICATE evaluates to false and *ENABLE-APPROXIMATE-COMPILATIO
 Both CENTER-CIRCUIT and the return value are lists of GATE-APPLICATIONs; A, D, and B are matrices; and Q1, Q0 are qubit indices."
   (multiple-value-bind (ua ub fidelity)
       (match-matrix-to-an-e-basis-diagonalization
-       (make-matrix-from-quil center-circuit :relabeling `((,q1 . 1) (,q0 . 0)))
+       (make-matrix-from-quil center-circuit :relabeling (standard-qubit-relabeler `(,q1 ,q0)))
```
stylewarning: wait since it's q1 -> 1, and q0 -> 0, shouldn't it be `(list q0 q1)` ?

stylewarning: also don't use backquote here it's not code

src/relabeling.lisp
```diff
@@ -0,0 +1,26 @@
+;;;; relabeling.lisp
```
stylewarning: idk if a new file was necessary but it's ok

src/gates.lisp
```diff
@@ -569,8 +569,8 @@ Note that this is a controlled version of a R_z gate multiplied by a phase."
                   m)))
       (magicl:multiply-complex-matrices mat m))))
 
-(defun make-matrix-from-quil (instruction-list &key relabeling)
-  "If possible, create a matrix out of the instructions INSTRUCTION-LIST, within the context of the environment ENVIRONS. If one can't be created, then return NIL.
+(defun make-matrix-from-quil (instruction-list &key (relabeling #'identity))
+  "If possible, create a matrix out of the instructions INSTRUCTION-LIST using the optional function RELABELLING that maps an input qubit index to an output qubit index. If one can't be created, then return NIL.
```
ecpeterson: I think "relabelling" and "relabeling" are both acceptable English, but since the key arg is `relabeling`, let's stick to that.

ecpeterson:
Pending the resolution of Robert's comments, this looks OK to me.

src/compilers/approx.lisp
notmgsk: it's reversed in standard-qubit-relabeler

PR #188, opened 2019-04-11
RPCQ changed from de/serializing lists to de/serializing vectors

ecpeterson:
When https://github.com/rigetti/rpcq/pull/45 gets merged, we will need to update some of the unpacking done in the quilc RPCQ server. This PR handles that change.


ecpeterson:
Now that the corresponding RPCQ PR is in, this needs to be merged.

Issue #116, opened 2019-02-09
print-instruction-to-string utility function


ecpeterson:
Repeatedly, as part of `print-instruction` itself and in support of error message printing, we make use of the `nil` stream functionality of `print-instruction` to extract a string from an instruction object. If there were a more direct utility wrapper for this functionality, we could incorporate all these calls (including some `mapcar` calls) into the relevant `format` string by using `~/fn-name/`. Investigate how much this cleans up the code.

appleby:
Can this issue be closed now, or should it morph into an issue for tracking the propagation of `"~/cl-quil::instruction-fmt/"` use throughout the code?

notmgsk:
I'm gonna close this. We can change stuff as it comes up in other PRs.

PR #194, opened 2019-04-12
add INSTRUCTION-FMT for FORMATting instructions

stylewarning:
This also changes some (current 1) call site of PRINT-INSTRUCTION to use the new format directive.

This addresses #116.

### DIFF
--- a/src/compressor/compressor.lisp
+++ b/src/compressor/compressor.lisp
@@ -48,11 +48,10 @@
 (defun format-quil-sequence (s instructions &optional prefix)
   "Nicely prints a sequence of INSTRUCTIONS to a stream S. If PREFIX is present, prepend it to the overall output."
   (when prefix
-    (format s prefix))
+    (format s prefix)
+    (terpri s))
   (dolist (instr instructions)
-    (write-string "    " s)
-    (print-instruction instr s)
-    (terpri s)))
+    (format s "    ~/cl-quil:instruction-fmt/~%" instr)))
 
 (defun qubits-in-instr-list (instructions)
   "Produces a list of all of the (unboxed) qubit indices appearing in INSTRUCTIONS, a list of applications."
--- a/src/package.lisp
+++ b/src/package.lisp
@@ -319,7 +319,8 @@
    #:parsed-program-executable-code     ; READER
 
    #:*print-fractional-radians*         ; PARAMETER
-   #:print-instruction                  ; GENERIC, METHOD
+   #:print-instruction                  ; FUNCTION
+   #:instruction-fmt                    ; FUNCTION (format directive)
    )
 
   ;; gates.lisp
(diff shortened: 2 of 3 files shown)


stylewarning:
Thoughts? Too long a name?

src/package.lisp
```diff
@@ -320,6 +320,7 @@
 
    #:*print-fractional-radians*         ; PARAMETER
    #:print-instruction                  ; GENERIC, METHOD
```
notmgsk: I don't think this is a generic

src/compressor/compressor.lisp
```diff
@@ -48,11 +48,10 @@
...
   "Nicely prints a sequence of INSTRUCTIONS to a stream S. If PREFIX is present, prepend it to the overall output."
   (when prefix
-    (format s prefix))
+    (format s prefix)
+    (terpri s))
   (dolist (instr instructions)
-    (write-string "    " s)
-    (print-instruction instr s)
-    (terpri s)))
+    (format s "    ~/cl-quil:pprint-isn/~%" instr)))
```
notmgsk: is this really the only place where this can be implemented

ecpeterson:
I think the name is fine: the package qualifier is nearly as long as the name.

src/compressor/compressor.lisp
stylewarning: no, it was one place i chose

stylewarning:
Maybe `pprint-` isn't really the right prefix. The `pprint-` functions in the `cl` package refer to interactions with the pretty printer, and this isn't so. Maybe I'll replace it with `fmt-`.

stylewarning:
Instead of bulk-changing code that has no visible effect, I'm going to suggest we do it piecemeal.

stylewarning:
@rigettizach @ecpeterson @notmgsk thoughts on the name? Ideally whatever convention is here should be followed elsewhere.

ecpeterson:
I like `fmt-` just fine.

rigettizach:
@stylewarning I don't like the abbreviation but I don't think the full expansion is a great name either; FORMAT-INSTRUCTION sounds like something you could call with just an instruction, but the format ~// conventions don't fit with that. INSTRUCTION-STRING likewise. 

Since I can't think of anything I like better, the current name is acceptable to me.

stylewarning:
> @stylewarning I don't like the abbreviation but I don't think the full expansion is a great name either; FORMAT-INSTRUCTION sounds like something you could call with just an instruction, but the format ~// conventions don't fit with that. INSTRUCTION-STRING likewise.
> 
> Since I can't think of anything I like better, the current name is acceptable to me.

I in general don't like short names either, but I also don't want 80 character format strings.

What about flipping the order @rigettizach ? `{isn, instruction}-fmt`?

rigettizach:
@stylewarning I think `instruction-fmt` would be better than `fmt-isn`.

PR #197, opened 2019-04-15
Extend format-complex to optionally print in polar form

### LINKED ISSUE #171: Add option to print complex numbers in polar form
I spend a lot of time staring at `print`-outs of trays of complex numbers, trying to decide if some collection of them is the same as some other collection up to phase.

We already have a complex number pretty-printer and it already has a global option to enable pretty-printing of certain fractions of pi. I'd like an option to extend it to have a global option to print polar forms of complex numbers instead of Cartesian forms.

### LINKED ISSUE #198: FORMAT-COMPLEX should be compatible with ~/.../ format strings
`~/cl-quil:format-complex/` would be very nice in `format`. If we did this, then we could use `@` or `:`-modifiers to make it print as cartesian or polar, maybe. Make it so!

karalekas:
Fixes #171 
and now also
Fixes #198

### DIFF
--- a/src/ast.lisp
+++ b/src/ast.lisp
@@ -950,6 +950,9 @@ If this slot is not supplied, then the gate is considered *anonymous*. If this i
 
 N.B., The fractions of pi will be printed up to a certain precision!")
 
+(defvar *print-polar-form* nil
+  "When true, FORMAT-COMPLEX prints our complex number in polar form with syntax r∠theta.")
+
 (defun format-complex (z stream)
   "Print the real or complex number Z nicely to the stream STREAM."
   (check-type z number)
@@ -986,6 +989,8 @@ N.B., The fractions of pi will be printed up to a certain precision!")
      (cond
        ((zerop (imagpart z))
         (format-complex (realpart z) stream))
+       (*print-polar-form*
+        (format stream "~A∠~A" (format-complex (abs z) nil) (format-complex (phase z) nil)))
        (t
         (format stream "~a~a~a"
                 (if (zerop (realpart z))
--- a/src/ast.lisp
+++ b/src/ast.lisp
@@ -953,6 +953,34 @@ N.B., The fractions of pi will be printed up to a certain precision!")
 (defvar *print-polar-form* nil
   "When true, FORMAT-COMPLEX prints out complex numbers in polar form with syntax AMPLITUDE∠PHASE.")
 
+(defun format-real (r stream)
+  "Print the real number R nicely to the stream STREAM."
+  (when *print-fractional-radians*
+    (when (double~ r pi)
+      (return-from format-real (format stream "pi")))
+    (when (double~ r (- pi))
+      (return-from format-real (format stream "-pi")))
+    (dolist (denom '(2 3 4 6 8 16))
+      (dotimes (numer (* 2 denom))
+        (when (/= numer denom)
+          (when (double~ r (/ (* pi numer) denom))
+            (return-from format-real
+              (cond
+                ((= numer 1)
+                 (format stream "pi/~d" denom))
+                ((zerop numer)
+                 (format stream "~F" r))
+                (t (format stream "~d*pi/~d" numer denom)))))
+          (when (double~ r (/ (* -1 pi numer) denom))
+            (return-from format-real
+              (cond
+                ((= numer 1)
+                 (format stream "-pi/~d" denom))
+                ((zerop numer)
+                 (format stream "~F" r))
+                (t (format stream "-~d*pi/~d" numer denom)))))))))
+  (format stream "~F" r))
+
 (defun format-complex (z stream)
   "Print the real or complex number Z nicely to the stream STREAM."
   (check-type z number)
@@ -960,37 +988,13 @@ N.B., The fractions of pi will be printed up to a certain precision!")
     ((zerop z)
      (format stream "0.0"))
     ((realp z)
-     (when *print-fractional-radians*
-       (when (double~ z pi)
-         (return-from format-complex (format stream "pi")))
-       (when (double~ z (- pi))
-         (return-from format-complex (format stream "-pi")))
-       (dolist (denom '(2 3 4 6 8 16))
-         (dotimes (numer (* 2 denom))
-           (when (/= numer denom)
-             (when (double~ z (/ (* pi numer) denom))
-               (return-from format-complex
-                 (cond
-                   ((= numer 1)
-                    (format stream "pi/~d" denom))
-                   ((zerop numer)
-                    (format stream "~F" z))
-                   (t (format stream "~d*pi/~d" numer denom)))))
-             (when (double~ z (/ (* -1 pi numer) denom))
-               (return-from format-complex
-                 (cond
-                   ((= numer 1)
-                    (format stream "-pi/~d" denom))
-                   ((zerop numer)
-                    (format stream "~F" z))
-                   (t (format stream "-~d*pi/~d" numer denom)))))))))
-     (format stream "~F" z))
+     (format-real z stream))
     ((complexp z)
      (cond
        ((zerop (imagpart z))
         (format-complex (realpart z) stream))
        (*print-polar-form*
-        (format stream "~A∠~A" (format-complex (abs z) nil) (format-complex (phase z) nil)))
+        (format stream "~F∠~A" (abs z) (format-real (phase z) nil)))
        (t
         (format stream "~a~a~a"
                 (if (zerop (realpart z))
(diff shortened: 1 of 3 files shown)


karalekas:
There's an issue with the `test-quilc` portion of the build, and I think it is unrelated to my changes. The error message in GitLab isn't super helpful, but running `(asdf:test-system :quilc)` gives the following:
```
(SB-KERNEL:ASSERT-ERROR
   (CL-QUIL::DOUBLE~ (ABS (MAGICL:REF CL-QUIL::MAT CL-QUIL::J 0))
   (ABS (MAGICL:REF CL-QUIL::REF-MAT CL-QUIL::J 0)))
   (((ABS (MAGICL:REF CL-QUIL::MAT CL-QUIL::J 0)) 4.927465022116596d-16)
   ((ABS (MAGICL:REF CL-QUIL::REF-MAT CL-QUIL::J 0)) 0.7071067811865475d0))
   NIL "Matrices do not lie in the same projective class.")
```

stylewarning:
@karalekas Re-run the tests. There are some randomized tests that fail sometimes due to what I believe is numerical stability.

src/ast.lisp
```diff
@@ -950,6 +950,9 @@ If this slot is not supplied, then the gate is considered *anonymous*. If this i
 
 N.B., The fractions of pi will be printed up to a certain precision!")
 
+(defvar *print-polar-form* nil
+  "When true, FORMAT-COMPLEX prints our complex number in polar form with syntax r∠theta.")
```
stylewarning:
```suggestion
  "When true, FORMAT-COMPLEX prints out complex numbers in polar form with syntax AMPLITUDE∠PHASE. The complex number 0 is normalized as 0∠0.")
```

src/ast.lisp
```diff
@@ -986,6 +989,8 @@ N.B., The fractions of pi will be printed up to a certain precision!")
      (cond
        ((zerop (imagpart z))
         (format-complex (realpart z) stream))
+       (*print-polar-form*
```
stylewarning: I suggest that polar form is _always_ printed when requested, even for real numbers. As such, I'd move this up one.

src/ast.lisp
```diff
@@ -986,6 +989,8 @@ N.B., The fractions of pi will be printed up to a certain precision!")
...
+        (format stream "~A∠~A" (format-complex (abs z) nil) (format-complex (phase z) nil)))
```
stylewarning: I suggest just doing:
```
(format stream "~F∠~F" (abs z) (phase z))
```

Do you see any downsides to this?

appleby:
> @karalekas Re-run the tests. There are some randomized tests that fail sometimes due to what I believe is numerical stability.

@stylewarning @karalekas 

I'm seeing the same assertion failure consistently in the faithfulness tests I recently enabled. Sorry about that! I've open #199 to track it.

src/ast.lisp
ecpeterson: FWIW, the current behavior is to print 0 as 0.

src/ast.lisp
ecpeterson: I'm mildly against this: if I were to turn this option on during a quilc run, I'd prefer than `RZ(0.2349808) 0` still get printed as `RZ(0.2349808) 0` rather than `RZ(0.2349808∠0) 0`.

src/ast.lisp
ecpeterson: Calling `format-complex` again gives it the opportunity to pretty-ify pi fractions, which sounds like a nice feature to me.

src/ast.lisp
stylewarning: @ecpeterson I'm OK with that in rectangular form, just because `+` is `+`. But it seems more odd to elide it with this angle ∠ notation. It could be justified by saying `∠x` == `e^(ix)` and it can be interpreted as multiplication.

If we allow ourselves to print `0`, then I would think we should allow ourselves to print `∠x` when the amplitude is `1` in the same way we allow `a+bi` to be printed as `bi` when `a==0`. But I don't think that's a good idea.

src/ast.lisp
stylewarning: I'll defer to you, since you read this more often than me.

src/ast.lisp
stylewarning: I would prefer to special case the angular part, and factor out the pi printing.

ecpeterson: `RZ(1.5707963267948966) 0` is then another motivating case. If you special case the pi printing on just the angular part, that will still print as `RZ(1.5707963267948966) 0` (or, if you prefer, `RZ(1.5707963267948966∠) 0`) when both flags are turned on, and not the expected `RZ(pi/2) 0`. If we're going to special-case the angular part alone, then I want one of two things to happen:

1. (Write and) call `format-real` instead of `format-complex` when you mean to be printing a real number. Use the pi pretty printer there.
2. Inside of the purely real branch of the complex pretty printer, still allow for pi fraction special casing and don't print the `∠0`.

karalekas:
Ok there has been some debate but I think my updates reflect the consensus? Let me know what you think

src/ast.lisp
stylewarning: That sounds good @ecpeterson

stylewarning:
> Ok there has been some debate but I think my updates reflect the consensus? Let me know what you think

I concede to @ecpeterson's motivations & reasoning.

src/ast.lisp
```diff
@@ -950,42 +950,51 @@ If this slot is not supplied, then the gate is considered *anonymous*. If this i
 
 N.B., The fractions of pi will be printed up to a certain precision!")
 
+(defvar *print-polar-form* nil
+  "When true, FORMAT-COMPLEX prints out complex numbers in polar form with syntax AMPLITUDE∠PHASE.")
+
+(defun format-real (r stream)
+  "Print the real number R nicely to the stream STREAM."
+  (when *print-fractional-radians*
```
stylewarning: Can we clean up the control flow here? Maybe start with a double-`cond`?

```
(cond
  (*print-fractional-radians*
   (cond
    ((double~ r pi) (format ...))
    ((double~ r (- pi)) (format ...))
    (t
      (dolist ...
```

I'm also not very keen on `return-from` on the `format` calls. The above suggested refactor removes those, but in the loop, I'd rather break out and just do `(return-from format-real nil)`.

The argument to `return` the `format` calls is to allow for `format nil`, but that's an anti-pattern when the calls are embedded like so.

src/ast.lisp
```diff
@@ -950,42 +950,51 @@ If this slot is not supplied, then the gate is considered *anonymous*. If this i
...
-                    (format stream "~F" z))
-                   (t (format stream "-~d*pi/~d" numer denom)))))))))
-     (format stream "~F" z))
+     (format-real z stream))
     ((complexp z)
      (cond
        ((zerop (imagpart z))
         (format-complex (realpart z) stream))
+       (*print-polar-form*
+        (format stream "~F∠~A" (abs z) (format-real (phase z) nil)))
```
stylewarning: If the above advice is followed, you'd write:
```
(format stream "~F∠" (abs z))
(format-real (phase z) stream)
```

We don't want to cons in the complex printer.

src/ast.lisp
```diff
@@ -950,42 +950,51 @@ If this slot is not supplied, then the gate is considered *anonymous*. If this i
 
 N.B., The fractions of pi will be printed up to a certain precision!")
 
+(defvar *print-polar-form* nil
+  "When true, FORMAT-COMPLEX prints out complex numbers in polar form with syntax AMPLITUDE∠PHASE.")
+
+(defun format-real (r stream)
+  "Print the real number R nicely to the stream STREAM."
```
stylewarning: We should
```
(check-type r real)
(check-type stream stream)
```

ecpeterson:
would you please rebase this off of master to make me feel better about the build error

karalekas:
Now also fixes #198

karalekas:
rebased

stylewarning:
There are still places to remove/fix `format-complex z nil`

src/ast.lisp
```diff
@@ -950,42 +950,61 @@ If this slot is not supplied, then the gate is considered *anonymous*. If this i
...
+  "Like the function format-real, but is compatible with format strings using the ~/.../ directive.
+For example,
+    (format t \"the number was ~/cl-quil:real-fmt/\" r)"
+  (declare (ignore colon-modifier at-modifier))
+  (format-real r stream))
+
 (defun format-complex (z stream)
   "Print the real or complex number Z nicely to the stream STREAM."
   (check-type z number)
+  (check-type stream stream)
```
ecpeterson: I'm surprised that this passes. If you'd asked me a week ago, I'd have confidently told you that we definitely call `format-complex` with `stream` set to `NIL` somewhere in the code base. Guess not!

stylewarning: There _were_ places in the codebase that used this. They're fixed.

karalekas:
🎉

PR #190, opened 2019-04-11
Rename parse-quil and parse-quil-string

notmgsk:
parse-quil -> parse-quil-into-raw-program
parse-quil-string -> parse-quil

Closes #146

Issue #69, opened 2019-02-08
Eliminate program-initial SWAPs


ecp-rigetti:
At the top of a ProtoQuil program, the compiler will automatically eliminate program-initial SWAPs that it inserts, choosing instead to modify the initial rewiring. We should also do this for user-supplied program-initial SWAPs.

PR #200, opened 2019-04-16
preserve the 'Exiting rewiring' comment when deleting the HALT instruction on a protoquil program

### LINKED ISSUE #199: quilc test-faithfulness fails assertion in cl-quil::scale-out-matrix-phases: Matrices do not lie in the same projective class
The faithfulness tests that were recently enabled in #189 are failing an assertion in `cl-quil::scale-out-matrix-phases`. The test failure appears to be a result of changes made in #159. At least, if I rebase and drop that commit, the test passes again. Note that #159 was merged just before #189, so @ecpeterson wouldn't have known.

Probably related to @karalekas comment here: https://github.com/rigetti/quilc/pull/197#issuecomment-483434459

Here is the backtrace:

```
Backtrace:
  0: (SB-KERNEL:ASSERT-ERROR (CL-QUIL::DOUBLE~ (ABS (MAGICL:REF CL-QUIL::MAT CL-QUIL::J 0)) (ABS (MAGICL:REF CL-QUIL::REF-MAT CL-QUIL::J 0))) (((ABS #) 2.1628161436138797d-16) ((ABS #) 0.7071067811865475d0..
      Locals:
        ARGS-AND-VALUES = (((ABS (MAGICL:REF CL-QUIL::MAT CL-QUIL::J 0)) 2.1628161436138797d-16) ((ABS (MAGICL:REF CL-QUIL::REF-MAT CL-QUIL::J 0)) 0.7071067811865475d0))
        ASSERTION = (CL-QUIL::DOUBLE~ (ABS (MAGICL:REF CL-QUIL::MAT CL-QUIL::J 0)) (ABS (MAGICL:REF CL-QUIL::REF-MAT CL-QUIL::J 0)))
        DATUM = "Matrices do not lie in the same projective class."
        PLACES = NIL
        SB-DEBUG::MORE = NIL
  1: (CL-QUIL::SCALE-OUT-MATRIX-PHASES #S(MAGICL:MATRIX :ROWS 64 :COLS 64 :DATA-TYPE MAGICL:Z :STORAGE-TYPE MAGICL::GE :DATA #(#C(0.1186426678685776d0 -0.6970824322567777d0) #C(1.0302206828327032d-16 -1.90..
      Locals:
        GOODNESS-VALUE = 0.7071067811865466d0
        J = 1
        MAT = #S(MAGICL:MATRIX ..)
        REF-MAT = #S(MAGICL:MATRIX ..)
        RESCALE-VALUE = #C(0.7071067811865475d0 0.7071067811865491d0)
  2: (QUILC::PROCESS-PROGRAM #<CL-QUIL:PARSED-PROGRAM {10042C2D33}> #<CL-QUIL::CHIP-SPECIFICATION of 8:8 objects>)
      Locals:
        CHIP-SPECIFICATION = #<CL-QUIL::CHIP-SPECIFICATION of 8:8 objects>
        ORIGINAL-MATRIX = #S(MAGICL:MATRIX ..)
        PROGRAM = #<CL-QUIL:PARSED-PROGRAM {10042C2D33}>
  3: (QUILC::PROCESS-OPTIONS :PREFER-GATE-LADDERS NIL :COMPUTE-GATE-DEPTH NIL :COMPUTE-GATE-VOLUME NIL :COMPUTE-RUNTIME NIL :COMPUTE-FIDELITY NIL :COMPUTE-MATRIX-REPS T :COMPUTE-2Q-GATE-DEPTH NIL :COMPUTE-..
  4: ((LABELS QUILC-TESTS::TEST-FAITHFULNESS :IN QUILC-TESTS::TEST-FAITHFULNESS))
...
```

ecpeterson:
This PR fixes a bad interaction between #159 and #189.

Closes #199, fixes the build failure in #197 (and probably also #190). H/t to @appleby for both spotting the error and correctly guessing the solution.

### DIFF
--- a/app/src/entry-point.lisp
+++ b/app/src/entry-point.lisp
@@ -420,6 +420,7 @@ HTTP server for good.
         (setf (quil::parsed-program-executable-code processed-program)
               (coerce
                (loop :for instr :across (quil::parsed-program-executable-code processed-program)
+                     :for j :from 0
                      :when (quil::comment instr)
                        :do (cond
                              ((uiop:string-prefix-p "Entering rewiring: " (quil::comment instr))
@@ -433,6 +434,10 @@ HTTP server for good.
                                                                               (/ (- (length comment) length) 2)
                                                                               2)))))
                              (t nil))
+                     :when (and (typep instr 'quil::halt) (quil::comment instr))
+                       :do (setf (quil::comment (aref (quil::parsed-program-executable-code processed-program)
+                                                      (1- j)))
+                                 (quil::comment instr))
                      :unless (typep instr 'quil::halt)
                        :collect instr)
                'vector)))
--- a/app/src/entry-point.lisp
+++ b/app/src/entry-point.lisp
@@ -421,11 +421,13 @@ HTTP server for good.
               (coerce
                (loop :for instr :across (quil::parsed-program-executable-code processed-program)
                      :for j :from 0
+                     ;; if there's a comment, consider extracting from it the final_rewiring
+                     ;; entry for use in the statistics dictionary.
                      :when (quil::comment instr)
                        :do (cond
-                             ((uiop:string-prefix-p "Entering rewiring: " (quil::comment instr))
+                             ((uiop:string-prefix-p "Exiting rewiring: " (quil::comment instr))
                               (setf (gethash "final_rewiring" *statistics-dictionary*)
-                                    (subseq (quil::comment instr) (length "Entering rewiring: "))))
+                                    (subseq (quil::comment instr) (length "Exiting rewiring: "))))
                              ((uiop:string-prefix-p "Entering/exiting rewiring: " (quil::comment instr))
                               (let ((comment (quil::comment instr))
                                     (length (length "Entering/exiting rewiring: (")))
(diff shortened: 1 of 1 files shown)


notmgsk:
What happened to the CI

ecpeterson:
Maybe it got sick of the build failures and shut itself off.

appleby:
Is there any situation in which the instruction preceding the HALT might have a comment of it's own? If so, would clobbering it matter? Is a pathological program containing only a HALT instruction supported?

ecpeterson:
All wise questions.

appleby:
I'm all questions and no answers, unfortunately. Is replacing the HALT with a NOP and attaching the comment there viable? Do these rewiring comments at the end of the program have any purpose other than to make `*compute-matrix-reps*` happy? If not, would it make sense to alter the processing of `original-matrix` to strip them off, rather than adding them back to `processed-program`?

stylewarning:
Replacing `HALT` with `NOP` sounds like a not to bad idea, if the rest of the stack tolerates it.

ecpeterson:
I don't think the rest of the QPU stack tolerates it, unfortunately.

ecpeterson:
Here's some more reasoned thought about this:

* This HALT removal only happens if the `*protoquil*` flag is on.
* If the `*protoquil*` flag is on, any input involving control flow, other HALTs, NOPs, ... gets rejected.
* Entering / exiting rewiring comments are only attached to the first / last instructions in a node in the control flow graph. If the input program is protoquil, then the original CFG has to be linear (with nodes delineated by PRESERVE_BLOCK boundaries), and the CFG after rewriting will have just one block. So, there's only one entering comment and one exiting comment to worry about.
* If the protoquil program is empty (except for the HALT instruction), then it doesn't matter what sort of rewiring you did: the unwired version of a rewired program is also empty. So, you can just discard the comments.
* If the protoquil program has just one non-HALT instruction, you'll have to deal with a collision of comments. (As an aside, in this case currently these comments will always be the same, but if #69 gets worked on I think we lose this guarantee.) Luckily, since there's only one Entering and only one Exiting comment, there's only one possible collision that can happen.

ecpeterson:
Here's a fix for this bug that doesn't itself introduce bugs. The code quality isn't great, but this area of quilc is already pretty gross, so 🤷‍♀️

PR #201, opened 2019-04-17
Configure host with --host

notmgsk:
e.g. using --host 127.0.0.1 restricts incoming connections to those
from localhost

### DIFF
--- a/app/src/entry-point.lisp
+++ b/app/src/entry-point.lisp
@@ -65,8 +65,8 @@
     (("help" #\h) :type boolean :optional t :documentation "print this help information and exit")
     (("server-mode-http" #\S) :type boolean :optional t :documentation "run as a web server *and* an RPCQ server. ignores --port and uses 6000 and 5555 respectively")
     (("server-mode-rpc" #\R) :type boolean :optional t :documentation "run as an RPCQ server")
-    (("host") :type string :optional t :documentation "host on which to run the RPCQ server")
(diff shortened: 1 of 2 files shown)


notmgsk:
@stylewarning those changes are less nice here, since there is some confusing between defaults for HTTP vs RPCQ.

When we get rid of HTTP, we can make those changes. Speaking of which, can we get rid of HTTP yet? @ecpeterson

ecpeterson:
It's been more than a month: we can get rid of quilc's HTTP.

stylewarning:
FWIW, someone on Twitter said they want the HTTP API. hth

ecpeterson:
What I'm hearing is a request for `rpcq.js`.

notmgsk:
@stevenheidel ^

stylewarning:
@notmgsk You haven't convinced me that you can't use `:initial-value` despite the existence of the HTTP server. How is doing
```
:initial-value "*"
```
different from no `:initial-value` and having 
```
(unless host
  (setf host "*"))
```
? Same for port, etc.

notmgsk:
@stylewarning both of those things are different in HTTP vs RPCQ, so how do you have a default?

stylewarning:
The docs say:
```
    (("host") :type string :optional t :documentation "host on which to run the RPCQ server")
    (("port" #\p) :type integer :optional t :documentation "port to run the RPCQ server on")
```

I thought it's RPCQ only.

notmgsk:
when you're right you're right

PR #204, opened 2019-04-20
Miscellaneous trivial fixes

appleby:
This PR contains miscellaneous fixes for typos, stale comments, minor refactorings and the like that caught my eye while reading over the code. The commits are independent and unrelated, but it seemed liked overkill to create separate pull requests for each, so I rolled them all up into one. If you prefer separate PRs, just say the word.

Individual commit messages copied/pasted here for easier perusal.

- Call `quilc::%entry-point` from `quil::entry-point`

  This removes a tiny bit of code duplication and makes it clear that `entry-point` and `%entry-point` are equivalent, modulo debugger setup and error reporting.

- Remove duplicate code from tests/utilities.lisp

  Copies of `matrix-equals-dwim`, `operator=`, and `first-column-operator=` appear in `cl-quil-tests` and `cl-quil`. Remove the versions in `cl-quil-tests` and add a `quil::` package prefix to the last remaining references to the versions in `cl-quil-tests`.

- Add /coverage-report/html/ to .gitignore
- Remove unnecessary double-colon for exported magicl symbols
- Make use of `quil` variable in `test-quil-roundtrip`
- Fix minor README typo
- Fix comment typo in compressor-configuration.lisp
- Remove stale comment in compiler-hook-tests

  The second qubit compression of the compiled program in `quil::parsed-program-to-logical-matrix` has been removed.

- Use consistent lower case spelling for `*isa-descriptor*`

  The variable is defined as `*ISA-descriptor*`, but referenced everywhere as `*isa-descriptor*`. If some lawless maverick were to set their readtable-case to `:preserve` or `:invert`, they might be momentarily confused.

- Loop over matrix rows not cols in `matrix-first-column-equality`

  Fixes a hypothetical bug that relied on the fact that `matrix-rows` == `matrix-cols` for square matrices.

- Do test output on `*standard-output*` not `*debug-io*`

  This change was requested in code review comments in #189, but that pull request was merged before I could make the change.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+/coverage-report/html/
 quilc
 .idea/
 system-index.txt
--- a/README.md
+++ b/README.md
@@ -229,7 +229,7 @@ We welcome and encourage community contributions! Peruse our
 [guidelines for contributing](CONTRIBUTING.md) to get you up to speed on
 expectations. Once that's clear, a good place to start is the
 [good first issue](https://github.com/rigetti/quilc/issues?q=is%3Aissue+is%3Aopen+label%3A%22good+first+issue%22)
-section. If you have any find bugs, please create an
+section. If you find any bugs, please create an
 [issue](https://github.com/rigetti/quilc/issues?q=is%3Aissue+is%3Aopen+label%3A%22good+first+issue%22).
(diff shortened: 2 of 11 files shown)


stylewarning:
@appleby Thanks! This is awesome.

PR #208, opened 2019-04-26
Fix the check for an available update

notmgsk:


app/src/entry-point.lisp
```diff
@@ -268,11 +268,12 @@
     (check-libraries))
   #+forest-sdk
   (unless skip-version-check
+    (format t "hi")
```
ecpeterson: get rid of this please

notmgsk: 😨

stylewarning:
can we be robust to not having a network connection?

app/src/versions.lisp
```diff
@@ -38,13 +38,17 @@
...
 (defun latest-sdk-version ()
   "Get the latest SDK quilc version, or NIL if unavailable."
-  (let* ((s (drakma:http-request "http://downloads.rigetti.com/qcs-sdk/version"
-                                 :want-stream t))
-         (p (yason:parse s)))
-    (multiple-value-bind (version success)
-        (gethash "quilc" p)
-      (when success
-        version))))
+  (handler-case
```
notmgsk: @stylewarning like this?

PR #211, opened 2019-04-29
Simplify gate parameter arithmetic expressions

### LINKED ISSUE #79: Simplify arithmetic expressions
The current output of parametric expressions can only such ugly expressions as `theta[0]*2/2+pi/2-pi/2`, which is unnecessarily ugly. Implement a rewriting ruleset, probably as a post-linear-algebraic program transformation, which reduces expressions to normal form.

karalekas:
Fixes #79.


stylewarning:
Heyo, this is an interesting and practical approach to a majority of the expressions that we encounter. I know this is a draft, but I recommend ensuring that the presence of `SIN` etc doesn't clobber your operation. In the case that there is a `SIN` or so, I would just return it unsimplified.

ecpeterson:
We're getting there: `simplify-arithmetic` will eventually skip parameters that throw an error.

src/analysis/simplify-arithmetic.lisp
```diff
@@ -0,0 +1,118 @@
...
+                       (make-delayed-expression (delayed-expression-params param)
+                                                (delayed-expression-lambda-params param)
+                                                (affine-representation->expression
+                                                 (expression->affine-representation
+                                                  (delayed-expression-expression param))))))
+                  ;; TODO: use custom error class
+                  (error () param)))
+              (application-parameters thing))))
+
+(defun simplify-arithmetics (parsed-prog)
```
notmgsk: Couldn't you fold this into `simplify-arithmetic`, dispatching on `parsed-program`?

karalekas: From my understanding, there is a collection of files in `src/analysis` that define transforms, and so the structure of this file is meant to mimic that of the others, like `resolve-applications.lisp`. I'm not sure if there are requirements on passing an ordinary function to `define-transform`, and adding this code as a method for `simplify-arithmetic` would then make that method call `simplify-arithmetic` recursively, which maybe reads a bit less nicely. But I'm not particularly opinionated about it at this point in my Lisp learnings.

ecpeterson: Dispatching on `parsed-program` is a decent idea.

src/analysis/simplify-arithmetic.lisp
```diff
@@ -0,0 +1,133 @@
...
+
+(define-condition expression-not-simplifiable (serious-condition)
+  ()
+  (:documentation "A condition that is signalled any time an expression cannot be simplified. In general, a sub-condition should be preferred over signalling this one."))
+
+(define-condition expression-not-linear (expression-not-simplifiable)
+  ()
+  (:documentation "A condition that is signalled any time an expression cannot be simplified due to being non-linear."))
+
+(defun give-up-simplification (&key (because ':unknown))
```
stylewarning: I'd prefer to just eliminate this function.

src/analysis/simplify-arithmetic.lisp
```diff
@@ -0,0 +1,133 @@
...
+  (ecase because
+    (:not-linear (error 'expression-not-linear))
+    (:unknown    (error 'expression-not-simplifiable))))
+
+(define-transform simplify-arithmetic (simplify-arithmetics)
+  "A transform which converts a parsed program with potentially complicated arithmetic to one that has simplified arithmetic expressions")
+
+(defstruct affine-representation
+  "Data structure that represents a linear expression."
```
stylewarning: `"A representation of a linear expression."`

But is there an even better way to describe this? Can you spell out what CONSTANT and COEFFICIENTS are as a mathematical expression?

src/analysis/simplify-arithmetic.lisp
```diff
@@ -0,0 +1,133 @@
...
+  (constant 0d0 :type double-float)
```
stylewarning: why `double-float`? Any reason not to have `number` or `real`?

src/analysis/simplify-arithmetic.lisp
```diff
@@ -0,0 +1,133 @@
...
+  (coefficients (make-hash-table :test #'equalp) :type hash-table))
```
stylewarning: Add a comment stating what the keys and values are, if not in the docstring above already.

src/analysis/simplify-arithmetic.lisp
```diff
@@ -0,0 +1,133 @@
...
+
+(defun expression->affine-representation (de)
+  "Recursively construct an AFFINE-REPRESENTATION from an EXPRESSION (of type CONS)."
```
stylewarning: Be more specific. Also `EXPRESSION` isn't a variable, it's `DE`. Also I'd refer to it as a tree, not of type `CONS`.

src/analysis/simplify-arithmetic.lisp
```diff
@@ -0,0 +1,133 @@
...
+  (typecase de
+    (double-float
+     (make-affine-representation :constant de))
+    (memory-ref
+     (let ((rep (make-affine-representation)))
```
stylewarning: It sounds better to have a better constructor for the affine reps. Something like

```
(make-affine-representation c a x b y c z ...)
```
or so, for `a`, `b`, `c` numbers and `x`, `y`, `z` mrefs. This would make the code a lot nicer, as opposed to hash table poking everywhere.

src/analysis/simplify-arithmetic.lisp
```diff
@@ -0,0 +1,133 @@
...
+       (setf (gethash de (affine-representation-coefficients rep)) 1.0)
+       rep))
+    (cons
+     (let ((left (expression->affine-representation (second de)))
+           (right (expression->affine-representation (third de))))
+       (ecase (car de)
+         (+
```
stylewarning: write a function
```
(combine-affine-representations operator left right)
```
and simplify this code along with its duplication.

src/analysis/simplify-arithmetic.lisp
```diff
@@ -0,0 +1,133 @@
...
+                    (/ coefficient (affine-representation-constant right))))
+            rep)))))
+    (otherwise (give-up-simplification))))
+
+(defun affine-representation->expression (rep)
+  "Build an EXPRESSION (of type CONS) from an AFFINE-REPRESENTATION, by first iterating through the memory references and their coefficients, and then adding the at the end."
+  (let ((expr nil))
+    (dohash ((ref coefficient) (affine-representation-coefficients rep))
+      (unless (double= 0 coefficient)
+        (setf expr (if expr
```
stylewarning: use quasiquoting to build up "template" expressions. e.g.:
```
`(+ (* ,coefficient ,ref) ,expr)
`(* ,coefficient ,ref)
```

src/analysis/simplify-arithmetic.lisp
```diff
@@ -0,0 +1,133 @@
...
+                       (list '+ (list '* coefficient ref) expr)
+                       (list '* coefficient ref)))))
+    (cond
+      ((null expr)
+       (affine-representation-constant rep))
+      ((double= 0 (affine-representation-constant rep))
+       expr)
+      (t
+       (list '+ (affine-representation-constant rep) expr)))))
```
stylewarning: same here

src/analysis/simplify-arithmetic.lisp
```diff
@@ -0,0 +1,133 @@
...
+    (map-into (application-parameters thing)
+              (lambda (param)
+                (handler-case
+                    (typecase param
+                      (constant
+                       param)
+                      (delayed-expression
+                       (make-delayed-expression (delayed-expression-params param)
+                                                (delayed-expression-lambda-params param)
+                                                (affine-representation->expression
```
stylewarning: write a function above which is something like `canonicalize-expression` which is just the composition of these two functions

src/analysis/simplify-arithmetic.lisp
```diff
@@ -0,0 +1,133 @@
...
+                                                 (expression->affine-representation
+                                                  (delayed-expression-expression param)))))
+                      (otherwise (give-up-simplification)))
```
stylewarning: `(otherwise param)`, no need to hand off to the `handler-case` god above

src/analysis/simplify-arithmetic.lisp
```diff
@@ -0,0 +1,133 @@
...
+                  (expression-not-linear () param)
+                  (expression-not-simplifiable () param)))
+              (application-parameters thing))))
+
+(defun simplify-arithmetics (parsed-prog)
+  "Simplify the arithmetic in all of the EXPRESSIONs in the APPLICATION-PARAMETERS of the GATE-APPLICATIONs of the EXECUTABLE-CODE of a PARSED-PROGRAM."
+  (map nil #'simplify-arithmetic (parsed-program-executable-code parsed-prog))
```
stylewarning: I would
```
(map-into (parsed-prog-exec-code pp) #'simplify-arithmetic (parsed-prog-exec-code pp))
```
to allow for this transform to be pure

stylewarning:
Also please add a lengthy descriptive header to the top of the file stating what it does, how it works, what its limitations are, and what possible alternatives might be

src/analysis/simplify-arithmetic.lisp
notmgsk: Isn't that the opposite of pure? It's mutating the input

src/analysis/simplify-arithmetic.lisp
notmgsk: @stylewarning what do you think about the above

src/analysis/simplify-arithmetic.lisp
stylewarning: it is, but it allows `simplify-arithmetic` to be pure, which i wasn't clear about

src/analysis/simplify-arithmetic.lisp
karalekas: I was following the convention at the top of `compilation-methods.lisp` to define a similar construct for arithmetic simplification. If I understand correctly, you'd like me to replace its uses with either `(error 'expression-not-linear)` or `(error 'expression-not-simplifiable)`?

src/analysis/simplify-arithmetic.lisp
karalekas: Took a stab at improving this, lmk what you think!

src/analysis/simplify-arithmetic.lisp
karalekas: I don't think we want to choose `number` as this constant should not be `complex`, but `real` seems reasonable to me assuming ratio types will be printed and interpreted correctly further down the stack. Happy to change to `real` if you agree.

src/analysis/simplify-arithmetic.lisp
stylewarning: the transform business just takes anything `function`-like

src/analysis/simplify-arithmetic.lisp
stylewarning: correct; we are trying to eliminate that weird `give-up` from the other file. Just signalling directly is preferable.

src/analysis/simplify-arithmetic.lisp
stylewarning: This is a little dangerous since if a memory-ref changes to (say) a class, this code silently breaks. Consider making a hash table that uses `memory-ref=` and `memory-ref-hash` as the equality and test functions. There might even be a function somewhere else in the code like `make-mref-hash-table`. Search around for how `memory-ref-hash` is used. (SLIME `C-c-w-c`)

src/analysis/simplify-arithmetic.lisp
stylewarning: I'm OK with double-float since that's used everywhere.

src/analysis/simplify-arithmetic.lisp
stylewarning: either/or is fine

src/analysis/simplify-arithmetic.lisp
stylewarning: @karalekas I changed my mind, use `number` to match whatever `constant` allows for from ast.lisp (currently that allows for `number` but also requires the type is tagged. If you insist these should not be complex, use `real`.

tests/analysis-tests.lisp
```diff
@@ -359,3 +359,33 @@ B(b) 0 1
              '(e 0 1 2)
              '(f 0 1)
              '(g 0))))
+
+;;; Simplify arithmetic tests
+
+(deftest test-simplify-arithmetic-linear ()
+  "Test that a linear expression can simplified"
+  (let ((in-p (quil::parse-quil "
```
notmgsk: `parse-quil` is exported from `cl-quil` now, so you only need a single colon.

src/analysis/simplify-arithmetic.lisp
```diff
@@ -0,0 +1,219 @@
...
+;;;;
+;;;; This implementation is certainly not without its limitations. For example, we can only simplify
+;;;; a particular subset of arithmetic expressions (ones that are linear). However, for the majority
+;;;; of our use cases at this time, this subset is sufficient.
+
+(in-package #:cl-quil)
+
+(define-condition expression-not-simplifiable (serious-condition)
+  ()
+  (:documentation "A condition that is signalled any time an expression cannot be simplified.
```
notmgsk: Don't line break docstrings.

src/analysis/simplify-arithmetic.lisp
```diff
@@ -0,0 +1,219 @@
...
+                              ====>           (+ (* 5.0 theta[1]) (* 2.0 theta[0]))
+   COEFFICIENTS:                                                |
+      theta[0]   2.0                                            v
+      theta[1]   5.0                      (+ 3.0 (+ (* 5.0 theta[1]) (* 2.0 theta[0])))
+"
+  (let ((expr nil))
+    (dohash ((ref coefficient) (affine-representation-coefficients rep))
+      (unless (double= 0 coefficient)
+        (setf expr (if expr
+                       `( + (* ,coefficient ,ref) ,expr)
```
notmgsk: space before `+`

src/analysis/simplify-arithmetic.lisp
```diff
@@ -0,0 +1,219 @@
...
+In general, a more specific conditionxs should be preferred over signalling this one."))
+
+(define-condition expression-not-linear (expression-not-simplifiable)
+  ()
+  (:documentation "A condition that is signalled any time an expression cannot be simplified due to
+being non-linear."))
+
+(define-transform simplify-arithmetic (simplify-arithmetic)
+  "A transform which converts a parsed program with potentially complicated arithmetic to one that
+has simplified arithmetic expressions")
```
notmgsk: period

src/analysis/simplify-arithmetic.lisp
```diff
@@ -0,0 +1,219 @@
...
+COEFFICIENT key-value pairs for building the COEFFICIENTS hash table. Thus, the function would be
+used by running
+
+   (make-affine-representation CONSTANT)
+
+which produces an AFFINE-REPRESENTATION with an empty COEFFICIENTS hash table, or
+
+   (make-affine-representation CONSTANT MEMORY-REF_0 COEFFICIENTS_0 ... MEMORY-REF_N COEFFICIENTS_N)
+
+which fills in the COEFFICIENTS hash table with the memory references and their corresponding coefficients"
```
notmgsk: period

src/analysis/simplify-arithmetic.lisp
```diff
@@ -0,0 +1,219 @@
...
+  (:documentation "Generic function that defines the underlying mechanics for the SIMPLIFY-ARITHMETIC
+transform. If this function is given a PARSED-PROGRAM, it recursively applies itself to the program's
+exectuable code. Otherwise, if this function is given a GATE-APPLICATION, it attempts to simplify
+the gate parameters by canonicalizing the arithmetic expressions they (potentially) contain.")
+  (:method ((thing t))
+    thing)
+  (:method ((thing gate-application))
+    (let ((new-gate (copy-instance thing)))
+      (setf (application-parameters new-gate)
+            (map 'cons (lambda (de)
```
notmgsk: I think it's more idiomatic to use the `'list` type here.

src/analysis/simplify-arithmetic.lisp
```diff
@@ -0,0 +1,219 @@
...
+                left: 2.0                           right: (* 2.0 theta[0])
+                                                    //                    \\
+                                            left: 2.0                     right: theta[0]
+"
+  (typecase de
+    (number
+     (make-affine-representation de))
+    (memory-ref
+     (make-affine-representation 0 de 1.0))
+    (cons
```
notmgsk: Probably more idiomatic to use `list` here.

src/analysis/simplify-arithmetic.lisp
```diff
@@ -0,0 +1,219 @@
...
+     (let ((left (expression->affine-representation (second de)))
+           (right (expression->affine-representation (third de))))
+       (combine-affine-representations (car de) left right)))
```
notmgsk: Use `first` rather than `cons`, since you're using `second`, `third`.

src/analysis/simplify-arithmetic.lisp
```diff
@@ -0,0 +1,219 @@
...
+        (setf expr (if expr
+                       `( + (* ,coefficient ,ref) ,expr)
+                       `(* ,coefficient ,ref)))))
+    (cond
+      ((null expr)
+       (affine-representation-constant rep))
+      ((double= 0 (affine-representation-constant rep))
+       expr)
+      (t
+       `(+ ,(affine-representation-constant rep) ,expr)))))
```
notmgsk: nice

tests/analysis-tests.lisp
```diff
@@ -359,3 +359,33 @@ B(b) 0 1
...
+
+(deftest test-simplify-arithmetic-non-linear ()
+  "Test that a non-linear expression is left alone"
+  (let ((in-p (quil::parse-quil "
+DECLARE theta REAL[1]
+RX(2.0+3.0*cos(theta[0])-3.0*theta[0]-2.0) 0
+"))
+        (out-p (quil::parse-quil "
+DECLARE theta REAL[1]
+RX(2.0+3.0*cos(theta[0])-3.0*theta[0]-2.0) 0
```
notmgsk: Can you simplify stuff inside `cos`, `sin`, etc?

src/analysis/simplify-arithmetic.lisp
karalekas: why? it's nice to be able to actually read things without having to rely on auto line wrapping

tests/analysis-tests.lisp
karalekas: @ecpeterson said that for now it's reasonable to just give up when we encounter nonlinear arithmetic expressions. maybe work for a follow-on PR?

src/analysis/simplify-arithmetic.lisp
karalekas: the expected return type is `CONS` -- is it ok to just replace that with `LIST`? this is what I get when I inspect these objects, so I figured I had to keep that the same:
```lisp
QUIL> (let ((pp (parse-quil "DECLARE theta REAL[1]
RX(2.0+3.0*theta[0]-3.0*theta[0]) 0")))
        (type-of (application-parameters (vnth 0 (parsed-program-executable-code pp)))))
CONS
```

src/analysis/simplify-arithmetic.lisp
notmgsk: My gripes are:
 1. It leads to inconsistent style. No one contributor has the same fill-column setting.
 2. Resizing your editor will give ugly lines if the window is narrower than your fill-column.
 3. Dunno.

I won't make more of a fuss, but I think it's worth working towards a consistent style. But maybe not. Maybe chaotic good is ... good?

src/analysis/simplify-arithmetic.lisp
notmgsk: All `cons` are `list`. So there's no issue there, just a comment RE what I think is idiomatic. Ignore it if you like :)

tests/analysis-tests.lisp
notmgsk: Totally. Let's get this one done first :D

src/analysis/simplify-arithmetic.lisp
karalekas: @notmgsk have u heard the new rap song by `(car de) B`

notmgsk: @karalekas i will have to give it a `(list n)`

karalekas:
Tests just failed or the first time (after responding to latest comments and rebasing off of `master`, commit `ba6c343`). Rerunning but here is the output:
```
TEST-SIMPLIFY-ARITHMETIC-NON-LINEARControl stack guard page temporarily disabled: proceed with caution
INFO: Control stack guard page unprotected
Unhandled SB-KERNEL::CONTROL-STACK-EXHAUSTED in thread #<SB-THREAD:THREAD "main thread" RUNNING
                                                          {10005984F3}>:
  Control stack exhausted (no more space for function call frames).
This is probably due to heavily nested or infinitely recursive function
calls, or a tail call that SBCL cannot or has not optimized away.

PROCEED WITH CAUTION.
```

src/analysis/simplify-arithmetic.lisp
notmgsk: > Arguably, natural text readability suffers even more than source code from really long unwrapped lines. 

I would argue the opposite. There can be a great deal of structure in a single line of source code that is difficult to parse, which is exactly why we often break a line when a new function argument begins, to show that structure. Text doesn't have that issue, IMO, since we're generally good at keeping sentences concise (if only we could do the same for code).

notmgsk: Something like soft-breaking docstrings would be nice.

karalekas:
@appleby re: `destructuring-bind` -- looks like we are not guaranteed that these things are of length 3 (ex: `(cos theta[0])`) so that doesn't work

src/analysis/simplify-arithmetic.lisp
```diff
@@ -0,0 +1,219 @@
...
+
+   (make-affine-representation CONSTANT MEMORY-REF_0 COEFFICIENTS_0 ... MEMORY-REF_N COEFFICIENTS_N)
+
+which fills in the COEFFICIENTS hash table with the memory references and their corresponding coefficients."
+  (let ((coefficients (make-hash-table :test 'memory-ref= :hash-function 'memory-ref-hash)))
+    (loop :for (ref coefficient) :on keyval :by #'cddr :do (setf (gethash ref coefficients) coefficient))
+    (%affine-representation :constant constant
+                            :coefficients coefficients)))
+
+(defun combine-affine-representations (operator left right)
```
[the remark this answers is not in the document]
karalekas: I took a stab -- lmk what you think

karalekas:
Also, the build is failing on master, so I don't think its my PR anymore.

appleby:
> @appleby re: `destructuring-bind` -- looks like we are not guaranteed that these things are of length 3 (ex: `(cos theta[0])`) so that doesn't work

Oh. That's a pretty good reason not use `destructuring-bind` then :-).

I'm confused though, wouldn't `combine-affine-representation` choke in that case, since it's `right` argument would wind up being `nil`? Maybe I should just download and try it...

src/analysis/simplify-arithmetic.lisp
jmbr: I agree with @notmgsk.

src/analysis/simplify-arithmetic.lisp
jmbr: Emacs' `M-x visual-line-mode` helps here.

src/analysis/simplify-arithmetic.lisp
appleby: I think changing `cons` to `list` here could result in infinite recursion if it's possible for `de` to be `nil`. Per [@karalekas comment](https://github.com/rigetti/quilc/pull/211#issuecomment-534712691) about the unsuitability of `destructuring-bind`, it seems like a `nil` `de` could be possible here?

src/analysis/simplify-arithmetic.lisp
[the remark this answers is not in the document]
appleby: works for me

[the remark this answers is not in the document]
jmbr: Suggestion (beware: I have not tested this code):

```
  (let ((rep (make-affine-representation
              (funcall operator
                       (affine-representation-constant left)
                       (affine-representation-constant right)))))
  
    (flet ((plus-or-minus (operator left right)
             (dohash ((ref coefficient) (affine-representation-coefficients left))
               (setf (gethash ref rep) coefficient))
             (dohash ((ref coefficient) (affine-representation-coefficients right))
               (a:if-let (x (gethash ref rep))
                 (setf (gethash ref rep) (funcall operator x coefficient))
                 (setf (gethash ref rep) coefficient)))))

      (ecase operator
        (+ (plus-or-minus operator left right))
        (- (plus-or-minus operator left right))
...
```

src/analysis/simplify-arithmetic.lisp
```diff
@@ -0,0 +1,211 @@
...
+   (make-affine-representation CONSTANT)
+
+which produces an AFFINE-REPRESENTATION with an empty COEFFICIENTS hash table, or
+
+   (make-affine-representation CONSTANT MEMORY-REF_0 COEFFICIENTS_0 ... MEMORY-REF_N COEFFICIENTS_N)
+
+which fills in the COEFFICIENTS hash table with the memory references and their corresponding coefficients."
+  (let ((coefficients (make-hash-table :test 'memory-ref= :hash-function 'memory-ref-hash)))
+    (loop :for (ref coefficient) :on keyval :by #'cddr :do (setf (gethash ref coefficients) coefficient))
+    (%affine-representation :constant constant
```
stylewarning: why not check the type here of coefficient?

src/analysis/simplify-arithmetic.lisp
```diff
@@ -0,0 +1,211 @@
...
+
+   (make-affine-representation CONSTANT)
+
+which produces an AFFINE-REPRESENTATION with an empty COEFFICIENTS hash table, or
+
+   (make-affine-representation CONSTANT MEMORY-REF_0 COEFFICIENTS_0 ... MEMORY-REF_N COEFFICIENTS_N)
+
+which fills in the COEFFICIENTS hash table with the memory references and their corresponding coefficients."
+  (let ((coefficients (make-hash-table :test 'memory-ref= :hash-function 'memory-ref-hash)))
+    (loop :for (ref coefficient) :on keyval :by #'cddr :do (setf (gethash ref coefficients) coefficient))
```
stylewarning: newline before `:do`

src/analysis/simplify-arithmetic.lisp
```diff
@@ -0,0 +1,211 @@
...
+                            :coefficients coefficients)))
+
+(defun combine-affine-representations (operator left right)
+  "Given an operator (e.g. +, -, *, /) and two operand expressions (LEFT and RIGHT, in AFFINE-REPRESENTATION
+form), combine them via the rules of arithmetic, enforcing the linearity of the output AFFINE-REPRESENTATION.
+When the OPERATOR is + or -, there are no restrictions on the contents of LEFT or RIGHT. If the OPERATOR
+is *, one of either LEFT or RIGHT must be simply a CONSTANT value, meaning that its COEFFICIENTS hash table
+is empty (otherwise we give up simplifying). If the OPERATOR is /, the RIGHT operand must be a CONSTANT
+value, meaning that its COEFFICIENTS hash table must be empty (otherwise we give up simplifying)."
+  (let ((rep (make-affine-representation
```
stylewarning: add `(check-type operator (member + - * /))`

appleby:
> > @appleby re: `destructuring-bind` -- looks like we are not guaranteed that these things are of length 3 (ex: `(cos theta[0])`) so that doesn't work
> 
> Oh. That's a pretty good reason not use `destructuring-bind` then :-).
> 
> I'm confused though, wouldn't `combine-affine-representation` choke in that case, since it's `right` argument would wind up being `nil`? Maybe I should just download and try it...

I see it now. It *is* an error to pass `nil` to `combine-affine-representation`, but the recursion in `expression->affine-representation` will bottom out in the `otherwise` clause and signal `expression-not-simplifiable`. I somehow missed that `otherwise` and assumed the recursive call would return `nil`. This reinforces that the type in the `typecase` should not be changed from `cons` to `list` unless you add a prior explicit `null` check.

src/analysis/simplify-arithmetic.lisp
```diff
@@ -0,0 +1,219 @@
...
+
+ where
+
+   CONSTANT is the constant offset (of type NUMBER) of the expression
+   COEFFICIENTS_i is the coefficient (of type NUMBER) for MEMORY-REF_i
+   MEMORY-REF_i is the MEMORY-REF with coefficient COEFFICIENTS_i
+"
+  (constant 0 :type number)
+  ;; This builds a hash table with MEMORY-REFs as keys and their coefficients (of type NUMBER) as values
+  (coefficients (make-hash-table :test 'memory-ref= :hash-function 'memory-ref-hash) :type hash-table))
```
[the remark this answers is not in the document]
appleby: yep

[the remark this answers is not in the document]
appleby: Filed #435

appleby:
> > > @appleby re: `destructuring-bind` -- looks like we are not guaranteed that these things are of length 3 (ex: `(cos theta[0])`) so that doesn't work
> > 
> > 
> > Oh. That's a pretty good reason not use `destructuring-bind` then :-).
> > I'm confused though, wouldn't `combine-affine-representation` choke in that case, since it's `right` argument would wind up being `nil`? Maybe I should just download and try it...
> 
> I see it now. It _is_ an error to pass `nil` to `combine-affine-representation`, but the recursion in `expression->affine-representation` will bottom out in the `otherwise` clause and signal `expression-not-simplifiable`. I somehow missed that `otherwise` and assumed the recursive call would return `nil`. This reinforces that the type in the `typecase` should not be changed from `cons` to `list` unless you add a prior explicit `null` check.

Hmmm, does this mean that an expression term like `(- theta[0])` anywhere would cause it to signal `expression-not-simplifiable`? For example: should `RX(-theta[0] + theta[0]) 0` simplify?

src/analysis/simplify-arithmetic.lisp
stylewarning: use mapcar

appleby:
> @appleby re: `destructuring-bind` -- looks like we are not guaranteed that these things are of length 3 (ex: `(cos theta[0])`) so that doesn't work

Not sure why this didn't occur to me earlier... got fixated on how the NULL was being handled... but `destructuring-bind`'s first argument is a destructuring lambda-list, so you can specify `&optional` for the optional third arg if you want. I think that would actually help to make it explicitly clear that `(third de)` might be NULL. Something like:

```
(destructuring-bind (op left &optional right) de
   (combine-affine-representations op
                                   (expression->affine-representation left)
                                   (expression->affine-representation right)))
```
Same goes for `left` if it's also optional.

Maybe add a comment indicating that when `right` is NULL, then `expression-not-simplifiable` will be signaled on the recursive invocation? Fooled me anyway :-D.

src/analysis/simplify-arithmetic.lisp
```diff
@@ -2,85 +2,83 @@
...
+;;;; expressions (ones that are linear). However, for the majority of
+;;;; our use cases at this time, this subset is sufficient.
 
 (in-package #:cl-quil)
 
 (define-condition expression-not-simplifiable (serious-condition)
   ()
-  (:documentation "A condition that is signaled any time an expression cannot be simplified.
-In general, a more specific conditions should be preferred over signaling this one."))
+  (:documentation "A condition that is signaled any time an expression cannot be simplified. In general, a more specific conditions should be preferred over signaling this one."))
```
notmgsk: Actually I prefer these with linebreaks.

karalekas:
@appleby you are correct that currently `RX(-theta[0] + theta[0]) 0` will not simplify
```
QUIL> (let ((pp (parse-quil "DECLARE theta REAL[1]
RX(-theta[0]+theta[0]) 0")))
        (transform 'simplify-arithmetic pp)
        (print-parsed-program pp))
DECLARE theta REAL

RX((-(theta[0])+theta[0])) 0
NIL
```

src/analysis/simplify-arithmetic.lisp
karalekas: I'd recommend you sleep with one eye open tonight

src/analysis/simplify-arithmetic.lisp
```diff
@@ -0,0 +1,198 @@
...
+;;;;    RZ(2.5*theta[1] + 5.0) 1
+;;;;
+;;;; This implementation is certainly not without its limitations. For
+;;;; example, we can only simplify a particular subset of arithmetic
+;;;; expressions (ones that are linear). However, for the majority of
+;;;; our use cases at this time, this subset is sufficient.
+
+(in-package #:cl-quil)
+
+(define-condition expression-not-simplifiable (serious-condition)
```
notmgsk: What is the motivation for `serious-condition` here?

karalekas: matches a pattern elsewhere in the source

stylewarning:
> @appleby you are correct that currently `RX(-theta[0] + theta[0]) 0` will not simplify
> 
> ```
> QUIL> (let ((pp (parse-quil "DECLARE theta REAL[1]
> RX(-theta[0]+theta[0]) 0")))
>         (transform 'simplify-arithmetic pp)
>         (print-parsed-program pp))
> DECLARE theta REAL
> 
> RX((-(theta[0])+theta[0])) 0
> NIL
> ```

this should be fixed and added as a test case

src/analysis/simplify-arithmetic.lisp
```diff
@@ -0,0 +1,205 @@
...
+which produces an AFFINE-REPRESENTATION with an empty COEFFICIENTS hash table, or
+
+   (make-affine-representation CONSTANT MREF_0 COEFF_0 ... MREF_N COEFF_N)
+
+which fills in the COEFFICIENTS hash table with the memory references and their corresponding coefficients."
+  (assert (evenp (length keyval)))
+  (let ((coefficients (make-hash-table :test 'memory-ref=
+                                       :hash-function 'memory-ref-hash)))
+    (loop :for (ref coefficient) :on keyval :by #'cddr
+          :do (assert (numberp coefficient))
```
ecpeterson: May as well also assert that `ref` is a memory reference?

src/analysis/simplify-arithmetic.lisp
```diff
@@ -0,0 +1,205 @@
...
+       (dohash ((ref coefficient) (affine-representation-coefficients right))
+           (if (gethash ref (affine-representation-coefficients rep))
+               (decf (gethash ref (affine-representation-coefficients rep)) coefficient)
+               (setf (gethash ref (affine-representation-coefficients rep)) (- coefficient))))
+       rep)
+      (*
+       (unless (or (zerop (hash-table-count (affine-representation-coefficients left)))
+                   (zerop (hash-table-count (affine-representation-coefficients right))))
+         (error 'expression-not-linear))
+       (dohash ((ref coefficient) (affine-representation-coefficients left))
```
ecpeterson: This pair of `dohash`es still tickles me.

src/analysis/simplify-arithmetic.lisp
karalekas: Happy to, but when I asked @stylewarning about this, he said that the custom test & hash functions above take care of that check

src/analysis/simplify-arithmetic.lisp
```diff
@@ -0,0 +1,205 @@
...
+                               left: 2.0                     right: theta[0]
+"
+  (typecase de
+    (number
+     (make-affine-representation de))
+    (memory-ref
+     (make-affine-representation 0 de 1.0))
+    (cons
+     (destructuring-bind (op left &optional right) de
+       (if (and (null right) (equalp op '-))
```
stylewarning: equalp => eq

ecpeterson:
I'm going to take Robert at his word that this was the final change and merge this.

PR #212, opened 2019-04-30
Reduce global state mutation in quilc

### LINKED ISSUE #192: Handle quilc's global variables in the tests more nicely
`%entry-point` mutates `quilc::*protoquil*` and `quilc::*compute-matrix-reps*`, which can cause future tests to be awry. We should audit this and make sure tests provide new bindings to mutate, *or* change the entry point to not mutate.

appleby:
Replace most global state mutation in quilc with let bindings.

This is "option 5" from the discussion in #192. Turns out the state mutation was almost entirely confined to `quilc::process-options`, so switching to let bindings everywhere wasn't too bad.

The only remaining global state mutation occurs in `quilc::setup-debugger` which is called from `quilc::entry-point`, but not `quilc::%entry-point`. Removing this last bit of state mutation is also possible, but slightly uglier since it requires either peeking under the covers of `sb-ext:disable-debugger` to figure out what additional vars to let-bind, or else running the body of `quilc::entry-point` in an `unwind-protect` that calls `quilc::enable-debugger` on exit. Since the whole point of `quilc::%entry-point` is presumably to provide a non-debugger-mangling entry point for interactive use, it didn't seem worth it to take this final step.

Fixes #192

### DIFF
--- a/app/src/entry-point.lisp
+++ b/app/src/entry-point.lisp
@@ -76,15 +76,6 @@
     #+forest-sdk
     (("skip-version-check") :type boolean :optional t :initial-value nil :documentation "Do not check for a new SDK version at launch.")))
 
-(defparameter *isa-descriptors*
-  (alexandria::alist-hash-table
-   `(("8Q" . ,(quil::build-8Q-chip))
-     ("20Q" . ,(quil::build-skew-rectangular-chip 0 4 5))
-     ("16QMUX" . ,(quil::build-16QMUX-chip))
-     ("bristlecone" . ,(quil::build-bristlecone-chip))
-     ("ibmqx5" . ,(quil::build-ibm-qx5)))
-   :test 'equal))
-
 (defun slurp-lines (&optional (stream *standard-input*))
   (flet ((line () (read-line stream nil nil nil)))
     (with-output-to-string (s)
@@ -114,14 +105,16 @@
            (coerce #(#\Newline #\#) 'string))))
 
 (defun lookup-isa-descriptor-for-name (isa)
-  (let ((isa-hash-value (gethash isa *isa-descriptors*)))
-    (cond
-      ((not (null isa-hash-value))
-       isa-hash-value)
-      ((probe-file isa)
-       (quil::read-chip-spec-file isa))
-      (t
-       (error "ISA descriptor does not name a known template or an extant file.")))))
+  (alexandria:switch (isa :test #'equal)
+    ("8Q" (quil::build-8Q-chip))
+    ("20Q" (quil::build-skew-rectangular-chip 0 4 5))
+    ("16QMUX" (quil::build-16QMUX-chip))
+    ("bristlecone" (quil::build-bristlecone-chip))
+    ("ibmqx5" (quil::build-ibm-qx5))
+    (t
+     (if (probe-file isa)
+         (quil::read-chip-spec-file isa)
+         (error "ISA descriptor does not name a known template or an extant file.")))))
 
 (defun log-level-string-to-symbol (log-level)
   (alexandria:eswitch (log-level :test #'string=)
--- a/app/src/entry-point.lisp
+++ b/app/src/entry-point.lisp
@@ -214,16 +214,14 @@
       (uiop:quit 1))))
 
 (defun %entry-point (argv)
-  ;; grab the CLI arguments
-  (setf *program-name* (pop argv))
-
-  (command-line-arguments:handle-command-line
-       *option-spec*
-       'process-options
-       :command-line argv
-       :name "quilc"
-       :positional-arity 0
-       :rest-arity nil))
+  (let ((*program-name* (pop argv)))
+    (command-line-arguments:handle-command-line
+     *option-spec*
+     'process-options
+     :command-line argv
+     :name *program-name*
+     :positional-arity 0
+     :rest-arity nil)))
 
 (defun process-options (&key
                           (prefer-gate-ladders nil)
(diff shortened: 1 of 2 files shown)


app/src/entry-point.lisp
```diff
@@ -342,52 +337,42 @@ HTTP server for good.
...
+	(t
+	 (cond
+	   (json-serialize
+	    (setf *json-stream* *standard-output*))
+	   (t
+	    (setf *human-readable-stream* *error-output*)
+	    (setf *quil-stream* *standard-output*)))
+         (when verbose
+           (setf *verbose* *human-readable-stream*))
+         (let ((*isa-descriptor* (lookup-isa-descriptor-for-name isa)))
```
appleby: The variable `*isa-descriptor*` is only referenced in `run-CLI-mode` where it is merely passed along to `process-program`. It would therefore be easy to remove this variable and replace it with an optional argument to `run-CLI-mode`, if desired.

stylewarning:
What about a macro called `with-debugger-disabled` or so?

appleby:
> What about a macro called `with-debugger-disabled` or so? CC @rigettizach

I did consider such a macro, but decided against it since I though it might seem like a Gratuitous Use of Macros (TM). Especially because I assumed that `entry-point` (as opposed to `%entry-point`) is only likely to be called when running the quilc executable and hence didn't seem like much point restoring the debugger state when the next thing that happens is process exit. I'm happy to add it though if maintainers feel it's a good idea.

stylewarning:
@appleby You're right, it is a bit of a gratuitous use of a macro, but then again, it's not a very confusing macro. I would look at @notmgsk's recent work on temporarily disabling the debugger with his `with-mocked-function-definitions` macro: https://github.com/rigetti/qvm/blob/master/app/tests/utils.lisp#L3

See its use here: https://github.com/rigetti/qvm/blob/master/app/tests/suite.lisp#L179

We ought to get this macro included in Fiasco itself, so everybody can use it.

appleby:
Personally, I'm all in favor of small macros that make your life better. I just erred on the side of caution since opinions vary and I wasn't 100% sure it was necessary.

There are some small complications. For example, `setup-debugger` only disables the debugger if the `forest-sdk` feature is not present; otherwise, it sets `*debugger-hook*` to `'command-line-debugger`. So it's not quite as simple as

``` common-lisp
(with-debugger-disabled
   ;; body of entry-point here)
```

One option is to rename the macro `with-debugger-setup`, but that has a slightly bad mouthfeel. As you say, `with-debugger-disabled` is general enough to possibly be included in `fiasco`, but `with-debugger-setup` feels like a single-use macro.

Alternatively, the +/- `forest-sdk` conditional feature logic of `setup-debugger` could be moved into `entry-point`, like so:

``` common-lisp
#+forest-sdk
(let ((*debugger-hook* 'command-line-debugger))
  (entry-point-thunk) ; or something with a better name
#-forest-sdk
(with-debugger-disabled
  (entry-point-thunk)))
```

One final complication I just noticed: `build-app.lisp` also calls `quilc::setup-debugger` when preparing the executable image. I'm not sure why. Possibly to ensure the debugger is setup even when `build-app` is passed the `--unsafe` option, which causes it to use `%entry-point` rather than `entry-point`, which would otherwise skip debugger setup?

Or maybe I've misunderstood your intention, and you mean to leave the debugger twiddling in `entry-point.lisp` alone, and that `with-debugger-disabled` is meant to be macro for use only in testing, similar to @notmgsk's `with-mocked-function-definitions`? And in that case, `setup-debugger` could be mocked to a no-op?

Clearly, I need to think about this a bit more. Posting rambles for comments in case anyone has better ideas.

stylewarning:
I would rather just setup the debugger in build-app.lisp, and avoid it in the entry-point. I don't think the debugger is re-enabled upon "boot", but maybe that's why both instances are there. I forgot.

appleby:
> I would rather just setup the debugger in build-app.lisp, and avoid it in the entry-point.

Ok, I will give that a try later and see how it goes.

appleby:
I just pushed a new commit that removes the call to `quilc::setup-debugger` from `quilc::entry-point`.

Here is the overly-long-winded description of this small change.

I manually tested the effect of this change with the following quilc variants / make targets:

- quilc
- quilc-unsafe
- quilc-sdk-barebones

I did not test quilc-sdk, because it apparently expects libs in `/usr/local/lib/rigetti/<whatever>`. I did try the docker-sdk target thinking that the docker build might put the libs in the right place for me, but that also fails at startup. See the end of this comment for the exact error message from running the docker-sdk build, in case that is not an expected error.

I tested this change by executing each of the following forms at the top of `quilc::%entry-point`, to make sure the behavior is the same before and after removing the call to `setup-debugger` in `entry-point`:

- `(break)`
- `(error ...)`
- `(invoke-debugger ...)`

The good news: behavior appears to be the same before and after. That is, whatever debugger state is modified by the call to `setup-debugger` in `build-app.lisp` does persist across the call to `sb-ext:save-lisp-and-die`. In no case was I dropped into a debugger. "Safe" variants that use `entry-point` print a one-line summary of errors, and print a backtrace for anything that would normally invoke the debugger. The "unsafe" variant that uses `%entry-point` prints a backtrace in all cases.

Two small surprises.

1) The call to `setup-debugger` can also be removed from `build-app.lisp`, apparently without consequence. I think the reason is that the make target invokes sbcl with the `--non-interactive` flag, which implies `--disable-debugger`. Apparently, the disabled debugger persists in the executable image.

2) as a consequence of (1), conditionally setting `*debugger-hook*` to `'command-line-debugger` when the `forest-sdk` feature is present appears to have no effect. Or at least I could not determine how to tickle it. To clarify, the definition of `setup-debugger` looks like this:

```
(defun setup-debugger ()
  #+forest-sdk
  (setf *debugger-hook* 'command-line-debugger)
  #-forest-sdk
  (disable-debugger))
```

When the `forest-sdk` feature is present (i.e. in the `{quilc,docker}-sdk{,-barebones}` variants), `*debugger-hook*` is indeed set to `'command-line-debugger`, but sbcl essentially ignores it because `sb-ext:*invoke-debugger-hook*` is already set to `sb-debug::debugger-disabled-hook`, due to (1).

Note that I didn't test any of this in ccl because I didn't see an easy way to produce a ccl executable since `build-app.lisp` is sbcl-specific. But I did notice that ccl has it's own implementations of `{enable,disable}-debugger` in `app/src/impl/clozure.lisp`, so it's possible the setting of `*debugger-hook*` is there for ccl.

# Error message when trying to run docker image produced by `make docker-sdk`

It seems libzmq.so is present in the docker image, but it's located at `/usr/lib/x86_64-linux-gnu/libzmq.so`.

```
$ docker run --rm -it quilc-sdk
Unhandled SIMPLE-ERROR in thread #<SB-THREAD:THREAD "main thread" RUNNING
                                    {10006B85B3}>:
  Error opening shared object "/usr/local/lib/rigetti/libzmq.so":
  /usr/local/lib/rigetti/libzmq.so: cannot open shared object file: No such file or directory.

Backtrace for: #<SB-THREAD:THREAD "main thread" RUNNING {10006B85B3}>
0: (SB-DEBUG::DEBUGGER-DISABLED-HOOK #<SIMPLE-ERROR "Error opening ~:[runtime~;shared object ~:*~S~]:
  ~A." {1000EC56D3}> #<unused argument> :QUIT T)
1: (SB-DEBUG::RUN-HOOK *INVOKE-DEBUGGER-HOOK* #<SIMPLE-ERROR "Error opening ~:[runtime~;shared object ~:*~S~]:
  ~A." {1000EC56D3}>)
2: (INVOKE-DEBUGGER #<SIMPLE-ERROR "Error opening ~:[runtime~;shared object ~:*~S~]:
  ~A." {1000EC56D3}>)
3: (ERROR "Error opening ~:[runtime~;shared object ~:*~S~]:
  ~A." "/usr/local/lib/rigetti/libzmq.so" "/usr/local/lib/rigetti/libzmq.so: cannot open shared object file: No such file or directory")
4: (SB-SYS:DLOPEN-OR-LOSE #S(SB-ALIEN::SHARED-OBJECT :PATHNAME #P"/usr/local/lib/rigetti/libzmq.so" :NAMESTRING "/usr/local/lib/rigetti/libzmq.so" :HANDLE NIL :DONT-SAVE NIL))
5: (SB-ALIEN::TRY-REOPEN-SHARED-OBJECT #S(SB-ALIEN::SHARED-OBJECT :PATHNAME #P"/usr/local/lib/rigetti/libzmq.so" :NAMESTRING "/usr/local/lib/rigetti/libzmq.so" :HANDLE NIL :DONT-SAVE NIL))
6: (SB-SYS:REOPEN-SHARED-OBJECTS)
7: (SB-IMPL::FOREIGN-REINIT)
8: (SB-IMPL::REINIT)
9: ((FLET SB-UNIX::BODY :IN SAVE-LISP-AND-DIE))
10: ((FLET "WITHOUT-INTERRUPTS-BODY-14" :IN SAVE-LISP-AND-DIE))
11: ((LABELS SB-IMPL::RESTART-LISP :IN SAVE-LISP-AND-DIE))

```

rigettizach:
`make docker-sdk` is for producing a binary to extract for our downloadable SDK. It's not meant to be run on its own.

appleby:
> `make docker-sdk` is for producing a binary to extract for our downloadable SDK. It's not meant to be run on its own.

Got it. Thanks.

app/src/entry-point.lisp
stylewarning: That sounds reasonable to me

appleby:
@ecpeterson Will do.

appleby:
Pushed additional commits removing both `*isa-descriptor*` and `*isa-descriptors*` hash-table. Not sure if the latter is wanted, but it was only referenced in `lookup-isa-descriptor-for-name`, so I just moved the moral equivalent into that function.

It also occurred to me that I could add some tests for option parsing, piggybacking on the aforementioned `with-mocked-function-definitions` macro. I actually started writing some, but figured I would check here first to see if such tests are even desired.  Personally, I think the tests could be useful for validating this PR, but would likely be low-value tests going forward. Adding the tests would presumably entail copy/pasting the definition of `with-mocked-function-definitions` from the qvm test package.

app/src/entry-point.lisp
```diff
@@ -115,14 +105,16 @@
...
 (defun lookup-isa-descriptor-for-name (isa)
-  (let ((isa-hash-value (gethash isa *isa-descriptors*)))
-    (cond
-      ((not (null isa-hash-value))
-       isa-hash-value)
-      ((probe-file isa)
-       (quil::read-chip-spec-file isa))
-      (t
-       (error "ISA descriptor does not name a known template or an extant file.")))))
+  (alexandria:switch (isa :test #'equal)
```
appleby: Changing the test to `#'equalp` here would allow users to specify the `--isa` option in either case, i.e., both `quilc --isa 8Q` and `quilc --isa 8q` would be acceptable.

stylewarning: I think it ought to be `string=` to be more type strict. If a non-string value is allowed, we should check for that beforehand.

app/src/entry-point.lisp
```diff
@@ -277,62 +265,59 @@ Version ~A is available from https://www.rigetti.com/forest~%"
...
-            *quil-stream* (make-broadcast-stream))
-
-      ;; launch the polling loop
-      (unless quiet
-        (show-banner))
-
-      (unless quiet
-        (format t "~%
+
+  (let* ((*log-level* (or (and log-level (log-level-string-to-symbol log-level))
```
stylewarning: not necessary in this PR, but we should aim to reduce these bindings in future PRs

app/src/entry-point.lisp
```diff
@@ -342,57 +327,46 @@ HTTP server for good.
...
-     (setf *isa-descriptor* (lookup-isa-descriptor-for-name isa))
-     (setf *verbose*
-           (cond
-            (verbose *human-readable-stream*)
-            (t (make-broadcast-stream))))
-     (run-CLI-mode)))))
-
-(defun run-CLI-mode ()
+	 (when port
+           (cl-syslog:rfc-log (*logger* :debug "WARNING: -p and -S are incompatible. Dropping -p.")
```
stylewarning: it's odd to see `:debug` then `WARNING`.

appleby: Log level change to `:warning` in the latest diff.

app/src/entry-point.lisp
appleby: Done.

app/src/entry-point.lisp
appleby: I opened #226 to track this.

PR #150, opened 2019-03-20
Classical logic gate compilation

notmgsk:
Classical logic gates (where each column contains one (and only one) one) when compiled by quilc do not produce a healthy-looking circuit. This is a result of LAPACK (which we use for matrix-related stuff) not dealing with these gates as a special-case, instead using a general-case algorithm which produces lacklustre results. [Tweedledum](https://github.com/boschmitt/tweedledum) is a C++ library to handle exactly this situtation. It produces shorter circuits with fewer 2Q gates, particularly as the gate size increases, when compared with quilc on the same gate.

This PR consists of the following items (please add more if you see fit):
- [x] CFFI to Tweedledum, cl-tweedledum.
- [x] Extension to the Quil language to support something like `DEFGATE AS (MATRIX|PERM)` where quilc will automatically invoke tweedledum for the `PERM` type.
- [x] Tests
- [ ] Analysis comparing tweedledum to quilc on range of inputs, pretty plots with labeleled axes.

Closes #120.

PR #219, opened 2019-05-01
treat stdgates uniformly with other DEFGATES (and more)

stylewarning:
This commit does a good helping of refactoring, further clarifying
confusion over what a gate is, how a standard gate rears its head, and
how CONTROLLED and DAGGER gates get simplified.

Three big notable things in this PR:

- Standard gates are read from a new file src/quil/stdgates.quil.

- GATE-APPLICATION-GATE is not an accessor anymore, but a function
      that computes and caches a gate object depending on the resolved
      name.

- Hooks to optimize the representation of a CONTROLLED gate and a
      DAGGER gate exist with the generic functions CONTROL-GATE and
      DAGGER-GATE, which are to be read as verbs.

This is intended to replace PR #217, and should **NOT** be merged only until @notmgsk's PR #150 is merged. (The reason to wait is because all of the standard gates are being boiled away to their Quil definitions, and thus will not be optimized any longer until that PR is merged.)

From Eric's PR:

- [x] `lookup-standard-gate` should return a `gate-definition` of sorts, _not_ a gate.
- [x] `resolve-applications` should be modified so that we stash away that definition in the `name-resolution` field.
- [x] We make generic functions `control-gate` and `dagger-gate` which are responsible for the kind of logic you wrote. This way the dispatch can be nicely specialized.
- [x] To `gate-application`, we add the `:writer` called `%set-gate-definition-gate`.
- [x] To `gate-application`, we add the `:reader` called `%get-gate-definition-gate`.
- [x] To `gate-application`, we make the `:initform` to the `gate` form `nil`.
- [x] We eliminate the `:accessor gate-definition-gate`, and instead make it a function that first tries to `%get-gate-definition-gate`, and if not, builds (but does not `%set`!) using the lifter.
- [x] We modify `operator-description-lifter` to make use of `control-gate` and `dagger-gate` GFs. This should make that function _much_ simpler, even simpler than before, since you're composing these GF's directly instead of making lambdas that are `make-instance`ing things all over.
- [x] Change the documentation to `operator-description-lifter` to be clear that it could construct any manner of `gate`-like object, but will try to produce a reasonable, direct, and efficient representation.
- [x] Make sure tests pass. (They're not right now!)
- [x] Fix the QVM in any way that this breaks it.

### DIFF
--- a/src/analysis/resolve-applications.lisp
+++ b/src/analysis/resolve-applications.lisp
@@ -21,13 +21,13 @@
       (let* ((operator (application-operator app))
              (addl-qubits (operator-description-additional-qubits operator))
              (name (operator-description-root-name operator))
-             (found-gate-defn (find name gate-definitions
-                                    :test #'string=
-                                    :key #'gate-definition-name))
+             (found-gate-defn (or (find name gate-definitions
+                                        :test #'string=
+                                        :key #'gate-definition-name)
+                                  (lookup-standard-gate name)))
              (found-circ-defn (find name circuit-definitions
                                     :test #'string=
-                                    :key #'circuit-definition-name))
-             (in-default-gateset (lookup-standard-gate name)))
+                                    :key #'circuit-definition-name)))
         (cond
           ;; Gate application
           (found-gate-defn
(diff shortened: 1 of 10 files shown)


stylewarning:
Note that this PR will cause some performance regressions (in terms of time to compile) because it reconstructs gate objects _much_ more frequently from the definitions, instead of having the definitions cached as they were before for standard gates.

stylewarning:
This last commit fixed the performance regression suitably well. 🙌

src/compilers/approx.lisp
```diff
@@ -632,8 +632,8 @@ Additionally, if PREDICATE evaluates to false and *ENABLE-APPROXIMATE-COMPILATIO
               (reduce #'magicl:multiply-complex-matrices
                       (list
                        m
-                       (su2-on-line 0 (gate-matrix (lookup-standard-gate "RY") sigma))
```
stylewarning: @ecpeterson be attentive to this change

src/analysis/resolve-applications.lisp
```diff
@@ -21,34 +23,37 @@
...
                       (+ addl-qubits
                          (ilog2
                           (isqrt
                            (length (gate-definition-entries found-gate-defn)))))))
                ;; Check that all arguments are distinct
-               (assert-and-print-instruction (= (length qubit-indices) (length distinct-indices))
+               (assert-and-print-instruction (= (length args) (length distinct-args))
                                              ()
-                                             "All arguments must have distinct indices. Check indices for duplicates.")
+                                             "All arguments must have distinct arguments. Check arguments for duplicates.")
```
ecpeterson: "must be distinct." ?

src/quil/stdgates.quil
```diff
@@ -0,0 +1,143 @@
+### Quil standard gate defintions.
```
ecpeterson: This is a cool file.

ecpeterson:
This is a cool PR.

stylewarning:
@ecpeterson done: https://github.com/rigetti/qvm/pull/84

PR #220, opened 2019-05-01
Define, signal, and handle specific conditions for compression issues

notmgsk:
This is mostly to remove the handler that has a `error` case. If an unrelated error was caused around the same time that compression occurs, it would be caught by this handler and not provide a particularly useful error message.

### DIFF
--- a/src/compressor/compressor.lisp
+++ b/src/compressor/compressor.lisp
@@ -449,6 +449,10 @@ other's."
         ;; otherwise, we're obligated to do full unitary compilation.
         (decompile-instructions-into-full-unitary)))))
 
+(define-condition quil-compression-error (simple-error))
+(define-condition quil-compression-matrices-disagree (quil-compression-error))
+(define-condition quil-compression-states-disagree (quil-compression-error))
+(define-condition quil-compression-matrices-not-close (quil-compression-error))
 
 (defun check-contextual-compression-was-well-behaved (instructions
                                                       decompiled-instructions
@@ -481,19 +485,20 @@ other's."
                   (stretched-matrix (or (make-matrix-from-quil instructions :relabeling relabeling)
                                         (return-from check-quil-agrees-as-matrices nil)))
                   (decompiled-matrix
-                   (make-matrix-from-quil decompiled-instructions
-                                          :relabeling relabeling))
+                    (make-matrix-from-quil decompiled-instructions
+                                           :relabeling relabeling))
                   (reduced-matrix
-                   (kron-matrix-up (make-matrix-from-quil reduced-instructions
-                                                          :relabeling relabeling)
-                                   (ilog2 (magicl:matrix-rows stretched-matrix))))
+                    (kron-matrix-up (make-matrix-from-quil reduced-instructions
+                                                           :relabeling relabeling)
+                                    (ilog2 (magicl:matrix-rows stretched-matrix))))
                   (reduced-decompiled-matrix
-                   (kron-matrix-up (make-matrix-from-quil reduced-decompiled-instructions
-                                                          :relabeling relabeling)
-                                   (ilog2 (magicl:matrix-rows stretched-matrix)))))
-             (assert (matrix-equality stretched-matrix
+                    (kron-matrix-up (make-matrix-from-quil reduced-decompiled-instructions
+                                                           :relabeling relabeling)
+                                    (ilog2 (magicl:matrix-rows stretched-matrix)))))
+             (unless (matrix-equality stretched-matrix
                                       (scale-out-matrix-phases reduced-matrix
-                                                               stretched-matrix)))
+                                                               stretched-matrix))
+               (signal ))
              (when decompiled-instructions
                (assert (matrix-equality stretched-matrix
                                         (scale-out-matrix-phases decompiled-matrix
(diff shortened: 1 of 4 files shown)


src/compressor/compressor.lisp
```diff
@@ -469,81 +473,78 @@ other's."
...
-                   ;; matrices don't take up quite so much space.
-                   (loop :for i :in (reduce #'union
-                                            (list instructions
-                                                  reduced-instructions
-                                                  decompiled-instructions
-                                                  reduced-decompiled-instructions)
-                                            :key #'qubits-in-instr-list)
-                         :for j :from 0
-                         :collect (cons i j)))
+                    ;; the actual reassignment we use is unimportant. this is
```
ecpeterson: mysterious spacing

stylewarning:
@notmgsk As discussed offline, consider a
```
(defmacro assert-compiler-correctness (condition-name &body check-forms)
  (check-type condition-name symbol)
  `(unless (and ,@check-forms) (error ',condition-name)))
```

src/compressor/compressor.lisp
notmgsk: yeah emacs is being weird -- not me trying to up my diff

src/compressor/compressor.lisp
```diff
@@ -449,6 +449,22 @@ other's."
         ;; otherwise, we're obligated to do full unitary compilation.
         (decompile-instructions-into-full-unitary)))))
 
+(defmacro assert-compiler-correctness (condition-name check-form &rest init-forms &key &allow-other-keys)
+  (check-type condition-name symbol)
+  `(unless ,check-form
+     (error ',condition-name ,@init-forms)))
+
+(define-condition quil-compression-error (simple-error) ())
```
stylewarning: Prefer `error` not `simple-error` if you're not using the format facilities.

src/compressor/compressor.lisp
```diff
@@ -449,6 +449,22 @@ other's."
...
+(define-condition quil-compression-matrices-disagree-error (quil-compression-error) ())
+(define-condition quil-compression-states-disagree-error (quil-compression-error)
+  ((final-wf :initarg :final-wf :accessor quil-compression-error-final-wf)
+   (final-wf-reduced :initarg :final-wf-reduced :accessor quil-compression-error-final-wf-reduced)
+   (type :initarg :type :accessor quil-compression-error-type))
+  (:report (lambda (c s) (format s "During careful checking of instruction compression, the wavefunction produced by by ~A was detected to not be collinear with the target wavefunction."
+                            (quil-compression-error-type c)))))
+(define-condition quil-compression-matrices-not-close-error (quil-compression-error)
+  ()
+  (:report (lambda (c s) (format s "During careful checking of instruction compression, the recomputed instruction sequence has an unreasonably large fidelity drop from the original sequence."))))
```
stylewarning: ignore `c`, use `write-string` if you're just doing that.

src/compressor/compressor.lisp
```diff
@@ -449,6 +449,22 @@ other's."
...
+  `(unless ,check-form
+     (error ',condition-name ,@init-forms)))
+
+(define-condition quil-compression-error (simple-error) ())
+(define-condition quil-compression-matrices-disagree-error (quil-compression-error) ())
+(define-condition quil-compression-states-disagree-error (quil-compression-error)
+  ((final-wf :initarg :final-wf :accessor quil-compression-error-final-wf)
+   (final-wf-reduced :initarg :final-wf-reduced :accessor quil-compression-error-final-wf-reduced)
+   (type :initarg :type :accessor quil-compression-error-type))
+  (:report (lambda (c s) (format s "During careful checking of instruction compression, the wavefunction produced by by ~A was detected to not be collinear with the target wavefunction."
```
stylewarning: put `format` on its own line

src/compressor/compressor.lisp
stylewarning: absolutely positively document each of these condition types.

stylewarning: also export them

src/compressor/compressor.lisp
```diff
@@ -449,6 +449,22 @@ other's."
         ;; otherwise, we're obligated to do full unitary compilation.
         (decompile-instructions-into-full-unitary)))))
 
+(defmacro assert-compiler-correctness (condition-name check-form &rest init-forms &key &allow-other-keys)
+  (check-type condition-name symbol)
+  `(unless ,check-form
+     (error ',condition-name ,@init-forms)))
+
+(define-condition quil-compression-error (simple-error) ())
+(define-condition quil-compression-matrices-disagree-error (quil-compression-error) ())
```
stylewarning: put slots on a new line

src/compressor/compressor.lisp
stylewarning: add some spacing between these definitions

src/compressor/compressor.lisp
notmgsk: can I just do `(:report "<...>")` here?

notmgsk: Can I just drop the lambda and use a string here?

notmgsk:
@stylewarning are you happy with these conditions being named `quil-compression-*` rather than `cl-quil-compression-*` or something else?

src/compressor/compressor.lisp
stylewarning: dont think so

stylewarning: no

src/compressor/compressor.lisp
```diff
@@ -449,6 +449,28 @@ other's."
         ;; otherwise, we're obligated to do full unitary compilation.
         (decompile-instructions-into-full-unitary)))))
 
+(defmacro assert-compiler-correctness (condition-name check-form &rest init-forms &key &allow-other-keys)
+  (check-type condition-name symbol)
+  `(unless ,check-form
+     (error ',condition-name ,@init-forms)))
+
+(define-condition quil-compression-error (error)
```
stylewarning: why not have `compiler-correctness-error` as the base class

stylewarning: or at least as a superclass

src/compressor/compressor.lisp
```diff
@@ -449,6 +449,28 @@ other's."
         ;; otherwise, we're obligated to do full unitary compilation.
         (decompile-instructions-into-full-unitary)))))
 
+(defmacro assert-compiler-correctness (condition-name check-form &rest init-forms &key &allow-other-keys)
```
stylewarning: we should move this function up somewhere so it's more accessible

stylewarning:
@notmgsk looks like you didn't push

notmgsk:
@ecpeterson sorry for asking you to review again. I mis-clicked.

notmgsk:
@ecpeterson also sorry for namedropping you above, during non-work hours.

notmgsk:
Changing this to work-in-progress; i.e. don't merge.

I haven't figured out how to test the individual conditions. I can test `compress-states-disagree-error` by providing two sets of instructions that do obviously different things. It's less obvious to me how to get, for example, `compression-matrices-disagree-error`: doing the same as above will cause a `matrices don't lie in the same projective class` error. If y'all have a suggestion for two instruction sequences that are in the same projective class but are not "close", I'm all ears.

ecpeterson:
The "projective class" error is only triggered when the leftmost columns of the two matrices aren't scalar multiples of one another. Some pair like `CNOT 0 1` and `CNOT 1 0` will pass that test but then fail the subsequent assert that their rescalings agree entrywise (which is, I think, what you're trying to catch).

PR #222, opened 2019-05-03
Saner version checking

notmgsk:
Previously, we said "1.2.0" was newer than "1.1.0" because we relied
on string= for kinda valid reasons.


app/src/versions.lisp
```diff
@@ -36,6 +36,10 @@
   :test #'string=
   :documentation "The git hash of the quilc repo.")
 
+(defun version-string-values (version &key (delimiter #\.))
```
stylewarning: delete if not used

PR #224, opened 2019-05-08
Move the DIMENSION slot from the GATE abstract base class into the PARAMETERIZED-GATE class

appleby:
I am not sure if this fix is semantically correct, but it fixed `make test` in the qvm repo :).

Note that there exists a wrapper function `cl-quil:make-permutation-gate`. As far as I can tell it's not used anywhere, but in theory the handful of individual calls to `(make-instance 'permutation-gate ...)` could be replaced with calls to `make-permutation-gate`, which handles setting `:dimension` for you.

Fixes rigetti/qvm#85

### DIFF
--- a/src/gates.lisp
+++ b/src/gates.lisp
@@ -104,10 +104,10 @@
 (defmethod initialize-instance :after ((gate simple-gate) &key)
   (with-slots (matrix) gate
     (assert (and (magicl:square-matrix-p matrix)
-                 (power-of-two-p (magicl:matrix-rows matrix)))
+                 (positive-power-of-two-p (magicl:matrix-rows matrix)))
             (matrix)
-            "MATRIX must be of dimension NxN where N is a power of two. ~
-             Actual dimensions are ~a x ~a."
+            "MATRIX must be square whose dimension is a positive power of two.~@
+             Actual dimensions are ~D x ~D."
             (magicl:matrix-rows matrix)
             (magicl:matrix-cols matrix))))
 
@@ -126,10 +126,10 @@
 
 (defmethod initialize-instance :after ((gate permutation-gate) &key)
   (with-slots (permutation) gate
-    (assert (power-of-two-p (length permutation))
+    (assert (positive-power-of-two-p (length permutation))
             (permutation)
-            "PERMUTATION length must be a power of two. ~@
-             PERMUTATION ~A has length ~a which is not a power of two."
+            "PERMUTATION length must be a positive power of two. ~@
+             PERMUTATION ~A has length ~D which is not a power of two."
             permutation (length permutation))
     (check-permutation permutation)))
 
@@ -179,9 +179,9 @@
 
 (defmethod initialize-instance :after ((gate parameterized-gate) &key)
   (with-slots (dimension) gate
-    (assert (power-of-two-p dimension)
+    (assert (positive-power-of-two-p dimension)
             (dimension)
-            "DIMENSION ~A must be a power of two."
+            "DIMENSION ~D must be a positive power of two."
             dimension)))
 
 (defmethod gate-matrix ((gate parameterized-gate) &rest parameters)
--- a/src/gates.lisp
+++ b/src/gates.lisp
@@ -152,14 +152,11 @@
          (bits (make-array n :element-type 'bit :initial-element 0)))
     (flet ((mark-bit (i)
              (cond
+               ((not (<= 0 i (1- n)))
+                (error "Invalid permutation ~A. Entry out of range [0, ~A): ~A" perm n i))
                ((zerop (sbit bits i)) (setf (sbit bits i) 1))
                (t (error "Invalid permutation ~A. Found duplicate entry: ~A" perm i)))))
-      ;; Mark all of the bits.
-      (mapc #'mark-bit perm)
-      ;; Check that they're all marked.
-      (dotimes (i n t)
-        (when (zerop (sbit bits i))
-          (error "Invalid permutation ~A. Missing entry: ~A" perm i))))))
+      (mapc #'mark-bit perm))))
 
 (defun make-permutation-gate (name &rest permutation)
   "Make a permutation gate from the given PERMUTATION."
--- a/src/gates.lisp
+++ b/src/gates.lisp
@@ -69,14 +69,12 @@
 ;;; made into a matrix with GATE-MATRIX.
 
 (defclass gate ()
-  ((dimension :initarg :dimension
-              :reader gate-dimension
-              :documentation "The minimal dimension of the space the gate acts on.")
-   ;; TODO move name into the gate-definition class?
+  (;; TODO move name into the gate-definition class?
    ;;
    ;; Eric sez that the name is useful for tracking anonymous gates.
    (name :initarg :name
          :reader gate-name
+         :type (or null string)
          :documentation "The name of the gate."))
   (:metaclass abstract-class)
   (:default-initargs :name nil)
@@ -103,6 +101,19 @@
            :documentation "The matrix of the gate."))
   (:documentation "Non-parameterized gate represented as a dense matrix."))
 
+(defmethod initialize-instance :after ((gate simple-gate) &key)
+  (with-slots (matrix) gate
+    (assert (and (magicl:square-matrix-p matrix)
+                 (power-of-two-p (magicl:matrix-rows matrix)))
+            (matrix)
+            "MATRIX must be of dimension NxN where N is a power of two. ~
+             Actual dimensions are ~a x ~a."
+            (magicl:matrix-rows matrix)
+            (magicl:matrix-cols matrix))))
+
+(defmethod gate-dimension ((gate simple-gate))
+  (magicl:matrix-rows (simple-gate-matrix gate)))
+
 (defmethod gate-matrix ((gate simple-gate) &rest parameters)
   (assert (null parameters))
   (simple-gate-matrix gate))
@@ -112,6 +123,18 @@
                 :reader permutation-gate-permutation
                 :documentation "The permutation representation of the gate."))
   (:documentation "A gate which could be realized as a permutation matrix."))
+
+(defmethod initialize-instance :after ((gate permutation-gate) &key)
+  (with-slots (permutation) gate
+    (assert (power-of-two-p (length permutation))
+            (permutation)
+            "PERMUTATION length must be a power of two. ~@
+             PERMUTATION ~A has length ~a which is not a power of two."
+            permutation (length permutation))
+    (check-permutation permutation)))
+
+(defmethod gate-dimension ((gate permutation-gate))
+  (length (permutation-gate-permutation gate)))
 
 (defmethod gate-matrix ((gate permutation-gate) &rest parameters)
   (assert (null parameters))
@@ -138,28 +161,28 @@
         (when (zerop (sbit bits i))
           (error "Invalid permutation ~A. Missing entry: ~A" perm i))))))
 
-(defun make-permutation-gate (name documentation &rest permutation)
-  "Make a permutation gate with the permut"
-  (check-type name (or nil string))
-  (check-type documentation (or null string))
-  (assert (zerop (logand (length permutation)
-                         (1- (length permutation))))
-          (permutation)
-          "Permutation ~A must be of a power-of-two number of elements."
-          permutation)
-  (check-permutation permutation)
+(defun make-permutation-gate (name &rest permutation)
+  "Make a permutation gate from the given PERMUTATION."
   (make-instance
    'permutation-gate
    :name name
-   :documentation documentation
-   :dimension (length permutation)
    :permutation permutation))
 
 (defclass parameterized-gate (dynamic-gate)
-  ((matrix-function :initarg :matrix-function
+  ((dimension :initarg :dimension
+              :reader gate-dimension
+              :documentation "The minimal dimension of the space the gate acts on.")
+   (matrix-function :initarg :matrix-function
                     :reader parameterized-gate-matrix-function
                     :documentation "Function mapping ARITY complex numbers to a DIMENSION x DIMENSION MAGICL matrix."))
   (:documentation "A gate parameterized by complex numbers."))
+
+(defmethod initialize-instance :after ((gate parameterized-gate) &key)
+  (with-slots (dimension) gate
+    (assert (power-of-two-p dimension)
+            (dimension)
+            "DIMENSION ~A must be a power of two."
+            dimension)))
 
 (defmethod gate-matrix ((gate parameterized-gate) &rest parameters)
   (apply (parameterized-gate-matrix-function gate) parameters))
@@ -284,16 +307,13 @@
          (dim (isqrt (length entries))))
     (make-instance 'simple-gate
                    :name name
-                   :dimension dim
                    :matrix (make-row-major-matrix dim dim entries))))
 
 (defmethod gate-definition-to-gate ((gate-def permutation-gate-definition))
-  (let* ((name (gate-definition-name gate-def))
-         (entries (permutation-gate-definition-permutation gate-def))
-         (dim (length entries)))
+  (let ((name (gate-definition-name gate-def))
+        (entries (permutation-gate-definition-permutation gate-def)))
     (make-instance 'permutation-gate
                    :name name
-                   :dimension dim
                    :permutation entries)))
 
 (defmethod gate-definition-to-gate ((gate-def parameterized-gate-definition))
@@ -430,9 +450,7 @@ as matrices."
                                            (application-arguments instr)))
                  u))))
     (make-instance 'gate-application
-                   :gate (make-instance 'simple-gate
-                                        :dimension (expt 2 (length qubits))
-                                        :matrix u)
+                   :gate (make-instance 'simple-gate :matrix u)
                    :operator (named-operator (format nil "FUSED-GATE-~D"
                                                      (incf-premultiplied-gate-count)))
                    :arguments (mapcar #'qubit qubits))))
--- a/src/utilities.lisp
+++ b/src/utilities.lisp
@@ -129,6 +129,6 @@ Return two values:
   "Compute integer logarithm of X to the base 2."
   (1- (integer-length x)))
 
-(defun power-of-two-p (n)
+(defun positive-power-of-two-p (n)
   "Given an INTEGER N, return true if N is a power of 2, greater than 1."
-  (zerop (logand n (1- n))))
+  (and (> n 1) (= 1 (logcount n))))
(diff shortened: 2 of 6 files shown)


notmgsk:
I imagine this would've been caught if the test coverage for `src/gates.lisp` wasn't poor.

appleby:
What about `:name`? Ok to be `nil` in these cases, or should I provide a value in the call to `make-instance`?

appleby:
Pushed a commit that initializes the `permutation-gate`'s dimension based on the permutation's length in `initialize-instance`. Couple of follow-up comments.

1) I limited this change to `permutation-gate`s, but it looks like `simple-gate` might also be a candidate? At least based on how `simple-gate`s are used in the test suite, it seems to always be the case that

```
(with-slots (dimension matrix) some-simple-gate
  (= dimension (magicl:matrix-rows matrix) (magicl:matrix-cols matrix)))
```

2) I was too timid to tinker with the class hierarchy, especially since these types are part of the exported API, but maybe `permutation-gate`s should not have a `dimension` slot, and instead just `(defmethod gate-dimension ...)` similar to `controlled-gate`, `dagger-gate`, etc. I gather from the fact that only `:reader`s are defined for these slots that they are meant to be read-only after initialization, so I guess `setf`ing the slot-values directly is a mostly theoretical concern, but it still seems a little awkward that `permutation-gate`s now have a vestigial `:dimension` initarg.

src/gates.lisp
```diff
@@ -113,6 +113,16 @@
                 :documentation "The permutation representation of the gate."))
   (:documentation "A gate which could be realized as a permutation matrix."))
 
+(defmethod initialize-instance :after ((gate permutation-gate) &key)
+  (with-slots (dimension permutation) gate
+    (let ((permutation-length (length permutation)))
+      (when (and (slot-boundp gate 'dimension)
+                 (/= dimension permutation-length))
+        (warn "Ignoring invalid :dimension specified when initializing ~a: ~
```
appleby: This is awkward, but as far as I can tell there is no way to "delete" the `:dimension` initarg that `permutation-gate` inherits from `gate`, so the best I could think to do is warn in this case. For example, even if you add a bare `dimension` slot-specifier to `permutation-gate`'s `defclass`, the [hyperspec](http://www.lispworks.com/documentation/HyperSpec/Body/07_ab.htm) says that:

> the set of valid initialization arguments that fill slots for a class is the union of the initialization arguments that fill slots declared as valid by that class and its superclasses.

src/gates.lisp
```diff
@@ -138,21 +148,17 @@
         (when (zerop (sbit bits i))
           (error "Invalid permutation ~A. Missing entry: ~A" perm i))))))
 
-(defun make-permutation-gate (name documentation &rest permutation)
```
appleby: The symbol `make-permutation-gate` is exported. Normally, I would not remove a required parameter from an exported function, but in this case `:documentation` is not a valid initarg for `permutation-matrix`, so calling this function previously resulted in an `sb-pcl::initarg-error`.

src/gates.lisp
```diff
@@ -113,6 +113,16 @@
                 :documentation "The permutation representation of the gate."))
   (:documentation "A gate which could be realized as a permutation matrix."))
 
+(defmethod initialize-instance :after ((gate permutation-gate) &key)
+  (with-slots (dimension permutation) gate
+    (let ((permutation-length (length permutation)))
```
appleby: Should we `(check-permutation permutation)` or `(assert (power-of-two-p permutation-length))` here, as is done in `make-permutation-gate`?

stylewarning: Yes, I think so. Then we can nix the stuff in `make-perm...`

src/gates.lisp
stylewarning: I wouldn't do this. If someone supplies a dimension, I'd respect their wishes. We do with with the `:amplitudes` slot of the `qvm`. In your other comment though you're right, we should not have `dimension` be a slot. Maybe we ought to try to fix that in this PR and add the generic function. WDYT?

We don't necessarily need to change the class hierarchy, just move around the slots.

appleby: I'm fine with making this change as long as maintainers are fine with a little extra review hand-holding to make sure I don't break something. I assume no one wants to review stuff over the weekend anyway, so I will take some time to familiarize myself with the code and maybe write some additional tests and hopefully push an update early next week. Based on quick glance, the number of affected classes  looks small. I just need to figure out where the `dimension` slot should be moved.

src/gates.lisp
stylewarning: This is OK.

src/gates.lisp
stylewarning: I'm happy for this to happen; shouldn't be hard to not totally break everything. It would be perfectly OK for the slot to be duplicated in the classes if that storage is actually needed. Often it's not needed, and can be inferred.

appleby: OK, I will push an update with these changes in the next few days.

src/gates.lisp
```diff
@@ -82,6 +79,10 @@
   (:default-initargs :name nil)
   (:documentation "Abstract class for gates."))
 
+(defmethod initialize-instance :after ((gate gate) &key)
```
stylewarning: I would nix this and add `:type (or null string)` to the definition. This can be spooky if the implementation doesn't check for it, but I don't like using `initialize-instance` for _just_ type checks.

src/gates.lisp
```diff
@@ -103,6 +104,19 @@
            :documentation "The matrix of the gate."))
   (:documentation "Non-parameterized gate represented as a dense matrix."))
 
+(defmethod initialize-instance :after ((gate simple-gate) &key)
```
stylewarning: I am, however, OK with this checking.

src/gates.lisp
```diff
@@ -103,6 +104,19 @@
...
+  (with-slots (matrix) gate
+    (assert (and (magicl:square-matrix-p matrix)
+                 (power-of-two-greater-than-1-p (magicl:matrix-rows matrix)))
+            (matrix)
+            "MATRIX must be of dimension NxN where N is a power of two greater than one. ~@
```
stylewarning: `MATRIX must be square whose dimension is a positive power of two.`

src/gates.lisp
```diff
@@ -113,6 +127,18 @@
                 :documentation "The permutation representation of the gate."))
   (:documentation "A gate which could be realized as a permutation matrix."))
 
+(defmethod initialize-instance :after ((gate permutation-gate) &key)
+  (with-slots (permutation) gate
+    (assert (power-of-two-greater-than-1-p (length permutation))
+            (permutation)
+            "PERMUTATION length must be a power of two greater than one. ~@
```
stylewarning: "... positive power of two"

src/gates.lisp
```diff
@@ -129,38 +155,35 @@
          (bits (make-array n :element-type 'bit :initial-element 0)))
     (flet ((mark-bit (i)
              (cond
+               ((or (< i 0) (>= i n))
```
stylewarning: replace with `(not (<= 1 i (1- n)))`

src/gates.lisp
```diff
@@ -129,38 +155,35 @@
...
+                (error "Invalid permutation ~A. Entry out of range [0, ~A): ~A" perm n i))
```
stylewarning: Prefer `~D` for integers, here and elsewhere.

src/gates.lisp
```diff
@@ -129,38 +155,35 @@
...
-  "Make a permutation gate with the permut"
-  (check-type name (or nil string))
-  (check-type documentation (or null string))
-  (assert (zerop (logand (length permutation)
-                         (1- (length permutation))))
-          (permutation)
-          "Permutation ~A must be of a power-of-two number of elements."
-          permutation)
-  (check-permutation permutation)
+      (mapc #'mark-bit perm))))
```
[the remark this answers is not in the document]
stylewarning: Understandable & I agree to not replace.

src/gates.lisp
```diff
@@ -103,6 +104,19 @@
...
+  (with-slots (matrix) gate
+    (assert (and (magicl:square-matrix-p matrix)
+                 (power-of-two-greater-than-1-p (magicl:matrix-rows matrix)))
+            (matrix)
+            "MATRIX must be of dimension NxN where N is a power of two greater than one. ~@
+             Actual dimensions are ~a x ~a."
+            (magicl:matrix-rows matrix)
+            (magicl:matrix-cols matrix))))
+
+(defmethod gate-dimension ((gate simple-gate))
```
stylewarning: Can we `defgeneric` this around the definition of `gate`?

src/gates.lisp
```diff
@@ -430,9 +450,7 @@ as matrices."
                                            (application-arguments instr)))
                  u))))
     (make-instance 'gate-application
-                   :gate (make-instance 'simple-gate
-                                        :dimension (expt 2 (length qubits))
```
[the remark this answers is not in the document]
stylewarning: it should be guaranteed when the gate-definition is created

src/utilities.lisp
```diff
@@ -129,6 +129,6 @@ Return two values:
   "Compute integer logarithm of X to the base 2."
   (1- (integer-length x)))
 
-(defun power-of-two-p (n)
+(defun power-of-two-greater-than-1-p (n)
```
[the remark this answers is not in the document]
stylewarning: I would name this `positive-power-of-two-p`, which, while slightly ambiguous, rolls off the tongue nicer

tests/misc-tests.lisp
```diff
@@ -231,3 +231,37 @@
     (is (= (* 11 11) (quil::chip-spec-n-links chip)))
     (is (plusp (length (parsed-program-executable-code prgm))))
     (is (plusp (length (parsed-program-executable-code (compiler-hook prgm chip)))))))
+
+(deftest test-power-of-two-greater-than-1-p ()
+  "Test that POWER-OF-TWO-GREATER-THAN-1-P does what it says on the tin."
+  (is (not (quil::power-of-two-greater-than-1-p -2)))
+  (is (not (quil::power-of-two-greater-than-1-p -1)))
+  (is (not (quil::power-of-two-greater-than-1-p 0)))
+  (do ((power-of-two 2 (* 2 power-of-two)))
```
stylewarning: whoa there cowboy, `do` is old school! Use `loop`
```
(loop :for power-of-two := 2 :then (* 2 power-of-two)
      :while ...
      :do ...)
```

src/utilities.lisp
```diff
@@ -129,6 +129,6 @@ Return two values:
...
   "Given an INTEGER N, return true if N is a power of 2, greater than 1."
-  (zerop (logand n (1- n))))
+  (and (> n 1) (= 1 (logcount n))))
```
stylewarning: note that `1 = logcount` is less efficient

src/gates.lisp
appleby: I'm not sure I understand the question, but there is a `(defgeneric gate-dimension ...)` near the top of the file.

tests/misc-tests.lisp
appleby: Heh. I saw that you advocated for `loop` --> `dotimes` in some other PR, and incorrectly assumed you were an anti-loop greybeard, when in fact you're probably just pro-dotimes/dolist.

src/utilities.lisp
appleby: > note that `1 = logcount` is less efficient

It's interesting you say that. `logand` probably is faster in general, but I assumed `logcount` would be faster because 1) it avoids the `(1- n)` and 2) I was taking for granted the awesomeness of SBCL on x86_64, which [defines a VOP](https://github.com/sbcl/sbcl/blob/cd7af0d5b15e98e21ace8ef164e0f39019e5ed4b/src/assembly/x86-64/arith.lisp#L291) that essentially compiles `logcount` down to a "single" `POPCNT` instruction (modulo fixnum tagging/untagging and a [workaround](https://www.pvk.ca/Blog/2014/08/16/how-to-define-new-intrinsics-in-sbcl/) for an [x86 POPCNT data-dependency bug](https://stackoverflow.com/a/25089720/3761797) on some processors). All told, the call to `logcount` for a fixnum looks to be ~ 22 instructions on my machine, according to `disassemble`.

At least on my crusty old laptop, there doesn't appear to be much difference between the two. The small delta is likely noise for this ghetto-benchmark.

```
(let ((n (random 1000)))
  (time (loop :repeat 1e10 :do (= 1 (logcount n))))
  (time (loop :repeat 1e10 :do (zerop (logand n (1- n))))))
Evaluation took:
  6.669 seconds of real time
  6.668647 seconds of total run time (6.666441 user, 0.002206 system)
  100.00% CPU
  16,852,062,060 processor cycles
  0 bytes consed

Evaluation took:
  6.677 seconds of real time
  6.676523 seconds of total run time (6.674737 user, 0.001786 system)
  100.00% CPU
  16,872,337,229 processor cycles
  0 bytes consed

NIL
```

Of course, for other compilers / processor architectures, all bets are off. Even SBCL falls back to an assembly routine if your CPU doesn't have `POPCNT`.

Anyway, I originally switched to `logcount` because it gives the correct answer for N=0, before I noticed the "N > 1" restriction in the docstring and added the `(> n 1)` clause. But since the function now checks for n > 1 explicitly, I guess the point is moot and I'm happy to revert to the `logand` version if you prefer.

appleby: Hmm... perhaps my "benchmark" is even more useless than I thought...

```
CL-USER> (time (loop :repeat 1e10))
Evaluation took:
  6.742 seconds of real time
  6.741053 seconds of total run time (6.740694 user, 0.000359 system)
  99.99% CPU
  17,034,776,501 processor cycles
  0 bytes consed
  
NIL
```

appleby: Found some [random benchmarking library](https://github.com/scymtym/sb-benchmarks) on github. Timing results are still jittery, but at least this version runs the test multiple times and reports the median results. I modified it slightly to also report the variance, which is so absurdly small it's probably meaningless.

According to this benchmark, both `logand` and `logcount` are in the same ballpark as `identity`, ~10-ish nanoseconds, which is a little hard to believe considering that presumably includes funcall overhead, bignum/fixnum dispatching, etc! Maybe the compiler is "sufficiently smart" to figure out that `n` is constant in these tests and it all gets constant-folded away...

```
CL-USER> (load "benchmark-util.lisp")
T
CL-USER> (let ((*load-pathname* #p"/tmp/slime-repl"))
	   (benchmark-util:with-benchmark (:name :identity :parameters ((n (0 1 7 1234))))
	     (benchmark-util:measuring (identity n)))
	   (benchmark-util:with-benchmark (:name :logcount :parameters ((n (0 1 7 1234))))
	     (benchmark-util:measuring (logcount n)))
	   (benchmark-util:with-benchmark (:name :logand :parameters ((n (0 1 7 1234))))
	     (benchmark-util:measuring (logand n (1- n)))))
::: Running :IDENTITY
:::   Running (N 0)
:::   Done (N 0) => 6.6757e-9 (+/- 1.2060e-17) s; 0 bytes consed; 340 bytes code [5 measurements]
:::   Running (N 1)
:::   Done (N 1) => 6.6757e-9 (+/- 3.0923e-19) s; 0 bytes consed; 340 bytes code [5 measurements]
:::   Running (N 7)
:::   Done (N 7) => 6.1989e-9 (+/- 5.4570e-19) s; 0 bytes consed; 340 bytes code [5 measurements]
:::   Running (N 1234)
:::   Done (N 1234) => 6.6757e-9 (+/- 2.4193e-18) s; 0 bytes consed; 340 bytes code [5 measurements]
::: Done :IDENTITY
::: Running :LOGCOUNT
:::   Running (N 0)
:::   Done (N 0) => 9.5367e-9 (+/- 5.2387e-18) s; 0 bytes consed; 23 bytes code [5 measurements]
:::   Running (N 1)
:::   Done (N 1) => 9.5367e-9 (+/- 2.5102e-18) s; 0 bytes consed; 23 bytes code [5 measurements]
:::   Running (N 7)
:::   Done (N 7) => 1.0490e-8 (+/- 1.9645e-18) s; 0 bytes consed; 23 bytes code [5 measurements]
:::   Running (N 1234)
:::   Done (N 1234) => 9.5367e-9 (+/- 8.5493e-19) s; 0 bytes consed; 23 bytes code [5 measurements]
::: Done :LOGCOUNT
::: Running :LOGAND
:::   Running (N 0)
:::   Done (N 0) => 1.4305e-8 (+/- 1.9645e-18) s; 0 bytes consed; 48 bytes code [5 measurements]
:::   Running (N 1)
:::   Done (N 1) => 1.4305e-8 (+/- 5.0932e-19) s; 0 bytes consed; 48 bytes code [5 measurements]
:::   Running (N 7)
:::   Done (N 7) => 1.4305e-8 (+/- 5.0932e-19) s; 0 bytes consed; 48 bytes code [5 measurements]
:::   Running (N 1234)
:::   Done (N 1234) => 1.4305e-8 (+/- 1.2369e-18) s; 0 bytes consed; 48 bytes code [5 measurements]
::: Done :LOGAND
NIL
```

appleby: Perhaps not constant-folded? The reported runtimes are shockingly close to the above, but the "bytes code" metric implies the above funcalls perhaps weren't completely optimized away.... Benchmarking optimizing compilers is hard, let's go shopping!

```
CL-USER> (let ((*load-pathname* #p"/tmp/slime-repl"))
	   (benchmark-util:with-benchmark (:name :identity)
	     (benchmark-util:measuring (identity 1)))
	   (benchmark-util:with-benchmark (:name :logcount)
	     (benchmark-util:measuring (logcount 1)))
	   (benchmark-util:with-benchmark (:name :logand)
	     (benchmark-util:measuring (logand 1 0))))
::: Running :IDENTITY
::: Done :IDENTITY => 7.6294e-9 (+/- 9.3133e-18) s; 0 bytes consed; 126 bytes code [5 measurements]
::: Running :LOGCOUNT
::: Done :LOGCOUNT => 7.6294e-9 (+/- 1.6371e-18) s; 0 bytes consed; 19 bytes code [5 measurements]
::: Running :LOGAND
::: Done :LOGAND => 7.6294e-9 (+/- 1.6735e-18) s; 0 bytes consed; 16 bytes code [5 measurements]
NIL
```

stylewarning: I'm ok with either implementation, and if I'm honest with myself, I don't actually know for certain which is more efficient.

src/gates.lisp
stylewarning: Ah, ok, missed that.

appleby:
> I conditionally approve on there being a PR to fix QVM issues.

I have a QVM diff ready that was just waiting on approval here. I will open a PR later after a final rebase and testing.

PR #227, opened 2019-05-10
Provide --proxy and catch more usocket errors when checking for update

notmgsk:
The changes here are three:

1. Catch `USOCKET:NS-ERROR` and `USOCKET:SOCKET-ERROR` when checking for an SDK update. Originally we caught `USOCKET:NS-HOST-NOT-FOUND-ERROR`, but this allowed other related errors through, causing the app to crash.
2. Provide `--proxy` to be used by `quilc::latest-sdk-version` and those behind network restrictions. This one makes me slightly uncomfortable, since it may be confused with providing a proxy for the server itself.
3. Replace the off-by-default `--skip-version-check` with the off-by-default `--check-sdk-version`. Network issues may cause the app to crash or be slow while waiting for the update check to finish; IMO this should be opt-in. Also, this option now causes the program to exit (with an appropriate return code).


app/src/entry-point.lisp
```diff
@@ -75,7 +75,9 @@
     (("log-level") :type string :optional t :initial-value "info" :documentation "maximum logging level (\"debug\", \"info\", \"notice\", \"warning\", \"err\", \"crit\", \"alert\", or \"emerg\") (default \"info\")")
     (("quiet") :type boolean :optional t :initial-value nil :documentation "Disable all non-logging output (banner, etc.)")
     #+forest-sdk
-    (("skip-version-check") :type boolean :optional t :initial-value nil :documentation "Do not check for a new SDK version at launch.")))
+    (("check-sdk-version") :type boolean :optional t :initial-value nil :documentation "Check for a new SDK version at launch.")
+    #+forest-sdk
```
stylewarning: let's nix the use of `forest-sdk`.

app/src/entry-point.lisp
```diff
@@ -290,6 +287,19 @@ Version ~A is available from https://www.rigetti.com/forest~%"
                                   (cl-syslog:syslog-log-writer "quilc" :local0)
                                   *error-output*)))
 
+  #+forest-sdk
```
stylewarning: same thing about #+forest-sdk here

app/src/versions.lisp
```diff
@@ -36,23 +36,33 @@
...
+                                     :want-stream t
+                                     :proxy proxy))
              (p (yason:parse s)))
         (multiple-value-bind (version success)
             (gethash "quilc" p)
           (when success
             version)))
-    (usocket:ns-host-not-found-error (condition)
-      (declare (ignore condition))
+    (usocket:ns-error (condition)
```
stylewarning: cool!

stylewarning:
@notmgsk please fix merge conflicts

PR #234, opened 2019-05-15
More thoroughly validate a permutation at parse time

notmgsk:
Test not only that the permutation is a power-of-two (in size) but that it truly is a permutation. Truly truly.

### DIFF
--- a/src/parser.lisp
+++ b/src/parser.lisp
@@ -913,7 +913,7 @@ INPUT-STRING that triggered the condition."
 
 (defun gate-permutation-p (entries)
   (and (power-of-two-p (length entries))
-       (equal (sort entries #'<)
+       (equal (sort (copy-list entries) #'<)
               (alexandria:iota (length entries)))))
 
 (defun parse-gate-entries-as-matrix (body-lines params name)
--- a/src/parser.lisp
+++ b/src/parser.lisp
@@ -904,15 +904,17 @@ INPUT-STRING that triggered the condition."
 (defun parse-gate-entries-as-permutation (body-lines name)
   (multiple-value-bind (parsed-entries rest-lines)
       (parse-permutation-gate-entries body-lines)
-    (validate-gate-permutation-size (length parsed-entries))
+    (unless (gate-permutation-p parsed-entries)
+      (quil-parse-error "Permutation gate entries do not represent a square matrix."))
     (values (make-instance 'permutation-gate-definition
                            :name name
                            :permutation parsed-entries)
             rest-lines)))
 
-(defun validate-gate-permutation-size (size)
-  (unless (positive-power-of-two-p size)
-    (quil-parse-error "Permutation gate entries do not represent a square matrix.")))
+(defun gate-permutation-p (entries)
+  (and (power-of-two-p (length entries))
+       (equal (sort entries #'<)
+              (alexandria:iota (length entries)))))
 
 (defun parse-gate-entries-as-matrix (body-lines params name)
   ;; Parse out the gate entries.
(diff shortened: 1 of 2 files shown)


appleby:
Just a heads-up that this change will have merge conflicts with the changes in #224. Fine to merge these in either order though as the conflicts should be easy to resolve.

ecpeterson:
I don't think there's a rush to merge this: it's a nice-to-have connected to a pyQuil PR, and the pyQuil PR ought to be merged very soon, but this doesn't block it.

src/parser.lisp
```diff
@@ -899,15 +899,17 @@ INPUT-STRING that triggered the condition."
...
     (values (make-instance 'permutation-gate-definition
                            :name name
                            :permutation parsed-entries)
             rest-lines)))
 
-(defun validate-gate-permutation-size-p (size)
-  (power-of-two-p size))
+(defun gate-permutation-p (entries)
+  (and (power-of-two-p (length entries))
+       (equal (sort entries #'<)
```
stylewarning: `sort` is a mutation. Use `(sort (copy-list entries ...`

appleby:
> Don't we test this validation at permutation construction time?

Once #224 is merged, the permutation will be checked via `check-permutation` in `initialize-instance` when creating a `permutation-gate`. Though I suppose reporting errors earlier is better.

As far as I can tell, there is no corresponding `initialize-instance` check for `permutation-gate-definition`s, though this appears to be the only place one is created with an unchecked permutation. (The only other place `(make-instance 'permutation-gate-definition ...)` is called, it's guaranteed that the permutation is valid).

notmgsk:
@appleby Maybe we can delete this PR then. I think validation of non-permutation gates is also deferred, so makes sense that permutation gates would have the same behaviour (i.e. an invalid permutation doesn't raise an error at parse time).

What do you think?

appleby:
> @appleby Maybe we can delete this PR then. I think validation of non-permutation gates is also deferred, so makes sense that permutation gates would have the same behaviour (i.e. an invalid permutation doesn't raise an error at parse time).
> 
> What do you think?

In my personal opinion, it's good to catch and report errors as early as reasonably possible. For example, the checks I added in #224 will only happen at gate construction time, so a `(quil:parse-quil some-quil-string-with-bogus-permutation)` will not complain.

Now in this case you could argue that a *parse* error should be limited to checking syntax, not semantics, and whether or not a list of integers represents a valid permutation is kind of on the border.

But checking for a permutation is not super expensive, so I think it's fine to do here. Another option would be to instead do the check in `initialize-instance` when initializing the `permutation-gate-definition`. That would have the advantage of catching the error both at parse time any other place someone tried to create a `permutation-gate-definition` with an invalid permutation.

I guess the `initialize-instance` could be practically identical to the one for `permutation-gate`:

```common-lisp
(defmethod initialize-instance :after ((gate permutation-gate) &key)
  (with-slots (permutation) gate
    (assert (positive-power-of-two-p (length permutation))
            (permutation)
            "PERMUTATION length must be a positive power of two. ~@
             PERMUTATION ~A has length ~D which is not a power of two."
            permutation (length permutation))
    (check-permutation permutation)))
```

appleby:
On further consideration, I guess it would be unlikely for someone to DEFGATE a permutation and then never actually apply the gate anywhere in their program. And, as you say, there is precedent since regular gate matrices are only checked for basic dimension sanity. Plus, I noticed that none of the other *GATE-DEFINTION classes do any validation of their slot initializations, so maybe this is overkill (or maybe they should all have checks).

There is also a balance to be struck between defensive programming and duplicate checks everywhere. I personally prefer to err on the side of caution and add duplicate checks unless I'm 100% sure no one will ever pass me a bad value, but I'm the nervous type.

What I'm trying to say is: if you want to drop this PR I think that's fine :).

ecpeterson:
I have a mild allergy to constantly re-checking for bad input, but checking for bad input at pinch points (like construction time or parse time) suits me just fine. I think we should retain this.

notmgsk:
Closing in favour of validating permutation at construction time.

PR #237, opened 2019-05-15
Feature: forked gates

ecpeterson:
Uniformly controlled rolls (UCRs) are emitted by a variety of compilation subroutines, but they've been treated as second-class citizens because the number of qubits they occupy (and the number of parameters they consume) varies from instance to instance, which Quil is unfriendly towards.

This is also the case for controlled gates, which we recently added language-level support for. This PR extends the Quil language in a similar direction by providing the keyword `FORKED` to imbue a special case of a multiplexed operator: `FORKED GATE(a1, ..., an, b1, ..., bn) c qm ... q0` indicates the block-sum of `GATE(a1, ..., an) qm ... q0` and `GATE(b1, ..., bn) qm ... q0` across the ground and excited states of qubit `c`. UCRs, then, are multiply `FORKED` rolls.

In addition to the language-level update, this PR includes a variety of internal updates that convert all our old uses of the UCR pseudoinstruction into legitimate uses of `FORKED`.

To-do:
* [x] Finish porting routines that emit UCRs to FORKED gates instead.
* [x] Address "chirality" in the expansion of FORKED expressions that specifically encode partially-expanded UCRs.
* [x] Robert's easter egg request.
* [x] Fix any test damage.


ecpeterson:
The two remaining tasks are: Sohaib's test breaks with state preparation turned on, and Robert's easter egg. Both of these will be minor, and the bulk of the changes that are going to be made have been made. This is ready for review.

cl-quil.asd
```diff
@@ -68,6 +68,7 @@
                (:module "compilers"
                 :serial t
                 :components ((:file "ucr-explode")
+                             #+ignore
```
stylewarning: why ignore? is it bit-rotting and you don't want to update it?

src/compilers/ucr-explode.lisp
```diff
@@ -7,117 +7,15 @@
 
 (in-package #:cl-quil)
 
+;; first, a helpful utility for constructing UCR gates
+(defun build-UCR (roll-name params &rest args)
```
stylewarning: this is a very eric-y function

src/compilers/ucr-explode.lisp
```diff
@@ -7,117 +7,15 @@
...
-                       ;; apply (* -1) to those indices j which bitmask with index to 1
-                       :when (logbitp (1- index) j)
-                         :do (setf (aref operating-vector j)
-                                   (- (aref operating-vector j))))
-                 (coerce operating-vector 'list))))))
-
-
-;;; now, the meat of UCR expansion
-
 ;; the idea here comes from the 2Q case: the single instruction
```
stylewarning: do we need to describe UCRs now that they've been obliterated?

src/compilers/ucr-explode.lisp
```diff
@@ -161,161 +59,210 @@
...
 ;;     X t, Z c, RY(pi/2) t, ISWAP c t, RY(r) c, ISWAP c t, RY(-pi/2) t
-;;          == UCRZ(r,-r) (c) t 
+;;          == UCRZ(r,-r) (c) t
 ;;
-(defun ucr-compiler (instr &key (target ':cz))
+
+(defun uniformly-controlled-roll-p (instr)
+  (labels ((op-walker (op)
+             (adt:match operator-description op
+               ((named-operator str) (if (or (string= str "RY") (string= str "RZ"))
```
stylewarning: add a `nil` branch to the `if`

src/compilers/ucr-explode.lisp
```diff
@@ -161,161 +59,210 @@
 ;;          == UCRY(r,-r) (c) t
 ;;
 ;;     X t, Z c, RY(pi/2) t, ISWAP c t, RY(r) c, ISWAP c t, RY(-pi/2) t
-;;          == UCRZ(r,-r) (c) t 
+;;          == UCRZ(r,-r) (c) t
 ;;
-(defun ucr-compiler (instr &key (target ':cz))
+
+(defun uniformly-controlled-roll-p (instr)
```
stylewarning: document that this returns the name of the gate if it is, otherwise nil

src/compilers/ucr-explode.lisp
```diff
@@ -161,161 +59,210 @@
...
-             (:down
-              (list (chiroll ':down averages) (cnot) (chiroll ':up differences)))
-             (:up
-              (list (chiroll ':down differences) (cnot) (chiroll ':up averages)))
-             ((nil)
+              (list avgroll (cnot) diffroll (cnot)))
+             ;; otherwise, we need to emit shorter UCRs encoded with CNOTs
+             ((member arch '(:cnot :cz))
+              ;; here's the obvious decomposition
+              #+ignore
```
stylewarning: `#+#:for-pedagogical-purposes` or so

src/compilers/ucr-explode.lisp
```diff
@@ -161,161 +59,210 @@
...
+              (list
+               (ucr averages rest)
+               (cnot)
+               (ucr differences rest)
+               (cnot))
+              ;; here's the non-obvious decomposition. in previous iterations of
+              ;; this trickery was called "chirality", but now we do this in-place.
```
stylewarning: maybe mention for future quilc archeologists what you meant by chirality?

src/compressor/compressor.lisp
```diff
@@ -364,39 +364,32 @@ other's."
...
-            (expand-to-native-instructions (append translation-results
-                                                   (rest instrs))
-                                           chip-specification
-                                           output-string))
-           (t
-            (error "Failed to expand ~a into native instructions."
-                   (print-instruction (first instrs) nil)))))))))
+         (when (and *compress-carefully*
+                    (not *enable-approximate-compilation*)
+                    (notany (lambda (instr) (typep instr 'state-prep-application))
```
stylewarning: could you document a little bit about why?

src/operator-bind.lisp
```diff
@@ -26,12 +25,30 @@
...
+                             ((forked-operator o)
+                              (recurse o)))))
+                  (recurse operator))))))
       (assert (not (null gate-def)) (operator) "BUILD-GATE only takes standard gate names.")
       (make-instance 'gate-application
-                     :operator (named-operator operator)
+                     :operator operator-adt
                      :name-resolution gate-def
-                     :gate (gate-definition-to-gate gate-def)
+                     :gate nil
```
stylewarning: I think (?) we can delete this line

src/ast.lisp
```diff
@@ -842,7 +842,8 @@ Each addressing mode will be a vector of symbols:
   ;; Note that reduction of consecutive dagger operators is not
   ;; performed here. Use the alternative constructor
   ;; INVOLUTIVE-DAGGER-OPERATOR.
-  (dagger-operator     operator-description))
```
stylewarning: for each operator description case, could you please provide a formal mathematical description, akin to the one you left as a comment  in this PR?

If you want to be super classy, after the DEFDATA, you can do

```
(setf (documentation 'named-operator 'function) "...")
(setf (documentation 'controlled-operator 'function) "...")
...
```

Please also reference the Quil syntax for these here.

Lastly (!) please improve the terrible docstring I wrote. `"A description of an operator that Quil takes."` is such junk.

cl-quil.asd
ecpeterson: correct

src/compilers/ucr-explode.lisp
ecpeterson: eric in, eric out

src/compilers/ucr-explode.lisp
ecpeterson: I think I'll rework the comment at the top of this function to explain all this & add a pointer to that here.

src/compressor/compressor.lisp
ecpeterson: I won't comment in the code, but I'm happy to comment here. This is a routine that escaped change during the conversion to `give-up-compilation`: it's checking for `NIL` as if that would signal total translation failure, whereas `NIL` now means that translation succeeded and gave a null-op. The only change here is to delete what's essentially a `(when translation-results ...)` guard, trusting that if there's an error it'll be thrown from within `apply-translation-compilers`.

cl-quil.asd
stylewarning: change it to `(:static-file "cs-compile")` and write a lil comment

src/compilers/ucr-explode.lisp
```diff
@@ -7,62 +7,248 @@
...
+  ;; if this isn't a UCR, skip it.
+  (adt:match operator-description (application-operator instr)
+    ((forked-operator op)
+     (let* ((roll-type (uniformly-controlled-roll-p instr))
+            (parameters (application-parameters instr))
+            (arguments (mapcar #'qubit-index (application-arguments instr)))
+            (control (first arguments))
+            (target (first (last arguments)))
+            (rest (rest arguments))
+            (pi/2 (constant (/ pi 2)))
```
stylewarning: i'd be down to just define these globally `(defconstant pi/2 (coerce (/ pi 2) 'double-float))` in utilities or w/e

ecpeterson: That would be very reasonable but I think it should be in a separate PR.

stylewarning: ok

PR #239, opened 2019-05-16
Add *protoquil* to bordeaux-threads:*default-special-bindings*

### LINKED ISSUE #238: RPCQ server does not respect `--protoquil` flag
When the RPCQ server is started with `--protoquil` present on the command line, it emits code as if the flag were absent. I suspect that this comes from a bad interaction between the RPCQ server's use of `bordeaux-threads`, which gives each thread its own dynamic context, and the recent merge of #212, which prefers dynamic binding to global state mutation. This specific bug would also be resolved by attending to #229, though by quite different means.

If we had more thorough testing of the application code, we would have caught this before merge.

notmgsk:
Items before merging:
- [x] Are we happy with this short-term approach (long-term is have these options configurable at run-time)
- [x] Are other variables missing too
- [x] Do the same for HTTP server? (or just let it rot)
- [ ] ~Tests~ Deferred to a later PR

Tagging @appleby since I can't add them as a reviewer directly.

Closes #238.

### DIFF
--- a/app/src/entry-point.lisp
+++ b/app/src/entry-point.lisp
@@ -268,10 +268,21 @@ Version ~A is available from https://www.rigetti.com/forest~%"
 
   (when (minusp time-limit)
     (error "A negative value (~D) was provided for the server time-limit." time-limit))
-
-  (let* ((*log-level* (or (and log-level (log-level-string-to-symbol log-level))
+  
+  (macrolet
+      ((special-bindings-let* (bindings &body body)
+         "Bind BINDINGS as in LET, and add those BINDINGS to BORDEAUX-THREADS:*DEFAULT-SPECIAL-BINDINGS* in the same LET."
+         `(let* (,@(loop :for (name value) :in bindings
+                         :collect `(,name ,value))
+                 (bordeaux-threads:*default-special-bindings*
+                   (list* ,@(loop :for (name value) :in bindings
+                                  :collect `(cons ',name ,name))
+                          bordeaux-threads:*default-special-bindings*)))
+            ,@body)))
+    (special-bindings-let*
+        ((*log-level* (or (and log-level (log-level-string-to-symbol log-level))
 			  *log-level*))
-	 (*logger* (make-instance 'cl-syslog:rfc5424-logger
+         (*logger* (make-instance 'cl-syslog:rfc5424-logger
 				  :app-name *program-name*
 				  :facility ':local0
 				  :maximum-priority *log-level*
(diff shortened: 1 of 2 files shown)


app/src/rpc-server.lisp
```diff
@@ -159,7 +159,9 @@
                            (port 5555)
                            (logger (make-instance 'cl-syslog:rfc5424-logger
                                                   :log-writer (cl-syslog:null-log-writer))))
-  (let ((dt (rpcq:make-dispatch-table)))
+  (let ((dt (rpcq:make-dispatch-table))
+        (bordeaux-threads:*default-special-bindings* (append bordeaux-threads:*default-special-bindings*
```
appleby: I think if this binding was done at the top of `quilc::process-options`, it would cover both web-server and rpc-server modes.

As a counter-point, I suppose it's nice to bind this as close to thread creation as possible.

Also not that when web-server mode is requested, `start-rpc-server` is actually called from inside a `make-thread`, so binding here might be too late for that case?

I'm also worried about other dynamic bindings that might need to be visible in the new threads, although fixing `*protoquil*` support is probably the most urgent.

ecpeterson: Good catch with the threaded behavior in pair-of-servers mode.

Unrelated to this PR, I wonder if `rpcq:start-server` should have a non-blocking variant that directly returns control to the parent thread.

notmgsk: OK. Moved the binding to earlier, immediately after other variables are bound and before anything thread-related happens.

ecpeterson:
Adding the other state variables is a good idea, and adding tests is a good idea, but neither are critical for getting 1.8 out the door.

app/src/entry-point.lisp
```diff
@@ -303,7 +303,9 @@ Version ~A is available from https://www.rigetti.com/forest~%"
          (*json-stream* (make-broadcast-stream))
          (*human-readable-stream* (make-broadcast-stream))
          (*quil-stream* (make-broadcast-stream))
-         (*verbose* (make-broadcast-stream)))
+         (*verbose* (make-broadcast-stream))
+         (bordeaux-threads:*default-special-bindings* (append bordeaux-threads:*default-special-bindings*
```
appleby: So I started trying to make a list of what other vars needed to be visible, then realized that both the web-server (in `compiler-post`) and the rpc server (in `quil-to-native-quil`) wind up calling `process-program`, which I think means that most (all?) of the above bindings need to be visible in the new threads.

Now in both cases, the servers appear to call `process-program` to get the `*statistics-dictionary*` that is returned, and I'm not certain that all the bindings are needed, but I guess it would be safest to err on the side of caution and include them.

Just spit-balling, not necessarily for this PR, but perhaps this entire `let*` could be replaced with a let-like macro that handles updating `*default-special-bindings*` automatically, to reduce boilerplate and the risk that someone later adds a binding here, but forgets to add it to `*default-special-bindings*`.

app/src/rpc-server.lisp
notmgsk: RE non-blocking, I think that is what I will need to be able to test this.

appleby:
Feels like I'm chiming in from the peanut gallery here a bit, but I'm also fine with whatever minimal fix the maintainers feel comfortable with, especially if you have an upcoming release you want to unblock.

As a follow-on, I can help with adding some tests, and possibly also figuring out a way to reduce boilerplate when specifying which bindings need to be added to `*default-special-bindings*`, in order to reduce the chance of future breakage.

appleby:
As a counter-point, if you have an upcoming release soon, I suppose the *safest* thing would be to revert the changes from #212, but I'm not sure how hard it would be to untangle if other stuff has been committed on top.

notmgsk:
@appleby your input is very much welcome and appreciated.

I'll tackle your boilerplate point in this PR, and we can return to it later if we need a better solution.

appleby:
@notmgsk Sounds good to me. And thanks for fixing my breakage!

stylewarning:
I think this should be seen as a somewhat quick fix, and we should also make a good effort to reduce the number of special variables. There are too many that don't actually need `special` treatment.

app/src/entry-point.lisp
```diff
@@ -268,10 +268,21 @@ Version ~A is available from https://www.rigetti.com/forest~%"
 
   (when (minusp time-limit)
     (error "A negative value (~D) was provided for the server time-limit." time-limit))
-
-  (let* ((*log-level* (or (and log-level (log-level-string-to-symbol log-level))
+  
+  (macrolet
```
stylewarning: I know I said `macrolet`, but I'd actually just prefer a global `defmacro`.

appleby:
Looks good to me. Note that `*program-name*` is also let-bound at the top of `%entry-point`, but as far as I can tell, it's never referenced in either of the server threads. But if you wind up making the macro global per @stylewarning's comment, then I guess it wouldn't hurt to use it for binding `*program-name*` as well, just to be safe.

appleby:
Here is a little test I wrote to ease my jangling nerves and validate that the protoquil flag is now respected after applying @notmgsk's changes from this PR. This is a long way from validating that all the bindings are respected, but it's a start.

I don't propose to actually include such a test in the test suite since it depends on an out-of-band "make" to build the quilc binary, just posting here in case this is a useful data-point. This test is based on the pre-existing `test-quil-roundtrip`, but modified to instead start quilc in a subprocess and compare the result of calling `quil-to-native-quil` both over the rpc socket and also via an in-process funcall.

``` common-lisp
(deftest test-quilc-subprocess-respects-protoquil-flag ()
  (let ((process-info (uiop:launch-program (list "./quilc" "-RP"))))
    (unwind-protect
        (rpcq:with-rpc-client (client "tcp://127.0.0.1:5555")
          (let* ((quilc::*protoquil* t)
                 (quil "H 0")
                 (isa (alexandria:plist-hash-table
                       (list
                        "1Q" (alexandria:plist-hash-table
                              (list
                               "0" (make-hash-table)))
                        "2Q" (make-hash-table))
                       :test #'equal))
                 (specs (make-hash-table))
                 (target-device (make-instance 'rpcq::|TargetDevice|
                                               :|isa| isa
                                               :|specs| specs))
                 (server-payload (make-instance 'rpcq::|NativeQuilRequest|
                                                :|quil| quil
                                                :|target_device| target-device))
                 (server-response (rpcq:rpc-call client "quil-to-native-quil" server-payload))
                 (local-response (quilc::quil-to-native-quil server-payload))
                 (lpp (quil::parse-quil (rpcq::|NativeQuilResponse-quil| local-response)))
                 (cpp (quil::parse-quil (rpcq::|NativeQuilResponse-quil| server-response))))
            (is (string= (with-output-to-string (s) (quil::print-parsed-program lpp s))
                         (with-output-to-string (s) (quil::print-parsed-program cpp s))))))
      (uiop:terminate-process process-info)
      (uiop:wait-process process-info))))
```

Prior to applying @notmgsk's changes, the test fails likes so (note the "HALT" at the end of RPC-server's version of the quil string).

```
CL-USER> (fiasco:run-tests 'quilc-tests::test-quilc-subprocess-respects-protoquil-flag)
TEST-QUILC-SUBPROCESS-RESPECTS-PROTOQUIL-FLAG.............................[FAIL]


Test run had 1 failure:

  Failure 1: FAILED-ASSERTION when running QUILC-TESTS::TEST-QUILC-SUBPROCESS-RESPECTS-PROTOQUIL-FLAG
    Binary predicate (STRING= X Y) failed.
    x: (WITH-OUTPUT-TO-STRING (QUILC-TESTS::S)
         (CL-QUIL::PRINT-PARSED-PROGRAM QUILC-TESTS::LPP QUILC-TESTS::S)) => "RZ(pi/2) 0
    RX(pi/2) 0
    RZ(pi/2) 0
    "
    y: (WITH-OUTPUT-TO-STRING (QUILC-TESTS::S)
         (CL-QUIL::PRINT-PARSED-PROGRAM QUILC-TESTS::CPP QUILC-TESTS::S)) => "RZ(pi/2) 0
    RX(pi/2) 0
    RZ(pi/2) 0
    HALT                                    # Exiting rewiring: #(0)
    "
NIL
(#<test-run of TEST-QUILC-SUBPROCESS-RESPECTS-PROTOQUIL-FLAG: 1 test, 1 assertion, 1 failure in 0.015 sec (1 failed assertion, 0 errors, none expected)>)
```

After apply the changes in this PR, the test passes.

PR #247, opened 2019-05-20
Clamp *program-name* to 48 chars

notmgsk:
Produces a name like `"<A>...<B>"` where `<A>` is a 15-char chunk from the
beginning of the name, and `<B>` is a 30-char chunk from the end of the
name.

### DIFF
--- a/app/src/entry-point.lisp
+++ b/app/src/entry-point.lisp
@@ -203,6 +203,17 @@
       (format *error-output* "~&! ! ! Error: ~A~%" c)
       (uiop:quit 1))))
 
+(defun clamp-syslog-name (name)
+  "Produces a name like \"<A>...<B>\" where <A> is a 15-char chunk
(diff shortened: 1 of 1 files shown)


notmgsk:
This again will lead to more code duplication since this needs to happen in qvm-app also.

app/src/utils.lisp
```diff
@@ -0,0 +1,10 @@
+(in-package :quilc)
+
+(defun clamp-syslog-name (name)
+  ""
```
ecpeterson: fill me out or delete me

notmgsk:
@ecpeterson I didn't mean to make this a "ready" PR. Mostly I wanted to prompt a discussion about duplication + whether this should live in cl-syslog. Good catch tho :D

notmgsk:
Just gonna go with having this duplicated in quilc and qvm for now. @stylewarning PLS REVIEW.

Issue #236, opened 2019-05-15
write more complete tests for stabilizer code


stylewarning:
The code for computing Clifford actions on the stabilizer group is woefully undertested. The main reason is that there's no function to compute the state vector from a stabilizer representation, so we can't test that behavior works for any and all Clifford elements.

See also issue #98.

PR #248, opened 2019-05-20
Add POWER-OF-TWO-P and call it in CLIFFORD-STABILIZER-ACTION

appleby:
POWER-OF-TWO-P was renamed in de7367c72b23a236ef16e37d2fa8218ac3396119, which was merged shortly after this code.

I am not sure how to test this function (see #236), but can confirm I no longer get a style-warning when compiling.

### DIFF
--- a/src/clifford/stabilizer.lisp
+++ b/src/clifford/stabilizer.lisp
@@ -274,7 +274,7 @@ but the symbols may be uninterned or named differently.
                         ;; record the X or Z variable in that bit's
                         ;; running list. In the end, the running list
                         ;; represents a disjunction of contributions.
-                        (when (power-of-two-p i) ; X and Z's will have power of two indexes.
+                        (when (quil::power-of-two-p i) ; X and Z's will have power of two indexes.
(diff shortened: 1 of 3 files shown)


src/clifford/stabilizer.lisp
```diff
@@ -274,7 +274,7 @@ but the symbols may be uninterned or named differently.
                         ;; record the X or Z variable in that bit's
                         ;; running list. In the end, the running list
                         ;; represents a disjunction of contributions.
-                        (when (power-of-two-p i) ; X and Z's will have power of two indexes.
+                        (when (quil::positive-power-of-two-p i) ; X and Z's will have power of two indexes.
```
appleby: Does this want a "positive power of two" or just any power of two?

I must admit I do not really understand what this code is doing, but something about the `let` binding for `var` just below this makes me suspect that perhaps it shouldn't ignore i = 1?

stylewarning: It should include `0`

appleby: Done.

PR #250, opened 2019-05-22
Load "contribs" during in build-app.lisp + do not hardcode shared lib paths

notmgsk:
Currently, to get access to the tweedledum compilation routine you must `(ql:quickload :cl-quil/tweedledum). We also want to load this during a `make quilc-sdk`, so it is available with `quilc < hella-big-permutation.quil`.

* [x] Load tweedledum in build-app.lisp
* [ ] Do not hardcode specific paths of shared libs

### DIFF
--- a/app/src/entry-point.lisp
+++ b/app/src/entry-point.lisp
@@ -193,6 +193,7 @@
   (disable-debugger))
 
 (defun entry-point (argv)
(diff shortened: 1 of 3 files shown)


appleby:
Leaving this here in case it's of some assistance:

https://sourceforge.net/p/sbcl/mailman/message/26164596/

notmgsk:
Closing in favour of #276.

PR #251, opened 2019-05-22
Respect final instrucion's wish to LINK-SUCCESSOR? in PROGRAM-CFG

### LINKED ISSUE #244: The slot CL-QUIL::OUTGOING is unbound in an instance of CL-QUIL::BASIC-BLOCK
I'm not sure this is even a valid quil program (is it required to terminate the program with a HALT?), but at least the error message could be better.

```
./quilc <<EOF
> DECLARE ro BIT[1]
> LABEL @START
> H 0
> MEASURE 0 ro[0]
> JUMP-UNLESS @START ro[0]
> EOF
! ! ! Error:
Error during printing.
The slot CL-QUIL::OUTGOING is unbound in an instance of CL-QUIL::BASIC-BLOCK.
```

Here is a minimal example with full backtrace.

```
./quilc-unsafe <<EOF
> DECLARE ro BIT[1]
> LABEL @START
> JUMP-UNLESS @START ro[0]
> EOF
Unhandled UNBOUND-SLOT in thread #<SB-THREAD:THREAD "main thread" RUNNING
                                    {10006185B3}>:
  The slot CL-QUIL::OUTGOING is unbound in the object
  #<BASIC-BLOCK START-597("START") len:0  in:1 out:#<error printing a CL-QUIL::BASIC-BLOCK: #<UNBOUND-SLOT OUTGOING {1005998403}>>.>

0: (SB-DEBUG::DEBUGGER-DISABLED-HOOK #<UNBOUND-SLOT OUTGOING {1005991E43}> #<unused argument> :QUIT T)
1: (SB-DEBUG::RUN-HOOK *INVOKE-DEBUGGER-HOOK* #<UNBOUND-SLOT OUTGOING {1005991E43}>)
2: (INVOKE-DEBUGGER #<UNBOUND-SLOT OUTGOING {1005991E43}>)
3: (ERROR #<UNBOUND-SLOT OUTGOING {1005991E43}>)
4: (SB-KERNEL:WITH-SIMPLE-CONDITION-RESTARTS ERROR NIL UNBOUND-SLOT :NAME CL-QUIL::OUTGOING :INSTANCE #<CL-QUIL::BASIC-BLOCK START-597("START") len:0  in:1 out:#<error printing a CL-QUIL::BASIC-BLOCK: #<UNBOUND-SLOT OUTGOING {10059A1B53}>>)
5: ((:METHOD SLOT-UNBOUND (T T T)) #<unused argument> #<CL-QUIL::BASIC-BLOCK START-597("START") len:0  in:1 out:#<error printing a CL-QUIL::BASIC-BLOCK: #<UNBOUND-SLOT OUTGOING {10059A50A3}>> CL-QUIL::OUTGOING) [fast-method]
6: (SB-PCL::SLOT-UNBOUND-INTERNAL #<CL-QUIL::BASIC-BLOCK START-597("START") len:0  in:1 out:#<error printing a CL-QUIL::BASIC-BLOCK: #<UNBOUND-SLOT OUTGOING {10059A7BC3}>> 2)
7: (CL-QUIL::CHILDREN #<CL-QUIL::BASIC-BLOCK START-597("START") len:0  in:1 out:#<error printing a CL-QUIL::BASIC-BLOCK: #<UNBOUND-SLOT OUTGOING {10059AA243}>>)
8: (CL-QUIL::ELIMINATE-DEAD-CODE #<CL-QUIL::CFG of 2 blocks {10056FE0D3}>)
9: (CL-QUIL::PROGRAM-CFG #<CL-QUIL:PARSED-PROGRAM {1004E31B33}> :DCE T :SIMPLIFY T)
10: (CL-QUIL:COMPILER-HOOK #<CL-QUIL:PARSED-PROGRAM {1004E31B33}> #<CL-QUIL::CHIP-SPECIFICATION of 8:8 objects> :PROTOQUIL NIL :REWIRING-TYPE NIL)
11: (QUILC::PROCESS-PROGRAM #<CL-QUIL:PARSED-PROGRAM {1004E31B33}> #<CL-QUIL::CHIP-SPECIFICATION of 8:8 objects>)
12: (QUILC::PROCESS-OPTIONS :PREFER-GATE-LADDERS NIL :COMPUTE-GATE-DEPTH NIL :COMPUTE-GATE-VOLUME NIL :COMPUTE-RUNTIME NIL :COMPUTE-FIDELITY NIL :COMPUTE-MATRIX-REPS NIL :COMPUTE-2Q-GATE-DEPTH NIL :COMPUTE-UNUSED-QUBITS NIL :SHOW-TOPOLOGICAL-OVERHEAD NIL :GATE-BLACKLIST NIL :GATE-WHITELIST NIL :WITHOUT-PRETTY-PRINTING NIL :PRINT-LOGICAL-SCHEDULE NIL :VERBOSE NIL :JSON-SERIALIZE NIL :ISA "8Q" :ENABLE-STATE-PREP-REDUCTIONS NIL :PROTOQUIL NIL :VERSION NIL :CHECK-LIBRARIES NIL :BENCHMARK NIL :SERVER-MODE-HTTP NIL :SERVER-MODE-RPC NIL :HOST "*" :PORT 5555 :TIME-LIMIT 0 :HELP NIL :LOG-LEVEL "info" :QUIET NIL :CHECK-SDK-VERSION NIL :PROXY NIL)
13: (QUILC::%ENTRY-POINT NIL)
14: ((LAMBDA NIL :IN MAKE-TOPLEVEL-FUNCTION))
15: ((FLET SB-UNIX::BODY :IN SAVE-LISP-AND-DIE))
16: ((FLET "WITHOUT-INTERRUPTS-BODY-14" :IN SAVE-LISP-AND-DIE))
17: ((LABELS SB-IMPL::RESTART-LISP :IN SAVE-LISP-AND-DIE))

```

If I slap a HALT onto the end (or `X 0` or whatever), then the error is avoided. The `JUMP @BLK-599` seemed suspicious to me since it appears to be JUMPing to an undefined LABEL, but I fed the output into QVM and it didn't complain, so it seems the bug is in my mental model, not the program.

```
./quilc <<EOF
> DECLARE ro BIT[1]
> LABEL @START
> H 0
> MEASURE 0 ro[0]
> JUMP-UNLESS @START ro[0]
> HALT
> EOF
DECLARE ro BIT

RZ(pi/2) 0                              # Entering rewiring: #(0 1 2 3 4 5 6 7)
RX(pi/2) 0
RZ(pi/2) 0
MEASURE 0 ro[0]
JUMP-WHEN @BLK-598 ro[0]
JUMP @BLK-599                           # Exiting rewiring: #(0 1 2 3 4 5 6 7)
LABEL @BLK-598                          # Entering rewiring: #(0 1 2 3 4 5 6 7)
```

appleby:
## Original commit message

When processing a JUMP-UNLESS or JUMP-WHEN instruction, PROCESS-INSTRUCTION will return a non-nil LINK-SUCCESSOR?, requesting the just-finished block to be linked to it's successor. Previously, if the conditional jump instruction was the final instruction in the program, successor linkage would not happen and the finished block's OUTGOING slot would be unbound. The compiler would later choke on said block, resulting in an UNBOUND-SLOT error.

With this change, ensure that successor linkage always happens, if requested, after PROCESS-INSTRUCTION returns.

Additionally, fix PRINT-OBJECT and the various routines for generating graphviz output to check for SLOT-BOUNDP before attempting to access the block's OUTGOING slot.

Fixes #244

## Additional comments

Note that this commit only focuses on fixing the "slot-unbound" error reported in #244. As @ecpeterson noted in the issue comments, there may be another bug in the compiled output (JUMP to undefined label), but I will address that in a separate PR when/if I'm able to determine that it's a genuine bug.

This is one approach to fixing the "slot-unbound" error reported in #244. I'm starting with this approach because it's the most localized change I could think of that would be least likely to break other things, but it may not be the "best" fix. Here are a couple of (untested) alternatives that come to mind. Feedback welcome.

1. Always canonicalize the instruction sequence at the start of compiler processing by slapping a HALT onto the end if one isn't there already. With this approach, no changes to PROGRAM-CFG would be required, because the program could never end in a conditional jump, so there would always be some succeeding block/instruction to process and link to. Note that the compiler already appends a HALT to every program, but does so near the end of processing, rather than at the start, in RECONSTITUTE-BASIC-BLOCK. I am not sure what the knock-on effects of such a change would be on the rest of the compiler, but usually adding an invariant helps more than it hurts. Or maybe it'll blow up in my face :).

2. This one is more speculative, but I think just initializing every BASIC-BLOCK's OUTGOING edge to TERMINATING-EDGE would fix this particular bug since the last BASIC-BLOCK containing the final jump instruction would then not need linkage because there is no succeeding block. But again, I'm unsure what this implies for the rest of the compiler. At a minimum, I'd have check all the places it's currently pattern matching on the OUTGOING slot and see what's affected.

If either of the above two alternatives sound more appealing (or some other option I haven't considered) I'm happy to test it out and report back.

### DIFF
--- a/src/cfg.lisp
+++ b/src/cfg.lisp
@@ -107,13 +107,15 @@
       (format stream "(~S)" (label-name (labeled obj))))
     (write-char #\Space stream)
     ;; instructions
-    (format stream "len:~D  " (length (basic-block-code obj)))
+    (format stream "len:~D " (length (basic-block-code obj)))
     ;; incoming and outgoing
     (format stream "in:~D out:" (length (incoming obj)))
-    (adt:match outgoing-edge (outgoing obj)
-      (terminating-edge (write-string "term" stream))
-      ((unconditional-edge _) (write-string "uncond" stream))
-      ((conditional-edge _ _ _) (write-string "cond" stream)))))
+    (if (not (slot-boundp obj 'outgoing))
+        (write-string "unbound" stream)
+        (adt:match outgoing-edge (outgoing obj)
+          (terminating-edge (write-string "term" stream))
+          ((unconditional-edge _) (write-string "uncond" stream))
+          ((conditional-edge _ _ _) (write-string "cond" stream))))))
 
 (defun make-block-from-label (label)
   "Make a new block from the label LABEL."
@@ -540,41 +542,38 @@ Return the following values:
 (defun program-cfg (pp &key (dce nil) (simplify nil))
   "Compute the control flow graph for the parsed program PP. Dead code in the CFG is eliminated if DCE is true (default: NIL). Measures are appended before HALT instructions if ADD-MEASURES is set to a CHIP-SPECIFICATION."
   (let* ((code (parsed-program-executable-code pp))
-         (entry (make-instance 'basic-block :name (gensym "ENTRY-BLK-")))
+         (entry (make-instance 'basic-block :name (gensym "ENTRY-BLK-") :outgoing terminating-edge))
          (cfg (make-instance 'cfg :entry-point entry
                                   :blocks (list entry))))
-    (loop :with current-block := (entry-point cfg)
-          :with finished-block := nil
-          :with link-successor? := nil
-          :for i :from 0
-          :for instr :across code
-          :do (progn
-                ;; Start a new block if necessary, and link it to its
-                ;; predecessor.
-                (when (null current-block)
-                  (setf current-block (make-instance 'basic-block)))
+    (flet ((next (current-block)
+             (or current-block (make-instance 'basic-block))))
+      (loop :with finished-block := nil
+            :with link-successor? := nil
+            :for instr :across code
+            :for current-block := (entry-point cfg) :then (next current-block)
+            :do (progn
+                  ;; Process the instruction.
+                  (multiple-value-setq (current-block finished-block link-successor?)
+                    (process-instruction cfg current-block instr))
 
-                ;; We processed a block before, and it needs to be
-                ;; linked to the current one.
-                (when link-successor?
-                  (assert (not (null finished-block)))
-                  (let ((edge (funcall link-successor? current-block)))
-                    (link-blocks finished-block edge)))
+                  (when link-successor?
+                    ;; Start a new block, and link it to its predecessor.
+                    (assert (null current-block))
+                    (assert (not (null finished-block)))
+                    (setf current-block (next current-block))
+                    (let ((edge (funcall link-successor? current-block)))
+                      (link-blocks finished-block edge)))
 
-                ;; Process the instruction.
-                (multiple-value-setq (current-block finished-block link-successor?)
-                  (process-instruction cfg current-block instr))
-
-                ;; Add the finished block if necessary.
-                (unless (null finished-block)
-                  (add-block-to-cfg finished-block cfg)))
-          :finally (progn
-                     ;; If we have a CURRENT-BLOCK that evidently
-                     ;; wasn't finished, we ought to add it to the
-                     ;; CFG, remembering to link it to the exit point.
-                     (unless (null current-block)
-                       (link-blocks current-block terminating-edge)
-                       (add-block-to-cfg current-block cfg))))
+                  ;; Add the finished block if necessary.
+                  (unless (null finished-block)
+                    (add-block-to-cfg finished-block cfg)))
+            :finally (progn
+                       ;; If we have a CURRENT-BLOCK that evidently
+                       ;; wasn't finished, we ought to add it to the
+                       ;; CFG, remembering to link it to the exit point.
+                       (unless (null current-block)
+                         (link-blocks current-block terminating-edge)
+                         (add-block-to-cfg current-block cfg)))))
 
     ;; Remove dead code when desired.
     (when dce
@@ -603,11 +602,12 @@ Return the following values:
     (loop :for x :across (basic-block-code blk) :do
       (print-instruction x s)
       (format s "\\l"))
-    (adt:match outgoing-edge (outgoing blk)
-      ((conditional-edge instr _ _)
-       (format s "\\nConditioned on ~a" (print-instruction instr nil))
-       (format s "\\l"))
-      (_ nil))))
+    (when (slot-boundp blk 'outgoing)
+      (adt:match outgoing-edge (outgoing blk)
+        ((conditional-edge instr _ _)
+         (format s "\\nConditioned on ~a" (print-instruction instr nil))
+         (format s "\\l"))
+        (_ nil)))))
 
 
 (defun generate-graphviz (cfg &optional (stream *standard-output*))
@@ -624,19 +624,20 @@ Return the following values:
 
   ;; All links
   (dolist (from-blk (cfg-blocks cfg))
-    (adt:match outgoing-edge (outgoing from-blk)
-      (terminating-edge nil)
-      ((unconditional-edge target)
-       (format stream "~A -> ~A;~%"
-               (underscorize (symbol-name (basic-block-name from-blk)))
-               (underscorize (symbol-name (basic-block-name target)))))
-      ((conditional-edge _ true-target false-target)
-       (format stream "~A -> ~A [label = \"1\", fontname = \"times-italic\"];~%"
-               (underscorize (symbol-name (basic-block-name from-blk)))
-               (underscorize (symbol-name (basic-block-name true-target))))
-       (format stream "~A -> ~A [label = \"0\", fontname = \"times-italic\"];~%"
-               (underscorize (symbol-name (basic-block-name from-blk)))
-               (underscorize (symbol-name (basic-block-name false-target)))))))
+    (when (slot-boundp from-blk 'outgoing)
+      (adt:match outgoing-edge (outgoing from-blk)
+        (terminating-edge nil)
+        ((unconditional-edge target)
+         (format stream "~A -> ~A;~%"
+                 (underscorize (symbol-name (basic-block-name from-blk)))
+                 (underscorize (symbol-name (basic-block-name target)))))
+        ((conditional-edge _ true-target false-target)
+         (format stream "~A -> ~A [label = \"1\", fontname = \"times-italic\"];~%"
+                 (underscorize (symbol-name (basic-block-name from-blk)))
+                 (underscorize (symbol-name (basic-block-name true-target))))
+         (format stream "~A -> ~A [label = \"0\", fontname = \"times-italic\"];~%"
+                 (underscorize (symbol-name (basic-block-name from-blk)))
+                 (underscorize (symbol-name (basic-block-name false-target))))))))
 
   ;; Footer
   (format stream "}~%")
--- a/tests/cfg-tests.lisp
+++ b/tests/cfg-tests.lisp
@@ -186,3 +186,11 @@
     ;; Originally this program has an unnecessary jump-when loop with 3 blocks when it could be completely removed,
     ;; as well as a trail of blocks to the end. This 8 block program should now be a 3 block program (at most).
     (is (>= 3 (length blocks-result)))))
+
+(deftest test-program-terminated-by-conditional-jump ()
+  "Verify that quil programs ending with a JUMP-WHEN or JUMP-UNLESS do not signal an UNBOUND-SLOT error.
+
+This is a regression test for https://github.com/rigetti/quilc/issues/244"
+  (dolist (jump-instr '("JUMP-WHEN" "JUMP-UNLESS"))
+    (let ((p (quil::parse-quil-into-raw-program (format nil "DECLARE ro BIT~%LABEL @START~%~A @START ro[0]" jump-instr))))
+      (not-signals error (quil::program-cfg p :dce t :simplify t)))))


src/cfg.lisp
```diff
@@ -540,41 +542,38 @@ Return the following values:
 (defun program-cfg (pp &key (dce nil) (simplify nil))
   "Compute the control flow graph for the parsed program PP. Dead code in the CFG is eliminated if DCE is true (default: NIL). Measures are appended before HALT instructions if ADD-MEASURES is set to a CHIP-SPECIFICATION."
   (let* ((code (parsed-program-executable-code pp))
-         (entry (make-instance 'basic-block :name (gensym "ENTRY-BLK-")))
+         (entry (make-instance 'basic-block :name (gensym "ENTRY-BLK-") :outgoing terminating-edge))
```
appleby: This is needed to handle the case of an empty program, i.e. `(zerop (length code))`. This change could be undone if I also revert the change for stepping `current-block` in the below `loop`. The reason being, if there no instructions, `current-block` now never gets set to the `entry` block, and hence `current-block` is `nil` in the loop's `:finally` clause and never gets linked to `terminating-edge`. This edge-case can be handled here by simply initializing the `entry` block's `outgoing` edge at the start. As far as I can tell, if any subsequent blocks are created, they will handle updating the outgoing edge.

tests/cfg-tests.lisp
```diff
@@ -186,3 +186,11 @@
     ;; Originally this program has an unnecessary jump-when loop with 3 blocks when it could be completely removed,
     ;; as well as a trail of blocks to the end. This 8 block program should now be a 3 block program (at most).
     (is (>= 3 (length blocks-result)))))
+
+(deftest test-program-terminated-by-conditional-jump ()
+  "Verify that quil programs ending with a JUMP-WHEN or JUMP-UNLESS to not signal an UNBOUND-SLOT error.
```
appleby: Whoops! Typo. s/to/do/

stylewarning:
```suggestion
  "Verify that quil programs ending with a JUMP-WHEN or JUMP-UNLESS do not signal an UNBOUND-SLOT error.
```

stylewarning:
My initial review is that this looks good.

tests/cfg-tests.lisp
appleby: Fixed

src/cfg.lisp
```diff
@@ -540,41 +542,38 @@ Return the following values:
...
-                     ;; CFG, remembering to link it to the exit point.
-                     (unless (null current-block)
-                       (link-blocks current-block terminating-edge)
-                       (add-block-to-cfg current-block cfg))))
+    (flet ((next (current-block)
+             (or current-block (make-instance 'basic-block))))
+      (loop :with finished-block := nil
+            :with link-successor? := nil
+            :for instr :across code
+            :for current-block := (entry-point cfg) :then (next current-block)
```
ecpeterson: I'm fuzzy on the behavior of `:for ... :then ...`. In the old version, it looks like `m-v-setq` was responsible for updating the contents of `current-block`. Since that `setq` is still happening here, is that in conflict with the `:then`?

ecpeterson:
I like this change. I don't view the proposed (1) and (2) as alternatives so much as things to consider doing in addition, since this change on its own improves the breadth of input of `program-cfg`, (1) improves narrows the kind of input that 'ought' to make it here, and similarly (2) with the output.

ecpeterson:
It's probably wise to wait for Robert for a merge, since he's the original author of this segment of the compiler.

src/cfg.lisp
appleby: It doesn't conflict because `(next current-block)` just returns the current block if non-nil, so it doesn't trample on the `current-block` returned by `process-instruction` or the new block created when `link-successor?` is non-nil. This `:then` clause is actually replacing the

```lisp
(when (null current-block)
  (setf current-block (make-instance 'basic-block)))
```
that used to appear at the top of the loop in the old version.

I believe this `:then` clause also preserves the semantics of the original version of the loop w.r.t. the case where the final `process-instruction` returns nil for both `current-block` and `link-successor?`, namely, that `current-block` will still be `nil` in the `:finally` clause.

Actually, I just realized that I'm assuming that the `for` loop is specified to process `:for.. :then` clauses in left-to-right order. For some reason I have this impression, but I should probably check on that...

appleby: [Hyperspec confirms](http://www.lispworks.com/documentation/lw51/CLHS/Body/06_aba.htm) sequential binding and stepping of `:for` clauses:

> If multiple iteration clauses are used to control iteration, variable initialization and stepping[1] occur sequentially by default. The and construct can be used to connect two or more iteration clauses when sequential binding and stepping[1] are not necessary. The iteration behavior of clauses joined by and is analogous to the behavior of the macro do with respect to do*.

And the sections for [`do*`](http://www.lispworks.com/documentation/lw51/CLHS/Body/m_do_do.htm#doST) and [`let*`](http://www.lispworks.com/documentation/lw51/CLHS/Body/s_let_l.htm#letST) imply left-to-right evaluation order for the `:for` clauses, as every sane person would expect.

appleby: Possibly too late because already merged, and definitely T.M.I., but in case you find yourself with nothing to do on some rainy Saturday...

There is still the question of whether the `:for` clauses are guaranteed to "short-circuit", i.e. whether the `:then` clause is guaranteed not to execute once the `:for instr :across code` runs out of instructions. If not, the processing could fail since `current-block` would never be non-nil in the `:finally` clause. I think all that is implied by sequential stepping mentioned above is the left-to-right order of evaluation (and visibility of updated bindings), but not necessarily that the termination conditions short-circuit, even though that is what you'd expect.

With regards to the termination conditions, the Hyperspec says:

> The for and as clauses iterate by using one or more local loop variables that are initialized to some value and that can be modified or stepped[1] after each iteration. For these clauses, iteration terminates when a local variable reaches some supplied value or when some other loop clause terminates iteration.

The phrase "some other loop clause terminates iteration" is a little vague, but I think taken together with the sequential-stepping paragraph above, the most reasonable interpretation would be to short-circuit termination.

All of SBCL, CCL, and ECL's LOOP implementations are apparently derived from the same Symbolics code, and their macro expansions are all very similar, and all appear to short-circuit, at least for this simple test.

```
;; sbcl
CL-USER> (pprint (sb-cltl2:macroexpand-all '(loop :for x :across some-vector :for y := 1 :then (1+ y) :collect y)))

(BLOCK NIL
  (LET ((X NIL)
        (#:LOOP-ACROSS-VECTOR-550 SOME-VECTOR)
        (#:LOOP-ACROSS-INDEX-551 0))
    (DECLARE (TYPE FIXNUM #:LOOP-ACROSS-INDEX-551)
             (TYPE VECTOR #:LOOP-ACROSS-VECTOR-550))
    (LET ((#:LOOP-ACROSS-LIMIT-552 (LENGTH #:LOOP-ACROSS-VECTOR-550)))
      (LET ((Y NIL))
        (DECLARE)
        (LET* ((#:LOOP-LIST-HEAD-553 (LIST NIL))
               (#:LOOP-LIST-TAIL-554 #:LOOP-LIST-HEAD-553))
          (TAGBODY
            (IF (>= #:LOOP-ACROSS-INDEX-551 #:LOOP-ACROSS-LIMIT-552)
                (GO SB-LOOP::END-LOOP))
            (SETQ X (AREF #:LOOP-ACROSS-VECTOR-550 #:LOOP-ACROSS-INDEX-551))
            (SETQ #:LOOP-ACROSS-INDEX-551 (1+ #:LOOP-ACROSS-INDEX-551))
            (SETQ Y 1)
           SB-LOOP::NEXT-LOOP
            (RPLACD #:LOOP-LIST-TAIL-554 (SETQ #:LOOP-LIST-TAIL-554 (LIST Y)))
            (IF (>= #:LOOP-ACROSS-INDEX-551 #:LOOP-ACROSS-LIMIT-552)
                (GO SB-LOOP::END-LOOP))
            (SETQ X (AREF #:LOOP-ACROSS-VECTOR-550 #:LOOP-ACROSS-INDEX-551))
            (SETQ #:LOOP-ACROSS-INDEX-551 (1+ #:LOOP-ACROSS-INDEX-551))
            (SETQ Y (1+ Y))
            (GO SB-LOOP::NEXT-LOOP)
           SB-LOOP::END-LOOP
            (RETURN-FROM NIL (CDR #:LOOP-LIST-HEAD-553))))))))
; No value
```


I suspect those are the implementations you're most likely to care about, but here is CLISP just for giggles, which does not appear to be derived from the aforementioned Symbolics code, but does also appear to short-circuit...

```
;; clisp
CL-USER> (pprint (macroexpand '(loop :for x :across some-vector :for y := 1 :then (1+ y) :collect y)))

(MACROLET ((LOOP-FINISH NIL (SYSTEM::LOOP-FINISH-ERROR)))
 (BLOCK NIL
  (LET
   ((#:VECTOR-21537
     (LOCALLY
      (DECLARE
       (SYSTEM::IMPLEMENTATION-DEPENDENT #:INDEX-21538 #:VECTOR-21537 X Y))
      SOME-VECTOR))
    (#:INDEX-21538 0))
   (PROGN
    (LET ((X NIL))
     (LET ((Y NIL))
      (LET ((#:ACCULIST-VAR-21539 NIL))
       (MACROLET ((LOOP-FINISH NIL '(GO SYSTEM::END-LOOP)))
        (TAGBODY
         (WHEN (>= #:INDEX-21538 (LENGTH #:VECTOR-21537)) (LOOP-FINISH))
         (SETQ X (AREF #:VECTOR-21537 #:INDEX-21538)) (SETQ Y 1)
         SYSTEM::BEGIN-LOOP
         (PROGN (SETQ #:ACCULIST-VAR-21539 (CONS Y #:ACCULIST-VAR-21539)))
         (PSETQ #:INDEX-21538 (1+ #:INDEX-21538))
         (WHEN (>= #:INDEX-21538 (LENGTH #:VECTOR-21537)) (LOOP-FINISH))
         (SETQ X (AREF #:VECTOR-21537 #:INDEX-21538)) (PSETQ Y (1+ Y))
         (GO SYSTEM::BEGIN-LOOP) SYSTEM::END-LOOP
         (MACROLET
          ((LOOP-FINISH NIL (SYSTEM::LOOP-FINISH-WARN) '(GO SYSTEM::END-LOOP)))
          (RETURN-FROM NIL
           (SYSTEM::LIST-NREVERSE #:ACCULIST-VAR-21539))))))))))))
; No value
```

stylewarning: Maybe @rigettizach has a thought?

appleby: It finally occurred to me to just use common sense rather than a "close" reading of the spec, and I guess the termination tests implicit in iteration control clauses must happen either before any variables are stepped or at least interleaved with sequential stepping; otherwise, something like

```lisp
(loop :for x :in '(1 2) :for y = (1+ x) ...)
```

would explode if both `x` and `y` were stepped prior to any termination test, since there wouldn't seem to be any sane guarantee for what the next value of `x` would be after 2 (presumably `nil`).

As a final juicy tidbit, however, in section [6.1.1.6 Order of Execution](http://www.lispworks.com/documentation/lw51/CLHS/Body/06_aaf.htm) the spec says:

> Iteration control clauses implicitly perform the following actions:
>     -- initialize variables;
>     -- step variables, generally between each execution of the loop body;
>     -- perform termination tests, generally just before the execution of the loop body.

This seems to imply that all step forms are executed *before* any termination tests! I suspect "termination tests" here means `while`, `until`, `always`, etc., and excludes the termination tests bundled in with iteration control clauses.

PR #253, opened 2019-05-23
Portability improvements

dkochmanski:
With these commits cl-quil loads without problems on ECL[1]. Some fixes are made to improve CCL support, but it still doesn't load cl-quil because of how ADT creates structuers (it uses uninterned symbols and make-load-form-saving-slots on CCL doesn't use slot offsets but names).

[1] Branch is not merged yet, but it introduces improvements to make-load-form-saving-slots and adds custom equivalence predicate capabilities to make-hash-table.

https://gitlab.com/embeddable-common-lisp/ecl/merge_requests/152

### DIFF
--- a/benchmarking/rewiring-analysis.lisp
+++ b/benchmarking/rewiring-analysis.lisp
@@ -134,13 +134,12 @@
             (read-quil-file prog-source))))
 
        (by-assignment (prog-source chip assn)
-         #+sbcl (sb-ext:gc :full t)
          (handler-bind
              ((simple-error
-                (lambda (e)
-                  (when break-on-error
-                    (break "~a" e))
-                  (return-from by-assignment (list nil nil 1)))))
+               (lambda (e)
+                 (when break-on-error
+                   (break "~a" e))
+                 (return-from by-assignment (list nil nil 1)))))
            (funcall assn (lambda ()
                            (rest (multiple-value-list (compiler-hook (get-prog prog-source) chip)))))))
 
@@ -191,7 +190,9 @@
                      (list (make-instance 'application-force-rewiring :target target)))))
 
 (defun generate-random-rewiring-prog (n-qubits state)
-  (let ((*random-state* (sb-ext:seed-random-state state)))
+  (let ((*random-state* #+sbcl (sb-ext:seed-random-state state)
+                        #+ecl  (make-random-state state)
+                        #-(or sbcl ecl) (error "don't know how to seed random state")))
     (make-rewiring-prog (quil::generate-random-rewiring n-qubits))))
 
 (defun measure-rewiring-swap-search (assn &rest args
@@ -212,7 +213,9 @@
 
 (defvar *basic-swap-search-assn*
     (make-assignments
-        ((*random-state* (sb-ext:seed-random-state 1))
+        ((*random-state* #+sbcl (sb-ext:seed-random-state 1)
+                         #+ecl  (make-random-state 1)
+                         #-(or sbcl ecl) (error "don't know how to seed random state"))
          (quil::*compressor-passes* 0))
         (quil::*addresser-swap-search-type*
          quil::*addresser-move-to-rewiring-swap-search-type*)
@@ -223,7 +226,9 @@
 
 (defvar *2q-tiers-assn*
   (make-assignments
-      ((*random-state* (sb-ext:seed-random-state 1))
+      ((*random-state* #+sbcl (sb-ext:seed-random-state 1)
+                       #+ecl  (make-random-state 1)
+                       #-(or sbcl ecl) (error "don't know how to seed random state"))
        (quil::*compressor-passes* 0))
       (quil::*initial-rewiring-default-type*
        quil::*addresser-swap-search-type*
@@ -243,7 +248,9 @@
 
 (defvar *swap-search-assn*
   (make-assignments
-      ((*random-state* (sb-ext:seed-random-state 1))
+      ((*random-state* #+sbcl (sb-ext:seed-random-state 1)
+                       #+ecl  (make-random-state 1)
+                       #-(or sbcl ecl) (error "don't know how to seed random state"))
        (quil::*compressor-passes* 0))
       (quil::*initial-rewiring-default-type*
        quil::*addresser-swap-search-type*
@@ -260,7 +267,9 @@
 
 (defvar *initial-rewiring-assn*
   (make-assignments
-      ((*random-state* (sb-ext:seed-random-state 1))
+      ((*random-state* #+sbcl (sb-ext:seed-random-state 1)
+                       #+ecl  (make-random-state 1)
+                       #-(or sbcl ecl) (error "don't know how to seed random state"))
        (quil::*compressor-passes* 0)
        (quil::*addresser-swap-search-type* :greedy-qubit))
       (quil::*initial-rewiring-default-type*)
@@ -272,7 +281,9 @@
 
 (defvar *depth-vs-swaps-assn*
   (make-assignments
-      ((*random-state* (sb-ext:seed-random-state 1))
+      ((*random-state* #+sbcl (sb-ext:seed-random-state 1)
+                       #+ecl  (make-random-state 1)
+                       #-(or sbcl ecl) (error "don't know how to seed random state"))
        (quil::*compressor-passes* 0))
       (quil::*initial-rewiring-default-type*
        quil::*addresser-swap-search-type*
--- a/src/analysis/rewrite-arithmetic.lisp
+++ b/src/analysis/rewrite-arithmetic.lisp
@@ -119,7 +119,8 @@ expressions, as elements of a single new memory descriptor.
          ;; just a single name (i.e., more than just __P). It adds
          ;; little bit of complexity in favor of a bit of additional
          ;; extensibility.
-         (recalculation-table (make-hash-table :test 'memory-ref=))
+         (recalculation-table (make-hash-table :test 'memory-ref=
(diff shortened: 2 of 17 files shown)


dkochmanski:
how complete is the test suite? when running tests against ecl I have some weird failures, most importantly:

```
Condition of type: SIMPLE-ERROR
No applicable method for CL-QUIL::PRINT-INSTRUCTION-GENERIC with arguments of types 
 LONG-FLOAT
 STRING-STREAM.
```

And indeed, there is no such specialization. When this error is hit (after many successfully passed tests), selecting restart `1. (TRY-NEXT-COMPILER) Ignore this error and try the next compiler in the list.` gives me the same error over and over again.

There is also this simple error earlier:

```
Condition of type: SIMPLE-ERROR
No keys match in SWITCH. Testing against (NIL :FLUSHING) with EQUAL.
```

which stems from `compress-instructions` -- `(old-state (governed-queue-state governed-queue))` old state is NIL.

dkochmanski:
cl-quil test results (for the reference)

[xxx.txt](https://github.com/rigetti/quilc/files/3212862/xxx.txt)

dkochmanski:
for quilc compilation I've made a PR to https://github.com/rigetti/rpcq/pull/69 which drops dependency on unicly (for uuid which is much better written and more portable).

ecpeterson:
You're probably right that quilc unwisely mixes `double-float` and `long-float`, leading to some of the problems here. However, I don't see where `print-instruction-generic` might be being called on an unboxed float of any sort. Any chance of sharing a call graph?

The `(NIL :FLUSHING)` error is more concerning. I'll think about that and get back to you.

benchmarking/rewiring-analysis.lisp
```diff
@@ -134,13 +134,13 @@
             (read-quil-file prog-source))))
 
        (by-assignment (prog-source chip assn)
-         #+sbcl (sb-ext:gc :full t)
+         (trivial-garbage:gc :full t)
```
stylewarning: @ecpeterson I wonder if this was a hold-over from Corwin? I wonder if we can nix this...

src/analysis/rewrite-arithmetic.lisp
```diff
@@ -119,7 +119,8 @@ expressions, as elements of a single new memory descriptor.
          ;; just a single name (i.e., more than just __P). It adds
          ;; little bit of complexity in favor of a bit of additional
          ;; extensibility.
-         (recalculation-table (make-hash-table :test 'memory-ref=))
+         (recalculation-table (make-hash-table :test 'memory-ref=
+                                               :hash-function 'memory-ref-hash))
```
stylewarning: is ECL ok with this extension?

src/package.lisp
```diff
@@ -422,7 +422,10 @@
   (:export
    #:standard-qubit-relabeler           ; FUNCTION
    )
-  )
+  #+ (or ccl ecl sbcl)
+  (:import-from
+   #+ccl #:ccl #+ecl #:ext #+sbcl #:sb-ext
+   double-float-positive-infinity))
```
stylewarning: Could we instead do a `defconstant` in the utilities file? I want to error at compile time if this isn't defined.

src/parser.lisp
```diff
@@ -95,9 +95,9 @@
    (return (tok :PARAMETER (param $1))))
   ((eager "{{STRING}}")
    (return (tok :STRING (read-from-string $@))))
-  ((eager #.(string #\DAGGER))
+  ((eager #.(string #-ccl #\DAGGER #+ccl #\u+2020))
```
stylewarning: haha :(

src/analysis/rewrite-arithmetic.lisp
dkochmanski: I've implemented it yesterday (see the linked branch). hash-table with custom test is also used in god-table.

src/parser.lisp
dkochmanski: I don't understand why this is funny (in a sad fashion ;), could you elaborate?

src/package.lisp
dkochmanski: fixed (I've used alexandria's defie-constant), if not implemented the value resolves to (error …).

dkochmanski:
@ecpeterson the problem with `(NIL :FLUSHING)` was caused by calling make-instance on a structure-class designator. Common Lisp is silent about the results of such operation. Last commit addresses this issue and `TEST-COMPILING-EMPTY-PROGRAM` passes now.

I'm going to look now into printing issue.

dkochmanski:
So the problem with printing is caused by the assumption that PI is of type double-float. `build-gate` takes the number and produces long float instead of a boxed value in this case. Shadowing pi with a constant #.(coerce cl:pi 'double-float) solves the issue (pushed). Now I'm facing a regression (in ECL) with weak hash tables.

dkochmanski:
With current state of this branch (most notably pi shadowed to be always double-float) ECL as compiled from branch `relax-reqs` passes all tests (except the test where c stack is exceeded and can't be grown any further). Alternative solution to shadowing pi would not be using this variable (but i.e a constant `+pi-df+`) or making quilc work with long floats too.

ECL performance is poor. Completing all tests took ~2h (compared to SBCL's ~15m). Everything was compiled on default optimization settings so it could be probably way better (I need to think about re-tweaking default compiler policies). Which operation takes most time depends on a test (sometimes it is gc, sometimes equal/string hashtable access, generic function dispatch, aref etc).

I'm attaching file with the test log. I don't know why fiasco nicely aligns test output on sbcl and not on ECL (format issue? using implementation extension on sbcl?).

[results-ecl.txt](https://github.com/rigetti/quilc/files/3217507/results-ecl.txt)
[results-sbcl.txt](https://github.com/rigetti/quilc/files/3217508/results-sbcl.txt)

benchmarking/rewiring-analysis.lisp
ecpeterson: This line indeed seems unimportant. You'd want it if you were measuring wall time inside of this loop, but we aren't.

src/compressor/compressor.lisp
```diff
@@ -752,7 +752,7 @@ This specific routine is the start of a giant dispatch mechanism. Its role is to
   (let* ((output nil)
          (n-qubits (chip-spec-n-qubits chip-specification))
          (governors (make-list (length (chip-specification-objects chip-specification))))
-         (global-governor (make-instance 'governed-queue))
+         (global-governor (make-governed-queue))
```
ecpeterson: 👍

src/parser.lisp
ecpeterson: I can't explain why it's funny, but the humor is in how several thousands of LoC doing the heavy lifting of quantum compilation mostly all just worked†, but this goofy easter egg that no one actually uses required special treatment.

† - "Just worked" only after a month of very serious work at getting MAGICL into line.

notmgsk: Wait - does this mean that the literal "DAGGER" won't work on CCL?

ecpeterson: CCL doesn't know the human-readable mnemonic `#\DAGGER` for 🗡, which we make available as an alias for the Quil-literal DAGGER.

notmgsk: Oh, the string "DAGGER" is caught elsewhere. My bad.

appleby:
@dkochmanski Regarding the stack overflow issue, does ECL do tail-call-optimization with default optimization settings?

This is speculation, but I have seen stack overflow in OUTER-INSTRUCTION-LOOP in the file `compressor.lisp` when running tests compiled with `(debug 3)` on SBCL, presumably because that disables TCO on SBCL.

benchmarking/rewiring-analysis.lisp
dkochmanski: in this case should it stay or should it go?

src/parser.lisp
dkochmanski: this is a minor issue (strings with dagger will still work whatsoever), the biggest (and as far as I'm aware the only) blocker for CCL is the fact that you do #.(foo) on structures defined with `adt:defdata`. This macro creates structures with uninterned slot names and CCL can't load them back (ECL couldn't neither but I've fixed it by using offsets in `make-load-data-using-slots`).

dkochmanski:
@appleby whenever ECL does TCO or not depends on the underlying C compiler. On GCC and Clang with speed >=2 flag "-O2" is set and optimization is present. That said I've build ECL with debugging enabled for this test run, so I'll rebuild all with more favourable optimize settings.

src/parser.lisp
stylewarning: @dkochmanski I made a PR to CL-ADT to fix that issue

benchmarking/rewiring-analysis.lisp
stylewarning: let's nix it

stylewarning:
Could we also state in the README the version needed for ECL and CCL?

dkochmanski:
I've removed call to gc.

regarding the README, Quilc does not compile on CCL (due to a reason I've mentioned with structure load-forms and uninterned slot names) and with ECL changes necessary for it to work are still in a branch. Even after merge to develop they won't be released yet, because develop branch is mid-release branch (so 16.1.3 is not enough to compile quilc and 16.2.0 is not released yet).

dkochmanski:
compilling with optimization options (speed 3) (debug 0) (safety 1) does not help with the stack overflow. problem happens in test-compiler-hook test (not right of the bat, some test cases pass until the one which consumes too much of the stack, resizing helps and few other pass until we hit a wall).

PR #255, opened 2019-05-23
improve look n feel of test output

stylewarning:


tests/compilation-tests.lisp
```diff
@@ -168,7 +168,7 @@
 
 (deftest test-cnot-rewiring-in-cnot-architecture ()
   "Test that all CNOTs on each pair of qubits compile into a single CNOT rewired appropriately."
-  (format t "[Test output: ~%")
+  (format t "~&[    Test output: ~%")
```
appleby: This line is space-after-open-bracket, whereas below it's space-before-open-bracket. Intentional?

tests/compilation-tests.lisp
```diff
@@ -186,7 +186,7 @@
 
 (deftest test-cz-compilation-and-rewiring-in-cnot-architecture ()
   "Test that all CZs on all qubit combinations compile to a single CNOT that's native."
-  (format t "[Test output: ~%")
+  (format t "~&    [Test output: ~%")
```
appleby: here

tests/compilation-tests.lisp
notmgsk: This is different to the diff below. This has `[    Test...`, other has `     [Test....`.

notmgsk: Got there before ya ;D

appleby: Blast! Too slow.

PR #257, opened 2019-05-24
Fix EXPT printing

### LINKED ISSUE #249: Parametric DEFGATE with exponent arithmetic producing erroneous EXPT in output
Witness

`test.quil`
```
DEFGATE TEST(%arg):
    COS(%arg)^2, 0
    0, SIN(%arg)^2

TEST(0.5) 0
```

`quilc < test.quil`
```
quilc 
DEFGATE TEST(%arg):
    (COS(%arg)EXPT(2.0)), 0.0
    0.0, (SIN(%arg)EXPT(2.0))


HALT                                    # Entering/exiting rewiring: (#(0 1 2 3 4 5 6 7) . #(0 1 2 3 4 5 6 7))
```

appleby:
This is a WIP, but I wanted to get early feedback to see how folks feel about the "golden file" testing approach used here, before I go crazy and add a bunch of tests.

Assuming no one hates the golden file approach, my plan was to write a small script that can be run to update the expected outputs whenever changes to the printer are made, and then to add a bunch of these tests (and probably some regular lisp-based tests to handle any edge-cases that are otherwise hard to tickle).

Pros of golden files:
  - Easy to add new test cases
  - Easy to automate updating expected output if the printer changes

Cons of golden files:
  - Have to remember to update expected output if the printer changes
  - Easy to automate updating expected output if the printer changes, and then blissfully ignore wrong updated output if the diff is not scrutinized sufficiently closely.

This is not really specific to the "golden file" approach, but basically whether or not it makes to add a bunch of printer tests depends on how stable the printed representations are. If they're constantly changing, then updating the test expectations will come to be a tedious chore. If they're relatively stable, only a minor tedious chore!

- [x] Determine whether golden files considered harmful
- [x] IF golden files not hateful THEN write program to automate regeneration of golden files
- [x] Add more tests
- [x] Simplify handling of trailing newlines
- [x] Add single-source-of-truth for bi-directional mapping between quil arthimetic operators & functions <--> lisp functions to prevent bugs when/if quil functions are added/removed. Note that `*valid-functions*` already exists in `parser.lisp`, but does not include the mappings for infix arithmetic ops.
- [x] Move generic lisp types and type predicates like `string-list-p` into a new `types.lisp` file
- [x] Write actual commit messages
- [ ] Profit!

Fixes #249


src/ast.lisp
```diff
@@ -112,7 +112,7 @@ EXPRESSION should be an arithetic (Lisp) form which refers to LAMBDA-PARAMS."
   "Evaluate the delayed expression DE to a numerical value (represented in a CONSTANT data structure). MEMORY-MODEL is an association list with keys MEMORY-REF structures and values the value stored at that location."
   (labels ((lookup-function (expr)
              (case expr
-               ((+ - * / cos sin tan) expr)
+               ((+ - * / expt cos sin sqrt exp cis) expr)
```
appleby: I assume that this list of functions should match what is accepted by the parser?

tests/printer-test-files/delayed-expressions.quil
```diff
@@ -0,0 +1,94 @@
...
+DEFCIRCUIT TEST(%arg1, %arg2) p:
+    RX(%arg1-%arg2) p
+
+TEST(pi, SIN(pi/2)) 0
+# Output
+DEFCIRCUIT TEST(%arg1, %arg2) p:
+    RX((%arg1-%arg2)) p
+
+
+RX(2.141592653589793) 0
```
appleby: Not sure how portable it is to depend on the default printed representation for `~F` here.

appleby:
Forgot to mention that another approach to the golden files is to have inputs and expected outputs live in separate files, with at most one input/output per file. Then `parse-golden-file` can go away. That approach is simpler and more machine-friendly, but I feel like the ability to have multiple inputs/outputs in a single file is more human-editor friendly. Debatable, though.

src/ast.lisp
notmgsk: That seems reasonable. But it seems so reasonable that I'm surprised this wasn't caught earlier -- so maybe there is something we're both missing. @ecpeterson @stylewarning

tests/printer-test-files/delayed-expressions.quil
notmgsk: That's a very good point, considering that we're going to be supporting ECL soon enough.

notmgsk:
> Forgot to mention that another approach to the golden files is to have inputs and expected outputs live in separate files, with at most one input/output per file. Then `parse-golden-file` can go away. That approach is simpler and more machine-friendly, but I feel like the ability to have multiple inputs/outputs in a single file is more human-editor friendly. Debatable, though.

If we're going the golden-file route, then my vote is for keeping input/output in the same file for exactly the reason you noted. I'll defer my vote on this golden-file idea for now.

@appleby you'll have to test the golden test stuff right? What do we call that? *Very* golden test?

src/ast.lisp
ecpeterson: This is correct. I’m surprised (1) that Quil uses `expt` to express exponentials, rather than `^` or `**`, and (2) that `tan` isn’t legal Quil. Shows how carefully I read!

notmgsk:
Then again, relying on `parse-golden-file` introduces more surface area for bugs. I'm on the fence.

src/ast.lisp
notmgsk: It *does* use `^`

appleby:
> @appleby you'll have to test the golden test stuff right? What do we call that? _Very_ golden test?

Ha! *Platinum* test files? Meta-Golden?

Actually I did consider writing some tests for `parse-golden-file`, but then thought maybe testing the tests was overkilll. The "parser" is quite simple, probably could be replaced with a regex or even a "split on substrings" function if I could find one (anyone? anyone?), plus the input file format is unlikely to change much, so I've just repl-tested it for now. But I'm happy to add some tests for `parse-golden-file`, if desired.

Actually, one reason I went with a simple parser rather than chunk-on-substrings is I had the idea that it could maybe be extended to also handle inputs that are meant to signal error via another magic comment cookie like "# Signals: foo-error" or the like.

tests/utilities.lisp
```diff
@@ -33,3 +33,102 @@
...
+                  :format-control format-control
+                  :format-arguments format-arguments)))
+    (loop :with state := :start
+          :with input-header := "# Input"
+          :with output-header := "# Output"
+          :with pending-lines := '()
+          :with inputs := '()
+          :with outputs := '()
+          :for line-number :upfrom 1
+          :for line :in (split-sequence:split-sequence
```
appleby: This probably qualifies as gratuitous consing, but I'm being being lazy for now. I might switch to `read-line` and just deal with edge cases myself.

notmgsk: IMO gratuitous is fine in the test suite.

notmgsk:
I wonder if this parsing business could be improved by marking gates with a pragma instead. E.g `PRAGMA TEST_{INPUT,OUTPUT}` or `PRAGMA BEGIN_TEST_INPUT_BLOCK` etc. `cl-quil` ought to have some pragma block logic you can co-opt.

src/ast.lisp
appleby: Correct. Quil uses infix "^", which maps to `cl:expt` on the lisp side. This is actually the only case where the name of the Quil operator and the lisp function differ, and was the source of this bug.

My idea is to add some abstraction as part of this PR to provide a single source of truth for the quil <--> lisp mapping, so that this `case` form won't have to duplicate the list of valid functions here.

appleby:
> I wonder if this parsing business could be improved by marking gates with a pragma instead. E.g `PRAGMA TEST_{INPUT,OUTPUT}` or `PRAGMA BEGIN_TEST_INPUT_BLOCK` etc. `cl-quil` ought to have some pragma block logic you can co-opt (someone has already implemented `PRAGMA PRESERVE_BLOCK ... PRAGMA END_PRESERVE_BLOCK`).

Personally, I would rather not sully the real parser with testing concerns if it can be avoided. Unless it's something that is expected to be generally useful elsewhere? Or are you saying there is some pre-existing PRAGMA facility that I could use instead?

notmgsk:
> > I wonder if this parsing business could be improved by marking gates with a pragma instead. E.g `PRAGMA TEST_{INPUT,OUTPUT}` or `PRAGMA BEGIN_TEST_INPUT_BLOCK` etc. `cl-quil` ought to have some pragma block logic you can co-opt (someone has already implemented `PRAGMA PRESERVE_BLOCK ... PRAGMA END_PRESERVE_BLOCK`).
> 
> Personally, I would rather not sully the real parser with testing concerns if it can be avoided. Unless it's something that is expected to be generally useful elsewhere? Or are you saying there is some pre-existing PRAGMA facility that I could use instead?

I expected that block logic to exist elsewhere, based on the existence of `PRAGMA (END)?_PRESERVE_BLOCK`. Seems to be not the case, however.

ecpeterson:
I, too, think that there should be some support for “region” PRAGMAs. I think there was an internal ticket for this, but it didn’t survive the public migration, & it isn’t very clear to me how it ought to be done anyhow.

stylewarning:
(this thread has used the word `golden` too much and has been semantically satiated)

tests/printer-test-files/delayed-expressions.quil
stylewarning: how about doing `TEST(pi, pi*SIN(pi/2))`

tests/printer-tests.lisp
```diff
@@ -0,0 +1,100 @@
...
+  (let* ((golden-files (uiop:directory-files
+                        (asdf:system-relative-pathname "cl-quil-tests" "tests/printer-test-files/")
+                        #P"*.quil")))
+    (is (not (null golden-files)))
+    (dolist (file golden-files)
+      (format t "~&    Testing file ~a" (pathname-name file))
+      (multiple-value-bind (golden-inputs golden-outputs) (parse-golden-file file)
+        (loop :for input :in golden-inputs
+              :for expected-output :in golden-outputs :do
+                (let* ((input-pp (quil:parse-quil input))
```
stylewarning: @ecpeterson @notmgsk Should we test `parse-quil` or `parse-quil-into-raw-program`?

tests/utilities.lisp
stylewarning: i wouldn't even mind if you did
```
:for line :in (loop :repeat 1000 :do (cons 'appleby 'appleby) :finally (return (split-sequence ...
```

tests/utilities.lisp
```diff
@@ -33,3 +33,102 @@
                  :operator (cl-quil::named-operator "TEST")
                  :arguments (mapcar #'cl-quil::qubit qubit-indices)
                  :gate matrix))
+
+
+;;;
+;;; Parsing of "golden" test files.
+;;;
+
```
stylewarning: ?can you write some commentary here about golden files, what they are, their purpose, etc.

tests/printer-tests.lisp
notmgsk: https://i.imgur.com/X1mVtuY.png

tests/printer-test-files/delayed-expressions.quil
appleby: I almost did precisely that before opening the PR, but figured it was better to have the conversation here since I thought it my crop up again.

tests/utilities.lisp
appleby: Tempting...

tests/printer-tests.lisp
appleby: I forgot to ask about this. I was considering doing both in a loop, but not sure if there is any point in that.

notmgsk: I don't think it matters

notmgsk: Maybe it does if `parse-quil` is simplifying arithmetic

tests/printer-test-files/delayed-expressions.quil
appleby: Done

tests/utilities.lisp
```diff
@@ -33,3 +33,102 @@
...
+          :with input-header := "# Input"
+          :with output-header := "# Output"
+          :with pending-lines := '()
+          :with inputs := '()
+          :with outputs := '()
+          :for line-number :upfrom 1
+          :for line :in (split-sequence:split-sequence
+                         #\Newline
+                         (alexandria:read-stream-content-into-string stream))
+          :do (progn
```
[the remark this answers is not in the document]
appleby: Good eye! Done.

tests/utilities.lisp
```diff
@@ -33,3 +33,102 @@
...
+                (ecase state
+                  (:start
```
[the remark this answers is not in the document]
appleby: Done.

tests/utilities.lisp
```diff
@@ -33,3 +33,102 @@
...
+                   (unless (string= line "# Input")
+                     (parse-error line-number line "Expected ~S" input-header))
+                   (setf state :reading-input))
```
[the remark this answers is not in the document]
appleby: Done. I noticed this pattern elsewhere in the code. Why quote a self-quoting symbol?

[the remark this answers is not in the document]
stylewarning: indicate that it's data, and not a keyword argument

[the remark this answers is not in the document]
appleby: self-evaluating?  you know what I mean...

tests/utilities.lisp
```diff
@@ -33,3 +33,102 @@
...
+                        input-header (length inputs)
+                        output-header (length outputs)))
+                     (return (values (nreverse inputs) (nreverse outputs)))))))
+
+(defun parse-golden-file (file-name &key (strip-final-blank-line t))
+  "Convenience wrapper around PARSE-GOLDEN-STREAM."
+  (with-open-file (f file-name)
+    (parse-golden-file-stream f :strip-final-blank-line strip-final-blank-line)))
+
+(define-condition golden-file-parse-error (alexandria:simple-parse-error)
```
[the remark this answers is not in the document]
appleby: Done. I actually moved this down to the bottom just before opening this PR because I was hoping the documentation string for `parse-golden-file-stream` would serve as a lazy-man's section intro comment.

tests/utilities.lisp
appleby: Done.

appleby:
Marking this DO NOT REVIEW until I finish the remaining tasklist.

tests/printer-tests.lisp
stylewarning: I think the simplification is happening @ parse time, not @ transformation time.

tests/printer-tests.lisp
```diff
@@ -0,0 +1,122 @@
...
+                  (not-signals error (quil:parse-quil actual-output))
+
+                  ;; Ensure expected-output is a fixed point of parse -> print.
+                  (is (string= expected-output
+                               (parse-and-print-quil-to-string expected-output)))))))))
+
+(deftest test-print-instruction ()
+  (is (string= "PRAGMA gate_time CNOT \"50 ns\""
+               (with-output-to-string (s)
+                 (cl-quil::print-instruction (make-instance 'quil::pragma
```
stylewarning: here's a Cool Tip, you can do

```
(format s "~/cl-quil:instruction-fmt/" x)
```
to print the `x` instruction to stream `s`. Use `s = NIL` to create a string.

tests/utilities.lisp
```diff
@@ -33,3 +33,233 @@
...
+
+OUTPUT-CALLBACK is function from STRING -> STRING. It will be called successively for each
+golden-file input section, and should return the corresponding output string for the given input.
+
+ADD-FINAL-NEWLINE is the dual of PARSE-GOLDEN-FILE-STREAM's STRIP-FINAL-NEWLINE. This means that if
+you update a golden file by calling this function, you should ensure that the value of
+ADD-FINAL-NEWLINE matches the value you plan to pass for STRIP-FINAL-NEWLINE when parsing the golden
+file in your tests. Both default to T.
+
+IF-EXISTS has the standard common lisp meaning. See http://l1sp.org/cl/open."
```
stylewarning:
```suggestion
IF-EXISTS has the standard Common Lisp meaning. See http://l1sp.org/cl/open."
```

tests/printer-tests.lisp
appleby: Nice. This was actually a pre-existing test that I moved from another file, but I will update it as part of this PR.

tests/utilities.lisp
```diff
@@ -33,3 +33,233 @@
...
+  "Convenience wrapper around PARSE-GOLDEN-STREAM."
+  (check-type file-name (or string pathname))
+  (with-open-file (f file-name)
+    (parse-golden-file-stream f :strip-final-newline strip-final-newline)))
+
+(defun update-golden-file-output-sections (file-paths output-callback
+                                           &key (add-final-newline t) (if-exists ':supersede))
+  "Update all output sections of the golden files at FILE-PATHS.
+
+If you call this function, examine the diffs of the old vs new output sections *carefully* before
```
stylewarning: you can also add a good ol `y-or-n-p` to this function if you want this to be a truly interactive state of affairs.

appleby: Sure. I did consider it, but then figured that the files being updated would likely be version controlled anyway. But I suppose it's safer to prompt.

tests/utilities.lisp
appleby: This is now fixed because, in the words of the Elephant Man: I AM NOT AN ANIMAL!

tests/printer-tests.lisp
appleby: Done

tests/utilities.lisp
appleby: Done

tests/utilities.lisp
appleby: Done

tests/utilities.lisp
```diff
@@ -33,3 +33,278 @@
...
+       (check-type element-test symbol)
+       (let ((docstring
+               (format nil "Return true if OBJECT is a ~A and every element satisfies ~A."
+                       container-type element-test)))
+         `(defun ,predicate-name (object)
+            ,docstring
+            (and (typep object ',container-type)
+                 (every #',element-test object))))))
+  (def string-list-p list stringp)
+  (def string-sequence-p sequence stringp))
```
appleby: There are a handful of other places that similar type predicates to these are defined throughout the code. For instance, there is `integer-sequence-p` in `define-pragma.lisp` and `integer-list-p` in `chip-reader.lisp`.

I considered collecting them all together in some new file named like `types.lisp` or some such, along with whatever other general-purpose type-related predicates or type aliases I might find.

On the one hand, it's nice to have them defined in the same file where they are used, and for highly-specific types that seems like the best way to go, but for general-purpose "list/sequence of foo" types like this, there is also something to be said for having them live in a single source file for easy discovery.

ecpeterson: I think the last paragraph is a sound idea. `optimal-2q-target-atom` (or whatever it's called) doesn't belong in `types.lisp`, but `string-list-p` sure would.

appleby: Ok, I've added it to my todo list, above. I will scrounge around in the sources and see what other generic types / type predicates I can find and move them into a new file.  Anyone care what the file is called?

- `types.lisp`
- `lisp-types.lisp`
- `generic-types.lisp`
- `generic-lisp-types.lisp`
- `types-of-the-generic-persuasion-that-pertain-to-lisp-usage-but-certainly-not-quil-types.lisp`

tests/utilities.lisp
appleby: The prompting was driving me crazy while testing, so I added a keyword argument to allow the caller to `:skip-prompt`, if desired.

tests/utilities.lisp
```diff
@@ -33,3 +33,281 @@
...
+  (with-open-file (f file-name)
+    (parse-golden-file-stream f)))
+
+(defun update-golden-file-output-sections (file-paths output-callback
+                                           &key (if-exists ':supersede) skip-prompt)
+  "Update all output sections of the golden files at FILE-PATHS.
+
+If you call this function, examine the diffs of the old vs new output sections *carefully* before
+committing the updated files, in order to ensure that all the changes are intended or expected.
+Golden files are precious, and their sanctity must be preserved!
```
appleby: I have tried to follow my own advice here and examine the "expected output" sections of the new golden files carefully to make sure they are correct, but I may have missed something. Extra care from reviewers in verifying that the new golden files contain correct outputs is greatly appreciated.

tests/printer-test-files/basics.quil
```diff
@@ -0,0 +1,143 @@
...
+
+# Output
+I 0
+X 1
+Y 2
+Z 0
+H 1
+RX(pi) 0
+RY(-1.0) 1
+RZ(pi/2) 2
```
appleby: Note that this `pi/2` has not been evaluated to a float, but an expression like `2*pi` does get evaluated to `6.283185307179586`. Expected? I assumed this was a readability convenience since fractional `pi` angles are probably common, but haven't looked into the source of it yet.

tests/printer-test-files/classical.quil
```diff
@@ -0,0 +1,65 @@
...
+NEG int[0]
+NOT ro[1]
+MOVE int[2] 1
+EXCHANGE int[0] int[1]
+CONVERT ro[0] int[0]
+LOAD b[0] b int[0]
+STORE b int[0] b[0]
+
+# Input
+# Stolen from tests/good-test-files/good-classical-binaries.quil
```
appleby: I am considering adding a function to pilfer golden inputs from other quil test files in the test suite (`good-test-files`, `compiler-hook-test-files`). The idea is that the function could be called interactively and would copy the quil source from those other test files and turn them into golden printer tests. I would call it something like `plunder-gold-from-yonder-files` (working title).

Does this sound like a good idea or overkill? What say ye?

- [ ] Yea!
- [ ] Nay!
- [ ] Meh.

tests/printer-test-files/decomposed-ccnot.quil
```diff
@@ -0,0 +1,55 @@
...
+    CNOT-DT-T-CNOT-T q1 q2 q3
+    CNOT-DT-T-CNOT-T q1 q2 q3
+    T q2
+    H q3
+    CNOT q1 q2
+    T q1
+    DAGGER T q2
+    CNOT q1 q2
+
+DECOMPOSED-CCNOT 0 1 2
```
appleby: Calling @ecpeterson 

So the current test is just parsing and printing the above Quil and making sure it does the expansion of the DEFCIRCUIT.

But I also noticed that the compiler is awesome and does indeed compile this DECOMPOSED-CCNOT into the equivalent of good ol' `CCNOT 0 1 2`, by which I mean the matrices returned by `parsed-program-to-logical-matrix` for the two programs are `matrix-equal-dwim`.

In fact, in this case, not only are the logical matrices "equal" but `compiler-hook` emits a nearly identical instruction sequence. (I think "equivalent up to phase" is the phrase I want? Not sure. Basically, the instruction sequences are identical, modulo the `%theta` parameters to the R* gates). The fact that the logical matrices were equal didn't surprise me (I guess that means the compiler is working!), but the fact that the instruction sequences were nearly identical did. But maybe the one implies the other?

Anyway, getting to the point: I looked around for tests that compare the output of `compiler-hook` when given two equivalent programs but with different Quil sources, but didn't find any such tests. Would it make sense to add some (assuming I didn't just miss them), or did I get lucky with `CCNOT` and `DECOMPOSED-CCNOT`?

appleby:
Removed the "do not review" tag from the headline. This PR is still not complete, but it's getting close enough that it's ready for review.

appleby:
The latest commit adds abstraction around the conversion of Quil functions and arithmetic operators to/from lisp symbols. Debatable whether the cure is worse than the disease here, but it at least centralizes the knowledge about how to map between Quil and lisp to a single place and avoids the possibility of `evaluate-delayed-expression`, say, getting out of sync with what the parser accepts.

src/parser.lisp
```diff
@@ -1515,25 +1515,79 @@ INPUT-STRING that triggered the condition."
...
+    (defun quil-infix-operator->lisp-symbol (quil-infix-operator)
+      (quil->lisp quil-infix-operator +quil<->lisp-infix-arithmetic-operators+))
+
+    (defun lisp-symbol->quil-function (symbol)
+      (lisp->quil symbol +quil<->lisp-functions+))
+
+    (defun quil-function->lisp-symbol (quil-function)
+      (quil->lisp quil-function +quil<->lisp-functions+))
+
+    (defun lisp-symbol->quil-function-or-prefix-operator (symbol)
```
appleby: If we're not happy with the spew of long-winded function names, another option is to replace all these quil->lisp and lisp->quil functions with a single function that dispatches on the type of it's first argument. If you pass it a string, it knows you want to do quil->lisp conversion. If you pass it a symbol, lisp->quil.  The function could accept an optional second argument to allow the caller to filter the results down to only the context they want, i.e. `:prefix`, `:infix`, or `:function`, or any combination of those.

The three separate alists could also be collapsed down to single list of plists which also contains the context where the function/operator is valid. Something like:

```lisp
(alexandria:define-constant +quil<->lisp-functions+
      '((:QUIL "SIN" :LISP cl:sin :CONTEXT :FUNCTION)
        (:QUIL "COS" :LISP cl:cos :CONTEXT :FUNCTION)
	...
        (:QUIL "-" :LISP cl:cis :CONTEXT (:PREFIX :INFIX)
        (:QUIL "+" :LISP cl:+ :CONTEXT :INFIX)
        ...)
    :test #'equal
    :documentation
    "Functions usable from within Quil, and their associated Lisp function symbols.")

(defun quil<->lisp-dwim (key &optional (context :any))
  (etypecase key
    (string
     ;; lookup lisp symbol for Quil string conditioned on context
     )
    (symbol
     ;; lookup Quil string for lisp symbol conditioned on context
     )))
```

src/ast.lisp
```diff
@@ -1151,11 +1155,11 @@ For example,
                     ((= (length expr) 3)
                      (format stream "(~a~a~a)"
                              (print-delayed-expression (second expr) nil)
-                             (first expr)
+                             (lisp-symbol->quil-infix-operator (first expr))
```
appleby: Note that we don't check for a valid operator here first (likewise just below). We are relying on the fact that the delayed-expression for sure :100:% has a lisp symbol that maps to a valid Quil operator in the CAR and not, say, CL:GET-INTERNAL-RUN-TIME. If lookup fails, `lisp-symbol->quil-infix-operator` will just return NIL.

tests/printer-test-files/decomposed-ccnot.quil
appleby: It seems I did not search carefully enough. It looks like `compiler-hook-tests.lisp` has a handful of such tests, e.g. `test-parsed-program-to-logical-matrix-cnot-rewiring`, `test-parsed-program-to-logical-matrix-swap-rewiring`, etc.

appleby: I take it back. The above mentioned tests seem to be concerned specifically with rewirings, and don't pass the parsed-programs through `compiler-hook`.

tests/utilities.lisp
ecpeterson: I like `types.lisp` just fine :)

tests/printer-test-files/decomposed-ccnot.quil
ecpeterson: I think the similarity in the instruction sequences that you've noticed is partially a matter of luck and partially a matter of implementation decisions. The compiler is typically in one of two situations: it either knows exactly what to do (e.g., it recognizes a CCNOT and that it has a decomposition for CCNOT in its back pocket) or it has no idea what to do (e.g., you supply it a random gate, and so it relies on some perfectly generic scheme). It's pretty rare that, during the course of a single compilation, you jump from Scenario A to Scenario B—i.e., the execution paths are usually either all A or all B. In either case, you're likely to get a similar pattern of gates out (up to the precise value of parameters, which gets influenced by all sorts of things, from numerical jitter to LAPACK's decision to sneak an extra reflection into a matrix decomposition for no particular reason).

It's not a bad idea to implement the test you mention, where you check that two distinct instruction sequences encode the same unitary, though I think it'd primarily be a check that `make-matrix-from-quil` is performing as expected—which is certainly in the critical path, but it's also pretty low-risk.

N.B.: "Equivalent up to phase" means that the two unitaries U, V encoded by the two instructions / instruction sequences have the property U = cV for some complex scalar c. In some sense, this definition compares with the parameter situation in the opposite way to what you describe: the more efficient a circuit template (i.e., gate sequence w/o prespecified parameters) is at encoding different unitaries, the harder it will be to find different choices of angle values that give the same total unitary up to phase. Irredundancy pushes you in the other direction.

appleby:
Latest commit moves the existing general-purpose types and type predicates I could find into a new `types.lisp` file. This is a minimal refactoring that just moves any pre-existing stuff into a new file. I resisted the urge to do any further refactoring, pending feedback here.

src/ast.lisp
```diff
@@ -100,17 +100,17 @@ EXPRESSION should be an arithetic (Lisp) form which refers to LAMBDA-PARAMS."
 
 (defun make-delayed-expression (params lambda-params expression)
   "Make a DELAYED-EXPRESSION object initially with parameters PARAMS (probably a list of PARAM objects), lambda parameters LAMBDA-PARAMS, and the form EXPRESSION."
-  (assert (every #'symbolp lambda-params))
```
appleby: I can roll these back If ASSERTs are preferable to CHECK-TYPE.

src/define-pragma.lisp
```diff
@@ -79,16 +79,6 @@ elements to ELT-TYPE.
     (alexa:lexer-match-error (c)
       (quil-parse-error "Lexer failure: ~A" c))))
 
-(defun integer-sequence-p (object)
-  (and (typep object 'sequence)
-       (every #'integerp object)))
-
-(deftype integer-vector ()
```
appleby: `integer-vector` wasn't used anywhere, so I nixed it.

src/types.lisp
```diff
@@ -0,0 +1,35 @@
...
+              ,docstring
+              (and (typep object ',container-type)
+                   (every #',element-test object))))))
+    (def integer-list-p list integerp)
+    (def integeropt-vector-p vector integer-or-nil-p)
+    (def number-list-p list numberp)
+    (def string-sequence-p sequence stringp)
+    (def symbol-list-p list symbolp)))
+
+(deftype integer-list ()
```
appleby: This `integer-list` was the only one with an existing `deftype`, so I kept it that way.  If folks prefer, I could instead modify the above `macrolet` to define both a predicate and the corresponding `deftype`, which allow writing, e.g.,  `(check-type object symbol-list)` rather than `(check-type object (satisfies symbol-list-p)` everywhere.

appleby:
Removed the [WIP] qualifier since everything from the tasklist above is now "done", pending review feedback.

tests/utilities.lisp
appleby: Done.

appleby:
Note that there are still a couple of generic-ish types defined in `cl-quil` subpackages, but I didn't touch those because I didn't want to move them across package boundaries, viz.

```lisp
(in-package #:cl-quil.clifford)
(deftype unsigned-fixnum ()
  `(and fixnum unsigned-byte))
```

```lisp
(in-package #:cl-quil.resource)

(deftype bit-set ()
  "Representation of a set of non-negative integers (specifically BIT-SET-ELEMENT) as bits. "
  'integer)

(deftype bit-set-element ()
  "An index into a bit set."
  '(and fixnum unsigned-byte))
```

tests/printer-test-files/decomposed-ccnot.quil
appleby: Haha. So... the exact opposite of what I said. Got it! ;-)

Let me see if I can clarify, and I'll open an issue to track this idea.

So do I understand correctly that what you are saying is that two semantically equivalent Quil programs with different sources will not necessarily compile down to parsed-programs whose logical matrices satisfy `matrix-equality` (or even `matrix-equal-dwim`)? And instead the valid test would be merely parse the programs, then pass their instruction sequences to `make-matrix-from-quil` and check if the resulting matrices were equal?

In other words, this is bogus:

```lisp
(quil::matrix-equality
 (quil::parsed-program-to-logical-matrix
  (quil:compiler-hook (quil:parse-quil program-a) some-chip-spec))
 (quil::parsed-program-to-logical-matrix
  (quil::compiler-hook (quil:parse-quil program-b) some-chip-spec)))
```

But this should work:

```lisp
(quil::matrix-equality
 (quil::make-matrix-from-quil
  (coerce (quil::parsed-program-executable-code
           (quil:parse-quil program-a))
          'list))
 (quil::make-matrix-from-quil
  (coerce (quil::parsed-program-executable-code
           (quil:parse-quil program-b))
          'list)))
```

stylewarning:
@appleby Changing cl-quil::unsigned-fixnum is OK, but changing bit-set-element wouldn't be OK.

src/ast.lisp
stylewarning: I don't like `check-type` for funky types. I'd prefer `assert` here _unless_ we are using `symbol-list-p` extensively elsewhere.

stylewarning: (And it looks like it's being used sufficiently much elsewhere, so it's OK.)

src/ast.lisp
```diff
@@ -152,8 +152,12 @@ EXPRESSION should be an arithetic (Lisp) form which refers to LAMBDA-PARAMS."
 
 ;;;;;;;;;;;;;;;;;;;;; Comment protocol for syntax tree objects  ;;;;;;;;;;;;;;;;;;;;
 
+(eval-when (:compile-toplevel :load-toplevel :execute)
+  (defun make-comment-table ()
```
stylewarning: Can we comment on the keys and values of this table?

src/parser.lisp
```diff
@@ -1515,25 +1515,79 @@ INPUT-STRING that triggered the condition."
...
+      '(("+" . cl:+)
+        ("-" . cl:-)
+        ("/" . cl:/)
+        ("*" . cl:*)
+        ("^" . cl:expt))
+    :test #'equal
+    :documentation
+    "Infix arithmetic operators usable from within Quil, and their associated Lisp function symbols.")
+
+  (flet ((lisp->quil (lisp-symbol alist)
```
stylewarning: I would just define these as toplevel functions prepended with `%`, instead of closing over the ones below.

src/parser.lisp
```diff
@@ -1515,25 +1515,79 @@ INPUT-STRING that triggered the condition."
...
+           (check-type lisp-symbol symbol)
+           (alexandria:when-let ((found (rassoc lisp-symbol alist :test #'eq)))
+             (car found)))
+         (quil->lisp (quil-string alist)
+           (check-type quil-string string)
+           (alexandria:when-let ((found (assoc quil-string alist :test #'string-equal)))
+             (cdr found))))
+
```
stylewarning: Could you write a longform comment about what's going on here, what "prefix operator" "lisp symbol" "infix operator" etc etc mean and why there are so many functions?

I'm not disagreeing with it at all, but I want to give the reader more of a hint as to what's about to go down.

src/parser.lisp
stylewarning: I'm OK _not_ doing this.

src/parser.lisp
```diff
@@ -1515,25 +1515,79 @@ INPUT-STRING that triggered the condition."
...
-    "Functions usable from within Quil, and their associated Lisp function symbols.")
-
   (defun validate-function (func-name)
-    "Validate the function named FUNC-NAME against *VALID-FUNCTIONS*. Signal a QUIL-PARSE-ERROR if it's invalid."
-    (let ((found (assoc func-name *valid-functions* :test #'string-equal)))
-      (if found
-          (second found)
-          (quil-parse-error "Invalid function name: ~A" func-name))))
+    "Return the lisp symbol that corresponds to the Quil function named FUNC-NAME, or signal a QUIL-PARSE-ERROR if FUNC-NAME is invalid."
+    (alexandria:if-let ((lisp-symbol (quil-function->lisp-symbol func-name)))
```
stylewarning: I'm ok with using `or` for this idiom here. `(or value error)`

appleby:
> @appleby Changing cl-quil::unsigned-fixnum is OK, but changing bit-set-element wouldn't be OK.

Ok, I will update the PR with this change. I left it out originally because I didn't see any references to `cl-quil` symbols in `cl-quil.clifford` at first, but on closer inspection I realize that I've already dirtied the waters with `cl-quil::positive-power-of-two-p` in `clifford/benchmarking-procedures.lisp`...

src/types.lisp
```diff
@@ -0,0 +1,35 @@
...
+;;;; Author: Appleby
+;;;;
+;;;; This file contains general-purpose types and type-predicates. Types that are more specialized
+;;;; do not belong here and should live in the file or package where they are most used,
+;;;; e.g. OPTIMAL-2Q-TARGET-ATOM in compilers/approx.lisp.
+
+(in-package #:cl-quil)
+
+(eval-when (:compile-toplevel :load-toplevel :execute)
+  (defun integer-or-nil-p (object)
```
stylewarning: The _type_ of this is `(or null integer)`. But I can see why it's not used below.

src/types.lisp
```diff
@@ -0,0 +1,35 @@
...
+    "Return true if OBJECT is NIL or satisfies INTEGERP."
+    (or (not object) (integerp object)))
```
stylewarning: use `null` not `not` to check if something is equal to `nil`.

src/types.lisp
```diff
@@ -0,0 +1,35 @@
...
+         (check-type container-type symbol)
+         (check-type element-test symbol)
+         (let ((docstring
+                 (format nil "Return true if OBJECT is a ~A and every element satisfies ~A."
+                         container-type element-test)))
+           `(defun ,predicate-name (object)
+              ,docstring
+              (and (typep object ',container-type)
+                   (every #',element-test object))))))
+    (def integer-list-p list integerp)
```
stylewarning: So, in Lisp, the `list` type is lousy. It's defined as
```
(deftype list () `(or null cons))
```
You should use `alexandria:proper-list`.

_HOWEVER_, since in _this case_, you're following it with a call to _every_, it's OK. As such, could you write a comment here saying why `list` us justified and `alexandria:proper-list` isn't strictly necessary?

src/types.lisp
```diff
@@ -0,0 +1,35 @@
...
+    (def integeropt-vector-p vector integer-or-nil-p)
```
stylewarning: could u align these columns plz

src/types.lisp
```diff
@@ -0,0 +1,35 @@
...
+
+(in-package #:cl-quil)
+
+(eval-when (:compile-toplevel :load-toplevel :execute)
+  (defun integer-or-nil-p (object)
+    "Return true if OBJECT is NIL or satisfies INTEGERP."
+    (or (not object) (integerp object)))
+
+  (macrolet
+      ((def (predicate-name container-type element-test)
```
stylewarning: I myself would have preferred
```
(def predicate name container-type element-type)
```
then used `typep` in the body. That way you can nix the function above and use `(or null integer)`.

src/types.lisp
```diff
@@ -0,0 +1,35 @@
...
+         (check-type predicate-name symbol)
+         (check-type container-type symbol)
+         (check-type element-test symbol)
+         (let ((docstring
+                 (format nil "Return true if OBJECT is a ~A and every element satisfies ~A."
```
stylewarning: I would prefer this be framed as a question:
```
"Is the OBJECT a ~A and ...?"
```

src/types.lisp
stylewarning: I think it would be cool to have that.
```
(def integer-list integer-list-p   list integer)
```
Make sure to actually state the type name and the predicate name in the `def`!

tests/utilities.lisp
```diff
@@ -33,3 +33,266 @@
...
+    (loop :with last := (length strings)
+          :with delimiter-string := (string delimiter)
+          :initially (and prefix-p (write-string delimiter-string stream))
+          :for str :in (coerce strings 'list)
+          :for i :upfrom 1
+          :do (write-string str stream)
+          :when (or (< i last) suffix-p)
+            :do (write-string delimiter-string stream))))
+
+;;;
```
stylewarning: nice comment

tests/printer-tests.lisp
```diff
@@ -0,0 +1,127 @@
...
+
+(defun reset-comment-table-hack ()
+  "Reset the CL-QUIL::**COMMENTS** hash table.
+
+  Resetting the **COMMENTS** table might be required for any test that wants to validate the output
+  of PRINT-PARSED-PROGRAM for programs that contain HALT, NOP, RESET, or other instructions that are
+  represented by singleton classes. This is because the **COMMENTS** hash table is keyed on object
+  identity, so if any prior tests have attached rewiring comments to any singleton instructions,
+  they will persist and be printed by PRINT-PARSED-PROGRAM."
+  (setf quil::**comments** (quil::make-comment-table)))
```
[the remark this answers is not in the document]
appleby: I've opened #270 to track fixing this.

src/types.lisp
appleby: I actually considered adding a

```lisp
(deftype nullable (type-spec)
  `(or null ,type-spec))
```
to this file and using it everywhere. But then I realized that `(or null ...)` is actually one character shorter than `(nullable ...)` and managed to contain myself...

src/types.lisp
appleby: Actually, I did that reflexively on the first pass, then rolled it back so as not to change the original definition. Now, with your blessing, I will make this change and secure a good night's rest for myself. Thank you.

src/types.lisp
appleby: I like the way you think. After all, you can always pass `(satisfies foo-p)` if you want a random predicate. I did it like this to most closely mimic the existing definitions, which were all using raw predicates rather than `typep` + type specifiers.

src/types.lisp
appleby: Hahaha. What, no magical finagling of the predicate name based on the type name, but making sure to do the right thing w.r.t. `*readtable-case*`? Or did you just want them both to be easily greppable? Either way, I'd like to subscribe to your newsletter.

tests/printer-test-files/decomposed-ccnot.quil
ecpeterson: `matrix-equality` isn't guaranteed work (in either case!), but `matrix-equals-dwim` (or `operator=`) will work. The names of these predicates could stand to be improved, but here's what they do:

* `double=` is used to test (complex) float equality throughout the code, except for when checking output.
* `double~` is used to test (complex) float equality on the output matrices. Its tolerance threshold is greater than that of `double=`: `double=` gets used to make thousands of substitution decisions through the course of a quilc run, and so `double=` is thousands of times more sensitive than `double~` to make up for it.
* `matrix-equality`: Tests whether A is entrywise `double~` to B.
* `operator=`: Checks whether there is a complex number c (the "global phase difference") so that c*A and B are entrywise `double~`. (This is the definition of operational equality in quantum mechanics; it's a fact of life that quantum operations admit their 'easiest' expression in terms of matrices, but in trade for this easy expression you have to use this clumsy notion of equality.)
* `matrix-equals-dwim`: The same as `operator=`, _except_ when `*enable-state-prep-compression*` is turned on, in which case the compressor is only guaranteed to emit a sequence of instructions whose corresponding operator passes `operator=` only with the leftmost column of the input matrix, and so `matrix-equals-dwim` tests only this column when this flag is high.

The last one is usually what you want.


While we're at it: `parsed-program-to-logical-matrix` is `make-matrix-from-quil` + sensitivity to rewiring comments. This extra caveat means that `parsed-program-to-logical-matrix` will always emit the same matrix for Quil passed through the addresser; all the other subroutines in quilc don't require this extra sensitivity, and `m-m-f-q` will work just fine.

src/ast.lisp
ecpeterson: The keys are instruction objects and the values are strings to be attached as comments to their parsed-program-printed representation. (I'm the one who didn't write a docstring; sorry, @appleby.)

src/ast.lisp
ecpeterson: Is there a reason not to throw an error in `l-s->q-i-o` if it tries to lookup a nonexistent key? That seems like a sane enough thing to do to me.

src/parser.lisp
ecpeterson: I also have a preference for tucking auxiliary functions into a local namespace, but I usually cede to Robert's counter-preference and it hasn't yet bitten me.

tests/printer-test-files/basics.quil
ecpeterson: The parser evaluates things to floats when it can, and (by default, though this is disable-able) we print out things that are near-ish to particular rational multiples of pi as such, since the (default) human user probably finds that helpful. The relevant flag is `*print-fractional-radians*`.

I'd indeed expect 2pi to be on such rational multiple of pi that we choose to pretty-print.

tests/printer-tests.lisp
ecpeterson: Unless there's a specific reason to choose the other, I'd prefer `parse-quil`, since it's what's used more commonly elsewhere.

ecpeterson:
I see that Robert has left some comments that are still unaddressed, but largely things look pretty good here!

appleby:
Latest commits address (I think) requested review changes, plus add some bells & whistles like tracking originating file and line numbers when parsing golden files for improved debugging messages on test failures.

tests/printer-test-files/decomposed-ccnot.quil
appleby: As usual, thank you for your patient explanation. I will go ahead and open an issue to track this idea, although I suspect I will struggle to come up with many examples of programs with different sources but equivalent "meaning". But that sounds like a fun exercise for me to get more familiar with Quil and quantum programming.

ecpeterson: You can get a head start by looking at `rewriting-rules.lisp`, which is full of little not-obviously-equal Quil snippets that are equivalent to each other in the `matrix-equals-dwim` sense.

tests/printer-tests.lisp
appleby: In the latest round of commits, I've added keyword options to `parse-and-print-quil-to-string` that allows the caller to specify the parser and printers they want, defaulting to `parse-quil` and `print-parsed-program`, respectively.

The golden file tests are still using the default `parse-quil`.  I tend to agree with @ecpeterson that `parse-quil` appears to be the user-visible exported interface and also winds up testing more code paths since it does the standard transforms. So far, I haven't run into any test cases where that is a problem.

tests/printer-test-files/basics.quil
appleby: I will make a note to look into this.

tests/printer-test-files/classical.quil
appleby: I think the thumbs-up means this has one "Yea!" vote. I assume no response from the other reviewers is an implicit "Meh." vote :).

Thinking about it more, I'm not sure this really needs to be an automated process, depending on how often tests cases get added to the other tests. I'm also not sure all of the test cases from those other directories will map to reasonable printer tests (might contain expressions that evaluate to crazy floats, for example).

This might be a "solution in search of a problem".  But there's no denying I really want to write a function called `plunder-gold-...`.  I will do a closer review of the test cases in question and circle back.

src/ast.lisp
appleby: It would be possible, though it would complicate the implementation `valid-quil-function-or-operator-p`, which currently just tries all the lookup functions in order.

```lisp
(defun valid-quil-function-or-operator-p (symbol)
    (not (null (or (lisp-symbol->quil-function symbol)
                   (lisp-symbol->quil-prefix-operator symbol)
                   (lisp-symbol->quil-infix-operator symbol)))))
```

In general, when adding an abstraction layer like this, I like to keep it purely functional if possible and allow the caller to decide how to handle errors if they want to. For example, if parsing fails, you probably want to signal a "parse-error", but in other places, maybe a different error.

I did consider adding another set of functions with the same name but with `-or-error` or `-or-lose` or some such appended, and using it here. In this case, that would map to `lisp-symbol->quil-infix-operator-or-error`, to make it easy for callers that are bothered about a custom error condition. I suppose the `on-error` variants could allow the caller to pass in the condition they want on failure.

src/types.lisp
appleby: Done.

appleby:
> @appleby Changing cl-quil::unsigned-fixnum is OK, but changing bit-set-element wouldn't be OK.

This is done.

src/parser.lisp
appleby: Done.

src/parser.lisp
appleby: Done.

src/types.lisp
appleby: I ended up going with your other suggestion of just passing a type spec rather than a predicate and nixing this altogether.

src/types.lisp
appleby: Done.

src/types.lisp
appleby: Note sure I understood correctly, but I think I did what you're asking in the latest round of commits.

src/types.lisp
appleby: Done.

src/types.lisp
appleby: Done.

tests/utilities.lisp
appleby: Thanks for noticing.

tests/printer-tests.lisp
```diff
@@ -0,0 +1,143 @@
...
+TEST(0.5) 0 1
+")
+         (after (parse-and-print-quil-to-string before :parser #'quil::parse-quil-into-raw-program)))
+    (is (string= before after))))
+
+(deftest test-jump-to-integer-label-printing ()
+  "Ensure that JUMP instructions with integer LABELs are printed correctly."
+  (is (string= (quil::print-instruction-to-string
+                (quil::make-instance 'quil::unconditional-jump :label 42))
+               "JUMP {absolute address 42}"))
```
appleby: I could not figure out how to ram these through the parser, so I'm just constructing and printing them directly.

tests/utilities.lisp
```diff
@@ -33,3 +33,335 @@
...
+                     (push pending-test-case test-cases)
+                     (return (nreverse test-cases))))))
+
+(defun parse-golden-file (file-name)
+  "Convenience wrapper around PARSE-GOLDEN-STREAM."
+  (check-type file-name (or string pathname))
+  (with-open-file (f file-name)
+    (parse-golden-file-stream f)))
+
+(defun map-golden-files-and-test-cases (test-case-function golden-files &optional (stream t))
```
appleby: I'm not totally satisfied with this function for various reasons, but it does clean up `test-print-parsed-program-golden-files` over in `printer-tests.lisp`, so leaving it in for consideration.

tests/printer-test-files/classical.quil
appleby: My pirate's raid to plunder golden inputs has yielded a potential bug! See #272.

I think this PR is big and hairy enough as it is (100+ comments!), so I've spun the idea of plundering gold out into a separate issue (#273) that can happen in a follow-on PR, assuming we decide it's worthwhile to do it.

tests/printer-test-files/decomposed-ccnot.quil
appleby: This is now #274

tests/utilities.lisp
```diff
@@ -33,3 +33,335 @@
...
+TEST-CASE-FUNCTION is function that takes a single GOLDEN-TEST-CASE argument.
+
+GOLDEN-FILES is a LIST of PATHNAMEs indicating golden files to test.
+
+STREAM is an optional OUTPUT-STREAM for writing debug messages indicating the current file and test
+case being processed. STREAM defaults to T."
+  (dolist (file golden-files)
+    (format stream "~&    Testing golden file ~A at line:" (pathname-name file))
+    (dolist (test-case (parse-golden-file file))
+      (format stream " ~D" (golden-test-case-line test-case))
```
appleby: Reporting the line number of the `test-case` here would not really be necessary if all of the `fiasco` assertions allowed specifying an additional message to be printed on failure like you can with the `fiasco:is` assertion. Currently, there doesn't appear to be a way to provide a detailed message for `fiasco:signals` / `not-signals`, so I report the test-case line number here in case one of those other assertions fails, to make it easier to track down the offending input section.

tests/printer-test-files/classical.quil
stylewarning: I agree, we should close this PR off and do follow ups

src/types.lisp
stylewarning: Easy grepping. I'm not a huge fan of making symbols that don't appear in the source, unless it's absolutely necessary.

src/parser.lisp
stylewarning: it's easier to debug when you got the functions available

tests/printer-tests.lisp
stylewarning: this is where transforms are biting, `patch-labels` is being called and it makes the Quil program unprintable.

src/parser.lisp
```diff
@@ -1515,25 +1515,113 @@ INPUT-STRING that triggered the condition."
...
+    (check-type lisp-symbol symbol)
+    (alexandria:when-let ((found (rassoc lisp-symbol alist :test #'eq)))
+      (car found)))
+
+  (defun %quil->lisp (quil-string alist)
+    (check-type quil-string string)
+    (alexandria:when-let ((found (assoc quil-string alist :test #'string-equal)))
+      (cdr found)))
+
+  ;; The following functions handle conversion between Quil's arithmetic operators/functions and the
```
stylewarning: Use 3-semicolons even tho we are in an eval when, we are still toplevel

src/types.lisp
```diff
@@ -0,0 +1,45 @@
...
+                              (typep element ',element-type))
+                            object)))
+              (deftype ,type-name ()
+                ,deftype-docstring
+                '(satisfies ,predicate-name))))))
+
+    ;; Normally, we'd want ALEXANDRIA:PROPER-LIST in place of LIST in the definitions
+    ;; below. However, because the predicate we define calls EVERY on the OBJECT (which will error
+    ;; if given an improper list), we're OK to use LISTs here.
+    (def integer-list      integer-list-p      list     integer)
```
stylewarning: look at the quality of this tabular data

notmgsk:
This PR really snowballed, but Let's go to LGTM.

stylewarning:
Also, now there are merge conflicts. RIP.

appleby:
> This PR really snowballed, but Let's go to LGTM.

My whole strategy to get code through review is wear you all down with too many comments in a war of attrition.

appleby:
> Also, now there are merge conflicts. RIP.

I was anticipating this, so I waited till after lunch.

src/ast.lisp
appleby: Anyone worried enough they want to see error checking here before merge? I suppose another option is to let the caller specify via keyword argument what behavior they want on lookup failure, i.e.: 1) return nil, 2) default error, 3) custom condition.

ecpeterson: 0 worries here

appleby: > 0 worries here

`mplayer ~/music/bob-marley/three-little-birds.mp3`

src/parser.lisp
appleby: Done.

src/ast.lisp
notmgsk: > > 0 worries here
> 
> `mplayer ~/music/bob-marley/three-little-birds.mp3`

Classic @appleby

stylewarning:
@appleby I hate to do it, but do you mind if i squash?

src/ast.lisp
appleby: > > > 0 worries here
> > 
> > 
> > `mplayer ~/music/bob-marley/three-little-birds.mp3`
> 
> Classic @appleby

Phew. For a second there I was worried no one would get the joke.

appleby:
> @appleby I hate to do it, but do you mind if i squash?

Please do. I only switched over from force-pushing to micro commits because I assumed everyone was tired of staring at giant mutating diff.

PR #260, opened 2019-05-27
buildsystem: add example makefile for interop C with Lisp

dkochmanski:
- shared library
- static library
- example run

### DIFF
--- a/elf-library/Makefile
+++ b/elf-library/Makefile
@@ -1,0 +1,26 @@
+all: quil_example.out quil_example_static.out
+
+clean:
+	rm *.so *.a *.out
+
+run_shared:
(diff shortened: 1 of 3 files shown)


dkochmanski:
This example code and Makefile show how to build static and shared libraries and how to interoperate between C and Lisp (with a handler case example). It is noteworthy that application startup is slow, because top-level forms of all dependencies must be replayed. It is also worth keeping in mind, that this to run requires:
- newest cffi from repository
- newest ecl from repository

dkochmanski:
Ah, also you may call strip on binaries, then they come down from 80MB to 23MB.

stylewarning:
highly cool! Will merge, and we can iterate/change names/whatever. Thanks!

PR #261, opened 2019-05-27
Convert EXHAUST-STACK in COMPILATION-HOOK from tail-recursion to an explicit LOOP

appleby:
Refactor OUTER-INSTRUCTION-LOOP in ALGEBRAICALLY-REDUCE-INSTRUCTIONS and EXHAUST-STACK in COMPILATION-HOOK from tail-recursive calls to explicit LOOPs.

I started working on this to try and get the test suite passing on ECL. However, it looks like there might be more places that need converting (e.g. OPTIMAL-2Q-COMPILER, EXPAND-TO-NATIVE-INSTRUCTIONS, possibly others) so I'm posting here to see if folks want to try and continue down this path of selectively converting tail-recursive functions into LOOPs wherever possible, or if we want to try another tack. If there are only a handful of these tail-recursive functions and this seems like a reasonable approach, then I can try refactoring EXPAND-TO-NATIVE-INSTRUCTIONS next. On the other hand, if we expect there are many such tail-recursive functions (or, indeed, non-tail recursions that will be harder to convert to LOOPs), then maybe some other approach makes more sense.

Note that in the case of OUTER-INSTRUCTION-LOOP, changing it to an explicit LOOP also avoids a stack overflow on SBCL for me when compiling with `(optimize debug)`, e.g. via `C-u C-c C-c` in SLIME.

EXHAUST-STACK was minding it's own business, but I happened to notice it while working on a different PR the other day and it was trivial to convert it to a LOOP, so I threw it in for good measure.

Here are the individual commit messages. These commits get riskier as you go. The first three commits seem relatively safe. The last one I am less sure of, both because I'm basically doing a mechanical refactoring with only a high-level understanding of what ALGEBRAICALLY-REDUCE-INSTRUCTIONS is doing, and because it's debatable whether this final refactoring helps or hurts readability. I left it as the last commit so it's easy to drop if nobody likes it.

**Cleanup whitespace in compilation-methods.lisp**

**Convert EXHAUST-STACK in COMPILER-HOOK to an explicit LOOP**

**Convert OUTER-INSTRUCTION-LOOP to an explicit LOOP**

Convert the locally defined function OUTER-INSTRUCTION-LOOP in ALGEBRAICALLY-REDUCE-INSTRUCTIONS from tail calls to an explicit loop.

When compiling with (optimize debug) on SBCL, certain tests (I'm looking at *you* TEST-ABSOLUTE-UNIT-CNOT-COMPILATION) would previously result in stack overflow on my machine, presumably because SBCL disables tail-call optimization when compiling with max debug settings. After this change, no stack overflow occurs.

This commit represents a minimal translation from tail calls to an explicit loop, changing as little as possible in the process in order to decrease the likelihood of introducing bugs.

**Further refactoring of OUTER-INSTRUCTION-LOOP**

Add an additional named inner LOOP and scrunch more of the bindings and control logic of OUTER-INSTRUCTION-LOOP into the inner LOOP's loop control clauses. This allows to get rid of some of the RETURN and RETURN-FROM forms, and to do away with the explicit outer BLOCK.

### DIFF
--- a/src/compilation-methods.lisp
+++ b/src/compilation-methods.lisp
@@ -52,7 +52,7 @@
 (defun apply-translation-compilers (instruction chip-spec hardware-object)
   "Wrapper function that calls the compilers associated to HARDWARE-OBJECT and the generic compilers associated to CHIP-SPEC in precedence order, returning the first found expansion of INSTRUCTION as a sequence."
   (labels ((try-compiler (compilation-method)
-             "Applies COMPILATION-METHOD to INSTRUCTION. If it succeeds, end 
+             "Applies COMPILATION-METHOD to INSTRUCTION. If it succeeds, end
               the whole procedure and return the resulting instruction sequence.
               If it fails, cede control by returning NIL."
              (restart-case
@@ -90,7 +90,7 @@
 (defun check-protoquil-program (program)
   "Checks that PROGRAM, an application-resolved parsed-program instance, conforms to the present definition of \"protoQuil\". Signals an error on failure."
   ;; a protoquil program carves up into 3 regions, each optional, of the form:
-  ;; (1) a RESET instruction 
+  ;; (1) a RESET instruction
   ;; (2) quantum gates, perhaps with classical memory references or classical arithmetic
   ;; (3) a collection of MEASURE instructions
   (let ((region-counter 1))
@@ -133,25 +133,25 @@ Returns a value list: (processed-program, of type parsed-program
                        topological-swaps, of type integer
                        unpreserved-block-duration, of type real)"
   (format *compiler-noise-stream* "COMPILER-HOOK: entrance.~%")
-  
+
   (warm-chip-spec-lookup-cache chip-specification)
-  
+
   ;; start by doing some basic expansion transformations
   (transform 'resolve-applications parsed-program)
-  
+
   ;; we disallow compilation of programs that use memory aliasing
   (loop :for mdesc :in (parsed-program-memory-definitions parsed-program)
         :when (memory-descriptor-sharing-parent mdesc)
           :do (error "Programs with aliased memory are currently unsupported."))
-  
+
   ;; check that the program obeys the dead qubit rule
   (when (eql ':naive rewiring-type)
     (check-program-skips-dead-qubits parsed-program chip-specification))
-  
+
   ;; check that a protoquil program is in fact protoquil
   (when protoquil
     (check-protoquil-program parsed-program))
-  
+
   ;; now we walk the CFG associated to the program
   (let* ((initial-rewiring (prog-initial-rewiring parsed-program chip-specification
                                                   :type rewiring-type))
@@ -160,7 +160,7 @@ Returns a value list: (processed-program, of type parsed-program
          (block-stack (list (list (entry-point cfg) nil)))
          (topological-swaps 0)
          (unpreserved-duration 0))
-    
+
     (let ((*print-pretty* nil))
       (format *compiler-noise-stream* "COMPILER-HOOK: initial rewiring ~a~%" initial-rewiring))
 
@@ -190,7 +190,7 @@ Returns a value list: (processed-program, of type parsed-program
         (push new-final-blk (cfg-blocks cfg))
         (link-blocks new-final-blk (make-instance 'terminating-edge))
         (link-blocks final-blk (unconditional-edge new-final-blk))))
-    
+
     ;; these local functions describe how we traverse / modify the CFG.
     (labels
         ;; this function introduces a new block that cajoles the compiler into
@@ -215,15 +215,15 @@ Returns a value list: (processed-program, of type parsed-program
              (push (list fresh-block registrant) block-stack)
              (format *compiler-noise-stream* "COMPILER-HOOK: Introduced ~a to deal with the rewiring.~%"
                      (basic-block-name fresh-block))))
-         
+
          (touch-preserved-block (blk)
            ;; if so, then we don't have any business compiling it. treat
            ;; it as marked, with the identity rewiring on both ends,
-           ;; and proceed 
+           ;; and proceed
            (setf (basic-block-in-rewiring blk) (make-rewiring (chip-spec-n-qubits chip-specification)))
            (setf (basic-block-out-rewiring blk) (make-rewiring (chip-spec-n-qubits chip-specification)))
            (change-class blk 'basic-block))
-         
+
          (touch-unpreserved-block (blk registrant)
            ;; actually process this block
            (multiple-value-bind (initial-l2p chip-schedule final-l2p)
@@ -252,7 +252,7 @@ Returns a value list: (processed-program, of type parsed-program
                (incf topological-swaps local-topological-swaps)
                (incf unpreserved-duration duration)
                (format *compiler-noise-stream* "COMPILER-HOOK: Done processing block ~a.~%" (basic-block-name blk)))))
-         
+
          (touch-reset-block (blk)
            ;; actually process this block
            (multiple-value-bind (initial-l2p chip-schedule final-l2p)
@@ -279,7 +279,7 @@ Returns a value list: (processed-program, of type parsed-program
                (incf topological-swaps local-topological-swaps)
                (incf unpreserved-duration duration)
                (format *compiler-noise-stream* "COMPILER-HOOK: Done processing block ~a.~%" (basic-block-name blk)))))
-         
+
          (process-block (blk registrant)
            ;; if this block is expecting a rewiring, we should make sure the
            ;; exit/enter rewirings match.
@@ -301,12 +301,12 @@ Returns a value list: (processed-program, of type parsed-program
                (unless (equalp final-l2p initial-l2p)
                  (return-from process-block
                    (edge-to-rewiring-block blk registrant initial-l2p)))))
-           
+
            ;; the source's exit rewiring now matches the target's entry rewiring.
            ;; if this block has already been visited, skip it.
            (when (basic-block-in-rewiring blk)
              (return-from process-block))
-           
+
            (let ((*print-pretty* nil))
              (format *compiler-noise-stream* "COMPILER-HOOK: Visiting ~a for the first time, coming from ~a (~a).~%"
                      (basic-block-name blk)
@@ -324,7 +324,7 @@ Returns a value list: (processed-program, of type parsed-program
              ((unconditional-edge target)
               (push (list target blk) block-stack))
              (terminating-edge nil))
-           
+
            ;; now fork based on whether the block is PRESERVEd.
            ;; note that touch-* will set block-initial-l2p, which indicates the block has been visited
            (typecase blk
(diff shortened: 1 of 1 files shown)


appleby:
Here is the stack overflow error and backtrace I get out of ECL when attempting to run the `cl-quil-tests` test suite with the changes from this PR applied. You can see that the culprit appears to be EXPAND-TO-NATIVE-INSTRUCTIONS.

```
  TEST-COMPILER-HOOK-RANDOM-4Q
    With *ENABLE-STATE-PREP-COMPRESSION* NIL
      Working on architecture CZ.


Condition of type: STACK-OVERFLOW
C-STACK overflow at size 1048576. Stack can probably be resized.
Proceed with caution.
Available restarts:

1. (CONTINUE) Extend stack size
2. (TRY-NEXT-COMPILER) Ignore this error and try the next compiler in the list.
3. (CONTINUE) Skip the rest of the test CL-QUIL-TESTS::TEST-COMPILER-HOOK-RANDOM-4Q and continue by returning (values)
4. (RETEST) Rerun the test CL-QUIL-TESTS::TEST-COMPILER-HOOK-RANDOM-4Q
5. (CONTINUE) Skip the rest of the test FIASCO-SUITES::CL-QUIL-TESTS and continue by returning (values)
6. (RETEST) Rerun the test FIASCO-SUITES::CL-QUIL-TESTS
7. (CONTINUE-WITHOUT-DEBUGGING) Turn off debugging for this test session and invoke the first CONTINUE restart
8. (CONTINUE-WITHOUT-DEBUGGING-ERRORS) Do not stop at unexpected errors for the rest of this test session and continue by invoking the first CONTINUE restart
9. (CONTINUE-WITHOUT-DEBUGGING-ASSERTIONS) Do not stop at failed assertions for the rest of this test session and continue by invoking the first CONTINUE restart
10. (ABORT-TESTING) Abort the entire test session
11. (RESTART-TOPLEVEL) Go back to Top-Level REPL.

Broken at CL-QUIL::GIVE-UP-COMPILATION. In: #<process TOP-LEVEL 0x5596c01b3f80>.
 File: "/home/ma/src/repos/rigetti/quilc/src/compilation-methods.lisp" (Position #1626)
>> :b

Backtrace:
  > CL-QUIL::GIVE-UP-COMPILATION
  > cl-quil::phase-to-rz
  > cl-quil::try-compiler
  > cl-quil::apply-translation-compilers
  > cl-quil::expand-to-native-instructions
  > cl-quil::expand-to-native-instructions
  > cl-quil::expand-to-native-instructions
  > cl-quil::expand-to-native-instructions
  > ...[many of these frames on stack]...
  > cl-quil::expand-to-native-instructions
  > cl-quil::touch-unpreserved-block
  > cl-quil::process-block
  > cl-quil::exhaust-stack
  > cl-quil:compiler-hook
  > cl-quil-tests::test-compiler-hook-random-4q
  > fiasco::run-test-body
  > #:g16
  > fiasco::call-with-test-handlers
  > fiasco::run-test-body-in-handlers
  > fiasco::pretty-run-test
  > #:body-sym256
  > cl-quil-tests::test-compiler-hook-random-4q
  > fiasco-suites::cl-quil-tests
  > fiasco::run-test-body
  > #:g16
  > fiasco::call-with-test-handlers
  > fiasco::run-test-body-in-handlers
  > fiasco::pretty-run-test
  > #:body-sym1
  > fiasco-suites::cl-quil-tests
  > fiasco:run-suite-tests
  > fiasco:run-tests
  > fiasco:run-package-tests
  > cl-quil-tests:run-cl-quil-tests
  > si:bytecodes [Evaluation of: (cl-quil-tests:run-cl-quil-tests)]
  > si:bytecodes [Evaluation of: (si:top-level t)]
```

appleby:
Here are the links if you want to directly compare the "minimal" vs "loop++" refactorings of 

[minimal first pass](https://github.com/rigetti/quilc/pull/261/commits/82c318e56f376176d7cb46787ba877fdd2bbe964)

[double the loops, double the fun](https://github.com/rigetti/quilc/pull/261/commits/97345e19e9d55d654415fb3fbf1f27c9d28f9346)

src/compressor/compressor.lisp
```diff
@@ -268,51 +268,45 @@ other's."
...
+                               ;; if we didn't find a new object but we haven't tried the
+                               ;; generic rules, try them now
+                               ((not exhausted-specializations)
+                                (setf exhausted-specializations t)
+                                (chip-specification-generic-rewriting-rules chip-specification))
+                               ;; otherwise, give up.
+                               (t nil)))
+                            (node-to-jump-to (apply-rules rewrite-rules nodes-for-inspection)))
+                         (setf node node-to-jump-to)
+                         (return-from inner-loop)))))))
```
appleby: Alternatively, I think this `(return-from inner-loop)` could be replaced with a simple `(loop-finish)`, which would also allow dropping the `:named` clause above. RETURN-FROM terminates the loop abruptly, skipping any `:finally` clause, whereas LOOP-FINISH does "normal" loop termination. This loop doesn't return any values or have a `:finally` clause, and I thought `(return-from inner-loop)` made it a little clearer which loop is being terminated, since there are nested loops.

appleby:
Note that this PR is essentially a subset of @dkochmanski's #262, which should probably be preferred since he has additionally converted EXPAND-TO-NATIVE-INSTRUCTIONS and WALK-GRAPH-WITH-CANDIDATES.

src/compilation-methods.lisp
```diff
@@ -334,14 +334,12 @@ Returns a value list: (processed-program, of type parsed-program
               (touch-reset-block blk))
              (otherwise
               (touch-unpreserved-block blk registrant))))
-         
+
          ;; this is the main loop that pushes through the CFG
          (exhaust-stack ()
-           (unless (endp block-stack)
```
stylewarning: This loop transformation I'm OK with. Both are equivalently readable to me, though I have heartfelt love for tail rec

src/compilation-methods.lisp
```diff
@@ -334,14 +334,12 @@ Returns a value list: (processed-program, of type parsed-program
...
-             (destructuring-bind (blk registrant) (pop block-stack)
-               (process-block blk registrant))
-             (exhaust-stack))))
-      
+           (loop :until (endp block-stack)
```
stylewarning: minor nit: for a single clause `:do` loop like this, put `:do` at the end of the line and indent the body as normal. (your editor should indent it like `DOTIMES` or similar)

src/compressor/compressor.lisp
```diff
@@ -268,51 +268,45 @@ other's."
...
-                         ;; otherwise, give up.
-                         (t nil))))
-                   (let ((node-to-jump-to (apply-rules rewrite-rules nodes-for-inspection)))
-                     (when node-to-jump-to
-                       (return-from outer-instruction-loop
-                         (outer-instruction-loop node-to-jump-to)))))))))
-         (outer-instruction-loop node)))
+       (outer-instruction-loop (input-node)
+         (loop
+           :for node = (peephole-rewriter-node-next input-node) :then (peephole-rewriter-node-next node)
```
stylewarning:
```suggestion
           :for node := (peephole-rewriter-node-next input-node) :then (peephole-rewriter-node-next node)
```

src/compressor/compressor.lisp
```diff
@@ -268,51 +268,45 @@ other's."
...
+           :while node
+           :do (update-context node)
+           :do (loop
```
stylewarning: I'm not too pleased about this loop but I don't have a lot of constructive criticism. What I do got is this:

- There's a lot of state being thrown around.
- There are multiple termination conditions: the `:for-:in`, the `:until`s, the `return-from`; makes for difficult reading.

I definitely feel like we've lost more than we've gained with this loop rewrite. Is there a better way we can refactor this? Maybe @ecpeterson can help.
Maybe @ecpeterson

src/compressor/compressor.lisp
```diff
@@ -268,51 +268,45 @@ other's."
...
+                 :named inner-loop
+                 :with nodes-for-inspection = (find-noncommuting-instructions node)
```
stylewarning:
```suggestion
                 :with nodes-for-inspection := (find-noncommuting-instructions node)
```

stylewarning: and the `= -> :=` below

src/compressor/compressor.lisp
```diff
@@ -268,51 +268,45 @@ other's."
...
+                 :with qubit-complex = nil
+                 :with exhausted-specializations = nil
+                 :for inspection-node :in nodes-for-inspection
+                 ;; for each instruction in the noncommuting tail...
+                 :for new-complex = (application-qubit-indices (peephole-rewriter-node-instr inspection-node))
+                 ;; if we've already hit the generic object, continue to the next list head element.
+                 :until exhausted-specializations
```
stylewarning: `:until (or ...)` instead of two `:until`s

src/compressor/compressor.lisp
```diff
@@ -268,51 +268,45 @@ other's."
...
+                 :until (subsetp new-complex qubit-complex)
+                 ;; enlarge the complex
+                 :do (setf qubit-complex (union qubit-complex new-complex))
+                 :do (let ((obj (nth-value 2 (lookup-hardware-address-by-qubits chip-specification qubit-complex))))
+                       ;; try to find an associated hardware object for this complex
+                       ;; if we can, then we want to loop over the object's rewrite rules.
+                       ;; if we can't, we fall through and do nothing.
+                       (alexandria:when-let*
+                           ((rewrite-rules
+                             (cond
```
stylewarning: i think this cond should be lifted out of the when-let, and we should duplicate the when-logic in the body of the cond

src/compressor/compressor.lisp
stylewarning: if it's renamed from `inner-loop` to something else, i'd be happy with it

src/compressor/compressor.lisp
notmgsk: Maybe @ecpeterson

stylewarning: Maybe @ecpeterson

appleby:
> Sorry, my previous review was meant to be "request changes."
> 
> I think this PR needs to have a more convincing, clearer piece of code to be worth it. While we get a benefit out of eliminating the tail recursion, I think we are losing some amount of comprehension.

I agree that the final refactoring of OUTER-INSTRUCTION-LOOP leaves a lot to be desired from a readability perspective. Perhaps it makes sense to drop this PR in favor of #262, which is basically a superset of these changes, minus the EXHAUST-STACK refactoring, which is not really necessary anyway.

stylewarning:
@appleby How about we just do the exhaust stack biz in this PR and leave the other changes to the other one. Can you reform this PR to just change exhaust-stack?

appleby:
> @appleby How about we just do the exhaust stack biz in this PR and leave the other changes to the other one. Can you reform this PR to just change exhaust-stack?

Will do.

src/compilation-methods.lisp
appleby: Done

src/compilation-methods.lisp
appleby: I feel your pain.

appleby:
This PR now only includes the changes to EXHAUST-STACK, plus some whitespace cleanup in the same file.

src/compilation-methods.lisp
```diff
@@ -334,14 +334,12 @@ Returns a value list: (processed-program, of type parsed-program
               (touch-reset-block blk))
              (otherwise
               (touch-unpreserved-block blk registrant))))
-         
+
          ;; this is the main loop that pushes through the CFG
          (exhaust-stack ()
```
appleby: The compiler is probably already doing it anyway, but EXHAUST-STACK is only called once below, so this loop could just as well be in-lined there. No need for a function call now that it's not recursive. The function name still servers a documentation purpose, I suppose. I'm not really advocating for this change, just pointing it out.

PR #262, opened 2019-05-27
replace recursion in functions with iteration

dkochmanski:
Functions in question are: walk-graph-with-candidates, outer-instruction-loop and expand-to-native-instructions.

Feel free to close the PR if you find recursion more elegant and it is not that much of a problem. With these changes sbcl passes tests on low speed / high safety and debug settings. I haven't encountered stack overflow on ECL so far either (but tests still run).

### DIFF
--- a/src/addresser/logical-schedule.lisp
+++ b/src/addresser/logical-schedule.lisp
@@ -479,11 +479,13 @@ mapping instructions to their tags. "
                (every (lambda (ancestor) (gethash ancestor visited-hash-table))
                       (gethash candidate (lscheduler-earlier-instrs lschedule))))
              (walk-graph-with-candidates (candidates)
-               (let* ((candidate (find-if #'candidate-has-no-ancestors
-                                          candidates))
-                      ;; TODO: use a queue to make this more like a normal topological sort
-                      (candidates (append (gethash candidate (lscheduler-later-instrs lschedule))
-                                          (remove candidate candidates))))
+               (do* (;; TODO: use a queue to make this more like a normal topological sort
+                     (candidates candidates
+                                 (append (gethash candidate (lscheduler-later-instrs lschedule))
+                                         (remove candidate candidates)))
+                     (candidate (find-if #'candidate-has-no-ancestors candidates)
+                                (find-if #'candidate-has-no-ancestors candidates)))
+                    ((null candidates))
                  (setf (gethash candidate visited-hash-table) t)
                  (let ((bumped-value (funcall bump-value candidate (gethash candidate distance-hash-table))))
                    (setf max-distance
@@ -492,9 +494,7 @@ mapping instructions to their tags. "
                      (setf (gethash child distance-hash-table)
                            (if (gethash child distance-hash-table)
                                (funcall test-values bumped-value (gethash child distance-hash-table))
-                               bumped-value))))
-                 (unless (endp candidates)
-                   (walk-graph-with-candidates candidates)))))
+                               bumped-value)))))))
       (dolist (instr (lscheduler-topmost-instructions lschedule))
         (setf (gethash instr distance-hash-table) base-value))
       (walk-graph-with-candidates (lscheduler-topmost-instructions lschedule))
--- a/src/compressor/compressor.lisp
+++ b/src/compressor/compressor.lisp
@@ -269,50 +269,48 @@ other's."
        ;; successfully, we rewind by the peephole window and try again. if it
        ;; fails, we fall through, step through to the next node, and try again.
        (outer-instruction-loop (node)
-         ;; for each instruction...
-         (setf node (peephole-rewriter-node-next node))
-         (unless node (return-from outer-instruction-loop))
-
-         (update-context node)
-
-         ;; build the noncommuting instruction list
-         (let* ((nodes-for-inspection (find-noncommuting-instructions node))
-                (qubit-complex nil)
-                (exhausted-specializations nil))
-           ;; for each instruction in the noncommuting tail...
-           (dolist (inspection-node nodes-for-inspection)
-             (let ((new-complex (application-qubit-indices (peephole-rewriter-node-instr inspection-node))))
-               ;; if we've already hit the generic object, continue to the next list head element.
-               (when exhausted-specializations
-                 (return-from outer-instruction-loop
-                   (outer-instruction-loop node)))
-               ;; does the next noncommuting instruction yield some new set of rules?
-               ;; if not, skip it.
-               (when (subsetp new-complex qubit-complex)
-                 (return))
-               ;; enlarge the complex
-               (setf qubit-complex (union qubit-complex new-complex))
-               ;; try to find an associated hardware object for this complex
-               (let ((obj (nth-value 2 (lookup-hardware-address-by-qubits chip-specification qubit-complex))))
-                 ;; if we can, then we want to loop over the object's rewrite rules.
-                 ;; if we can't, we fall through and do nothing.
-                 (alexandria:when-let
-                     ((rewrite-rules
-                       (cond
-                         ;; if we found a new hardware object, then use its rules
-                         (obj (hardware-object-rewriting-rules obj))
-                         ;; if we didn't find a new object but we haven't tried the
-                         ;; generic rules, try them now
-                         ((not exhausted-specializations)
-                          (setf exhausted-specializations t)
-                          (chip-specification-generic-rewriting-rules chip-specification))
-                         ;; otherwise, give up.
-                         (t nil))))
-                   (let ((node-to-jump-to (apply-rules rewrite-rules nodes-for-inspection)))
-                     (when node-to-jump-to
-                       (return-from outer-instruction-loop
-                         (outer-instruction-loop node-to-jump-to)))))))))
-         (outer-instruction-loop node)))
+         (do (;; for each instruction...
+              (node #1=(peephole-rewriter-node-next node) #1#))
+             ((null node))
+           (update-context node)
+           ;; build the noncommuting instruction list
+           (let* ((nodes-for-inspection (find-noncommuting-instructions node))
+                  (qubit-complex nil)
+                  (exhausted-specializations nil))
+             ;; for each instruction in the noncommuting tail...
+             (dolist (inspection-node nodes-for-inspection)
+               (let ((new-complex (application-qubit-indices (peephole-rewriter-node-instr inspection-node))))
+                 ;; if we've already hit the generic object, continue to the next list head element.
+                 (when exhausted-specializations
+                   (go :ADVANCE-NODE))
+                 ;; does the next noncommuting instruction yield some new set of rules?
+                 ;; if not, skip it.
+                 (when (subsetp new-complex qubit-complex)
+                   (return))
+                 ;; enlarge the complex
+                 (setf qubit-complex (union qubit-complex new-complex))
+                 ;; try to find an associated hardware object for this complex
+                 (let ((obj (nth-value 2 (lookup-hardware-address-by-qubits chip-specification qubit-complex))))
+                   ;; if we can, then we want to loop over the object's rewrite rules.
+                   ;; if we can't, we fall through and do nothing.
+                   (alexandria:when-let
+                       ((rewrite-rules
+                         (cond
+                           ;; if we found a new hardware object, then use its rules
+                           (obj (hardware-object-rewriting-rules obj))
+                           ;; if we didn't find a new object but we haven't tried the
+                           ;; generic rules, try them now
+                           ((not exhausted-specializations)
+                            (setf exhausted-specializations t)
+                            (chip-specification-generic-rewriting-rules chip-specification))
+                           ;; otherwise, give up.
+                           (t nil))))
+                     (let ((node-to-jump-to (apply-rules rewrite-rules nodes-for-inspection)))
+                       (when node-to-jump-to
+                         (setf node node-to-jump-to)
+                         (go :ADVANCE-NODE))))))))
+          :ADVANCE-NODE
+          nil)))
 
     ;; strip out all the NOPs.
     (let* ((instructions (remove-if (lambda (x) (typep x 'no-operation)) instructions))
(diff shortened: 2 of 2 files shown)


src/compressor/compressor.lisp
```diff
@@ -269,50 +269,47 @@ other's."
...
+         (do (;; for each instruction...
+              (node #1=(peephole-rewriter-node-next node) #1#))
+             ((null node))
+           (update-context node)
+           ;; build the noncommuting instruction list
+           (let* ((nodes-for-inspection (find-noncommuting-instructions node))
+                  (qubit-complex nil)
+                  (exhausted-specializations nil))
+             ;; for each instruction in the noncommuting tail...
+             (dolist (inspection-node nodes-for-inspection)
```
dkochmanski: these two loops (do and dolist) could probably be rewritten as one

src/compressor/compressor.lisp
```diff
@@ -269,50 +269,47 @@ other's."
...
+               (let ((new-complex (application-qubit-indices (peephole-rewriter-node-instr inspection-node))))
+                 ;; if we've already hit the generic object, continue to the next list head element.
+                 (when exhausted-specializations
+                   (go :continue))
```
dkochmanski: escape with do's implicit tagbody

src/compressor/compressor.lisp
```diff
@@ -330,6 +327,66 @@ other's."
...
+            (push instr processed))
+           ;; otherwise, we are nonnative. translate us. error on null.
+           ((null (setf translation-results
+                        (apply-translation-compilers instr chip obj)))
+            (error "Failed to expand ~a into native instructions."
+                   (print-instruction instr nil)))
+           ;; we managed a translation, use these instructions instead.
+           (t
+            ;; elaborate assertion follows
+            (when (and *compress-carefully*
```
dkochmanski: maybe this assertion should be refactored as a separate function?

src/compressor/compressor.lisp
```diff
@@ -330,55 +327,51 @@ other's."
...
          (cond
-           (translation-results
+           ;; are we native? then stick this instruction onto the output.
+           ((and obj
+                 (funcall (hardware-object-native-instructions obj) instr))
+            (pop instructions)
+            (push instr processed))
+           ;; otherwise, we are nonnative. translate us. error on null.
+           ((null (setf translation-results
+                        (apply-translation-compilers instr chip obj)))
```
dkochmanski: note the side-effect in the test clause

appleby:
Whoops, was just about to ping you to make sure you weren't also working on this. Guess I should have done that first! See also #261.

src/addresser/logical-schedule.lisp
```diff
@@ -479,11 +479,13 @@ mapping instructions to their tags. "
                (every (lambda (ancestor) (gethash ancestor visited-hash-table))
                       (gethash candidate (lscheduler-earlier-instrs lschedule))))
              (walk-graph-with-candidates (candidates)
-               (let* ((candidate (find-if #'candidate-has-no-ancestors
-                                          candidates))
-                      ;; TODO: use a queue to make this more like a normal topological sort
-                      (candidates (append (gethash candidate (lscheduler-later-instrs lschedule))
-                                          (remove candidate candidates))))
+               (do* (;; TODO: use a queue to make this more like a normal topological sort
+                     (candidates candidates
```
appleby: Why swap the order of the `candidates` / `candidate` binding here?

src/compressor/compressor.lisp
```diff
@@ -330,55 +327,51 @@ other's."
...
-           (expand-to-native-instructions (rest instrs)
-                                          chip-specification
-                                          (cons (first instrs) output-string))))
-       ;; otherwise, we are nonnative. translate us.
-       (let ((translation-results (apply-translation-compilers (first instrs)
-                                                               chip-specification
-                                                               obj)))
-         ;; if we managed a translation, use these instructions instead.
-         ;; otherwise, throw an error: we failed to perform translation.
+(defun expand-to-native-instructions (instructions chip)
```
appleby: Documentation string has disappeared.

dkochmanski: fixed

src/addresser/logical-schedule.lisp
dkochmanski: if it were the other way around, then:

1st iteration

candidate = first candidates
candidates = candidates

2nd iteration
candidate = first candidates
candidates = (whatever candidates)

so candidate would be bound twice to the same thing (unlike the original function). in this scenario we have

candidates = candidates
candidate = first candidates

candidates = whatsoever candidates
candidate = first candidates

appleby: Apologies. My confusion here was because I thought `candidates` might be getting a different value the first time through the loop. Specifically,

**1st iteration prior to change**

candidate = `(find-if #'candidate-has-no-ancestors candidates)`
candidates = `(append (blah..) (remove candidate candidates))`

**1st iteration after change**

candidates = candidates
candidate = same as above

But on closer inspection is seems `candidates` is not referenced in the loop body, but only `candidate`, and since you now do the termination test at the top of the loop rather than at the bottom, I guess you don't actually want to update `candidates` until the second time through the loop.

dkochmanski:
n.b all tests pass on ecl with this change

src/compressor/compressor.lisp
ecpeterson: I think this is actually a bug in the original: `apply-translation-compilers` will always yield a valid result (perhaps including `NIL`, should `instr` be a null-operation), and in the event of an error it'll actually signal an error condition. If you're brave, we can make that change here; if you'd rather I do this separately, that's fine too.

ecpeterson: (cf. this similar change in another open PR https://github.com/rigetti/quilc/pull/237/files#diff-c69b2b4238cc0d73cdab132c180d7b54L364 )

dkochmanski: I've only transformed code and preserved its semantics / comments. I'm not that brave to suggest change which I would not understand ;-).

ecpeterson: Very reasonable position. I’ll earmark this as something to investigate in the other PR instead.

src/compressor/compressor.lisp
```diff
@@ -269,50 +269,47 @@ other's."
...
+                   ;; if we can't, we fall through and do nothing.
+                   (alexandria:when-let
+                       ((rewrite-rules
+                         (cond
+                           ;; if we found a new hardware object, then use its rules
+                           (obj (hardware-object-rewriting-rules obj))
+                           ;; if we didn't find a new object but we haven't tried the
+                           ;; generic rules, try them now
+                           ((not exhausted-specializations)
+                            (setf exhausted-specializations t)
```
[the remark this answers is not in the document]
stylewarning: maybe we need this final call to `update-context`?

src/compressor/compressor.lisp
```diff
@@ -330,55 +327,52 @@ other's."
...
-           (expand-to-native-instructions (rest instrs)
-                                          chip-specification
-                                          (cons (first instrs) output-string))))
-       ;; otherwise, we are nonnative. translate us.
-       (let ((translation-results (apply-translation-compilers (first instrs)
-                                                               chip-specification
-                                                               obj)))
-         ;; if we managed a translation, use these instructions instead.
-         ;; otherwise, throw an error: we failed to perform translation.
+  (do* ((processed nil)
```
stylewarning: can we change _this_ `do` to a `loop` just to be idiomatic with the rest of the code?

src/compressor/compressor.lisp
```diff
@@ -330,55 +327,52 @@ other's."
...
-           (translation-results
+           ;; are we native? then stick this instruction onto the output.
+           ((and obj
+                 (funcall (hardware-object-native-instructions obj) instr))
+            (pop instructions)
+            (push instr processed))
+           ;; otherwise, we are nonnative. translate us. error on null.
+           ((null (setf translation-results
+                        (apply-translation-compilers instr chip obj)))
+            (error "Failed to expand ~a into native instructions."
```
stylewarning: use `"Failed to expand ~/cl-quil:instruction-fmt/ into native instructions." instr`

src/compressor/compressor.lisp
```diff
@@ -269,50 +269,47 @@ other's."
...
+                           ((not exhausted-specializations)
+                            (setf exhausted-specializations t)
+                            (chip-specification-generic-rewriting-rules chip-specification))
+                           ;; otherwise, give up.
+                           (t nil))))
+                     (let ((node-to-jump-to (apply-rules rewrite-rules nodes-for-inspection)))
+                       (when node-to-jump-to
+                         (setf node node-to-jump-to)
+                         (go :continue))))))))
+          :continue)))
```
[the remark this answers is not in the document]
ecpeterson: `:ADVANCE-NODE` or `:PROCEED-TO-NEXT-NODE` would both be fine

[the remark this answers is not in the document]
notmgsk: I like the latter, it has some character. Maybe `:PROCEED-CALMLY-TO-NEXT-NODE`.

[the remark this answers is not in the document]
ecpeterson: `:KEEP-CALM-AND-CARRY-ON-TO-NEXT-NODE`

src/addresser/logical-schedule.lisp
```diff
@@ -479,11 +479,13 @@ mapping instructions to their tags. "
                (every (lambda (ancestor) (gethash ancestor visited-hash-table))
                       (gethash candidate (lscheduler-earlier-instrs lschedule))))
              (walk-graph-with-candidates (candidates)
-               (let* ((candidate (find-if #'candidate-has-no-ancestors
-                                          candidates))
-                      ;; TODO: use a queue to make this more like a normal topological sort
-                      (candidates (append (gethash candidate (lscheduler-later-instrs lschedule))
-                                          (remove candidate candidates))))
+               (do* (;; TODO: use a queue to make this more like a normal topological sort
```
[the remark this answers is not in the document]
appleby: There are literally dozens of us!

src/compressor/compressor.lisp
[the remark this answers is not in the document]
stylewarning: i like a bit of character if i do say so myself

src/compressor/compressor.lisp
```diff
@@ -269,50 +269,47 @@ other's."
...
+           (update-context node)
+           ;; build the noncommuting instruction list
+           (let* ((nodes-for-inspection (find-noncommuting-instructions node))
+                  (qubit-complex nil)
+                  (exhausted-specializations nil))
+             ;; for each instruction in the noncommuting tail...
+             (dolist (inspection-node nodes-for-inspection)
+               (let ((new-complex (application-qubit-indices (peephole-rewriter-node-instr inspection-node))))
+                 ;; if we've already hit the generic object, continue to the next list head element.
+                 (when exhausted-specializations
```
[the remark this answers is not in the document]
dkochmanski: it is not obvious to me how to do that so I'd rather leave it as it is now

src/compressor/compressor.lisp
```diff
@@ -269,50 +269,47 @@ other's."
...
+                   (return))
+                 ;; enlarge the complex
+                 (setf qubit-complex (union qubit-complex new-complex))
+                 ;; try to find an associated hardware object for this complex
+                 (let ((obj (nth-value 2 (lookup-hardware-address-by-qubits chip-specification qubit-complex))))
+                   ;; if we can, then we want to loop over the object's rewrite rules.
+                   ;; if we can't, we fall through and do nothing.
+                   (alexandria:when-let
+                       ((rewrite-rules
+                         (cond
```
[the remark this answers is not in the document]
dkochmanski: ditto

src/compressor/compressor.lisp
dkochmanski: that was my first approach but maybe my loop-fu is weak because the code grew in complexity

src/compressor/compressor.lisp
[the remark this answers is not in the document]
stylewarning: ok

src/compressor/compressor.lisp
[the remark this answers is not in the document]
dkochmanski: which final call? this change doesn't change anything from the computation point of view as far as I'm aware.

src/compressor/compressor.lisp
dkochmanski: fixed

src/compressor/compressor.lisp
```diff
@@ -269,50 +269,47 @@ other's."
...
+                           ;; generic rules, try them now
+                           ((not exhausted-specializations)
+                            (setf exhausted-specializations t)
+                            (chip-specification-generic-rewriting-rules chip-specification))
+                           ;; otherwise, give up.
+                           (t nil))))
+                     (let ((node-to-jump-to (apply-rules rewrite-rules nodes-for-inspection)))
+                       (when node-to-jump-to
+                         (setf node node-to-jump-to)
+                         (go :continue))))))))
```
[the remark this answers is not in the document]
dkochmanski: fixed

src/compressor/compressor.lisp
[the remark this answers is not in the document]
dkochmanski: fixed

src/compressor/compressor.lisp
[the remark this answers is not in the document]
dkochmanski: fixed

src/addresser/logical-schedule.lisp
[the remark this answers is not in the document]
dkochmanski: it is worth noting, that while `DO` is often labelled as an iteration operator (and is used to define some other iteration operators) in fact it is a wonderful syntactic sugar for tail recursion. Think about it: recursive function has: initial arguments, stop condition and argument modification step. In do:

```
(DO ((argument :initial-value :step)*)
    (stop-condition cleanup)
  (optional-side-effects))
```

so if you have a soft spot for tail recursion then you should love do :-)

src/compressor/compressor.lisp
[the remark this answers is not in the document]
stylewarning: I'm saying if we lift out the cond, embed the body of the `when` in the cond, etc.

src/addresser/logical-schedule.lisp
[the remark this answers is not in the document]
stylewarning: @dkochmanski You've not convinced me!

dkochmanski:
given addressed remarks (and OK on not lifting the cond), are we good to merge?

stylewarning:
@dkochmanski I'm good. Any last things you'd like to do?

dkochmanski:
no, I'm good. thank you all for the review.

PR #263, opened 2019-05-28
Test MEASURE semantics

### LINKED ISSUE #77: Add MEASURE semantics tests
The compiler has very limited support for manipulating MEASURE instructions, but this is essentially untested because all the tests of quantum semantics are built around pure state simulation. (Look into density matrix simulation methods in the QVM and) introduce some tests that check correctness for programs involving mid-program MEASURE instructions.

notmgsk:
Closes #77

### DIFF
--- a/cl-quil-tests.asd
+++ b/cl-quil-tests.asd
@@ -14,6 +14,7 @@
                #:uiop
                #:yacc                     ; for the conditions
                (:version #:alexa "1.0.1") ; for the conditions
+               #:cl-permutation
                )
   :perform (asdf:test-op (o s)
                          (uiop:symbol-call ':cl-quil-tests
--- a/cl-quil-tests.asd
+++ b/cl-quil-tests.asd
@@ -7,7 +7,7 @@
   :author "Robert Smith <robert@rigetti.com>"
   :license "Apache License 2.0 (See LICENSE.txt)"
   :depends-on (#:cl-quil
-               (:version #:qvm "1.8.2")
+               #:qvm
                #:magicl
                #:alexandria
                #:fiasco
--- a/cl-quil-tests.asd
+++ b/cl-quil-tests.asd
@@ -7,7 +7,7 @@
   :author "Robert Smith <robert@rigetti.com>"
   :license "Apache License 2.0 (See LICENSE.txt)"
   :depends-on (#:cl-quil
-               #:qvm
+               (:version #:qvm "1.8.2")
                #:magicl
                #:alexandria
                #:fiasco
--- a/cl-quil-tests.asd
+++ b/cl-quil-tests.asd
@@ -7,6 +7,7 @@
   :author "Robert Smith <robert@rigetti.com>"
   :license "Apache License 2.0 (See LICENSE.txt)"
   :depends-on (#:cl-quil
+               #:qvm
                #:magicl
                #:alexandria
                #:fiasco
--- a/tests/misc-tests.lisp
+++ b/tests/misc-tests.lisp
@@ -336,13 +336,15 @@ but that is outside our scope of interest."
 To put the density matrix into the basis state, e.g., |01><11|, we
 would choose BASIS-INDEX = 7. In general, the basis state |a><b| is
 prepared by choosing BASIS-INDEX = (2^N * a + b)."
-  (let ((amps (make-array 64 :element-type 'qvm:cflonum :initial-element (qvm:cflonum 0))))
+  (let ((amps (make-array (expt 2 (* 2 num-qubits))
+                          :element-type 'qvm:cflonum
+                          :initial-element (qvm:cflonum 0))))
     (setf (aref amps basis-index) (qvm:cflonum 1))
     (qvm:make-density-qvm num-qubits :amplitudes amps)))
 
 (deftest test-measure-semantics ()
-  "Test that artifacts of compilation (namely moving MEASUREs) does
-not change the semantics of the program."
+  "Test that artifacts of compilation (namely moving and rewiring
+MEASUREs) does not change the semantics of the program."
   (let* ((p-str "H 0
 CNOT 0 2
 MEASURE 0
--- a/tests/misc-tests.lisp
+++ b/tests/misc-tests.lisp
@@ -262,3 +262,13 @@
       (is (eq lisp-symbol (quil::quil-function->lisp-symbol quil-string)))
       (is (string= quil-string (quil::lisp-symbol->quil-function lisp-symbol)))
       (is (string= quil-string (quil::lisp-symbol->quil-function-or-prefix-operator lisp-symbol))))))
+     (a:iota n))))
+
+(deftest test-measure-semantics ()
+  "Test that something or other is what it is"
+  (let ((qvm (qvm:make-density-qvm 1))
+        (p (parse-quil "H 0
+MEAUSURE 0
+H 0")))
+    (qvm:load-program qvm :supersede-memory-subsystem t)
+    (setf (qvm::program-compiled-p qvm) t)))
--- a/tests/misc-tests.lisp
+++ b/tests/misc-tests.lisp
@@ -264,24 +264,6 @@
       (is (string= quil-string (quil::lisp-symbol->quil-function-or-prefix-operator lisp-symbol))))))
      (a:iota n))))
 
-;; Helper function for below. Should probably be made more general and
-;; useful elsewhere.
-(defun cheeky-relabeler (parsed-program qubits)
-  "Send all application/measurement arguments to their position in the
-list of qubits QUBITS."
-  (check-type parsed-program parsed-program)
-  (dolist (instr (coerce (parsed-program-executable-code parsed-program) 'list) parsed-program)
-    (typecase instr
-      (quil::measure
-       (setf (quil::measurement-qubit instr)
-             (quil::qubit (position (quil::qubit-index (quil::measurement-qubit instr))
-                                    qubits))))
-      (quil::application
-       (setf (quil::application-arguments instr)
-             (mapcar (lambda (arg) (quil::qubit (position (quil::qubit-index arg)
-                                                     qubits)))
-                     (quil::application-arguments instr)))))))
-
 (deftest test-measure-discard-semantics ()
   "Test that artifacts of compilation (namely moving MEASUREs) does
 not change the semantics of the program."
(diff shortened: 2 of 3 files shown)


cl-quil-tests.asd
```diff
@@ -7,6 +7,7 @@
   :author "Robert Smith <robert@rigetti.com>"
   :license "Apache License 2.0 (See LICENSE.txt)"
   :depends-on (#:cl-quil
+               (:version #:qvm "1.8.2")
```
stylewarning: bold move

tests/misc-tests.lisp
```diff
@@ -275,3 +275,19 @@
      (lambda (permutation)
        (not-signals simple-error (quil::check-permutation permutation)))
      (alexandria:iota n))))
+
+(deftest test-measure-semantics ()
+  "Test that something or other is what it is"
```
stylewarning: nice doc string

tests/misc-tests.lisp
```diff
@@ -275,3 +275,19 @@
      (lambda (permutation)
        (not-signals simple-error (quil::check-permutation permutation)))
      (alexandria:iota n))))
+
+(deftest test-measure-semantics ()
```
stylewarning: `test-mid-program-measure`

tests/misc-tests.lisp
```diff
@@ -275,3 +275,19 @@
...
+  "Test that something or other is what it is"
+  (let* ((qvm (qvm:make-density-qvm 1))
+         (qvm-comp (qvm:make-density-qvm 1))
+         (p (parse-quil "H 0
```
stylewarning: can you make a _list_ of different programs to try? For instance, MEASURE at the start, in the middle, at the end? And at least one with more than one measure and more than one qubit?

stylewarning:
(sorry didn't notice this was a draft)

cl-quil-tests.asd
notmgsk: That was to see if I could force quicklisp or the build tool to pull the right version. No luck.

stylewarning: ASDF only tests that the version it pulls is `>=` than that. It won't pull a specific or special version, it just blocks loading anything smaller than that v.

notmgsk:
> (sorry didn't notice this was a draft)

that's because you're daft

tests/misc-tests.lisp
notmgsk: I've left this as is since it now does pre- mid- and post-measure

tests/misc-tests.lisp
```diff
@@ -295,6 +313,57 @@ MEASURE 0")))
...
+                   (qvm::amplitudes qvm)
+                   (qvm::amplitudes qvm-comp)))))))
+
+(deftest test-measure-semantics ()
+  (let* ((p (parse-quil "
+CNOT 0 2
+MEASURE 0
+MEASURE 1
+MEASURE 2"))
+         (p-comp (parse-quil "
```
ecpeterson: I'd prefer this to be a call to `compiler-hook` over a literal.

tests/misc-tests.lisp
```diff
@@ -276,7 +276,25 @@
        (not-signals simple-error (quil::check-permutation permutation)))
      (alexandria:iota n))))
 
-(deftest test-measure-semantics ()
+;; Helper function for below. Should probably be made more general and
+;; useful elsewhere.
+(defun cheeky-relabeler (parsed-program qubits)
```
ecpeterson: Could I convince you to name this something other than `relabeler`, since that's a technical noun in CL-QUIL?

tests/misc-tests.lisp
```diff
@@ -295,6 +313,57 @@ MEASURE 0")))
...
+RZ(2.078959324140279) 2
+RX(-pi/2) 2
+RZ(-1.8815794754856947) 2")))
+    (let* ((qvm (qvm:make-density-qvm 3))
+           (qvm-comp (qvm:make-density-qvm 3)))
+      (qvm:load-program qvm p :supersede-memory-subsystem t)
+      (qvm:load-program qvm-comp p-comp :supersede-memory-subsystem t)
+      (qvm:run qvm)
+      (qvm:run qvm-comp)
+      (is (every #'quil::double=
```
ecpeterson: That you can make this comparison successfully without taking into account the final rewiring must be an artifact of the particular program we're using (and/or the fact that this QVM starts in the ground state). IMO this means we should use a more complicated program.

tests/misc-tests.lisp
```diff
@@ -295,6 +313,57 @@ MEASURE 0")))
...
+RZ(pi/2) 1
+MEASURE 1
+RZ(pi) 0
+MEASURE 0
+RZ(1.651484668090844) 2
+RX(pi/2) 2
+RZ(2.078959324140279) 2
+RX(-pi/2) 2
+RZ(-1.8815794754856947) 2")))
+    (let* ((qvm (qvm:make-density-qvm 3))
```
ecpeterson: You should loop the simulation over a spanning set of density matrices as initial states; otherwise this is testing only that the compiler does the right thing when the QVM starts in the ground state.

notmgsk:
@ecpeterson I think I have addressed your comments. Would you take another look? Thanks chap

tests/misc-tests.lisp
```diff
@@ -275,3 +275,73 @@
...
+To put the density matrix into the basis state, e.g., |01><11|, we
+would choose BASIS-INDEX = 7. In general, the basis state |a><b| is
+prepared by choosing BASIS-INDEX = (2^N * a + b)."
+  (let ((amps (make-array (expt 2 (* 2 num-qubits))
+                          :element-type 'qvm:cflonum
+                          :initial-element (qvm:cflonum 0))))
+    (setf (aref amps basis-index) (qvm:cflonum 1))
+    (qvm:make-density-qvm num-qubits :amplitudes amps)))
+
+;; Disclaimer: the below is maybe incomplete. Should we test not just
```
notmgsk: I don't think this really elucidates the stuff we talked about offline @ecpeterson. Do you have a suggestion for improvement

tests/misc-tests.lisp
```diff
@@ -42,6 +42,42 @@
   (is (equal '(a b c d) (quil::reduce-append '((a) (b c) nil (d)))))
   (is (equal '(a b c d e f) (quil::reduce-append '((a) (b c) nil (d) (e f))))))
 
+(deftest test-print-instruction ()
+  (is (string= "PRAGMA gate_time CNOT \"50 ns\""
+               (with-output-to-string (s)
+                 (cl-quil::print-instruction (make-instance 'quil::pragma
```
stylewarning: can we `(format nil "~/cl-quil:instruction-fmt/ (make-instance ...))` these?

tests/misc-tests.lisp
```diff
@@ -262,3 +298,59 @@
...
+(defun %extract-trivial-exit-rewiring (pp)
+  "Extract the exit rewiring comment from parsed program PP. Trivial
+here means PP is expected to have a single exit rewiring. A more
+complicated CFG could produce multiple exit rewirings in a program,
+but that is outside our scope of interest."
+  (declare (type parsed-program pp))
+  (loop :with code := (parsed-program-executable-code pp)
+        :for i :below (length code)
+        :for comment := (quil::comment (elt code i))
+        :when (and comment
```
stylewarning: in this case, I'd prefer `:do (when ...)`

tests/misc-tests.lisp
```diff
@@ -262,3 +298,59 @@
...
+                   (uiop:string-prefix-p "Exiting rewiring: " comment)) :do
+                     (return (quil::rewiring-l2p
+                              (quil::make-rewiring-from-string
+                               (subseq comment (length "Exiting rewiring: ")))))))
+
+(defun %make-density-qvm-initialized-in-basis (num-qubits basis-index)
+  "Make a DENSITY-QVM that is initialized in the basis state described by BASIS-INDEX.
+
+To put the density matrix into the basis state, e.g., |01><11|, we
```
stylewarning: cool comment, except please don't break lines

tests/misc-tests.lisp
```diff
@@ -262,3 +298,59 @@
...
+    (qvm:make-density-qvm num-qubits :amplitudes amps)))
+
+;; Disclaimer: the below is maybe incomplete. Should we test not just
+;; for measure-discard? If so, then more thought will be required. It
+;; will not always be the case that passing in a non-density matrix to
+;; the qvm will work as it does below.
+(deftest test-measure-semantics ()
+  "Test that artifacts of compilation (namely moving and rewiring
+MEASUREs) does not change the semantics of the program."
+  (let* ((p-str "H 0
```
stylewarning: maybe choose a program that doesn't create a uniform superposition?

tests/misc-tests.lisp
ecpeterson: when about `:when ... :return ...`

tests/misc-tests.lisp
notmgsk: why's that

tests/misc-tests.lisp
ecpeterson: I suggest typing a lot more—all the things we talked about could be relevant to a future maintainer.
```
;; Our test below is built on the following idea: rather than directly
;; calculating the entire superoperator encoding the pre- and post-compilation
;; programs and comparing the results, we instead merely calculate the behavior
;; of the superoperator (itself necessarily a linear operator) on a maximal
;; linearly independent set of inputs and check that they match individually.
;;
;; However, we cheat a bit while doing this: we select a maximal linearly
;; independent set drawn from __all matrices__, without regards to the three
;; defining characteristics of density operators (Hermitian, unit trace, and
;; positive semi-definite), and so we are actually feeding invalid inputs to the
;; QVM. Provided the QVM does not closely inspect its input, this is fine---but
;; it would matter when simulating, e.g., a measure-and-store.
;;
;; Should this test begin to misbehave, it might be resolved by enforcing these
;; constraints on our generated probe operators.  I believe the constraints to
;; be of descending order of severity: it is easy to imagine two distinct
;; extensions to the space of all matrices of the "same" superoperator, in the
;; sense that they give the same behavior on density operators but differ on the
;; complement of Hermitian matrices.  I don't believe that this is possible for
;; the subclasses of unit-trace or positive-definite operators, but it could
;; still be that (for something other than pure operations and MEASURE-DISCARDs)
;; the QVM makes a calculation based on these assumptions and gives garbage
;; output if they're violated.  Caveat programmer.
```

tests/misc-tests.lisp
stylewarning: @notmgsk to make the statistics a bit more difficult to recreate. If every number is 0.5, then it's invariant to permutation

tests/misc-tests.lisp
stylewarning: tru too

tests/misc-tests.lisp
stylewarning: ping on this

notmgsk: done

tests/misc-tests.lisp
notmgsk: also done, thought it was sort of tangential (`o r t h o g o n a l`) to this PR

tests/misc-tests.lisp
```diff
@@ -262,3 +293,78 @@
...
+RZ(1.0) 0
+H 0
+CNOT 0 2
+RZ(1.2) 1
+RX(0.8) 1
+MEASURE 0
+MEASURE 1
+MEASURE 2")
+         (p (parse-quil p-str)))
+    (loop :for i :below (expt 2 6) :do
```
stylewarning: @notmgsk Can we randomly permute the program a few times?

```
CL-USER> (defun randomly-permute-program! (pp)
           (let* ((code #1=(quil:parsed-program-executable-code pp))
                  (perm (perm:random-perm (length code))))
             (setf #1# (perm:permute perm code :type 'vector))
             pp))
RANDOMLY-PERMUTE-PROGRAM!
CL-USER> (quil::print-parsed-program
          (randomly-permute-program!
           (quil:parse-quil "H 0
CNOT 0 1
CNOT 1 2")))
CNOT 0 1
H 0
CNOT 1 2
NIL
CL-USER> (quil::print-parsed-program
          (randomly-permute-program!
           (quil:parse-quil "H 0
CNOT 0 1
CNOT 1 2")))
CNOT 1 2
CNOT 0 1
H 0
NIL
```

stylewarning:
good job with the test

PR #267, opened 2019-05-30
rewrite-arithmetic-for-isn matches all quil types

### LINKED ISSUE #266: REWRITE-ARITHMETIC fails on JUMPs
A user ran into this bug when running on the QPU. The program
```
PRAGMA INITIAL_REWIRING "PARTIAL"
RESET
DECLARE ro BIT[2]
RY(pi/3) 0
RY(pi/3) 1
RY(pi/3) 2
RY(pi/3) 3
H 0
MEASURE 0 ro[1]
```
causes the error (during the `rewrite-arithmetic` transform)
```
RPCError: There is no applicable method for the generic function
  #<STANDARD-GENERIC-FUNCTION CL-QUIL::REWRITE-ARITHMETIC-FOR-ISN (5)>
when called with arguments
  (#<CL-QUIL:JUMP-TARGET {10070DE623}> "__P3157837055997390270" 0).
See also:
  The ANSI Standard, Section 7.6.6
```
when targeting the lattice `Aspen-4-4Q-A`.

Exploring this locally, I find that a JUMP/LABEL is produced by the compiler:
```
QUIL> (print-parsed-program (compiler-hook (parse-quil "PRAGMA INITIAL_REWIRING \"PARTIAL\"
RESET
DECLARE ro BIT[2]
RY(pi/3) 0
RY(pi/3) 1
RY(pi/3) 2
RY(pi/3) 3
H 0
MEASURE 0 ro[1]") (build-8q-chip)))
DECLARE ro BIT[2]

PRAGMA INITIAL_REWIRING "PARTIAL"
JUMP @RESET-2676                        # Entering/exiting rewiring: (#(0 1 2 3 4 5 6 7) . #(0 1 2 3 4 5 6 7))
LABEL @RESET-2676                       # Entering rewiring: #(0 1 7 2 3 4 5 6)
RX(pi/2) 0
RZ(5*pi/6) 0
RX(pi/2) 0
MEASURE 0 ro[1]
RX(pi/2) 1
RZ(pi/3) 1
RX(-pi/2) 1
RX(pi/2) 2
RZ(pi/3) 2
RX(-pi/2) 2
RX(pi/2) 7
RZ(pi/3) 7
RX(-pi/2) 7
HALT                                    # Exiting rewiring: #(0 1 7 2 3 4 5 6)
```
Calling `rewrite-arithmetic` on the above causes the aforementioned error
```
NIL
QUIL> (rewrite-arithmetic (compiler-hook (parse-quil "PRAGMA INITIAL_REWIRING \"PARTIAL\"
RESET
DECLARE ro BIT[2]
RY(pi/3) 0
RY(pi/3) 1
RY(pi/3) 2
RY(pi/3) 3
H 0
MEASURE 0 ro[1]") (build-8q-chip)))
; Evaluation aborted on #<SB-PCL::NO-APPLICABLE-METHOD-ERROR {1005572DA3}>.
```
However, I notice that the JUMP/LABEL only appear for *some* lattices. If I instead target a fully-connected device, the (seemingly superfluous) JUMP/LABEL goes away
```
QUIL> (print-parsed-program (compiler-hook (parse-quil "PRAGMA INITIAL_REWIRING \"PARTIAL\"
RESET
DECLARE ro BIT[2]
RY(pi/3) 0
RY(pi/3) 1
RY(pi/3) 2
RY(pi/3) 3
H 0
MEASURE 0 ro[1]") (build-nq-fully-connected-chip 8)))
DECLARE ro BIT[2]

PRAGMA INITIAL_REWIRING "PARTIAL"
RX(pi/2) 0                              # Entering rewiring: #(0 1 2 3 4 5 6 7)
RZ(5*pi/6) 0
RX(pi/2) 0
MEASURE 0 ro[1]
RX(pi/2) 1
RZ(pi/3) 1
RX(-pi/2) 1
RX(pi/2) 2
RZ(pi/3) 2
RX(-pi/2) 2
RX(pi/2) 3
RZ(pi/3) 3
RX(-pi/2) 3
```
And the above can enjoy a `rewrite-arithmetic` without error.

The error occurs because `rewrite-arithmetic` does not include a method for `jump-target` objects. Why this issue is coming up now, I don't know.

notmgsk:
Previously this would error if `isn` was anything but an `instruction` or `application`. This closes #266.

### DIFF
--- a/src/analysis/rewrite-arithmetic.lisp
+++ b/src/analysis/rewrite-arithmetic.lisp
@@ -48,6 +48,9 @@ NOTE: This function does *not* cause side effects.")
   (:method ((isn instruction) mref-name index)
     (values isn '() index))
 
+  (:method ((isn jump-target) mref-name index)
+    (values isn '() index))
+
   (:method ((isn application) mref-name index)
     (check-type mref-name string)
     (check-type index unsigned-byte)


notmgsk:
> I don't like the looseness of the types here. Can we opt to add more cases instead?

Why

stylewarning:
i think it's bad taste to accept _any_ type for something that is intended to operate more strictly

notmgsk:
@stylewarning This was prompted by an error where the compiler was producing bad output, and then passing that onto `rewrite-arithmetic`. First thought was to allow `rewrite-arithmetic` to handle non-`instruction` objects (I've created a `pseudo-instruction` base class which `jump-target` and `include` now derive from), but maybe `rewrite-arithmetic` really should error on bad compiler output. Not sure. What do you think?

stylewarning:
I do not agree with this. `pseudo-instruction` is not a meaningful class that should exist in the AST. Could we just add the two cases, `include` and `jump-target` to this?

In fact, I'd suggest that `include` _errors_ in this instance. (Rewrite arithmetic should have `process-includes` as a prereq anyway.)

PR #268, opened 2019-05-30
Fix: RESET block merge bug

ecpeterson:
This PR addresses a pair of bugs introduced in #159 :

1. When processing a CFG block containing a RESET instruction—which gets dropped and the block retyped as a `reset-block`—the RESET instruction was not reintroduced when processing the CFG back into a single vector of instructions.
2. `reset-block`s, which are prefixed with a RESET, do not care about their initial rewiring: they can be freely rewired. Accordingly, during a merge of a `reset-block` into a preceding block, one need not check that their adjacent rewirings match before performing the merge. Without this caveat, the emitted Quil would have an extra (and superfluous) JUMP/LABEL pair near RESET instructions.

I am **not** confident that I haven't introduced more bugs in the course of this, and I'd appreciate a very careful review.


src/compilation-methods.lisp
```diff
@@ -342,12 +342,20 @@ Returns a value list: (processed-program, of type parsed-program
...
-      (dolist (blk (cfg-blocks cfg))
-        (when (typep blk 'reset-block)
-          (change-class blk 'basic-block)))
+      
       ;; one more pass of CFG collapse
       (simplify-cfg cfg)
+      
+      ;; if there are any lingering reset blocks, declassify them
+      (dolist (blk (cfg-blocks cfg))
+        (format t "I'm looking at (~a) ~a:~%" (type-of blk) blk)
```
stylewarning: this won;t be bueno

ecpeterson: haha

ecpeterson: i edited this out, but i didn't save before commit and push

src/cfg.lisp
```diff
@@ -396,16 +398,21 @@ Return the following values:
                   (eq parent (first (children blk)))
```
notmgsk: Gods there's a lot going on here

ecpeterson: you can see why it's easy to mess up and not notice

src/cfg.lisp
```diff
@@ -354,14 +354,16 @@ Return the following values:
 
 (defun merge-sequentially (blk1 blk2)
   "Returns a new BASIC-BLOCK representing a merged BLK1 and BLK2, such that its incoming list is a copy of BLK1's incoming list, its outgoing edge is a copy of BLK2's outgoing edge, and its code is the concatentation of BLK1's code and BLK2's code. Any reference to BLK1 or BLK2 within this new block's edge or incoming list is replaced with a reference to itself."
-  (assert (and (typep blk1 (type-of blk2))
-               (typep blk2 (type-of blk1))))
```
appleby: Would it make sense to leave this assertion in place, but add an `(or (typep blk2 'reset-block) ...)`? That is, either the second block is a reset block or both blocks have the same type. Something like:

```lisp
(assert (or (typep blk2 'reset-block)
            (and (typep blk1 (type-of blk2))
                 (typep blk2 (type-of blk1)))))
```

Is ever possible for `blk1` to be a reset block?

src/cfg.lisp
```diff
@@ -354,14 +354,16 @@ Return the following values:
...
+              (not (basic-block-out-rewiring blk1))
               (not (basic-block-in-rewiring blk2))
               (equalp (basic-block-out-rewiring blk1)
                       (basic-block-in-rewiring blk2))))
   (let ((new-blk (make-instance (type-of blk1)
-                                :code (concatenate 'vector (basic-block-code blk1)
+                                :code (concatenate 'vector
+                                                   (basic-block-code blk1)
+                                                   (when (typep blk2 'reset-block)
+                                                     (list (make-instance 'reset)))
```
appleby: Where/why is the reset instruction dropped in the first place? The `process-instruction` method specialized on `(instr reset)` appears to `vector-push-extend` the RESET instruction onto the `basic-block-code` of the new block...

ecpeterson: The (not especially well-founded) idea was to keep the addresser from having to deal with mid-block RESET instructions, which completely reset the addresser's state—instead, having RESET cause a block to get segmented would naturally cause the addresser's state to get reset. There's then a follow-up choice: either RESET always actually appears in a block and the addresser has to utilize that fact somehow (e.g.: require RESET to appear at the beginning, search for it, and if found set itself up so that it's able to freely rewire qubits) or RESET is always implicit and whoever's responsible for calling the addresser (viz., `compiler-hook`) is responsible for tracking this and priming the addresser to do what it ought.

I went with the second option because of some *other* logic related to blocks having multiple input branches: if it's possible when talking a sort of the CFG to return to a block that has already been visited, it may be required that a new block get inserted in order to match the final rewiring state of the 'lower' block with the initial rewiring state of the 'earlier' block. Since that induces some required RESET-related headache in the CFG crawler anyway, it made sense to put the whole business there.

By consequence, though, this is all a lot of elaborate special-casing on a very particular construction, and it's not evident what other `basic-block` subclasses could arise in the future, how they might interact, ... , which makes it hard (for me) to make very principled and robust decisions about what belongs where, what functionality ought to be broken out into generics, ... .

This is all to explain this hyperlocal decision: the only two types are `basic-block` and `reset-block`; `basic-block` doesn't mean anything special; `reset-block` means "I have an implicit RESET at the start"; and merging anything into a later `reset-block` will mean reintroducing that implicit instruction (and then retaining the type of the earlier block). Gross, but there it is.

src/cfg.lisp
```diff
@@ -396,16 +398,21 @@ Return the following values:
...
+                      (not (basic-block-out-rewiring parent))
                       (not (basic-block-in-rewiring blk))
                       (equalp (basic-block-out-rewiring parent)
                               (basic-block-in-rewiring blk)))
-                  (or (not (basic-block-out-rewiring parent))
-                      (not (basic-block-in-rewiring blk))
-                      (equalp (basic-block-out-rewiring parent)
-                              (basic-block-in-rewiring blk))))
+                  (or (typep parent 'reset-block)
+                      (not (basic-block-in-rewiring parent))
```
ecpeterson: N.B.: This inversion of `in` and `out` is, I think, an unrelated bug in the original. The two `or` clauses used to be identical!

src/cfg.lisp
```diff
@@ -396,16 +398,21 @@ Return the following values:
...
+                      (not (basic-block-out-rewiring blk))
+                      (equalp (basic-block-in-rewiring parent)
+                              (basic-block-out-rewiring blk))))
+                 
                  ;; update the rewiring data
-                 (setf (basic-block-in-rewiring parent) (basic-block-out-rewiring parent))
+                 (setf (basic-block-in-rewiring parent) (or (basic-block-in-rewiring parent)
```
ecpeterson: Ditto here, this just seemed like gibberish in the original, so I tried to write something that made more sense—for now.

src/cfg.lisp
ecpeterson: It's definitely possible: see the longer comment below, where I tried to recount the thin justification for how this is set up.

ecpeterson:
🎶 this PR gives me migraine 🎶

src/cfg.lisp
appleby: I won't pretend I understood that completely, but that you for taking the time to explain it anyway :).

I am still confused about where exactly the RESET instruction gets dropped, but I guess this is a good excuse to go RTFS and educate myself.

One minor point: there also appears to be a `preserved-block` type in addition to `basic-block` and `reset-block`, but maybe the whole point is that those blocks don't get touched and therefore are uninteresting in this context?

ecpeterson: You're right, I totally forgot about `preserved-block`s! You're right that because of how `compiler-hook` treats them (viz.: it marks them as visited and them resets their type to `basic-block`) and at what point it decides to call `simplify-cfg` (strictly after all traces of `preserved-block`s are removed), it was OK for me to forget about them. If I want `simplify-cfg` to operate on CFGs that might still have `preserved-block`s in them, then I should go back and consider adding more `(typep blk 'reset-block)` guards—and that flexibility was definitely my intent a few hours ago. Let me think about how obnoxious it would be to respect this.

ecpeterson: P.S.: `RESET` gets dropped right here: https://github.com/rigetti/quilc/blob/master/src/cfg.lisp#L201 .

appleby: > P.S.: `RESET` gets dropped right here: https://github.com/rigetti/quilc/blob/master/src/cfg.lisp#L201 .

Uh-oh! I am apparently more confused than I thought (happens all the time). I was thinking that the `(vector-push-extend instr (basic-block-code blk))` at the end of that method was explicitly preserving, *not* dropping the RESET instruction. Although I admit I do not understand what the "upcall to push the RESET instruction on" comment means... I will have to meditate on this function longer to attain enlightenment.

ecpeterson: Oh, gosh, I think I need to meditate more too.

ecpeterson: Yup, this PR needs work: the explicit `RESET` instruction does continue to live at the head of a `reset-block`, and `compiler-hook` manages its removal before (and, before #159, its readdition after) processing by the rest of the compiler. I think the least headache at this point is to make the code confirm to my description in these comments and remove this "upcall".

notmgsk: `not-so-simply-simplify-cfg`

ecpeterson:
OK: I've now also pushed a couple of commits that make `contract-paths` correctly handle `preserved-block` instances. This is a nice-to-have, but if it has to be rolled back for some reason, I also don't think it blocks the PR.

src/cfg.lisp
```diff
@@ -352,11 +352,15 @@ Return the following values:
...
-                      (basic-block-in-rewiring blk2))))
+  (assert (and (or (and (typep blk1 'preserved-block)
+                        (typep blk2 'preserved-block))
+                   (and (not (typep blk1 'preserved-block))
+                        (not (typep blk2 'preserved-block))))
+               (or (typep blk2 'reset-block)
+                   (not (basic-block-out-rewiring blk1))
+                   (not (basic-block-in-rewiring blk2))
+                   (equalp (basic-block-out-rewiring blk1)
+                           (basic-block-in-rewiring blk2)))))
```
appleby: Absolute unit of an assert! In awe of the size of this lad!

You could de-chonk this bad-boy slightly with a predicate like `blocks-agree-wrt-preservedness-p` (or something with a less silly name) and then reuse it below. This is admittedly a weirdly specific predicate, but at least it could get a handful of uses in this file.

src/cfg.lisp
```diff
@@ -354,14 +352,20 @@ Return the following values:
...
   (let ((new-blk (make-instance (type-of blk1)
-                                :code (concatenate 'vector (basic-block-code blk1)
+                                :code (concatenate 'vector
+                                                   (basic-block-code blk1)
```
appleby: So in the case where `blk1` is a `reset-block`, it doesn't need to have its RESET instruction spliced back in? Is this because it's guaranteed to have already been through `merge-sequentially` as the `blk2` argument, i.e. the cfg always has some non-reset entry-point block prepended?

appleby:
This LGTM. No further objections your honor. Plus, it's time to go walk the dog.

Oh, and thanks for playing Appleby's Annoying 20-Questions Game! We're all out of time for today, but be sure to tune in next week when I pepper you with questions like "WTF is a qubit, anyway?"

notmgsk:
> "WTF is a qubit, anyway?"

It is what it is.

My take away from this PR is a strong urge to refactor stuff, which generally means I want to procrastinate because the problem is too difficult. I will defer to smarter folk.

src/cfg.lisp
appleby: Back from dog walk. I think I understand now. If `blk1` is a `reset-block`, then `new-blk` will be as well and will have it's RESET instruction added back during reconstitution (per the latest commit). Hence, only `blk2` needs to be dealt with here?

ecpeterson: That's right. Thanks to you ^ and to Mark for pointing out that there was a stray bit of reconstitution that was happening manually in `compiler-hook` instead of where it ought to be.

notmgsk:
LGTM for a :fire: hotfix :fire:

src/cfg.lisp
ecpeterson: I agree with this. Offline discussion is in favor of an even more serious refactor, where we take a bunch of the special-casing happening all over this file and break it apart into appropriate GFs.

tests/compiler-hook-tests.lisp
```diff
@@ -181,6 +181,25 @@ RX(pi) 2
                                         (quil::parsed-program-to-logical-matrix processed-program))))))))
 
 
+(deftest test-compiler-hook-preserves-RESETs ()
+  (let* ((pp (quil::parse-quil "
+PRAGMA INITIAL_REWIRING \"PARTIAL\"
+RESET
```
appleby: Final parting thought: does it make sense to add a handful of test cases with RESETs at different points in the program, or does this one pretty much cover it?

ecpeterson: Definitely this test doesn't cover all possible Quil input configurations (though it does cover the only case that's IRL applicable in June 2019. I'd be up for adding more test inputs, but it's not super clear to me how to validate the output in a way that stays robust to future compiler improvements. I do think it's something to earmark.

P.S.: We're only now starting to add serious test coverage for MEASURE instructions ( https://github.com/rigetti/quilc/pull/263 ), for sort-of the same reasons: in June 2019, IRL MEASURE instruction usage is so constrained that you can test quilc correctness almost without involving them, and properly validating them in more complex configurations requires serious work (in this case: the involvement of the density matrix QVM).

appleby: Yes, I noticed that when quilc is in "protoquil" mode, it restricts the input to single RESET at the top, and all MEASUREs must occur at the end of the program.

appleby: So I'm finally getting around to reading the forest sdk docs, and now I see that "protoquil" is [all that's allowed](http://docs.rigetti.com/en/stable/basics.html#qpu-allowable-quil) on the QPUs. I had been putting off reading the docs because I incorrectly assumed they were limited to documenting the pyQuil API, but I now see they are in fact a goldmine of general Quil and quantum computing topics. I guess I should have read them sooner!

ecpeterson:
Tests pass locally, even if the test runners are still broken because of the QVM version mismatch. Gonna merge.

PR #269, opened 2019-05-31
Package local nickname for alexandria

notmgsk:
For the sake of reasonable line-length when currying and more!

I will add the same to `quilc.asd` etc. and replace `(alexandria:` with `(a:` if y'all concur.

### DIFF
--- a/src/package.lisp
+++ b/src/package.lisp
@@ -30,7 +30,9 @@
         #:parse-float
         #:abstract-classes
         #:singleton-classes
-        #:cl-quil.resource)
+        #:cl-quil.resource
+        #:alexandria)
+  (:local-nicknames (:a :alexandria))
   ;; options.lisp
   (:export
    #:*allow-unresolved-applications*    ; VARIABLE
--- a/src/package.lisp
+++ b/src/package.lisp
@@ -4,6 +4,7 @@
 
 (defpackage #:cl-quil.resource
   (:use #:cl)
+  (:local-nicknames (:a :alexandria))
   (:export
    #:make-resource-collection           ; FUNCTION
    #:make-qubit-resource                ; FUNCTION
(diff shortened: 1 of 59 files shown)


src/package.lisp
```diff
@@ -30,7 +30,9 @@
         #:parse-float
         #:abstract-classes
         #:singleton-classes
-        #:cl-quil.resource)
+        #:cl-quil.resource
+        #:alexandria)
```
notmgsk: This is a mistake and will be removed. Only want the line below.

appleby:
I admire your chutzpah. I don't think I get a vote, but I support your quest to reduce or eliminate repetitive strain injuries.

If `a` is too short, there is always `ax` or `alex` or even `alexandri` in honor of [the hummingbirds](https://en.wikipedia.org/wiki/Black-chinned_hummingbird) (*A. alexandri*) I sometimes see outside my window.

appleby:
> i like `a` just fine, but i'd like to add `a8a` to the list of cute alternatives

I feel like there is some quantum and/or "infinity" math joke here that I'm too thick to get.

Also, that ocho is awfully far from the home row... but maybe I'm just a poor typist.

stylewarning:
```
[package cl-quil-tests].........; 
; caught ERROR:
;   READ error during COMPILE-FILE:
;   
;     Package A does not exist.
;   
;       Line: 151, Column: 26, File-Position: 5634
;   
;       Stream: #<SB-INT:FORM-TRACKING-STREAM for "file /builds/rigetti/forest/quilc/tests/parser-tests.lisp" {100D6B0C33}>
Unhandled UIOP/LISP-BUILD:COMPILE-FILE-ERROR in thread #<SB-THREAD:THREAD "main thread" RUNNING
                                                          {10005C85B3}>:
  COMPILE-FILE-ERROR while
  compiling #<CL-SOURCE-FILE "cl-quil-tests" "parser-tests">
```

stylewarning:
@notmgsk ^

notmgsk:
I don't understand how there ended up being more additions than removals in this PR. But it is what it is.

PR #271, opened 2019-06-01
Adds compiler-rewrite-does-not-apply error condition

matt-y:
Updates relevant call sites previously using the non-specific `GIVE-UP-COMPILATION` to directly signal the new condition when applicable. 

please see rigetti/quilc/issues/46

### DIFF
--- a/src/compilation-methods.lisp
+++ b/src/compilation-methods.lisp
@@ -38,13 +38,17 @@
   ()
   (:documentation "This is signaled when a compiler is technically applicable, but would act as an identity."))
 
+(define-condition compiler-rewrite-does-not-apply (compiler-does-not-apply)
+  ()
+  (:documentation "This is signaled when a rewriting rule cannot be applied to an instruction sequence."))
 
 
 (defun give-up-compilation (&key (because ':unknown))
   (ecase because
-    (:invalid-domain (error 'compiler-invalid-domain))
-    (:acts-trivially (error 'compiler-acts-trivially))
-    (:unknown        (error 'compiler-does-not-apply))))
+    (:invalid-domain         (error 'compiler-invalid-domain))
+    (:acts-trivially         (error 'compiler-acts-trivially))
+    (:rewrite-does-not-apply (error 'compiler-rewrite-does-not-apply))
+    (:unknown                (error 'compiler-does-not-apply))))
 
 
 ;;; Core routines governing how a chip-specification's compiler list is walked
--- a/src/compilation-methods.lisp
+++ b/src/compilation-methods.lisp
@@ -47,7 +47,6 @@
   (ecase because
     (:invalid-domain         (error 'compiler-invalid-domain))
     (:acts-trivially         (error 'compiler-acts-trivially))
-    (:rewrite-does-not-apply (error 'compiler-rewrite-does-not-apply))
     (:unknown                (error 'compiler-does-not-apply))))
 
 
--- a/src/compressor/rewriting-rule-data-type.lisp
+++ b/src/compressor/rewriting-rule-data-type.lisp
@@ -14,7 +14,7 @@
 READABLE-NAME is a string used to identify the rewriting rule during debugging.
 
 COUNT is an integer that specifies the number of adjacent instructions the rule takes as input.
- 
+
 CONSUMER is a function that consumes a COMPRESSOR-CONTEXT followed by a list of instructions.  Its possible results are:
    * The error condition COMPILER-DOES-NOT-APPLY if the rule is inapplicable.
    * a LIST of instructions if the rule applies.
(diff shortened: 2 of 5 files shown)


src/compressor/rewriting-rule-data-type.lisp
```diff
@@ -23,7 +23,7 @@ CONSUMER is a function that consumes a COMPRESSOR-CONTEXT followed by a list of
   (count 1 :read-only t :type (and fixnum unsigned-byte))
   (consumer (lambda (context item)
               (declare (ignore context item))
-              (give-up-compilation))
+              (error 'compiler-rewrite-does-not-apply))
```
notmgsk: `(give-up-compilation :because ':rewrite-does-not-apply)`

src/compressor/rewriting-rule-data-type.lisp
```diff
@@ -108,12 +108,12 @@ BODY is a list of forms that, by the end, should construct a list of instruction
 ;;; dumb little macros for cond guards that we were repeating a million times
 
 (defmacro give-up-compilation-unless (test &body body)
-  "If TEST passes, proceed to execute BODY (and return the value of the final expression). If TEST fails, call GIVE-UP-COMPILATION."
+  "If TEST passes, proceed to execute BODY (and return the value of the final expression). If TEST fails, signal compiler-rewrite-does-not-apply."
```
notmgsk: I'm not convinced this macro shouldn't be `(defmacro give-up-compilation-unless (test because &body body) ...)` to mimic `give-up-compilation`. Likewise this docstring should be reverted + include description of the `because` arg.

src/compressor/rewriting-rules.lisp
```diff
@@ -391,7 +391,7 @@
                  ;; the control bit is on: 1100
                  ;; all the bits are nonzero, so we can't reduce: 1111
                  (otherwise
-                  (give-up-compilation)))))))))))
+                  (error 'compiler-rewrite-does-not-apply)))))))))))
```
notmgsk: See above.

src/compilation-methods.lisp
```diff
@@ -38,13 +38,17 @@
   ()
   (:documentation "This is signaled when a compiler is technically applicable, but would act as an identity."))
 
+(define-condition compiler-rewrite-does-not-apply (compiler-does-not-apply)
+  ()
+  (:documentation "This is signaled when a rewriting rule cannot be applied to an instruction sequence."))
 
 
 (defun give-up-compilation (&key (because ':unknown))
```
stylewarning: I think we ought to deprecate this function. I added it as a stop-gap measure for portability but that was a mistake I think. Doesn't mean you have to delete it in _this_ PR, but we shouldn't add to it.

(If you want to change the call sites that have something _other_ than `:unknown`, you're welcome to.

src/compilation-methods.lisp
```diff
@@ -38,13 +38,17 @@
   ()
   (:documentation "This is signaled when a compiler is technically applicable, but would act as an identity."))
 
+(define-condition compiler-rewrite-does-not-apply (compiler-does-not-apply)
+  ()
```
stylewarning: We could add the rewriting rule to this condition if we wanted to. But not necessary for the PR.

src/compressor/rewriting-rule-data-type.lisp
stylewarning: Maybe `give-up-rewriting-unless` provided it's _only_ used for rewriting?

src/compressor/rewriting-rule-data-type.lisp
```diff
@@ -108,12 +108,12 @@ BODY is a list of forms that, by the end, should construct a list of instruction
...
   `(cond
      (,test
       ,@body)
      (t
-      (give-up-compilation))))
+      (give-up-compilation :because ':rewrite-does-not-apply))))
```
stylewarning: signal the condition directly

src/compilation-methods.lisp
notmgsk: that would be neat

matt-y:
@notmgsk @stylewarning Hello I've tidied up the following: 

1. removed my addition to the `GIVE-UP-COMPILATION` method as it appears this is to be deprecated 
2. Renamed the `GIVE-UP-COMPILATION-UNLESS` macro used in the rewrite data type and rules files to be called `GIVE-UP-REWRITING-UNLESS` and updated call sites. 
3. Went back to directly signaling `'compiler-rewrite-does-not-apply` in place of calling `GIVE-UP-COMPILATION` 
4. Updated two other uses of the keyword argument `GIVE-UP-COMPILATION` to directly signal the respective error being specified by the keyword arg

src/compilation-methods.lisp
```diff
@@ -38,13 +38,16 @@
...
+  (:documentation "This is signaled when a rewriting rule cannot be applied to an instruction sequence."))
 
 
 (defun give-up-compilation (&key (because ':unknown))
   (ecase because
-    (:invalid-domain (error 'compiler-invalid-domain))
-    (:acts-trivially (error 'compiler-acts-trivially))
-    (:unknown        (error 'compiler-does-not-apply))))
+    (:invalid-domain         (error 'compiler-invalid-domain))
```
stylewarning: oof the new spacing

notmgsk:
@stylewarning tests pass. Let's go to LGTM.

Issue #254, opened 2019-05-23
Path to libtweedledum.{so,dylib} is hardcode into binary


notmgsk:
Steps to reproduce
1. Build quilc with tweedledum support
2. Move `libtweedledum.{so,dylib}` to somewhere else (e.g. to `./quilc/`) where it ought to be found by the load-lib process
3. Run quilc, see that library cannot be found at specific path

joachimneu:
The current Forest SDK for Linux (bare bones, downloaded from rigetti.com) also has this issue:
```
Unhandled SIMPLE-ERROR in thread #<SB-THREAD:THREAD "main thread" RUNNING
                                    {10006B85B3}>:
  Error opening shared object "/src/quilc/src/contrib/tweedledum/libtweedledum.so":
  /src/quilc/src/contrib/tweedledum/libtweedledum.so: cannot open shared object file: No such file or directory.
```

joachimneu:
(Creating the path and symlinking `libtweedledum.so` is a workaround.)

jmbr:
Is it admissible to rely on the LD_LIBRARY_PATH environment variable during build and when running quilc?

notmgsk:
@jmbr I expect that the folk who choose to install the barebones package are the type of folk who would not be averse to managing their `LD_LIBRARY_PATH`. As long as this does not affect users of the other distributions, I would be on board. We would of course have to explicity state this in documentation / READMEs where appropriate.

PR #276, opened 2019-06-05
Rely on the dynamic linker for libtweedledum (#254)

jmbr:
This is a simple way to fix the issue. Its drawback is that it requires that the user fiddles with LD_LIBRARY_PATH.

I tried different alternatives such as [1] `setf`-ing `(uiop:getenv "LD_LIBRARY_PATH")` from within the method `perform ((operation load-op) (component c->so))` and [2] trying to remove the hardcoded path from `sb-sys:*shared-objects*` before calling `save-lisp-and-die`, but this was the cleanest fix.

I've tested this patch using make test-tweedledum.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,6 +14,7 @@ QUICKLISP_BOOTSTRAP_URL=https://beta.quicklisp.org/quicklisp.lisp
 UNAME_S=$(shell uname -s)
 ZMQ_REPO=https://download.opensuse.org/repositories/network:/messaging:/zeromq:/release-stable/xUbuntu_16.04/
 PREFIX ?= /usr/local
+TWEEDLEDUM ?= $(join $(realpath $(shell pwd)), /src/contrib/tweedledum)
 
 all: quilc
 
@@ -140,6 +141,7 @@ test-quilc:
 # You can specify a different c++17-compatible compiler via the CXX
 # variable. For example: make CXX=/usr/bin/clang++ test-tweedledum
 test-tweedledum:
+	LD_LIBRARY_PATH=$(TWEEDLEDUM):$(LD_LIBRARY_PATH) \
 	$(QUICKLISP) \
 		--eval "(ql:quickload :cl-quil/tweedledum-tests)" \
 		--eval "(asdf:test-system :cl-quil/tweedledum-tests)"
--- a/Makefile
+++ b/Makefile
@@ -1,10 +1,8 @@
 COMMIT_HASH=$(shell git rev-parse --short HEAD)
 LISP_CACHE ?= $(HOME)/.cache/common-lisp
 RIGETTI_LISP_LIBRARY_HOME=../
-TWEEDLEDUM ?= $(join $(CURDIR), /src/contrib/tweedledum)
-LD_LIBRARY_PATH := $(TWEEDLEDUM):$(LD_LIBRARY_PATH)
 SBCL_BIN=sbcl
-SBCL=LD_LIBRARY_PATH=$(LD_LIBRARY_PATH) $(SBCL_BIN) --noinform --no-userinit --no-sysinit --non-interactive
+SBCL=$(SBCL_BIN) --noinform --no-userinit --no-sysinit --non-interactive
 QUICKLISP_HOME=$(HOME)/quicklisp
 QUICKLISP_SETUP=$(QUICKLISP_HOME)/setup.lisp
 QUICKLISP=$(SBCL) --load $(QUICKLISP_HOME)/setup.lisp \
@@ -16,13 +14,6 @@ QUICKLISP_BOOTSTRAP_URL=https://beta.quicklisp.org/quicklisp.lisp
 UNAME_S=$(shell uname -s)
 ZMQ_REPO=https://download.opensuse.org/repositories/network:/messaging:/zeromq:/release-stable/xUbuntu_16.04/
 PREFIX ?= /usr/local
-ifeq ($(UNAME_S),Linux)
-SOEXT := .so
-endif
-ifeq ($(UNAME_S),Darwin)
-SOEXT := .dylib
-endif
-LIBTWEEDLEDUM := $(TWEEDLEDUM)/libtweedledum$(SOEXT)
 
 all: quilc
 
@@ -75,7 +66,7 @@ endif
 ###############################################################################
 
 .PHONY: quilc
-quilc $(LIBTWEEDLEDUM): system-index.txt
+quilc: system-index.txt
 	$(SBCL) $(FOREST_SDK_FEATURE) \
 	        --eval "(setf sb-ext:\*on-package-variance\* '(:warn (:swank :swank-backend :swank-repl) :error t))" \
 		--eval '(push :hunchentoot-no-ssl *features*)' \
@@ -117,12 +108,9 @@ docker-sdk-barebones: docker
 # INSTALL/UNINSTALL
 ###############################################################################
 
-.PHONY: install install-libtweedledum
-install: quilc install-libtweedledum
+.PHONY: install
+install: quilc
 	install quilc $(DESTDIR)$(PREFIX)/bin
-
-install-libtweedledum: $(LIBTWEEDLEDUM)
-	install $< $(DESTDIR)$(PREFIX)/lib
 
 .PHONY: uninstall
 uninstall:
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,7 @@
 COMMIT_HASH=$(shell git rev-parse --short HEAD)
 LISP_CACHE ?= $(HOME)/.cache/common-lisp
 RIGETTI_LISP_LIBRARY_HOME=../
-TWEEDLEDUM ?= $(join $(realpath $(shell pwd)), /src/contrib/tweedledum)
+TWEEDLEDUM ?= $(join $(CURDIR), /src/contrib/tweedledum)
 LD_LIBRARY_PATH := $(TWEEDLEDUM):$(LD_LIBRARY_PATH)
 SBCL_BIN=sbcl
 SBCL=LD_LIBRARY_PATH=$(LD_LIBRARY_PATH) $(SBCL_BIN) --noinform --no-userinit --no-sysinit --non-interactive
--- a/Makefile
+++ b/Makefile
@@ -1,8 +1,10 @@
 COMMIT_HASH=$(shell git rev-parse --short HEAD)
 LISP_CACHE ?= $(HOME)/.cache/common-lisp
 RIGETTI_LISP_LIBRARY_HOME=../
+TWEEDLEDUM ?= $(join $(realpath $(shell pwd)), /src/contrib/tweedledum)
+LD_LIBRARY_PATH := $(TWEEDLEDUM):$(LD_LIBRARY_PATH)
 SBCL_BIN=sbcl
-SBCL=$(SBCL_BIN) --noinform --no-userinit --no-sysinit --non-interactive
+SBCL=LD_LIBRARY_PATH=$(LD_LIBRARY_PATH) $(SBCL_BIN) --noinform --no-userinit --no-sysinit --non-interactive
 QUICKLISP_HOME=$(HOME)/quicklisp
 QUICKLISP_SETUP=$(QUICKLISP_HOME)/setup.lisp
 QUICKLISP=$(SBCL) --load $(QUICKLISP_HOME)/setup.lisp \
@@ -14,7 +16,13 @@ QUICKLISP_BOOTSTRAP_URL=https://beta.quicklisp.org/quicklisp.lisp
 UNAME_S=$(shell uname -s)
 ZMQ_REPO=https://download.opensuse.org/repositories/network:/messaging:/zeromq:/release-stable/xUbuntu_16.04/
 PREFIX ?= /usr/local
-TWEEDLEDUM ?= $(join $(realpath $(shell pwd)), /src/contrib/tweedledum)
+ifeq ($(UNAME_S),Linux)
+SOEXT := .so
+endif
+ifeq ($(UNAME_S),Darwin)
+SOEXT := .dylib
+endif
+LIBTWEEDLEDUM := $(TWEEDLEDUM)/libtweedledum$(SOEXT)
 
 all: quilc
 
@@ -67,7 +75,7 @@ endif
 ###############################################################################
 
 .PHONY: quilc
-quilc: system-index.txt
+quilc $(LIBTWEEDLEDUM): system-index.txt
 	$(SBCL) $(FOREST_SDK_FEATURE) \
 	        --eval "(setf sb-ext:\*on-package-variance\* '(:warn (:swank :swank-backend :swank-repl) :error t))" \
 		--eval '(push :hunchentoot-no-ssl *features*)' \
@@ -109,9 +117,12 @@ docker-sdk-barebones: docker
 # INSTALL/UNINSTALL
 ###############################################################################
 
-.PHONY: install
-install: quilc
+.PHONY: install install-libtweedledum
+install: quilc install-libtweedledum
 	install quilc $(DESTDIR)$(PREFIX)/bin
+
+install-libtweedledum: $(LIBTWEEDLEDUM)
+	install $< $(DESTDIR)$(PREFIX)/lib
 
 .PHONY: uninstall
 uninstall:
@@ -141,7 +152,6 @@ test-quilc:
 # You can specify a different c++17-compatible compiler via the CXX
 # variable. For example: make CXX=/usr/bin/clang++ test-tweedledum
 test-tweedledum:
-	LD_LIBRARY_PATH=$(TWEEDLEDUM):$(LD_LIBRARY_PATH) \
 	$(QUICKLISP) \
 		--eval "(ql:quickload :cl-quil/tweedledum-tests)" \
 		--eval "(asdf:test-system :cl-quil/tweedledum-tests)"
--- a/app/src/entry-point.lisp
+++ b/app/src/entry-point.lisp
@@ -193,6 +193,7 @@
   (disable-debugger))
 
 (defun entry-point (argv)
+  #-win32
   (uiop:symbol-call ':cl-quil.tweedledum '#:load-tweedledum)
   (handler-case
       (%entry-point argv)
--- a/app/src/entry-point.lisp
+++ b/app/src/entry-point.lisp
@@ -193,7 +193,7 @@
   (disable-debugger))
 
 (defun entry-point (argv)
-  (cffi:load-foreign-library (read-from-string "CL-QUIL.TWEEDLEDUM::LIBTWEEDLEDUM"))
+  (uiop:symbol-call ':cl-quil.tweedledum '#:load-tweedledum)
   (handler-case
       (%entry-point argv)
     (interactive-interrupt (c)
--- a/src/contrib/tweedledum/tweedledum.lisp
+++ b/src/contrib/tweedledum/tweedledum.lisp
@@ -9,17 +9,8 @@
 (in-package #:cl-quil.tweedledum)
 
 (cffi:define-foreign-library libtweedledum
-  (:darwin (:or #.(merge-pathnames "libtweedledum.dylib"
-                                   (or *compile-file-truename*
-                                       *load-truename*))
-                "libtweedledum.dylib"
-                "tweedledum.dylib"))
-  (:unix  (:or #.(merge-pathnames "libtweedledum.so"
-                                  (or *compile-file-truename*
-                                      *load-truename*))
-               "libtweedledum.so"
-               "tweedledum.so"))
-
+  (:darwin (:or "libtweedledum.dylib" "tweedledum.dylib"))
+  (:unix (:or "libtweedledum.so" "tweedledum.so"))
   (t (:default "libtweedledum")))
 
 (defvar *tweedledum-libs-loaded* nil
--- a/src/contrib/tweedledum/tweedledum.lisp
+++ b/src/contrib/tweedledum/tweedledum.lisp
@@ -8,7 +8,8 @@
 
 (in-package #:cl-quil.tweedledum)
 
-(cffi:define-foreign-library libtweedledum
+(cffi:define-foreign-library
+    (libtweedledum :search-path "/usr/local/lib/rigetti/")
   (:darwin (:or #.(merge-pathnames "libtweedledum.dylib"
                                    (or *compile-file-truename*
                                        *load-truename*))
--- a/src/contrib/tweedledum/tweedledum.lisp
+++ b/src/contrib/tweedledum/tweedledum.lisp
@@ -4,13 +4,22 @@
   (:use #:common-lisp
         #:cffi)
   (:local-nicknames (:a :alexandria))
-  (:export #:synthesis-dbs))
+  (:export #:load-tweedledum #:synthesis-dbs))
 
 (in-package #:cl-quil.tweedledum)
 
 (cffi:define-foreign-library libtweedledum
-  (:darwin (:or "libtweedledum.dylib" "tweedledum.dylib"))
-  (:unix (:or "libtweedledum.so" "tweedledum.so"))
+  (:darwin (:or #.(merge-pathnames "libtweedledum.dylib"
+                                   (or *compile-file-truename*
+                                       *load-truename*))
+                "libtweedledum.dylib"
+                "tweedledum.dylib"))
+  (:unix  (:or #.(merge-pathnames "libtweedledum.so"
+                                  (or *compile-file-truename*
+                                      *load-truename*))
+               "libtweedledum.so"
+               "tweedledum.so"))
+
   (t (:default "libtweedledum")))
 
 (defvar *tweedledum-libs-loaded* nil
@@ -59,14 +68,15 @@ GIVE-UP-COMPILATION if INSTR is not a permutation gate."
         (t
          (quil::give-up-compilation))))))
 
-(defun run-tweedledum-tests ()
+(defun load-tweedledum ()
+  (cffi:load-foreign-library 'libtweedledum)
+  ;; TODO Some error handling here
   (unless *tweedledum-libs-loaded*
-    (push (constantly 'cl-quil.tweedledum::compile-perm-gate-with-tweedledum)
+    (push (constantly 'compile-perm-gate-with-tweedledum)
           cl-quil::*global-compilers*))
+  (setf *tweedledum-libs-loaded* t))
+
+(defun run-tweedledum-tests ()
+  (load-tweedledum)
   (uiop:symbol-call ':cl-quil-tests
                     '#:run-cl-quil-tests))
-
-;; TODO Some error handling here
-(unless *tweedledum-libs-loaded*
-  (push (constantly 'compile-perm-gate-with-tweedledum) cl-quil::*global-compilers*)
-  (setf *tweedledum-libs-loaded* t))
--- a/src/contrib/tweedledum/README.md
+++ b/src/contrib/tweedledum/README.md
@@ -116,7 +116,14 @@ $ git submodule init
 $ git submodule update --init
 ```
 
-You can then load the `cl-quil/tweedledum` package
+You must inform the dynamic linker of the path where the shared library,
+libtweedledum, will be placed once it is compiled (see below). For example,
+using the Bash shell,
+```
+export LD_LIBRARY_PATH=/path/to/quilc/src/contrib/tweedledum
+```
+
+Then, run sbcl and load the `cl-quil/tweedledum` package
 ```
 CL-USER> (ql:quickload :cl-quil/tweedledum)
 CL-USER> (cl-quil.tweedledum:synthesis-dbs '(0 1 3 2))
--- a/src/contrib/tweedledum/README.md
+++ b/src/contrib/tweedledum/README.md
@@ -120,7 +120,7 @@ You must inform the dynamic linker of the path where the shared library,
 libtweedledum, will be placed once it is compiled (see below). For example,
 using the Bash shell,
 ```
-export LD_LIBRARY_PATH=/path/to/quilc/src/contrib/tweedledum
+export LD_LIBRARY_PATH=/path/to/quilc/src/contrib/tweedledum:$LD_LIBRARY_PATH
 ```
 
 Then, run sbcl and load the `cl-quil/tweedledum` package
--- a/src/contrib/tweedledum/README.md
+++ b/src/contrib/tweedledum/README.md
@@ -116,14 +116,7 @@ $ git submodule init
 $ git submodule update --init
 ```
 
-You must inform the dynamic linker of the path where the shared library,
-libtweedledum, will be placed once it is compiled (see below). For example,
-using the Bash shell,
-```
-export LD_LIBRARY_PATH=/path/to/quilc/src/contrib/tweedledum:$LD_LIBRARY_PATH
-```
-
-Then, run sbcl and load the `cl-quil/tweedledum` package
+You can then load the `cl-quil/tweedledum` package
 ```
 CL-USER> (ql:quickload :cl-quil/tweedledum)
 CL-USER> (cl-quil.tweedledum:synthesis-dbs '(0 1 3 2))


jmbr:
Also, for the record, there's this relevant piece of info: https://sourceforge.net/p/sbcl/mailman/message/26164596/

notmgsk:
I think you should also copy the changes made in this PR: https://github.com/rigetti/quilc/pull/250 I will close that PR and we can work solely on this one.

As it currently stands, building the quilc binary does not load the tweedledum shared library. The above PR fixes that (though we may not settle on that exact solution).

src/contrib/tweedledum/tweedledum.lisp
```diff
@@ -9,17 +9,8 @@
...
-                                       *load-truename*))
-                "libtweedledum.dylib"
-                "tweedledum.dylib"))
-  (:unix  (:or #.(merge-pathnames "libtweedledum.so"
-                                  (or *compile-file-truename*
-                                      *load-truename*))
-               "libtweedledum.so"
-               "tweedledum.so"))
-
+  (:darwin (:or "libtweedledum.dylib" "tweedledum.dylib"))
```
notmgsk: I don't think we want to keep this change. It breaks the development use case where we simply do `make quilc` for local testing. If we revert the above change while keeping the rest, I don't have to futz with `LD_LIBRARY_PATH` unless I really need to.

We might instead conditionalise these out with `#-forest-sdk`.

Let me know if the above makes sense / you disagree.

jmbr: I've uploaded another commit that is a WIP. The new commit let's you do make quilc without problems.

Let me know if that goes in a more palatable direction. Otherwise, I'm OK adopting your suggestion.

notmgsk: `make quilc` *does* work, however invoking `./quilc` errors since it cannot find the library still.

jmbr: Yes, that's why I included the export LD_LIBRARY_PATH=... in the updated documentation.

jmbr:
```
export LD_LIBRARY_PATH=/home/jmbr/rigetti/quilc/src/contrib/tweedledum/tweedledum:$LD_LIBRARY_PATH
./quilc
```

appleby: My $0.02: I tend to agree with @notmgsk. If there is a way to have `make` followed by `./quilc` continue to work without requiring extra steps, it's a small hackability win, especially for new or infrequent contributors. I also don't really want to install quilc under `/usr/local` on my laptop. I guess I could change `PREFIX` to somewhere under `$HOME`, but then I'm probably back to having to twiddle `LD_LIBRARY_PATH` anyway.

I'm basing this on the assumption that getting `make && ./quilc` to work is as simple as reverting this one change. If that doesn't work or it requires gross hacks, then by all means do whatever is sane.

Makefile
```diff
@@ -1,8 +1,10 @@
 COMMIT_HASH=$(shell git rev-parse --short HEAD)
 LISP_CACHE ?= $(HOME)/.cache/common-lisp
 RIGETTI_LISP_LIBRARY_HOME=../
+TWEEDLEDUM ?= $(join $(realpath $(shell pwd)), /src/contrib/tweedledum)
```
appleby: Dubious nitpick: I think you could replace this `$(shell pwd)` with `$(CURDIR)` and save yourself a shell-out to `pwd`.  You know, to keep the Makefile blazing fast!

Feel free to ignore this suggestion, especially if doesn't work!

src/contrib/tweedledum/tweedledum.lisp
jmbr: Got it. Thanks to both.

Makefile
jmbr: Changed. I had forgotten about $(CURDIR)! I'm sure this will significantly boost speed, haha.

appleby: I just occurred to me, I think `CURDIR` is spelled `.CURDIR` in bsd make. It looks like this Makefile is already using other GNUmake-isms, but if it's intended to be portable (I suspect not), maybe `shell pwd` would work in either case.

Another small detail: I think `CURDIR` / `pwd` gives you the directory where make was invoked, which in rare circumstances may not be the same directory where the Makefile resides (e.g. `make -f foo/bar/quilc/Makefile`). If you want a path relative to the Makefile, I think you can use something like:

```
$(join $(shell dirname $(realpath $(lastword $(MAKEFILE_LIST)))), /src/contrib/tweedledum)
```

https://www.gnu.org/software/make/manual/make.html#Special-Variables

Personally, I think it's fine to use `CURDIR` and assume make will be invoked from the quilc directory, since `CURDIR` declares the intent reasonably well, whereas the `MAKEFILE_LIST` incantation is less clear.

jmbr: Thank you, @appleby. I had already thought of the GNU-specificness but noticed other constructs that are only valid in GNU makefiles and decided it was therefore OK.

src/contrib/tweedledum/tweedledum.lisp
jmbr: OK. So the latest patch loads libtweedledum from `entry-point`. There's no need to conditionalize on the `forest-sdk` feature: quilc will pick up the default location automatically either from the local build directory or rely on the dynamic linker for this (i.e., to load from a system directory such as /usr/local/lib or from a directory in LD_LIBRARY_PATH).

I tinkered a bit more with CFFI and SBCL internals and found another solution that I didn't include in the patch because it turned out not to be necessary. I'm pasting it below in case it might be useful in the future:
```(defun mark-dont-save (&rest names)
  "Ensure that a loaded shared object appearing in NAMES does not persist
after a call to SAVE-LISP-AND-DIE."
  (flet ((in-names-p (shared-object)
           (let* ((so (sb-alien::shared-object-pathname shared-object))
                  (name (pathname-name so)))
             (some (lambda (x) (string= name x)) names))))
    (let ((x (find-if #'in-names-p sb-sys:*shared-objects*)))
      (setf (sb-alien::shared-object-dont-save x) t))))
```
So, writing `(mark-dont-save "libtweedledum" "tweedledum")` will ensure that libtweedledum is dropped from the Lisp image stored by `save-lisp-and-die`.

app/src/entry-point.lisp
```diff
@@ -193,6 +193,8 @@
   (disable-debugger))
 
 (defun entry-point (argv)
+  #-win32
+  (uiop:symbol-call ':cl-quil.tweedledum '#:load-tweedledum)
```
appleby: What's the rationale for loading libtweedledum here, rather than, say, at the top of `%entry-point` or even in the body of the below `handler-case`? Not disagreeing, just trying to understand.

notmgsk: I don't know if there's a stronger reason than: it works. :)

appleby: A perfectly reasonable reason!

I guess I'm thinking about the difference between the `make quilc` vs `make quilc-unsafe` targets. Currently, I think they produce more-or-less equivalent binaries, except for the suppressing of backtraces and handling Ctrl-C for the "safe" variant. I guess after this change, `make quilc` (which uses `entry-point` as it's entry-point) produces a binary with tweedledum enabled, whereas `make quilc-unsafe` (which uses `%entry-point`) doesn't. Also, since loading tweedledum is done outside the `handler-case` in `entry-point`, you'll presumably get the full backtrace if it fails, whereas previously `make quilc` gave you a binary that only reported a brief summary of failures, and you had to build `quilc-unsafe` to get full backtraces.

I have no idea whether any of this actually matters :).

I guess another consideration is testing, since there are handful of tests that call `%entry-point`, that may be another reasons not to load tweedledum there.

notmgsk:
> This looks good to me, but I am also confused. It's OK for me to be confused, so if the explanation is complicated or tiresome to write, feel free to skip it or just respond with a soothing `:tropical_drink:` emoji.
> 
> Here are some things I am confused about:
> 
>     1. Is it true that tweedledum is only loaded/enabled when running the quilc binary (and the tweedledum tests)? If so, why not load/enable by default when loading `cl-quil` or maybe just `quilc`? To not foist the dependency on downstream users of `cl-quil`? Or because doing it at load-time was confusing the quilc binary and it needs to happen at runtime? My thinking here is that I want those sweet-sweet optimizations, but don't want to have to load tweedledum manually when hacking :). Plus, it's nice if the default setup for hacking in slime-repl is as similar as possible to whatever the quilc binary is doing.

That is correct. The choice to load tweedledum when building the binaries is that those binaries are expected to be distributed to our end-users who will naturally prefer the many benefits tweedledum provides. When we're working locally, however, that is less of a interest and so we don't bother. This keeps working locally simpler.

You might `(ql:quickload :cl-quil/tweedledum)` rather than `(ql:quickload :cl-quil)`. :)

> 
>     2. Similar to (1), would it make sense to have tweedledum enabled for the default `make test` target, if it's now enabled for the quilc binary?

I would say no but maybe. I lean towards no because we want to be sure that the tests pass without relying on this optional extension. Likewise, if the tests fail it might be confusing whether that is the base test suite failing, or tweedledum failing.

> 
>     3. How does this all work in the case of the `make quilc-sdk` / `make docker-sdk` builds? I think these are for producing the forest sdk images? I see that `libtweedledum.so` is accounted for in `app/src/mangle-shared-objects.lisp`, but I'm curious who or what is installing libtweedledum in the expected `/usr/local/lib/rigetti` location that `mangle-shared-objects.lisp` appears to expect. Also, is it true that merely loading `cl-quil/tweedledum` no longer loads the foreign library unless you call `load-tweedledum`, so that when `mangle-shared-objects.lisp` is loaded by `build-app.lisp`, it won't find libtweedledum in `sb-sys:*shared-objects*` in any case?

That happens with our SDK build tooling (not public). `(ql:quickload :cl-quil/tweedledum)` *does* load the library.

> 
>     4. Where are my glasses?

They are where they are.

notmgsk:
I have tested and verified this works in the following make targets:

1. `CXX=g++ make quilc`. 
2. `CXX=g++ make quilc-sdk`.

These all work well if libtweedledum is (a) in its original location (where it was built); (b) in the same directory as the quilc binary; or (c) LD_LIBRARY_PATH is provided. There is a ~0.3s overhead in case (1) that will maybe upset @stylewarning. 

Tests also pass.

I think we can explore alternatives for loading tweedledum in the future to mitigate the slower startup. But this PR seems to achieve its goal. LGTM.

stylewarning:
0.3s = two human reactions to a stimulus

notmgsk:
> 0.3s = two human reactions to a stimulus

Review or approve.

appleby:
> > ```
> > 1. Is it true that tweedledum is only loaded/enabled when running the quilc binary (and the tweedledum tests)? If so, why not load/enable by default when loading `cl-quil` or maybe just `quilc`? ...
> > ```
>
> That is correct. The choice to load tweedledum when building the binaries is that those binaries are expected to be distributed to our end-users who will naturally prefer the many benefits tweedledum provides. When we're working locally, however, that is less of a interest and so we don't bother. This keeps working locally simpler.
>
> You might `(ql:quickload :cl-quil/tweedledum)` rather than `(ql:quickload :cl-quil)`. :)
>

Fair enough. Personally, I would say if the optimizations are good enough for end-users, they're good enough for me while working locally, especially because my laptop is getting long-in-the-tooth, but it's easy enough for me to enable tweedledum locally. I will probably just stick it in a special .rigetti-sbclrc file that I load automatically when hacking on rigetti stuff.

> > ```
> > 2. Similar to (1), would it make sense to have tweedledum enabled for the default `make test` target, if it's now enabled for the quilc binary?
> > ```
> 
> I would say no but maybe. I lean towards no because we want to be sure that the tests pass without relying on this optional extension. Likewise, if the tests fail it might be confusing whether that is the base test suite failing, or tweedledum failing.
>

I don't think this is a big deal either way as long as both sets of tests are getting run by some build automation somewhere. But again, if the tests fail due to tweedledum related stuff, presumably the quilc binary will now exhibit the same failure mode, so I think that's a feature, not a bug. But again, it's easy for me to type `make test-tweedledum` :).

> > ```
> > 3. How does this all work in the case of the `make quilc-sdk` / `make docker-sdk` builds? ...
> > ```
> 
> That happens with our SDK build tooling (not public).

Ah, that explains it!

> `(ql:quickload :cl-quil/tweedledum)` _does_ load the library.

I don't think that's the case with the latest round of diffs? I think that used to be true since loading previously happened in the top-level, but now requires calling `cl-quil.tweedledum:load-tweedledum`. I don't see where that function gets called except at the top of `entry-point` and in `run-tweedledum-tests`, but maybe that also happens in whatever in-house build produces the sdk images?

I guess this doesn't really matter as long as the in-house sdk build is putting the lib somewhere the dynamic linker can find it at runtime, but I think the code related to libtweedledum in `mangle-shared-objects.lisp` is now a no-op?

The below REPL session is from a local git branch where I've check out the latest changes from this PR:

```
CL-USER> (ql:quickload :cl-quil/tweedledum)
To load "cl-quil/tweedledum":
  Load 1 ASDF system:
    cl-quil/tweedledum
; Loading "cl-quil/tweedledum"
..................................................
..................................................
..................................................
......................Enter path to C++17 compiler and restart: clang++
clang-8: warning: treating 'c' input as 'c++' when in C++ mode, this behavior is deprecated [-Wdeprecated]
In file included from /home/ma/src/repos/rigetti/quilc/src/contrib/tweedledum/tweedledum.c:8:
In file included from /home/ma/src/repos/rigetti/quilc/src/contrib/tweedledum/tweedledum/include/tweedledum/algorithms/synthesis/dbs.hpp:18:
/home/ma/src/repos/rigetti/quilc/src/contrib/tweedledum/tweedledum/libs/kitty/kitty/esop.hpp:36:2: warning: "DEPRECATED: the functions in this file are marked as deprecated.  Most recent implementation can be found in https://github.com/hriener/easy/ in the file src/esop/constructors.hpp" [-W#warnings]
#warning "DEPRECATED: the functions in this file are marked as deprecated.  Most recent implementation can be found in https://github.com/hriener/easy/ in the file src/esop/constructors.hpp"
 ^
1 warning generated.
............................
[package cl-quil.tweedledum].
(:CL-QUIL/TWEEDLEDUM)
CL-USER> sb-sys:*shared-objects*
(#S(SB-ALIEN::SHARED-OBJECT
    :PATHNAME #P"libffi.so.6"
    :NAMESTRING "libffi.so.6"
    :HANDLE #.(SB-SYS:INT-SAP #X7FA6F0000DD0)
    :DONT-SAVE NIL)
 #S(SB-ALIEN::SHARED-OBJECT
    :PATHNAME #P"libblas.so"
    :NAMESTRING "libblas.so"
    :HANDLE #.(SB-SYS:INT-SAP #X7FA6F0001430)
    :DONT-SAVE NIL)
 #S(SB-ALIEN::SHARED-OBJECT
    :PATHNAME #P"liblapack.so"
    :NAMESTRING "liblapack.so"
    :HANDLE #.(SB-SYS:INT-SAP #X7FA6F0006580)
    :DONT-SAVE NIL))
CL-USER> (cl-quil.tweedledum:load-tweedledum)
T
CL-USER> sb-sys:*shared-objects*
(#S(SB-ALIEN::SHARED-OBJECT
    :PATHNAME #P"libffi.so.6"
    :NAMESTRING "libffi.so.6"
    :HANDLE #.(SB-SYS:INT-SAP #X7FA6F0000DD0)
    :DONT-SAVE NIL)
 #S(SB-ALIEN::SHARED-OBJECT
    :PATHNAME #P"libblas.so"
    :NAMESTRING "libblas.so"
    :HANDLE #.(SB-SYS:INT-SAP #X7FA6F0001430)
    :DONT-SAVE NIL)
 #S(SB-ALIEN::SHARED-OBJECT
    :PATHNAME #P"liblapack.so"
    :NAMESTRING "liblapack.so"
    :HANDLE #.(SB-SYS:INT-SAP #X7FA6F0006580)
    :DONT-SAVE NIL)
 #S(SB-ALIEN::SHARED-OBJECT
    :PATHNAME #P"/home/ma/src/repos/rigetti/quilc/src/contrib/tweedledum/libtweedledum.so"
    :NAMESTRING "/home/ma/src/repos/rigetti/quilc/src/contrib/tweedledum/libtweedledum.so"
    :HANDLE #.(SB-SYS:INT-SAP #X7FA6F0007580)
    :DONT-SAVE NIL))
CL-USER>
```

notmgsk:
> > > ```
> > > 1. Is it true that tweedledum is only loaded/enabled when running the quilc binary (and the tweedledum tests)? If so, why not load/enable by default when loading `cl-quil` or maybe just `quilc`? ...
> > > ```
> > 
> > 
> > That is correct. The choice to load tweedledum when building the binaries is that those binaries are expected to be distributed to our end-users who will naturally prefer the many benefits tweedledum provides. When we're working locally, however, that is less of a interest and so we don't bother. This keeps working locally simpler.
> > You might `(ql:quickload :cl-quil/tweedledum)` rather than `(ql:quickload :cl-quil)`. :)
> 
> Fair enough. Personally, I would say if the optimizations are good enough for end-users, they're good enough for me while working locally, especially because my laptop is getting long-in-the-tooth, but it's easy enough for me to enable tweedledum locally. I will probably just stick it in a special .rigetti-sbclrc file that I load automatically when hacking on rigetti stuff.
> 

As it matures we can revisit this idea, I think. 

> > > ```
> > > 2. Similar to (1), would it make sense to have tweedledum enabled for the default `make test` target, if it's now enabled for the quilc binary?
> > > ```
> > 
> > 
> > I would say no but maybe. I lean towards no because we want to be sure that the tests pass without relying on this optional extension. Likewise, if the tests fail it might be confusing whether that is the base test suite failing, or tweedledum failing.
> 
> I don't think this is a big deal either way as long as both sets of tests are getting run by some build automation somewhere. But again, if the tests fail due to tweedledum related stuff, presumably the quilc binary will now exhibit the same failure mode, so I think that's a feature, not a bug. But again, it's easy for me to type `make test-tweedledum` :).
> 

Likewise above. Don't tell anyone, but tweedledum's tests are only run infrequently, and not as part of any CI/CD. We ought to fix this, but there are bigger fish to fry.

> > > ```
> > > 3. How does this all work in the case of the `make quilc-sdk` / `make docker-sdk` builds? ...
> > > ```
> > 
> > 
> > That happens with our SDK build tooling (not public).
> 
> Ah, that explains it!
> 
> > `(ql:quickload :cl-quil/tweedledum)` _does_ load the library.
> 
> I don't think that's the case with the latest round of diffs? I think that used to be true since loading previously happened in the top-level, but now requires calling `cl-quil.tweedledum:load-tweedledum`. I don't see where that function gets called except at the top of `entry-point` and in `run-tweedledum-tests`, but maybe that also happens in whatever in-house build produces the sdk images?
> 

Hm. You're right. The binaries are built with `build-app.sdk`. The SDK build only makes sure packages are packaged + available in particular locations when installed. Not sure if we need to fix that in this PR, however. Again something to think about.

> I guess this doesn't really matter as long as the in-house sdk build is putting the lib somewhere the dynamic linker can find it at runtime, but I think the code related to libtweedledum in `mangle-shared-objects.lisp` is now a no-op?

This is actually a blocker for the PR. Thank you for pointing it out. (Something about eyes, bugs, and shallow.) For the non-barebones packages, our libraries are installed into a location that we can manage `/usr/local/lib/rigetti/` -- tweedledum will also be placed here, so we need to make sure that that path is available to CFFI. This was originally done by patching the binary's search path with `mangle-shared-objects.lisp`, but that relies on CFFI having information about the build-time location of the libraries and now it looks like that is not available with the above changes.

@jmbr I think we can fix this by adding `/usr/local/lib/rigetti/` to the CFFI search path directly when defining the foreign lib: 

```
(cffi:define-foreign-library
    (libtweedledum :search-path "/usr/local/lib/rigetti/")
  (:darwin (:or #.(merge-pathnames "libtweedledum.dylib"
                                   (or *compile-file-truename*
                                       *load-truename*))
                "libtweedledum.dylib"
                "tweedledum.dylib"))
  (:unix  (:or #.(merge-pathnames "libtweedledum.so"
                                  (or *compile-file-truename*
                                      *load-truename*))
               "libtweedledum.so"
               "tweedledum.so"))

  (t (:default "libtweedledum")))
```

> 
> The below REPL session is from a local git branch where I've check out the latest changes from this PR:
> 
> ```
> CL-USER> (ql:quickload :cl-quil/tweedledum)
> To load "cl-quil/tweedledum":
>   Load 1 ASDF system:
>     cl-quil/tweedledum
> ; Loading "cl-quil/tweedledum"
> ..................................................
> ..................................................
> ..................................................
> ......................Enter path to C++17 compiler and restart: clang++
> clang-8: warning: treating 'c' input as 'c++' when in C++ mode, this behavior is deprecated [-Wdeprecated]
> In file included from /home/ma/src/repos/rigetti/quilc/src/contrib/tweedledum/tweedledum.c:8:
> In file included from /home/ma/src/repos/rigetti/quilc/src/contrib/tweedledum/tweedledum/include/tweedledum/algorithms/synthesis/dbs.hpp:18:
> /home/ma/src/repos/rigetti/quilc/src/contrib/tweedledum/tweedledum/libs/kitty/kitty/esop.hpp:36:2: warning: "DEPRECATED: the functions in this file are marked as deprecated.  Most recent implementation can be found in https://github.com/hriener/easy/ in the file src/esop/constructors.hpp" [-W#warnings]
> #warning "DEPRECATED: the functions in this file are marked as deprecated.  Most recent implementation can be found in https://github.com/hriener/easy/ in the file src/esop/constructors.hpp"
>  ^
> 1 warning generated.
> ............................
> [package cl-quil.tweedledum].
> (:CL-QUIL/TWEEDLEDUM)
> CL-USER> sb-sys:*shared-objects*
> (#S(SB-ALIEN::SHARED-OBJECT
>     :PATHNAME #P"libffi.so.6"
>     :NAMESTRING "libffi.so.6"
>     :HANDLE #.(SB-SYS:INT-SAP #X7FA6F0000DD0)
>     :DONT-SAVE NIL)
>  #S(SB-ALIEN::SHARED-OBJECT
>     :PATHNAME #P"libblas.so"
>     :NAMESTRING "libblas.so"
>     :HANDLE #.(SB-SYS:INT-SAP #X7FA6F0001430)
>     :DONT-SAVE NIL)
>  #S(SB-ALIEN::SHARED-OBJECT
>     :PATHNAME #P"liblapack.so"
>     :NAMESTRING "liblapack.so"
>     :HANDLE #.(SB-SYS:INT-SAP #X7FA6F0006580)
>     :DONT-SAVE NIL))
> CL-USER> (cl-quil.tweedledum:load-tweedledum)
> T
> CL-USER> sb-sys:*shared-objects*
> (#S(SB-ALIEN::SHARED-OBJECT
>     :PATHNAME #P"libffi.so.6"
>     :NAMESTRING "libffi.so.6"
>     :HANDLE #.(SB-SYS:INT-SAP #X7FA6F0000DD0)
>     :DONT-SAVE NIL)
>  #S(SB-ALIEN::SHARED-OBJECT
>     :PATHNAME #P"libblas.so"
>     :NAMESTRING "libblas.so"
>     :HANDLE #.(SB-SYS:INT-SAP #X7FA6F0001430)
>     :DONT-SAVE NIL)
>  #S(SB-ALIEN::SHARED-OBJECT
>     :PATHNAME #P"liblapack.so"
>     :NAMESTRING "liblapack.so"
>     :HANDLE #.(SB-SYS:INT-SAP #X7FA6F0006580)
>     :DONT-SAVE NIL)
>  #S(SB-ALIEN::SHARED-OBJECT
>     :PATHNAME #P"/home/ma/src/repos/rigetti/quilc/src/contrib/tweedledum/libtweedledum.so"
>     :NAMESTRING "/home/ma/src/repos/rigetti/quilc/src/contrib/tweedledum/libtweedledum.so"
>     :HANDLE #.(SB-SYS:INT-SAP #X7FA6F0007580)
>     :DONT-SAVE NIL))
> CL-USER>
> ```

appleby:
Just thinking out loud here... would it work if we add a call to `(cl-quil.tweedledum:load-tweedledum)` in `build-app.lisp`, just after loading the tweedledum package, then let `mangle-shared-objects.lisp` do it's thing for the non-barebones sdk builds (with appropriate updates to make sure `mangle-shared-objects` knows how to mangle the full `/whatever/.../quilc/src/contrib/tweedledum/libtweedledum.so` path, rather than just a bare `libtweedledum.so` like it currently appears to expect)?

Or am I missing the whole point of this PR and that doesn't work for CFFI-loaded libs for some reason? :)

notmgsk:
> Just thinking out loud here... would it work if we add a call to `(cl-quil.tweedledum:load-tweedledum)` in `build-app.lisp`, just after loading the tweedledum package, then let `mangle-shared-objects.lisp` do it's thing for the non-barebones sdk builds (with appropriate updates to make sure `mangle-shared-objects` knows how to mangle the full `/whatever/.../quilc/src/contrib/tweedledum/libtweedledum.so` path, rather than just a bare `libtweedledum.so` like it currently appears to expect)?
> 

Might be, but that seems like more work than the above fix. Not apparent to me what the benefit would be vs the cost of hard-coding `/usr/local/lib/rigetti/` in the foreign lib definition.

appleby:
Fair point. I don't think there's anything wrong with specifying the path in foreign lib definition. That approach has the advantage that everything you need to know about libtweedledum is contained in one place.  I guess the advantage of handling it in `mange-shared-objects.lisp` (assuming that even works), is that all the shared objects are then handled in the same way in the same place.

And maybe if the mangling works, would it be possible to just load the lib in `build-app.lisp`, and then drop the call to `load-tweedledum` in `entry-point`?  Again, I'm speculating here. I should probably just try it and see, but without access to the sdk build it makes testing that portion of it awkward.

notmgsk:
> Fair point. I don't think there's anything wrong with specifying the path in foreign lib definition. That approach has the advantage that everything you need to know about libtweedledum is contained in one place. I guess the advantage of handling it in `mange-shared-objects.lisp` (assuming that even works), is that all the shared objects are then handled in the same way in the same place.
> 

That's a good point in it's favour.

> And maybe if the mangling works, would it be possible to just load the lib in `build-app.lisp`, and then drop the call to `load-tweedledum` in `entry-point`? Again, I'm speculating here. I should probably just try it and see, but without access to the sdk build it makes testing that portion of it awkward.

You can test this without knowledge of the SDK packaging stuff. I think you just need to build with `make quilc-sdk`, move libtweedledum to `/usr/local/lib/rigetti`, and then test that everything works. My favourite way of testing that tweedledum is loaded is by doing the following: `time ./quilc < tests/compiler-hook-test-files/sohaib.quil`. Without tweedledum that test takes ~6s, and with tweedledum it takes ~0.5s.

Likewise test the above for `make quilc`.

appleby:
Testing in progress... please standby!

appleby:
Jeeze. ~6s *without* tweedledum? I think I need a new laptop! Mine takes ~50s without the tweeds and ~3s with. :( 

My suggested changes appear to "work on my machine", but now I am confused because I'm not sure how this state of affairs differs from what was already happening before the changes in this PR...

Scroll to the bottom of this comment to see the diff I am testing.

`make test-tweedledum` passes as well.

# `make quilc` tests

## No tweedledum at all :(

```
$ time ./quilc < tests/compiler-hook-test-files/sohaib.quil

...snip...

real    0m49.440s
user    0m24.467s
sys     0m0.242s
```

## "Developer mode"

Load tweedledum in toplevel of `tweedledum.lisp`, i.e. when loading the `cl-quil/tweedledum` ASDF system. Don't load in `entry-point`. Don't move shared object from build location.

```
$ time ./quilc < tests/compiler-hook-test-files/sohaib.quil

...snip...

real    0m3.290s
user    0m1.595s
sys     0m0.095s
```

# `make quilc-sdk` test

Same as "developer mode", above, but also move `libtweedledum.so` out of build location and into `/usr/local/lib/rigetti` and let `mangle-shared-objects.lisp` do it's thing.

```
$ time ./quilc-sdk < tests/compiler-hook-test-files/sohaib.quil

...snip...

real    0m2.874s
user    0m1.384s
sys     0m0.064s
```


# git diff

Here is the diff I am testing, applied on top of the changes in this PR.

```
diff --git a/app/src/entry-point.lisp b/app/src/entry-point.lisp
index 334714e..e15b6d5 100644
--- a/app/src/entry-point.lisp
+++ b/app/src/entry-point.lisp
@@ -193,8 +193,6 @@
   (disable-debugger))
 
 (defun entry-point (argv)
-  #-win32
-  (uiop:symbol-call ':cl-quil.tweedledum '#:load-tweedledum)
   (handler-case
       (%entry-point argv)
     (interactive-interrupt (c)
diff --git a/src/contrib/tweedledum/tweedledum.lisp b/src/contrib/tweedledum/tweedledum.lisp
index d2f6416..3e2f884 100644
--- a/src/contrib/tweedledum/tweedledum.lisp
+++ b/src/contrib/tweedledum/tweedledum.lisp
@@ -69,12 +69,14 @@ GIVE-UP-COMPILATION if INSTR is not a permutation gate."
          (quil::give-up-compilation))))))
 
 (defun load-tweedledum ()
-  (cffi:load-foreign-library 'libtweedledum)
   ;; TODO Some error handling here
   (unless *tweedledum-libs-loaded*
+    (cffi:load-foreign-library 'libtweedledum)
     (push (constantly 'compile-perm-gate-with-tweedledum)
-          cl-quil::*global-compilers*))
-  (setf *tweedledum-libs-loaded* t))
+          cl-quil::*global-compilers*)
+    (setf *tweedledum-libs-loaded* t)))
+
+(load-tweedledum)
 
 (defun run-tweedledum-tests ()
   (load-tweedledum)
```

notmgsk:
@appleby What about if you move libtweedledum into the same location as the quilc binary. Does the above test still work? I could test this myself but I'm feeling lazy :)

appleby:
> @appleby What about if you move libtweedledum into the same location as the quilc binary. Does the above test still work? I could test this myself but I'm feeling lazy :)

I'm guessing no, but let's see!

appleby:
sadtrumpet.wav

Basically, I think CFFI has to be able to find the lib at the time `cffi:cffi:load-foreign-library` is called. If you're planning to move it after it's been loaded into the lisp image, some provision equivalent to what `mangle-shared-objects.lisp` does needs to happen to tell SBCL where to find it.

Is there some specific workflow for which this use-case needs to be supported?

```
[ma@march quilc]? pwd
/home/ma/src/repos/rigetti/quilc
[ma@march quilc]? mv src/contrib/tweedledum/libtweedledum.so .
[ma@march quilc]? time ./quilc < tests/compiler-hook-test-files/sohaib.quil
Unhandled SIMPLE-ERROR in thread #<SB-THREAD:THREAD "main thread" RUNNING
                                    {10006185B3}>:
  Error opening shared object "/home/ma/src/repos/rigetti/quilc/src/contrib/tweedledum/libtweedledum.so":
  /home/ma/src/repos/rigetti/quilc/src/contrib/tweedledum/libtweedledum.so: cannot open shared object file: No such file or directory.

0: (SB-DEBUG::DEBUGGER-DISABLED-HOOK #<SIMPLE-ERROR "Error opening ~:[runtime~;shared object ~:*~S~]:
  ~A." {1000D957C3}> #<unused argument> :QUIT T)
1: (SB-DEBUG::RUN-HOOK *INVOKE-DEBUGGER-HOOK* #<SIMPLE-ERROR "Error opening ~:[runtime~;shared object ~:*~S~]:
  ~A." {1000D957C3}>)
2: (INVOKE-DEBUGGER #<SIMPLE-ERROR "Error opening ~:[runtime~;shared object ~:*~S~]:
  ~A." {1000D957C3}>)
3: (ERROR "Error opening ~:[runtime~;shared object ~:*~S~]:
  ~A." "/home/ma/src/repos/rigetti/quilc/src/contrib/tweedledum/libtweedledum.so" "/home/ma/src/repos/rigetti/quilc/src/contrib/tweedledum/libtweedledum.so: cannot open shared object file: No such file or directory")
4: (SB-SYS:DLOPEN-OR-LOSE #S(SB-ALIEN::SHARED-OBJECT :PATHNAME #P"/home/ma/src/repos/rigetti/quilc/src/contrib/tweedledum/libtweedledum.so" :NAMESTRING "/home/ma/src/repos/rigetti/quilc/src/contrib/tweedledum/libtweedledum.so" :HANDLE NIL :DONT-SAVE NIL))
5: (SB-ALIEN::TRY-REOPEN-SHARED-OBJECT #S(SB-ALIEN::SHARED-OBJECT :PATHNAME #P"/home/ma/src/repos/rigetti/quilc/src/contrib/tweedledum/libtweedledum.so" :NAMESTRING "/home/ma/src/repos/rigetti/quilc/src/contrib/tweedledum/libtweedledum.so" :HANDLE NIL :DONT-SAVE NIL))
6: (SB-SYS:REOPEN-SHARED-OBJECTS)
7: (SB-IMPL::FOREIGN-REINIT)
8: (SB-IMPL::REINIT)
9: ((FLET SB-UNIX::BODY :IN SAVE-LISP-AND-DIE))
10: ((FLET "WITHOUT-INTERRUPTS-BODY-14" :IN SAVE-LISP-AND-DIE))
11: ((LABELS SB-IMPL::RESTART-LISP :IN SAVE-LISP-AND-DIE))

real    0m1.610s
user    0m0.787s
sys     0m0.044s

```

notmgsk:
> sadtrumpet.wav

womp womp

> Is there some specific workflow for which this use-case needs to be supported?
> 

Yes, the barebones package case. For the barebones package, we do not distribute other libraries and so we do not install libraries into `/usr/local/lib/rigetti/`. We instead assume that the user is maintaining those libraries herself. We do not make the same assumption about libtweedledum, however, since that's kinda niche (not available in package repos etc).

appleby:
Ok, now I see why the lib is being loaded at runtime. I have an idea for the barebones case. Will give it a try and circle back.

appleby:
Power is out here and my laptop battery only lasts 3 minutes, so maybe I
won't circle back! Depends how long it takes to get power back...

I know it will be hard, but please carry on without me as I slowly ramp up
to speed on what this PR is even about offline :)

appleby:
Took me a while, but I think I'm starting to get it.

For what it's worth my scheme to get mangle-shared-objects to work with the barebones sdk build (just re-write the SO path to `"./libtweedledum.so"`) does appear to work. (Probably not portably, but I guess we already don't load tweedledum on win32...)

However I realize now that I may have been missing the point. Basically, different quilc build variants (quilc, quilc-sdk, quilc-sdk-barebones) might need to find libtweedledum in various different places. If we're OK with each one having exactly one place it looks for the lib, then loading the foreign lib and relying on mangle-shared-objects (or not) at binary build-time is workable. This results in a static path to the SO embedded in the lisp image, which is less flexible but allows skipping the call to `cffi:load-foreign-library` at startup in `entry-point` and means that the behavior of the quilc and quilc-unsafe variants remains the same.

For more flexibility, or if certain builds need to be able find the SO in multiple places, then a static path is useless, and you have to resort to loading the library at runtime.

appleby:
In case it wasn't clear from the above, I'm mostly out of ideas and defer to the wisdom of @notmgsk and @jmbr who have clearly already thought this through! Apologies for derailing the PR while I caught up :).

As an aside, I now realize why @jmbr was advocating to not special-case `make && ./quilc` since it does seem cleaner to just treat all the libs the same and rely on `LD_LIBRARY_PATH` (but I still think it's better if `make && ./quilc` "just works").

Taking a step back, I guess what I was indirectly advocating for above was that it would be swell if there was some way to make pre-loading tweedledum work and to reduce the diff between how libtweedledum and other shared libs are handled. It also makes me a little uneasy that there is now this extra difference between the `quilc` and `quilc-unsafe` builds, and I feel like in 5 years someone is going to look at it and wonder why disabling tweedledum is "unsafe". But these things have probably already been considered, and I apparently don't have any better ideas on how to satisfy the existing constraints.

notmgsk:
> Taking a step back, I guess what I was indirectly advocating for above was that it would be swell if there was some way to make pre-loading tweedledum work and to reduce the diff between how libtweedledum and other shared libs are handled. It also makes me a little uneasy that there is now this extra difference between the `quilc` and `quilc-unsafe` builds, and I feel like in 5 years someone is going to look at it and wonder why disabling tweedledum is "unsafe". But these things have probably already been considered, and I apparently don't have any better ideas on how to satisfy the existing constraints.

@stylewarning suggested that we could read the shared library into the quilc binary, and then at run-time spit the library out to the file system where it can be loaded by CFFI. That way we don't have to worry about distributing the library or worry about where the library is located. There's going to be run-time overhead with this, though. Maybe something to explore.

appleby:
One other (untested) idea: If there's a way to coax the ASDF build to plop `libtweedledum.so` in the root directory of the build next to the quilc binary, then all of the quilc, quilc-unsafe, and quilc-sdk-barebones variants can be reduce to the same case, namely load the lib from `"./libtweedledum.so"`. Then, only the quilc-sdk build is special and can be handled by mangle-shared-objects.

The obvious glaring problem with re-writing the SO path to "./libtweedledum.so" is that the path is actually relative to the directory where you run quilc, not the directory where the binary sits on the filesystem. So if you do something like `/foo/bar/quilc` it will break, which is clearly terrible.

I'm finally beginning to see that @jmbr had the right idea all along, and that relying on `LD_LIBRARY_PATH` like virtually every other executable is in some sense the sanest option, but I guess if we can continue to support the common cases that already work without requiring the user to fiddle with `LD_LIBRARY_PATH`, that would be neat.

I guess another idea along the lines of what @stylewarning is suggesting would be to make use of `cffi`s [static linking](https://common-lisp.net/project/cffi/manual/html_node/Static-Linking.html) option, but now I am once again hi-jacking the thread with untested conjecture, and if the history of this PR has anything to teach us it's that this idea has probably already been considered and rejected for technical reasons that I just don't understand yet :).

stylewarning:
@jmbr @appleby @notmgsk What about us loading the library by default when we quickload, but unloading the library after we asdf-load it in build-app? That way we don't have to mangle, and we can rely on LD_LIBRARY_PATH & co.?

notmgsk:
Stop giving more ideas. LGTM ASAP.

appleby:
> @jmbr @appleby @notmgsk What about us loading the library by default when we quickload, but unloading the library after we asdf-load it in build-app? That way we don't have to mangle, and we can rely on LD_LIBRARY_PATH & co.?

Considered responding, but afraid of a @notmgsk scolding...  I'm respecting his wish to keep my drive-by suggestions to myself.

ecpeterson:
I don't want to drag this out any longer than it already has been, but if it counts for anything, my opinion lands strongly on the side of searching along `LD_LIBRARY_PATH`.

jmbr:
I like the solutions suggested above and I would like to implement them. In the meantime, I've submitted a commit adding the specific search path for CFFI so that we can close this issue and proceed with further improvements later on.

appleby:
> I like the solutions suggested above and I would like to implement them. In the meantime, I've submitted a commit adding the specific search path for CFFI so that we can close this issue and proceed with further improvements later on.

That sounds reasonable to me. I have a few more ideas / questions, but I'll hold my tongue for now and wait to chime in on whatever follow-on issue gets created.

stylewarning:
So I'm not sure what folks are finally suggestion. Are we saying we ought to merge this, accept our setting of the search path to the `/rigetti/` dir, and only load at runtime (not compile-time), and wait to implement further changes to rely on `LD_LIBRARY_PATH`?

@jmbr @notmgsk @appleby @ecpeterson

notmgsk:
@stylewarning yes, that is my wish. I have tested these changes locally, looks good to me.

jmbr:
Just a quick update on this front. The idea I liked the most was @appleby 's suggestion of using static linking support within CFFI. I tried it a bit and, for the moment, my SBCL 1.5.3.55 on Mac OS complains saying that "Your SBCL doesn't support :SB-LINKABLE-RUNTIME"

appleby:
> Just a quick update on this front. The idea I liked the most was @appleby 's suggestion of using static linking support within CFFI. I tried it a bit and, for the moment, my SBCL 1.5.3.55 on Mac OS complains saying that "Your SBCL doesn't support :SB-LINKABLE-RUNTIME"

:(

I also like the static linking idea, but don't think it needs to block this PR. If we can make it work in a follow-on PR, that would be neat, but only if it's not overly painful.

@stylewarning : I agree with what @notmgsk and @jmbr said, above.

jmbr:
So here's the recipe to bundle libtweedledum with sbcl:

1. Compile libtweedledum as a **static** library.
2. Combine libsbcl.a with libtweedledum.a to obtain a **new** sbcl executable that has built-in access to the `tweedledum_synthesis_dbs` function using:

```gcc-9 -g -mmacosx-version-min=10.6 -arch x86_64 -dynamic -twolevel_namespace -bind_at_load -pagezero_size 0x100000 -Wall -o sbcl-tweedledum -Wl,-force_load libsbcl.a -Wl,-force_load libtweedledum.a -lSystem -lc -ldl -lpthread -lz -lm -lstdc++```

All we have to do is to use the new sbcl binary during hacking sessions and to invoke `save-lisp-and-die` from it when building our binaries. No quilc code needs to be changed except for, perhaps, removing the `define-foreign-library` declaration in tweedledum.lisp.

What's left is to automate the process (my preference is with CMake) of producing the static library and the new sbcl binary.

appleby:
Nice work! I'm sure figuring out the exact gcc options needed to make it work was "fun" :).

I have mixed feelings since I like the idea of statically linking libtweedledum into the quilc binary, but don't love the idea of requiring a custom sbcl build.  If we go down this route, will it still be possible to hack on quilc using stock sbcl (but maybe just without tweedledum support)?  That would make it easier for random internet riffraff (such as myself) to get involved.

Other questions that I am too lazy to put into paragraph format:
- How much bigger is the resulting quilc binary?
- Does this also statically link libc and all the rest, or just tweedledum?
- I guess this is sbcl-specific?  Maybe not the biggest deal since the current binary build is already sbcl-specific, but might be a show-stopper if it means tweedledum-enabled CCL builds are off the table (this is only a hypothetical concern).  If I understand correctly, it looks like this wouldn't affect the recently added ECL build example in `elf-library`. (This was not really a question, was it?)

jmbr:
Thanks for your comments, @appleby 

Yes, you would be able to have libtweedledum.so somewhere where your dynamic linker could find it (i.e., the same way as before, even if you don't use the special sbcl binary).

The new sbcl image (not the quilc binary) with the quilc asdf module loaded seems to be just 1M larger than the one without libtweedledum.a

This approach only links statically the tweedledum library. The rest works as in "vanilla" sbcl.

This is totally sbcl-specific but it does **not** require changes to the source code, so everything else should work as before. The only thing that changes is that we won't have to ship libtweedledum.so in binary distributions of quilc (or to place libtweedledum.so anywhere in particular if we choose to develop our code using the new sbcl binary).

notmgsk:
> What's left is to automate the process (my preference is with CMake) of producing the static library and the new sbcl binary.

Get to work ;)

appleby:
Awesome. Being able to use a stock sbcl was my main concern.

Pffft. A mere megabyte? The quilc binary is already like 26M, so I think we can live with that (assuming the 1MB increase also applies to quilc).

My only remaining concern would be to make sure the CMake parts of the build can be seamlessly integrated with the ASDF portion, quicklisp installs still work, etc.  I guess "don't break the build" goes without saying, but stating the obvious is what I do best!

appleby:
> I don't want to drag this out any longer than it already has been, but if it counts for anything, my opinion lands strongly on the side of searching along `LD_LIBRARY_PATH`.

I think it's fitting at this point to invite @ecpeterson to see if he has any objections to the static linking approach outlined above, since his opinion definitely counts for something :-).

PR #280, opened 2019-06-06
Fix numerical instability in diagonalizer-in-e-basis

### LINKED ISSUE #278: `diagonalizer-in-e-basis` suffers numerical instability
The following Quil snippet triggers an error during compression in quilc 1.8.2-hotfix:
```
DECLARE ro BIT[2]
H 0
H 1
X 0
PHASE(-13.823007675781957) 0
X 0
PHASE(-13.823007675781957) 0
RZ(-17.592918860086126) 0
RZ(-22.61946710582502) 1
CNOT 0 1
RZ(22.61946710582502) 1
CNOT 0 1
X 0
PHASE(-1.2566370614347233) 0
X 0
PHASE(-1.2566370614347233) 0
RZ(-2.5132741228694466) 1
RZ(-2.5132741228694466) 0
CNOT 0 1
RZ(2.5132741228694466) 1
CNOT 0 1
X 0
PHASE(-6.283185307173617) 0
X 0
PHASE(-6.283185307173617) 0
RZ(-7.539822368608339) 1
RZ(-7.539822368608339) 0
CNOT 0 1
RZ(12.566370614347234) 1
CNOT 0 1
H 0
RZ(-1.2566370614278395) 0
H 0
H 1
RZ(-1.2566370614278395) 1
H 1
MEASURE 0 ro[0]
MEASURE 1 ro[1]
```
The ultimate source of the error appears to be in `diagonalizer-in-e-basis`, which attempts to group the eigenvectors emitted by `magicl:eig` into common eigenspaces. The eigenspaces in the above sequence are completely degenerate: everything ought to belong to the eigenspace of weight 1, but `d-i-e-b` sorts them into pairs of weight 1±eps, which causes some cascading failure further down the line (specifically: `find-real-spanning-set` goes off the rails).

I propose we first look into Gavin's randomized algorithm for performing this computation. If his approach only deals with instability in `find-real-spanning-set`, then that's no help here; but if it obviates us from having to sort the eigenvectors as well, then that could be a good approach. Failing that, I don't have an immediate idea for an approach to this.

notmgsk:
Closes #278.

### DIFF
--- a/src/compilers/approx.lisp
+++ b/src/compilers/approx.lisp
@@ -212,9 +212,7 @@ Signals DIAGONALIZER-NOT-FOUND if the diagonalizer is not found.
 
 Three self-explanatory restarts are offered:
 
-    - RETURN-NIL,
-
-    - TRY-AGAIN, and
+    - TRY-AGAIN,
 
     - GIVE-UP-COMPILATION."
   (restart-case (find-diagonalizer-in-e-basis m +diagonalizer-max-attempts+)
--- a/src/compilers/approx.lisp
+++ b/src/compilers/approx.lisp
@@ -154,7 +154,7 @@
   :documentation "This is a precomputed Hermitian transpose of +E-BASIS+.")
 
 (defun ensure-positive-determinant (m)
-  (if (double~ -1d0 (magicl:det m))
+  (if (double= -1d0 (magicl:det m))
       (m* m (magicl:diag 4 4 (list -1 1 1 1)))
       m))
 
--- a/src/compilers/approx.lisp
+++ b/src/compilers/approx.lisp
@@ -153,6 +153,11 @@
   :test #'matrix-equality
   :documentation "This is a precomputed Hermitian transpose of +E-BASIS+.")
 
+(defun maybe-permute-basis (m)
+  (if (double~ -1d0 (magicl:det m))
+      (m* m (magicl:diag 4 4 (list -1 1 1 1)))
+      m))
+
 ;; specifically this would replace the guts of both
 ;; `diagonalizer-in-e-basis` and `real-spanning-set`. the only thing
 ;; you’d keep from `d-i-e-b` is the definition of `m` as a product
@@ -224,6 +229,14 @@
                                evecs
                                (magicl:diag (length evals) (length evals) evals)
                                (magicl:transpose evecs)))))
+        (multiple-value-bind (_ evecs)
+            (magicl:eig matrix)
+          (declare (ignore _))
+          (let* ((evecs (maybe-permute-basis (orthonormalize-matrix evecs)))
+                 (evals (magicl:matrix-diagonal
+                         (m* (magicl:transpose evecs)
+                             gammag
+                             evecs)))
               ;; assert np.allclose(np.eye(4), O2.transpose() @ O2) # Sanity check
               (assert (matrix-every #'double=
                                     (magicl:make-identity-matrix 4)
--- a/src/compilers/approx.lisp
+++ b/src/compilers/approx.lisp
@@ -183,7 +183,6 @@
                         (magicl:diag (length evals) (length evals) evals)
                         (magicl:transpose evecs))))
             (when (matrix-every #'double= gammag v)
-              ;; assert np.allclose(np.eye(4), O2.transpose() @ O2) # Sanity check
               (assert (matrix-every #'double=
                                     (magicl:make-identity-matrix 4)
                                     (m* (magicl:transpose evecs)
--- a/src/compilers/approx.lisp
+++ b/src/compilers/approx.lisp
@@ -182,8 +182,9 @@
          (gammag (m* u (magicl:transpose u))))
     (loop :repeat num-attempts :do
       (let* ((rand-coeff (random 1.0d0))
-             (matrix (matrix-map (lambda (z) (+ (* rand-coeff       (realpart z))
-                                           (* (- 1 rand-coeff) (imagpart z))))
+             (matrix (matrix-map (lambda (z)
+                                   (+ (* rand-coeff       (realpart z))
+                                      (* (- 1 rand-coeff) (imagpart z))))
                                  gammag))
              (evecs (ensure-positive-determinant
                      (orthonormalize-matrix
@@ -210,11 +211,7 @@
 
 Signals DIAGONALIZER-NOT-FOUND if the diagonalizer is not found.
 
-Three self-explanatory restarts are offered:
-
-    - TRY-AGAIN,
-
-    - GIVE-UP-COMPILATION."
+Three self-explanatory restarts are offered: TRY-AGAIN, and GIVE-UP-COMPILATION."
   (restart-case (find-diagonalizer-in-e-basis m +diagonalizer-max-attempts+)
     (try-again ()
       :report "Continue searching for the diagonlizer using random perturbations."
@@ -224,7 +221,7 @@ Three self-explanatory restarts are offered:
       (give-up-compilation))))
 
 (defun orthogonal-decomposition (m)
-  "Extracts from M a decomposition of E^* M E into A * D * B, where A and B are orthogonal and D is diagonal.  Returns the results as the VALUES triple (VALUES A D B)."
+  "Extracts from M a decomposition of E^* M E into A * D * B, where A and B are orthogonal and D is diagonal. Returns the results as the VALUES triple (VALUES A D B)."
   (let* ((m (magicl:scale (expt (magicl:det m) -1/4) m))
          (a (diagonalizer-in-e-basis m))
          (db (m* (magicl:transpose a) +edag-basis+ m +e-basis+))
@@ -238,8 +235,13 @@ Three self-explanatory restarts are offered:
                                 (cis phase))))
          (d (magicl:diag 4 4 diag))
          (b (m* (magicl:conjugate-transpose d) db)))
-    ;; it could be the case that b has negative determinant. if that's the case, we'll
-    ;; swap two of its columns that live in the same eigenspace.
+    ;; it could be the case that b has negative determinant. if that's
+    ;; the case, we'll swap two of its columns that live in the same
+    ;; eigenspace.  We want to preserve the equation M = ADB and D's
+    ;; diagonal form, so in our scheme to insert an orthogonal matrix
+    ;; O like M = A(DO)(O^T B), we need to pick O so that (1) O^T B
+    ;; has determinant 1 and (2) DO is again diagonal. The second
+    ;; condition excludes permutation matrices. - ecp
     (when (double~ -1d0 (magicl:det b))
       (setf d (m* d (magicl:diag 4 4 (list -1 1 1 1))))
       (setf b (m* (magicl:diag 4 4 (list -1 1 1 1)) b)))
@@ -283,15 +285,15 @@ Three self-explanatory restarts are offered:
       (orthogonal-decomposition (magicl:scale (expt (magicl:det mprime) -1/4) mprime))
     (let (o
           oT
-          (d-as-list (loop :for j :below 4 :collect (magicl:ref d j j)))
-          (dprime-as-list (loop :for j :below 4 :collect (magicl:ref dprime j j)))
+          (d-as-list (matrix-diagonal-entries d))
+          (dprime-as-list (matrix-diagonal-entries dprime))
           (max-fidelity 0d0))
       ;; maximize the trace over signed permutations
       (cl-permutation:doperms (sigma 4)
-        (dolist (signs (list (list 1 1 1 1)
-                             (list -1 -1 1 1)
-                             (list -1 1 -1 1)
-                             (list -1 1 1 -1)))
+        (dolist (signs (list (list  1  1  1  1)
+                             (list -1 -1  1  1)
+                             (list -1  1 -1  1)
+                             (list -1  1  1 -1)))
           (let* ((new-trace
                    (loop
                      :for x :in (cl-permutation:permute sigma d-as-list)
@@ -384,7 +386,7 @@ One can show (cf., e.g., the formulas in arXiv:0205035 with U = M2, E(rho) = V r
                  ((member #.(/ pi 2) intermediate-value :test #'double=)
                   (sort (mapcar #'abs intermediate-value) #'>))
                  (t intermediate-value)))))
-    (let* ((angles (mapcar #'phase (loop :for i :below 4 :collect (magicl:ref d i i))))
+    (let* ((angles (mapcar #'phase (matrix-diagonal-entries d)))
            (first  (mod    (+ (third angles) (fourth angles)) pi))
            (second (mod (- (+ (third angles) (first angles))) pi))
            (third  (mod    (+ (third angles) (second angles)) pi))
--- a/src/compilers/approx.lisp
+++ b/src/compilers/approx.lisp
@@ -158,10 +158,9 @@
       (m* m (magicl:diag 4 4 (list -1 1 1 1)))
       m))
 
-(a:define-constant
-    +diagonalizer-max-attempts+
+(defconstant +diagonalizer-max-attempts+
   16
-  :documentation "Maximum number of attempts DIAGONALIZER-IN-E-BASIS should make to diagonalize the input matrix using a random perturbation.")
+  "Maximum number of attempts DIAGONALIZER-IN-E-BASIS should make to diagonalize the input matrix using a random perturbation.")
 
 ;; this is a support routine for optimal-2q-compile (which explains the funny
 ;; prefactor multiplication it does).
--- a/src/compilers/approx.lisp
+++ b/src/compilers/approx.lisp
@@ -731,8 +731,7 @@ NOTE: This routine degenerates to an optimal 2Q compiler when *ENABLE-APPROXIMAT
                       (sandwich-with-local-gates center-circuit a d b q1 q0)
                     (push (cons (* circuit-cost fidelity) sandwiched-circuit)
                           candidate-pairs)))
-              (compiler-does-not-apply () nil)
-              (diagonalizer-not-found () (give-up-compilation)))))
+              (compiler-does-not-apply () nil))))
         ;; now vomit the results
         (cond
           ((endp candidate-pairs)
--- a/src/compilers/approx.lisp
+++ b/src/compilers/approx.lisp
@@ -192,8 +192,9 @@
                                     evecs))
                   (evecs)
                   "Calculated eigenvectors were not found to be orthonormal.")
-          (return-from diagonalizer-in-e-basis evecs))))))
-
+          (return-from diagonalizer-in-e-basis evecs)))))
+  (error "Could not find diagonalizer for ~%~A~% in ~D attempts."
+         m +diagonalizer-max-attempts+))
 
 (defun orthogonal-decomposition (m)
   "Extracts from M a decomposition of E^* M E into A * D * B, where A and B are orthogonal and D is diagonal.  Returns the results as the VALUES triple (VALUES A D B)."
--- a/src/compilers/approx.lisp
+++ b/src/compilers/approx.lisp
@@ -161,20 +161,30 @@
 (defconstant +diagonalizer-max-attempts+ 16
   "Maximum number of attempts DIAGONALIZER-IN-E-BASIS should make to diagonalize the input matrix using a random perturbation.")
 
+(define-condition diagonalizer-not-found (error)
+  ((matrix :initarg :matrix :reader diagonalizer-not-found-matrix)
+   (attempts :initarg :attempts :reader diagonalizer-not-found-attempts))
+  (:report (lambda (c s)
+             (format s "Could not find diagonalizer for matrix ~%~A~%after ~D attempt~:P."
+                     (diagonalizer-not-found-matrix c)
+                     (diagonalizer-not-found-attempts c))))
+  (:documentation "The diagonalizer for the given matrix was not found after a number of attempts."))
+
 ;; this is a support routine for optimal-2q-compile (which explains the funny
 ;; prefactor multiplication it does).
 ;;
 ;; This implementation is based on the function
 ;; _eig_complex_symmetric() in QuantumFlow's decompositions.py.
-(defun diagonalizer-in-e-basis (m)
-  "For M in SU(4), compute an SO(4) column matrix of eigenvectors of E^* M E (E^* M E)^T."
+(defun find-diagonalizer-in-e-basis (m num-attempts)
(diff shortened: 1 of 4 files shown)


src/compilers/approx.lisp
```diff
@@ -153,109 +153,48 @@
...
-                               (length vectors)
-                               (reduce-append vectors))
-                              reassemble-matrix)
-                             (kernel coeff-matrix)))
-         (backsolved-vectors
-           (loop :for i :below (magicl:matrix-cols backsolved-matrix)
-                 :collect (loop :for j :below (magicl:matrix-rows backsolved-matrix)
-                                :collect (magicl:ref backsolved-matrix j i)))))
-    (gram-schmidt backsolved-vectors)))
+(defun maybe-permute-basis (m)
```
stylewarning: Can we describe more precisely what this is doing? Normally this looks like it would be flipping the determinant, but we can do that with an exchange of two columns. What more is this doing?

src/compilers/approx.lisp
```diff
@@ -153,109 +153,48 @@
...
+  (if (double~ -1d0 (magicl:det m))
+      (m* m (magicl:diag 4 4 (list -1 1 1 1)))
+      m))
 
 ;; this is a support routine for optimal-2q-compile (which explains the funny
 ;; prefactor multiplication it does).
 (defun diagonalizer-in-e-basis (m)
   "For M in SU(4), compute an SO(4) column matrix of eigenvectors of E^* M E (E^* M E)^T."
+  (declare (optimize (debug 3)))
```
stylewarning: delete

src/compilers/approx.lisp
```diff
@@ -153,109 +153,48 @@
...
-                         (setf current-evects (list (second pair)))))))))))
-        ;; form a matrix out of the results so far.
-        (setf a (magicl:make-complex-matrix 4 4 real-data)))
-      ;; lastly, fix the determinant (if there's anything left to fix)
-      (when (minusp (realpart (magicl:det a)))
-        (setf a (magicl:multiply-complex-matrices a (magicl:diag 4 4 '(-1d0 1d0 1d0 1d0)))))
-      a)))
+  (let* ((u (m* +edag-basis+ m +e-basis+))
+         (gammag (m* u (magicl:transpose u)))
+         (max-tries 16))
```
stylewarning: make this a toplevel `(global-vars:define-global-var **max-tries** 16)`

src/compilers/approx.lisp
```diff
@@ -153,109 +153,48 @@
...
+    (loop :repeat max-tries :do
+      (let* ((rand-coeff (random 1.0d0))
+             (matrix (m+ (magicl:scale rand-coeff (matrix-real-part gammag))
+                         (magicl:scale (- 1 rand-coeff) (matrix-imag-part gammag)))))
+        (multiple-value-bind (_ evecs)
```
stylewarning: `(let ((evecs (nth-value 1 (magicl...`

stylewarning: even better, do away with the MVB/let entirely, and just embed this in the `let*` below

src/compilers/approx.lisp
```diff
@@ -153,109 +153,48 @@
...
+                             gammag
+                             evecs)))
+                 (v (m* evecs
+                        (magicl:diag (length evals) (length evals) evals)
+                        (magicl:transpose evecs))))
+            (when (matrix-every #'double= gammag v)
+              (assert (matrix-every #'double=
+                                    (magicl:make-identity-matrix 4)
+                                    (m* (magicl:transpose evecs)
+                                        evecs)))
```
stylewarning: write a descriptive message here.

src/compilers/approx.lisp
```diff
@@ -153,109 +153,48 @@
...
-        ;; form a matrix out of the results so far.
-        (setf a (magicl:make-complex-matrix 4 4 real-data)))
-      ;; lastly, fix the determinant (if there's anything left to fix)
-      (when (minusp (realpart (magicl:det a)))
-        (setf a (magicl:multiply-complex-matrices a (magicl:diag 4 4 '(-1d0 1d0 1d0 1d0)))))
-      a)))
+  (let* ((u (m* +edag-basis+ m +e-basis+))
+         (gammag (m* u (magicl:transpose u)))
+         (max-tries 16))
+    (loop :repeat max-tries :do
```
stylewarning: error if we exceed max tries after this `loop` with a descriptive message

src/compilers/approx.lisp
```diff
@@ -265,30 +204,23 @@
                                     (setf phase (mod (phase (magicl:ref db j i)) pi))))
                                 (cis phase))))
          (d (magicl:diag 4 4 diag))
-         (b (magicl:multiply-complex-matrices
-             (magicl:conjugate-transpose d) db)))
+         (b (m* (magicl:conjugate-transpose d) db)))
     ;; it could be the case that b has negative determinant. if that's the case, we'll
```
stylewarning: why aren't we swapping columns here?

src/compilers/approx.lisp
```diff
@@ -316,18 +248,19 @@
   ;; last line, as well as the fidelity as a values triple.
   (multiple-value-bind (aprime dprime bprime)
       (orthogonal-decomposition (magicl:scale (expt (magicl:det mprime) -1/4) mprime))
-    (let (o oT
-            (d-as-list      (loop :for j :below 4 :collect (magicl:ref d j j)))
-            (dprime-as-list (loop :for j :below 4 :collect (magicl:ref dprime j j)))
-            (max-fidelity 0d0))
+    (let (o
+          oT
+          (d-as-list (loop :for j :below 4 :collect (magicl:ref d j j)))
```
stylewarning: write a function to get the diagonal, maybe call it `diagonal-entries`

src/compilers/approx.lisp
```diff
@@ -316,18 +248,19 @@
...
+          (dprime-as-list (loop :for j :below 4 :collect (magicl:ref dprime j j)))
+          (max-fidelity 0d0))
       ;; maximize the trace over signed permutations
       (cl-permutation:doperms (sigma 4)
-        (dolist (signs (list (list  1  1  1  1)
```
stylewarning: oof why ruin the alignment

src/matrix-operations.lisp
```diff
@@ -317,3 +317,45 @@ as needed so that they are the same size."
...
+                      (magicl:matrix-cols m)
+                      (lambda (i j)
+                        (unless (apply predicate
+                                       (magicl:ref m i j)
+                                       (mapcar (a:rcurry #'magicl:ref i j)
+                                               more-m))
+                          (return-from matrix-every nil))))
+  t)
+
+(defun matrix-real-part (m)
```
stylewarning: `realpart` for consistency

src/matrix-operations.lisp
```diff
@@ -317,3 +317,45 @@ as needed so that they are the same size."
...
+
+(defun m* (&rest m)
+  (reduce #'magicl:multiply-complex-matrices
+          m))
+
+(defun m+ (&rest m)
+  (reduce #'magicl:add-matrix
+          m))
+
+(defun m- (&rest m)
```
stylewarning: this won't handle 0 or 1 argument gracefully

src/matrix-operations.lisp
```diff
@@ -317,3 +317,45 @@ as needed so that they are the same size."
...
+
+(defun matrix-imag-part (m)
+  "Returns a matrix X_ij = Im(M_ij)."
+  (matrix-map #'imagpart m))
+
+(defun m* (&rest m)
+  (reduce #'magicl:multiply-complex-matrices
+          m))
+
+(defun m+ (&rest m)
```
stylewarning: this wont handle 0 args gracefully

src/matrix-operations.lisp
```diff
@@ -317,3 +317,45 @@ as needed so that they are the same size."
...
+
+(defun matrix-real-part (m)
+  "Returns a matrix X_ij = Re(M_ij)."
+  (matrix-map #'realpart m))
+
+(defun matrix-imag-part (m)
+  "Returns a matrix X_ij = Im(M_ij)."
+  (matrix-map #'imagpart m))
+
+(defun m* (&rest m)
```
stylewarning: this wont handle 1 arg gracefully

src/compilers/approx.lisp
stylewarning: maybe `ensure-positive-determinant`, or perhaps take out the `double~` call and call it `flip-determinant`

src/compilers/approx.lisp
```diff
@@ -153,109 +153,48 @@
...
-                               (reduce-append vectors))
-                              reassemble-matrix)
-                             (kernel coeff-matrix)))
-         (backsolved-vectors
-           (loop :for i :below (magicl:matrix-cols backsolved-matrix)
-                 :collect (loop :for j :below (magicl:matrix-rows backsolved-matrix)
-                                :collect (magicl:ref backsolved-matrix j i)))))
-    (gram-schmidt backsolved-vectors)))
+(defun maybe-permute-basis (m)
+  (if (double~ -1d0 (magicl:det m))
```
stylewarning: @ecpeterson @notmgsk Can we do `minusp` instead of checking if it's approx 1?

src/compilers/approx.lisp
```diff
@@ -153,109 +153,48 @@
...
+         (gammag (m* u (magicl:transpose u)))
+         (max-tries 16))
+    (loop :repeat max-tries :do
+      (let* ((rand-coeff (random 1.0d0))
+             (matrix (m+ (magicl:scale rand-coeff (matrix-real-part gammag))
+                         (magicl:scale (- 1 rand-coeff) (matrix-imag-part gammag)))))
+        (multiple-value-bind (_ evecs)
+            (magicl:eig matrix)
+          (declare (ignore _))
+          (let* ((evecs (maybe-permute-basis (orthonormalize-matrix evecs)))
```
stylewarning: see previous comment about `maybe-permute-basis`, and possibly putting the `if` here

src/matrix-operations.lisp
```diff
@@ -317,3 +317,45 @@ as needed so that they are the same size."
...
       (find-root f new-guess (1- depth-bound)))))
+
+(defun matrix-map (f m &rest more-m)
+  "Returns a matrix X_ij = f(M_ij, ...)."
+  (magicl:tabulate (magicl:matrix-rows m)
+                   (magicl:matrix-cols m)
+                   (lambda (i j)
+                     (apply #'funcall f
+                            (magicl:ref m i j)
+                            (mapcar (a:rcurry #'magicl:ref i j)
```
stylewarning: "I ain't afraid of no `cons`!"

src/compilers/approx.lisp
ecpeterson: It doesn't matter: multiplying by a matrix that exchanges two columns would also be fine, and `ensure-positive-determinant` is a better name for this other choice of fixer.

src/compilers/approx.lisp
stylewarning: (also choose a good name)

src/compilers/approx.lisp
ecpeterson: We want to preserve the equation M = ADB and D's diagonal form, so in our scheme to insert an orthogonal matrix O like M = A(DO)(O^T B), we need to pick O so that (1) O^T B has determinant 1 and (2) DO is again diagonal. The second condition excludes permutation matrices.

src/compilers/approx.lisp
ecpeterson: I think I was convincing offline that I'd prefer `double=` over `double-` or (something like) `minusp`.

src/compilers/approx.lisp
stylewarning: I see, let's document this!

src/compilers/approx.lisp
stylewarning: so we should change this to `double=`

src/compilers/approx.lisp
```diff
@@ -153,41 +153,43 @@
...
   :documentation "This is a precomputed Hermitian transpose of +E-BASIS+.")
 
-(defun maybe-permute-basis (m)
+(defun ensure-positive-determinant (m)
   (if (double~ -1d0 (magicl:det m))
       (m* m (magicl:diag 4 4 (list -1 1 1 1)))
       m))
 
+(a:define-constant
+    +diagonalizer-max-attempts+
```
stylewarning: can you put this on the same line as the `define-constant`?

Also, no need for Alexandria here, use `defconstant` (CL's `defconstant` uses `eql` for equality, and 16 will always eql 16.)

src/compilers/approx.lisp
```diff
@@ -153,41 +153,43 @@
...
+                         gammag
+                         evecs)))
+             (v (m* evecs
+                    (magicl:diag (length evals) (length evals) evals)
+                    (magicl:transpose evecs))))
+        (when (matrix-every #'double= gammag v)
+          (assert (matrix-every #'double=
+                                (magicl:make-identity-matrix 4)
+                                (m* (magicl:transpose evecs)
+                                    evecs)))
```
stylewarning: helpful error message?

src/compilers/approx.lisp
```diff
@@ -153,41 +153,43 @@
...
+                     (orthonormalize-matrix
+                      (nth-value 1 (magicl:eig matrix)))))
+             (evals (magicl:matrix-diagonal
+                     (m* (magicl:transpose evecs)
+                         gammag
+                         evecs)))
+             (v (m* evecs
+                    (magicl:diag (length evals) (length evals) evals)
+                    (magicl:transpose evecs))))
+        (when (matrix-every #'double= gammag v)
```
stylewarning: can we be looser in checking here with double noodle? If not in the `when`, definitely with the `assert`. Agree? Disagree?

ecpeterson: I want the `when` to stay as-is: any kind of comparison that guards a code substitution should use `double=`. I don't care so much about which version the `assert` uses, but I would like careful `assert`s like this to be guarded by `*compress-carefully*`.

src/matrix-operations.lisp
```diff
@@ -348,14 +348,17 @@ as needed so that they are the same size."
...
 
-(defun m+ (&rest m)
+(defun m+ (m &rest more-m)
   (reduce #'magicl:add-matrix
-          m))
+          more-m
+          :initial-value m))
 
-(defun m- (&rest m)
+(defun m- (m &rest more-m)
```
stylewarning: m- should negate in the case of 1 arg

src/compilers/approx.lisp
```diff
@@ -154,14 +154,13 @@
...
 (defun ensure-positive-determinant (m)
-  (if (double~ -1d0 (magicl:det m))
+  (if (double= -1d0 (magicl:det m))
       (m* m (magicl:diag 4 4 (list -1 1 1 1)))
       m))
 
-(a:define-constant
-    +diagonalizer-max-attempts+
+(defconstant +diagonalizer-max-attempts+
   16
```
stylewarning: put on same line as defconstant

src/compilers/approx.lisp
```diff
@@ -153,109 +153,50 @@
...
-        (setf a (magicl:make-complex-matrix 4 4 real-data)))
-      ;; lastly, fix the determinant (if there's anything left to fix)
-      (when (minusp (realpart (magicl:det a)))
-        (setf a (magicl:multiply-complex-matrices a (magicl:diag 4 4 '(-1d0 1d0 1d0 1d0)))))
-      a)))
+  (let* ((u (m* +edag-basis+ m +e-basis+))
+         (gammag (m* u (magicl:transpose u))))
+    (loop :repeat +diagonalizer-max-attempts+ :do
+      (let* ((rand-coeff (random 1.0d0))
+             (matrix (m+ (magicl:scale rand-coeff (matrix-real-part gammag))
```
ecpeterson: It would be appropriate to attribute this incarnation of this procedure to Gavin / to QuantumFlow.

src/compilers/approx.lisp
```diff
@@ -153,109 +153,50 @@
...
-                              reassemble-matrix)
-                             (kernel coeff-matrix)))
-         (backsolved-vectors
-           (loop :for i :below (magicl:matrix-cols backsolved-matrix)
-                 :collect (loop :for j :below (magicl:matrix-rows backsolved-matrix)
-                                :collect (magicl:ref backsolved-matrix j i)))))
-    (gram-schmidt backsolved-vectors)))
+(defun ensure-positive-determinant (m)
+  (if (double= -1d0 (magicl:det m))
+      (m* m (magicl:diag 4 4 (list -1 1 1 1)))
```
ecpeterson: I find it a little hard to read the function `m*` next to the variable `m`. I don't have a superior suggestion, though.

src/compilers/approx.lisp
notmgsk: Yeah! My bad. I did intend to do that way back when.

src/compilers/approx.lisp
stylewarning: `m:mbop`

ecpeterson: omg

notmgsk: i could change the matrix `m` to something else if you really prefer

ecpeterson: I don't have that preference; `m` is used all over, and I don't want to deal with a mass renaming.

notmgsk: good because i was not serious

src/compilers/approx.lisp
```diff
@@ -153,109 +153,83 @@
...
+  (error 'diagonalizer-not-found :matrix m :attempts num-attempts))
+
+(defun diagonalizer-in-e-basis (m)
+  "For M in SU(4), compute an SO(4) column matrix of eigenvectors of E^* M E (E^* M E)^T.
+
+Signals DIAGONALIZER-NOT-FOUND if the diagonalizer is not found.
+
+Three self-explanatory restarts are offered:
+
+    - RETURN-NIL,
```
stylewarning: delete

src/compilers/approx.lisp
notmgsk: because emacs

src/compilers/approx.lisp
```diff
@@ -153,109 +153,81 @@
...
-      ;; lastly, fix the determinant (if there's anything left to fix)
-      (when (minusp (realpart (magicl:det a)))
-        (setf a (magicl:multiply-complex-matrices a (magicl:diag 4 4 '(-1d0 1d0 1d0 1d0)))))
-      a)))
+  (let* ((u (m* +edag-basis+ m +e-basis+))
+         (gammag (m* u (magicl:transpose u))))
+    (loop :repeat num-attempts :do
+      (let* ((rand-coeff (random 1.0d0))
+             (matrix (matrix-map (lambda (z) (+ (* rand-coeff       (realpart z))
+                                           (* (- 1 rand-coeff) (imagpart z))))
```
stylewarning: spacing

stylewarning: indentation

src/compilers/approx.lisp
```diff
@@ -153,109 +153,81 @@
...
+                  "Calculated eigenvectors were not found to be orthonormal.")
+          (return-from find-diagonalizer-in-e-basis evecs)))))
+  (error 'diagonalizer-not-found :matrix m :attempts num-attempts))
+
+(defun diagonalizer-in-e-basis (m)
+  "For M in SU(4), compute an SO(4) column matrix of eigenvectors of E^* M E (E^* M E)^T.
+
+Signals DIAGONALIZER-NOT-FOUND if the diagonalizer is not found.
+
+Three self-explanatory restarts are offered:
```
stylewarning: put restarts on the same line as the colon

src/compilers/approx.lisp
stylewarning: add this comment

src/compilers/approx.lisp
stylewarning: add this function

src/compilers/approx.lisp
stylewarning: :(

src/compilers/approx.lisp
stylewarning: your classic 1 char lambda ruining indentation

src/compilers/approx.lisp
```diff
@@ -763,7 +731,8 @@ NOTE: This routine degenerates to an optimal 2Q compiler when *ENABLE-APPROXIMAT
                       (sandwich-with-local-gates center-circuit a d b q1 q0)
                     (push (cons (* circuit-cost fidelity) sandwiched-circuit)
                           candidate-pairs)))
-              (compiler-does-not-apply () nil))))
+              (compiler-does-not-apply () nil)
+              (diagonalizer-not-found () (give-up-compilation)))))
```
stylewarning: @ecpeterson What do you think here?

ecpeterson: This isn't fully unreasonable, but it doesn't have the behavior I was imagining: `give-up-compilation` often signals an expected error that gets gracefully handled by the caller. `diagonalizer-not-found` is an unexpected error, and I'd like the caller to have the option to do something about it—including proceed "as normal", as if `give-up-compilation` had been emitted instead—but I don't like the idea of the unexpected error getting silently and automatically converted into the meant-to-be-recovered-from error.

But, again, it's ultimately not that bad in context. If this routine flags `give-up-compilation` to its usual caller, what will happen in 99% of cases is catastrophic failure shortly thereafter. So, stack traces will be murkier, but we won't start emitting bad output.

notmgsk: So should we remove this handler, and just let the error bubble up?

ecpeterson: Yeah, I suppose so. You could consider trapping it in apply-translation-compilers with the same option to restart with the next compiler, but it’d be fine to not bother.

src/compilers/approx.lisp
```diff
@@ -153,109 +153,78 @@
...
+    (try-again ()
+      :report "Continue searching for the diagonlizer using random perturbations."
+      (diagonalizer-in-e-basis m))
+    (give-up-compilation ()
+      :report "Give up compilation."
+      (give-up-compilation))))
 
 (defun orthogonal-decomposition (m)
-  "Extracts from M a decomposition of E^* M E into A * D * B, where A and B are orthogonal and D is diagonal.  Returns the results as the VALUES triple (VALUES A D B)."
+  "Extracts from M a decomposition of E^* M E into A * D * B, where A and B are orthogonal and D is diagonal. Returns the results as the VALUES triple (VALUES A D B)."
```
ecpeterson: :'(

PR #281, opened 2019-06-10
Remove singleton metaclass for NOP, HALT, RESET, and WAIT

### LINKED ISSUE #270: Singleton instructions share rewiring comments
Rewiring comments are currently stored in the `**comments**` hash table, keyed on object identity. This means that if a program contains multiple instructions that are represented as singleton classes in `cl-quil`, all occurrences of those instructions in the program will share the same rewiring comment.

The current list of singleton instructions is: HALT, NOP, RESET, and WAIT.

I am not sure what this means for program correctness, but at a minimum this means that any test that wants to verify the output of `print-instruction` for a singleton instruction has to clear the `**comments**` table first, otherwise prior test runs may have attached a rewiring comment which will get printed by `print-instruction`.

The desired outcome of this issue to investigate and determine the best way to fix this without breaking other stuff.

## Example 1: PRINT-PARSED-PROGRAM output affected by intervening call to COMPILER-HOOK.

```
CL-USER> (progn
	   (remhash (make-instance 'quil::wait) quil::**comments**)
	   (quil::print-parsed-program (quil:parse-quil "WAIT"))
	   (quil:compiler-hook (quil:parse-quil "WAIT") (quilc::lookup-isa-descriptor-for-name "8Q"))
	   (quil::print-parsed-program (quil:parse-quil "WAIT")))
WAIT
WAIT                                    # Entering rewiring: #(0 1 2 3 4 5 6 7)
NIL
CL-USER> (progn
	   (remhash (make-instance 'quil::halt) quil::**comments**)
	   (quil::print-parsed-program (quil:parse-quil "HALT"))
	   (quil:compiler-hook (quil:parse-quil "HALT") (quilc::lookup-isa-descriptor-for-name "8Q"))
	   (quil::print-parsed-program (quil:parse-quil "HALT")))
HALT
HALT                                    # Entering/exiting rewiring: (#(0 1 2 3 4 5 6 7) . #(0 1 2 3 4 5 6 7))
NIL
```

## Example 2: Program with multiple WAIT instructions share same rewiring

To be fair, I'm not certain this a bug. Maybe both WAITs are supposed to have the same rewiring.

```
./quilc <<EOF
> WAIT
> RESET
> WAIT
> EOF
WAIT                                    # Entering rewiring: #(0 1 2 3 4 5 6 7)
WAIT                                    # Entering rewiring: #(0 1 2 3 4 5 6 7)
```

This was originally reported as "Curio No. 2" in #245.

The testing consequences became evident while working on #257, e.g. see [this comment thread](https://github.com/rigetti/quilc/pull/257/#discussion_r288601185).

appleby:
- Drop singleton metaclass for the above instructions
- Remove RESET-COMMENT-TABLE-HACK, which is no longer needed.
- Update a parser test that broke as a result of this change.

Fixes #270

The self-flagellating feature expressions are to serve as a reminder that, in a better world, referential transparency would be preserved.

### DIFF
--- a/src/ast.lisp
+++ b/src/ast.lisp
@@ -326,6 +326,11 @@ as a permutation."
 (defclass no-operation (instruction)
   ()
   (:documentation "The \"do-nothing\" instruction.")
+  ;; The singleton-class is disabled rather than removed here (and elsewhere) as a reminder that
+  ;; this is a quick fix. Ideally, we'd like to find a way to preserve the singleton nature of these
+  ;; AST classes, but still work with rewiring comments and the guts of logical-scheduler. See
+  ;; https://github.com/rigetti/quilc/issues/270 for more context.
+  #+#:appleby-sufficiently-classy
   (:metaclass singleton-class))
 
 (defmethod arguments ((instruction no-operation)) #())
@@ -358,6 +363,7 @@ numbers. It must start with a string.")
 (defclass halt (instruction)
   ()
   (:documentation "An instruction to immediately halt all execution.")
+  #+#:appleby-sufficiently-classy
   (:metaclass singleton-class))
 
 (defmethod arguments ((instruction halt)) #())
@@ -366,6 +372,7 @@ numbers. It must start with a string.")
 (defclass reset (instruction)
   ()
   (:documentation "An instruction to reset all qubits to the |0>-state.")
+  #+#:appleby-sufficiently-classy
   (:metaclass singleton-class))
 
 (defmethod arguments ((instruction reset)) #())
@@ -384,6 +391,7 @@ as the reset is formally equivalent to measuring the qubit and then conditionall
 (defclass wait (instruction)
   ()
   (:documentation "An instruction to wait for some signal from a classical processor.")
+  #+#:appleby-sufficiently-classy
   (:metaclass singleton-class))
 
 (defmethod arguments ((instruction wait)) #())
--- a/tests/parser-tests.lisp
+++ b/tests/parser-tests.lisp
@@ -134,8 +134,7 @@
               "    RESET qq"
               "RESET"))
          (code (parsed-program-executable-code p)))
-    (is (equalp (aref code 0)
-                (make-instance 'reset)))
+    (is (typep (aref code 0) 'reset))
     (is (equalp (reset-qubit-target (aref code 1))
                 (qubit 5)))))
 
--- a/tests/printer-tests.lisp
+++ b/tests/printer-tests.lisp
@@ -15,22 +15,11 @@
   (with-output-to-string (s)
     (funcall printer (funcall parser input) s)))
 
-(defun reset-comment-table-hack ()
-  "Reset the CL-QUIL::**COMMENTS** hash table.
-
-  Resetting the **COMMENTS** table might be required for any test that wants to validate the output
-  of PRINT-PARSED-PROGRAM for programs that contain HALT, NOP, RESET, or other instructions that are
-  represented by singleton classes. This is because the **COMMENTS** hash table is keyed on object
-  identity, so if any prior tests have attached rewiring comments to any singleton instructions,
-  they will persist and be printed by PRINT-PARSED-PROGRAM."
-  (setf quil::**comments** (quil::make-comment-table)))
-
 (defun update-print-parsed-program-golden-files (&key skip-prompt)
   "Call UPDATE-PRINT-PARSED-PROGRAM-GOLDEN-FILES on all the files in *PRINTER-TEST-FILES-DIRECTORY*.
 
 See the documentation string for UPDATE-PRINT-PARSED-PROGRAM-GOLDEN-FILES for more info and an
 admonition against carelessness."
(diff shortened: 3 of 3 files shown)


notmgsk:
@appleby if you have time, for my sake could you provide a digest (short but more detail than the above) of the what and why of this PR. Reading through the referenced issue makes my eyes cross.

src/ast.lisp
```diff
@@ -326,6 +326,7 @@ as a permutation."
 (defclass no-operation (instruction)
   ()
   (:documentation "The \"do-nothing\" instruction.")
+  #+#:appleby-sufficiently-classy
```
stylewarning: please write a detailed comment about this wackiness

appleby:
> @appleby if you have time, for my sake could you provide a digest (short but more detail than the above) of the what and why of this PR. Reading through the referenced issue makes me go cross-eyed.

Sure! I'm not sure I met the requirement for keeping it short, but here goes :).

Basically, rewiring comments live in a global `**comments**` (weak) hash table where the hash keys are instruction instances and the values are strings representing the rewiring comments. The hash table test is `eq`, so identical objects map to the same value. In this case you can think of "identical objects" as meaning instruction instances getting compared by pointer identity. For "normal" non-singleton instructions, this works fine. You want a 1-to-1 mapping between instructions and comments, and every time you call `make-instance` for a non-singleton class, you get a fresh instance allocated with a distinct object identity. But for `singleton` instructions (HALT, NOP, WAIT, RESET), you only ever get one instance of the class no matter how many copies of that instruction you have in your program, or indeed even across programs.

Now my understanding is that this doesn't come up much in "real world" usage since protoquil I think is limited to a single RESET at the beginning and a single HALT (?) that gets inserted at the end. But, in general, in full-blooded Quil you could have multiple of these instructions in a single program, in which case all occurrences of, say, WAIT instructions will share identical rewiring comments in that program.

This also comes up in testing or REPL hacking. Suppose you run a `compiler-hook` test that ends up attaching a rewiring comment to a RESET instruction in the given parsed-program under test. Then suppose a later test runs that also contains a RESET instruction, only this one shouldn't have any rewiring comment attached to it. In this case, the new program's RESET will still have the old rewiring comment hanging around unless you cleared the hash table between the two tests.

An example is worth a thousand words, so here is a little code snippet to help clarify. Notice that the first call to `print-parsed-program` just prints `HALT`, whereas after calling `compiler-hook` (which attaches an entering/exiting rewiring comment to the HALT instruction) calling `print-parsed-program` on the exact same Quil program now prints the HALT + the rewiring comment that was attached by the previous call to `compiler-hook`. This is problem for the recently-merged `test-print-parsed-program-golden-files`, but could conceivably also affect correctness of other tests or even real Quil compilations processed by quilc.

```lisp
(progn
	   (remhash (make-instance 'quil::halt) quil::**comments**)
	   (quil::print-parsed-program (quil:parse-quil "HALT"))
	   (quil:compiler-hook (quil:parse-quil "HALT") (quilc::lookup-isa-descriptor-for-name "8Q"))
	   (quil::print-parsed-program (quil:parse-quil "HALT")))
HALT
HALT                                    # Entering/exiting rewiring: (#(0 1 2 3 4 5 6 7) . #(0 1 2 3 4 5 6 7))
NIL
```

appleby:
> @appleby if you have time, for my sake could you provide a digest (short but more detail than the above) of the what and why of this PR. Reading through the referenced issue makes me go cross-eyed.

I realized my above response was mostly focused on the "why" and not much about the "what".

One simple but perhaps not-so-aesthetically-pleasing way to fix the issue described above is to just drop the `singleton` metaclass for HALT, RESET, NOP, and WAIT, in which case the 1-to-1 mapping between instructions and comments is trivially restored.

In the issue discussion for #270, it was suggested that ideally some other fix could be found that doesn't rely on object identity. To be fair this is not just about aesthetics, but about referential transparency, which is a really nice property for programs to have if you can get it, that often makes reasoning about them simpler.

Long story short: while it might be possible to find some scheme for rewiring comments that doesn't depend on object identity, there are similar assumptions made elsewhere that would be harder to re-work, hence the `(:metaclass singleton-class)` forms are disabled via a feature expression rather than removed outright, as a reminder to future appleby to go back and fix it good.

src/ast.lisp
stylewarning: (in the source code of course)

appleby: Understood :).  I'll most likely push an update later today. Currently wrapped up in other stuff.

appleby: This three-line comment took nearly three days to fully mature. Please, handle it with care.

appleby:
I should mention that I had to disable `test-measure-semantics` to get the tests to pass, but that was also the case on the master branch for me, so I don't think it's related to these changes.

appleby:
> I should mention that I had to disable `test-measure-semantics` to get the tests to pass, but that was also the case on the master branch for me, so I don't think it's related to these changes.

This turned out to be user error. All tests are passing now.

PR #287, opened 2019-06-13
Replace references to cl-quil.clifford::extract-qubits-used

notmgsk:
For details: https://github.com/rigetti/quilc/pull/157

PR #288, opened 2019-06-14
Test quilc endpoints

notmgsk:
quilc endpoints are mostly untested, which allowed a bug to creep in (see #287).

The tests here could be better, if I knew more about cliffords etc.

### DIFF
--- a/app/src/rpc-server.lisp
+++ b/app/src/rpc-server.lisp
@@ -64,15 +64,12 @@
       (error "Currently no more than two qubit randomized benchmarking is supported."))
     (let* ((cliffords (mapcar #'quil.clifford::clifford-from-quil gateset))
            (qubits-used (mapcar (a:compose
-                                 (a:curry #'reduce #'union)
-                                 #'cl-quil.clifford::extract-qubits-used
+                                 #'cl-quil:qubits-used
                                  #'cl-quil:parse-quil)
                                 gateset))
            (qubits-used-by-interleaver
              (when interleaver
-               (reduce #'union
-                       (cl-quil.clifford::extract-qubits-used
-                        (cl-quil:parse-quil interleaver)))))
+               (cl-quil:qubits-used (cl-quil:parse-quil interleaver))))
            (qubits (union qubits-used-by-interleaver (reduce #'union qubits-used)))
            (embedded-cliffords (loop :for clifford :in cliffords
                                      :for i :from 0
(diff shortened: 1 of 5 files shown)


app/src/rpc-server.lisp
```diff
@@ -64,15 +64,12 @@
       (error "Currently no more than two qubit randomized benchmarking is supported."))
     (let* ((cliffords (mapcar #'quil.clifford::clifford-from-quil gateset))
            (qubits-used (mapcar (a:compose
-                                 (a:curry #'reduce #'union)
```
ecpeterson: nice clean up of some ancient anthony code

app/tests/rpcq-tests.lisp
```diff
@@ -46,3 +46,78 @@
...
                (is (quil::matrix-equality mat1 mat2)))))
       (bt:destroy-thread server-thread))))
+
+(deftest test-native-quil-to-binary-endpoint ()
+  (let* ((server-function (lambda ()
+                            (quilc::start-rpc-server)))
+         (server-thread (bt:make-thread server-function)))
+    (sleep 1)
+    (unwind-protect
+         (rpcq:with-rpc-client (client "tcp://127.0.0.1:5555")
```
ecpeterson: If you're feeling brave, you can modify the `host` kwarg of `start-rpc-server` to include the protocol, then use an initialization more like https://github.com/rigetti/rpcq/blob/master/src/server.lisp#L281 to keep the tests from failing if there happens to be a preexisting quilc server.

src/package.lisp
```diff
@@ -381,6 +381,13 @@
    #:rewrite-arithmetic                 ; FUNCTION/TRANSFORMATION
    )
 
+  ;; compressor.lisp
```
appleby: `analysis/compress-qubits.lisp`

app/tests/rpcq-tests.lisp
notmgsk: Oh neat. I thought about doing something like randomizing ports, but felt it was hacky. I'll do this instead

src/package.lisp
```diff
@@ -381,7 +381,7 @@
    #:rewrite-arithmetic                 ; FUNCTION/TRANSFORMATION
    )
 
-  ;; compressor.lisp
+  ;; analysis/compressor.lisp
```
appleby: lol. `analysis/compress-qubits.lisp`

notmgsk: i may be smiling but inside im crying

app/tests/rpcq-tests.lisp
```diff
@@ -4,120 +4,102 @@
 
 (in-package #:quilc-tests)
 
+(defmacro with-random-rpc-client ((client) &body body)
```
appleby: This is beautiful macroing right here.

notmgsk: I was expecting a comment about it being leaky.

appleby: Cleaned up the rest of the code so much, this thing could be Niagra Falls and I wouldn't care.

src/package.lisp
appleby: Third time's a charm

notmgsk:
Tests fail still because quicklisp's QVM is still a point-release behind.

Issue #226, opened 2019-05-10
Reduce the number of special variables bound in quilc::process-options


appleby:
Per review comments in #212, it would be nice to reduce the number of special variables in `app/src/entry-point.lisp` wherever that is possible and makes sense.

Most of the bindings occur in `quilc::process-options`.

notmgsk:
Can we close this now?

appleby:
No, this is a (perhaps poorly worded) separate issue to the one addressed in #212. This is about getting rid of the special vars altogether wherever possible, whereas #212 was about moving away from global mutation via `setf` and friends to dynamic let bindings.

PR #289, opened 2019-06-14
Configurable protoquil

### LINKED ISSUE #95: Fix or abandon lscheduler serialization
Previously, `logical-scheduler` objects had 'print to stream' functionality, which we used in the HTTP reply package with the intention of offloading program statistics calculation to pyQuil / the end-user. Serialization modes for the new `logical-scheduler` structure are less obvious, so it's current ly left unimplemented—with the consequence that this quilc option flag no longer works.

Either implement a serializer ( :) ) or abandon the flag ( :( ).

notmgsk:
~## Blocked by https://github.com/rigetti/rpcq/pull/85~

The changes here are many.

* **Removed many global variables.** All statistics-related variables are no longer globals. Instead they are passed around as local variables (i.e. as inputs to `quilc::process-program`).
* **Statistics dictionary removed.** Instead I pack the statistics into a new `metadata` slot for the `cl-quil:parsed-program` class. Seems appropriate that they would be carried along with their respective program, rather than a global variable.
* **Removed `#-forest-sdk` for `--verbose` and `--print-logical-schedule`.** I think these are probably useful enough that they can be exposed to SDK users.
* **Removed a bunch of `quil::` prefixes, and exported more packages from `cl-quil`.**
* **Configurable protoquil.** The server required the `-P` flag to enable protoquil restrictions. It is conceivable that the end-user could want to compile both proto- and non-protoquil programs with the same server. This PR enables that by passing a `:protoquil t` keyword to the `quil-to-native-quil` endpoint.
* **Statistics.** Statistics are calculated whenever `-P` or `:protoquil t` are supplied.

`TODO`:
  - [x] Test `process-options`.
  - [x] Check man pages / docs / etc for required updates.
  - [x] ~Make a decision about the HTTP server. This PR removes / changes stuff, but hasn't made appropriate changes to the HTTP server.~ Removed.

~Blocked by https://github.com/rigetti/rpcq/pull/80.~

Closes #95, #229


ecpeterson:
so far so good

app/src/entry-point.lisp
```diff
@@ -10,19 +10,8 @@
...
-(defparameter *compute-gate-depth* nil)
-(defparameter *compute-runtime* nil)
-(defparameter *compute-fidelity* nil)
-(defparameter *compute-matrix-reps* nil)
-(defparameter *topological-swaps* nil)
-(defparameter *compute-gate-volume* nil)
-(defparameter *compute-2Q-gate-depth* nil)
-(defparameter *compute-unused-qubits* nil)
-(defparameter *gate-whitelist* nil)
-(defparameter *gate-blacklist* nil)
```
appleby: Only ten lines in, and I love this PR already.

:fire: #226

app/src/entry-point.lisp
```diff
@@ -52,17 +39,16 @@
     (("gate-blacklist") :type string :optional t :documentation "when calculating statistics, ignore these (comma-separated) gates")
     (("gate-whitelist") :type string :optional t :documentation "when calculating statistics, consider only these (comma-separated) gates")
     (("without-pretty-printing") :type boolean :optional t :documentation "turns off pretty-printing features")
-    #-forest-sdk
     (("verbose") :type boolean :optional t :documentation "verbose compiler trace output")
+    ;; TODO decide if this is useful, if not let's nix it. It it is, then let's give SDK users the option.
```
ecpeterson: This option predates the web server (!), when users were expected to start `quilc` and feed it via / consume from its pipes. Now that the servers reply with structured data, this console output option is not worth very much.

notmgsk:
OK. This is about ready I think. There are a bunch of changes, so a careful review would be much appreciated.

app/src/package.lisp
```diff
@@ -5,5 +5,6 @@
 (in-package #:cl-user)
 
 (defpackage #:quilc
-  (:use #:cl)
+  (:use #:cl #:cl-quil)
+  (:shadow #:rewrite-arithmetic)
```
appleby: Do I understand correctly that `rpcq:dispatch-table-add-handler` will optionally let you specify the handler name if you want something other than `(symbol-name handler)`? If so, the other option is to rename the handler in this package to something like `rewrite-arithmetic-handler`, rather than shadowing here. I don't think it matter too much, but my slight preference would be to avoid shadowing, if possible.

app/tests/rpcq-tests.lisp
```diff
@@ -20,11 +21,13 @@
          (bt:destroy-thread server-thread)))))
 
 (deftest test-easy-version-call ()
+  "Test that the `get-version-info` endpoint works."
```
appleby: Minor nitpick: this markdown formatting seems inconsistent with other docstring conventions, where UPPERCASE is used to denote symbols. But part of the problem is we don't really have a consistently-applied convention, e.g. #8.

app/tests/rpcq-tests.lisp
```diff
@@ -51,7 +54,32 @@
...
+                                         :|isa| isa
+                                         :|specs| specs))
+           (server-payload (make-instance 'rpcq::|NativeQuilRequest|
+                                          :|quil| quil
+                                          :|target_device| target-device))
+           (server-response (rpcq:rpc-call client "quil-to-native-quil" server-payload))
+           (server-response-protoquil (rpcq:rpc-call client "quil-to-native-quil" server-payload :protoquil t))
+           (cpp (quil:parse-quil (rpcq::|NativeQuilResponse-quil| server-response)))
+           (cpp-protoquil (quil:parse-quil (rpcq::|NativeQuilResponse-quil| server-response-protoquil))))
+      (is (not (quil:protoquil-program-p cpp)))
```
appleby: I think the answer is "no", but I will ask anyway: is there any case where `quil-to-native-quil :protoquil nil` might coincidentally return protoquil here?  Or does does `:protoquil nil` guarantee non-protoquil output?

notmgsk:
> First round of superficial comments. Will take a closer look later.
> 
> I would feel better if `process-options` had more tests around it, but I admit those tests are tedious to write and low-value once/after this PR is merged, so I guess manual testing is ok. I was also lazy to write them when I worked on #212...

I will also feel uncomfortable until those tests exist. Going to add a todo item for that.

app/src/entry-point.lisp
notmgsk: OK. I think I'll take the bold action and remove it.

app/src/package.lisp
notmgsk: Good spot. I didn't notice that keyword argument. Changed.

app/tests/rpcq-tests.lisp
notmgsk: My bad. I've written some julia recently that uses this convention.

notmgsk: I've changed this, but used quotes instead since I'm not refering to the symbol, but the actual endpoint name registered with the RPCQ server.

app/tests/rpcq-tests.lisp
notmgsk: With these changes, it should explicitly *not* return protoquil. Even if the input satisfied the restrictions of protoquil, the compiler would append a `HALT` instruction making it non-protoquil. I've added a `:protoquil nil` to make it clearer.

app/tests/rpcq-tests.lisp
```diff
@@ -51,7 +54,32 @@
         (setf mat1 (quil::scale-out-matrix-phases mat1 mat2))
         (is (quil::matrix-equality mat1 mat2))))))
 
+(deftest test-quil-to-native-quil-protoquil-endpoint ()
+  "Test that the \"quil-to-native-quil\" will compile protoquil when given `:protoquil t`."
```
appleby: Still got backticks around the `:protoquil t` :eyes: here.

notmgsk: oof

app/src/package.lisp
appleby: LGTM

app/tests/rpcq-tests.lisp
```diff
@@ -55,7 +55,7 @@
         (is (quil::matrix-equality mat1 mat2))))))
 
 (deftest test-quil-to-native-quil-protoquil-endpoint ()
-  "Test that the \"quil-to-native-quil\" will compile protoquil when given `:protoquil t`."
+  "Test that the \"quil-to-native-quil\" will compile protoquil when given :PROTOQUIL T."
```
appleby: Final nitpick: maybe drop the "the" or else add "endpoint" after "quil-to-native-quil"?

app/src/entry-point.lisp
```diff
@@ -86,7 +68,7 @@
 (defun print-quil-list (executable-code stream)
   (loop :for instr :in executable-code :do
      (progn
-       (quil::print-instruction instr stream)
+       (print-instruction instr stream)
```
stylewarning: we can do 1 format call here:
```
(format stream "~/cl-quil:instruction-fmt/~%" instr)
```

app/src/entry-point.lisp
```diff
@@ -324,25 +305,14 @@ Version ~A is available from https://www.rigetti.com/forest~%"
...
        (*without-pretty-printing* without-pretty-printing)
-       (*print-logical-schedule* print-logical-schedule)
-       (*gate-blacklist* (and gate-blacklist
-			      (split-sequence:split-sequence #\, (remove #\Space gate-blacklist))))
-       (*gate-whitelist* (and gate-whitelist
-			      (split-sequence:split-sequence #\, (remove #\Space gate-whitelist))))
-       (*topological-swaps* show-topological-overhead)
-       (*protoquil* protoquil)
+       (gate-blacklist (and gate-blacklist
+                            (split-sequence:split-sequence #\, (remove #\Space gate-blacklist))))
```
stylewarning: `:remove-empty-subseqs t`

app/src/entry-point.lisp
```diff
@@ -324,25 +305,14 @@ Version ~A is available from https://www.rigetti.com/forest~%"
...
+       (gate-whitelist (and gate-whitelist
+                            (split-sequence:split-sequence #\, (remove #\Space gate-whitelist))))
```
stylewarning: same here

app/src/entry-point.lisp
```diff
@@ -395,150 +365,209 @@ HTTP server for good.
...
+                            (setf (gethash "final_rewiring" stats)
+                                  (read-from-string
+                                   (subseq (comment instr) length (- (length comment)
+                                                                     (/ (- (length comment) length) 2)
+                                                                     2))))))
+                         (t nil))
+                       ;; if there's a HALT instruction with a rewiring comment on
+                       ;; it, we need to migrate the comment up one instruction
+                       ;; (and perhaps merge it with any comment already present)
                      :when (and (< 0 j)
```
stylewarning: plusp

app/src/entry-point.lisp
```diff
@@ -395,150 +365,209 @@ HTTP server for good.
...
                        :collect instr)
                'vector)))
 
       ;; now that we've compiled the program, we have various things to output
       ;; one thing we're always going to want to output is the program itself.
       (print-program processed-program *quil-stream*)
 
-      (when *topological-swaps*
+      (when print-topological-swap-count
+        (setf (gethash "topological_swaps" (parsed-program-metadata processed-program))
```
stylewarning: see my note elsewhere about getters/setters for this info

app/src/entry-point.lisp
```diff
@@ -395,150 +365,209 @@ HTTP server for good.
...
                 lschedule)))
 
-      (when (and *protoquil*
-                 (or *compute-gate-depth*
-                     *compute-gate-volume*
-                     *compute-runtime*
-                     *compute-fidelity*
-                     *compute-unused-qubits*))
+      (when (and protoquil
+                 (or compute-gate-depth
```
stylewarning: can we simplify this logic? can we just compute all of them?

stylewarning: is it costly to do so? if not, it would massively simplify all of this file

app/src/printers.lisp
```diff
@@ -9,94 +9,65 @@
...
 
-(defun print-gate-depth (lschedule)
-  (let ((depth (quil::lscheduler-calculate-depth lschedule)))
-    (setf (gethash "gate_depth" *statistics-dictionary*) depth)
-    (format *human-readable-stream*
-            "# Compiled gate depth: ~d~%"
-            depth)))
+(defun print-gate-depth (depth)
+  (format *human-readable-stream*
+          "# Compiled gate depth: ~d~%"
```
stylewarning: capital D imho

app/src/printers.lisp
```diff
@@ -9,94 +9,65 @@
...
-         (unused-qubits
-           (loop :for i :below (length (quil::vnth 0 (quil::chip-specification-objects chip-specification)))
-                 :unless (quil::resources-intersect-p (quil::make-qubit-resource i)
-                                                      lscheduler-resources)
-                   :collect i)))
-    (setf (gethash "unused_qubits" *statistics-dictionary*) unused-qubits)
-    (format *human-readable-stream*
+(defun print-unused-qubits (unused-qubits)
+  (format *human-readable-stream*
             "# Unused qubit list: ~{~a~^, ~}~%"
```
stylewarning: indentation looks a lil spooked

app/src/printers.lisp
```diff
@@ -9,94 +9,65 @@
...
           "# SWAPs incurred by topological considerations: ~d~%"
           topological-swaps))
 
+(defun print-matrix-comparision (m1 m2)
+  (format *human-readable-stream* "~%#Matrix read off from input code~%")
+  (print-matrix-with-comment-hashes m1 *human-readable-stream*)
+  (format *human-readable-stream* "~%#Matrix read off from compiled code~%")
+  (print-matrix-with-comment-hashes m2 *human-readable-stream*)
+  (format *human-readable-stream* "~%")
+  (format *human-readable-stream* "#Matrices are~a equal~%" (if (quil::matrix-equals-dwim m1 m2) "" " not"))
```
stylewarning: use the cool format directive
```
QVM> (format t "~:[no~;yes~]" t)
yes
NIL
QVM> (format t "~:[no~;yes~]" nil)
no
NIL
```

app/src/printers.lisp
```diff
@@ -9,94 +9,65 @@
...
   (format *human-readable-stream*
           "# SWAPs incurred by topological considerations: ~d~%"
           topological-swaps))
 
+(defun print-matrix-comparision (m1 m2)
+  (format *human-readable-stream* "~%#Matrix read off from input code~%")
+  (print-matrix-with-comment-hashes m1 *human-readable-stream*)
+  (format *human-readable-stream* "~%#Matrix read off from compiled code~%")
+  (print-matrix-with-comment-hashes m2 *human-readable-stream*)
+  (format *human-readable-stream* "~%")
```
stylewarning: put this in with the next call to format

src/ast.lisp
```diff
@@ -1319,12 +1319,15 @@ For example,
    (memory-definitions :initarg :memory-definitions
                        :accessor parsed-program-memory-definitions)
    (executable-program :initarg :executable-code
-                       :accessor parsed-program-executable-code))
+                       :accessor parsed-program-executable-code)
+   (metadata           :initarg :metadata
```
stylewarning: barf

but I will allow it under the following conditions:

1. Its type is documented in the doc string of the object, and it describes the possible keys and values (e.g., keys are strings)
2. Every possible piece of metadata must have an explicit, exported reader function. Example, `metadata-duration (parsed-program)`, or whatever. No accessing this slot as a raw hash table

I don't want this metadata slot to fester into "put anything u want here and deal w/it".

src/ast.lisp
```diff
@@ -1319,12 +1319,15 @@ For example,
...
+                       :accessor parsed-program-metadata))
```
stylewarning: make this `%parsed-program-metadata` to disincentivise ppl

src/package.lisp
```diff
@@ -320,12 +320,15 @@
    #:parsed-program-circuit-definitions ; READER
    #:parsed-program-memory-definitions  ; READER
    #:parsed-program-executable-code     ; READER
+   #:parsed-program-metadata            ; READER
```
stylewarning: i would nix this and put the special accessors

app/src/entry-point.lisp
ecpeterson: `(format stream "~{~/cl-quil:instruction-fmt/~%~}" executable-code)`

src/ast.lisp
notmgsk: Maybe I should rethink storing this in a `parsed-program`.

notmgsk: What about a `quilc:processed-program` class?

app/src/entry-point.lisp
notmgsk: Turns out this function isn't even used anywhere. ❌

app/src/entry-point.lisp
notmgsk:
```
CL-USER> (split-sequence:split-sequence #\, (remove #\Space "H, S, T"))
("H" "S" "T")
5
CL-USER> (split-sequence:split-sequence #\, "H, S, T" :remove-empty-subseqs t)
("H" " S" " T")
```

app/src/printers.lisp
notmgsk: Likewise in the other depth printers?

app/src/printers.lisp
notmgsk: I would expect that the "true" branch came first, but w/e.

app/src/entry-point.lisp
stylewarning: `"H,,S,T"`

src/ast.lisp
stylewarning: I'm not sure the distinction is ultimately meaningful; I think it can be ok to put metadata here as long as we are careful.

app/src/entry-point.lisp
stylewarning: then delete it (as long as it's not used in tests or dependent systems)

app/src/entry-point.lisp
notmgsk: Ah. I thought you were suggesting to replace the whitespace removal.

stylewarning: A string-trim doens't hurt either

notmgsk: That is handled by the `(remove #\Space ...)`

app/src/entry-point.lisp
```diff
@@ -395,150 +361,210 @@ HTTP server for good.
...
-            (print-gate-volume lschedule))
-          (when *compute-runtime*
-            (print-program-runtime lschedule chip-specification))
-          (when *compute-fidelity*
-            (print-program-fidelity lschedule chip-specification))
-          (when *compute-unused-qubits*
-            (print-unused-qubits lschedule chip-specification))))
-
-      (when (and *protoquil* *compute-2Q-gate-depth*)
+          (when compute-gate-depth
```
appleby: You could reduce boilerplate here with something like

```list
(flet ((maybe (predicate print-function key value-thunk)
         (when predicate
           (let ((value (value-thunk)))
             (setf (gethash key metadata) value)
             (funcall print-function value)))))
  (maybe compute-gate-depth #'print-gate-depth "gate_depth"
         (lambda () (quil::lscheduler-calculate-depth lschedule)))
  (maybe compute-gate-volume #'print-gate-volume "gate_volume"
         (lambda () (quil::lscheduler-calculate-volume lschedule)))
  (maybe compute-runtime #'print-program-runtime "program_duration"
         (lambda () (quil::lscheduler-calculate-duration lschedule chip-specification)))
  ...)
```

If you don't like having to wrap the computations in a thunk, you could always `symbol-macrolet`, instead.

Or just don't bother, if we aren't worried about efficiency as @stylewarning points out above.

app/src/entry-point.lisp
```diff
@@ -395,150 +361,210 @@ HTTP server for good.
...
+              topological-swaps)
         (print-topological-swap-count topological-swaps))
 
-      (when (and *protoquil*
-                 *print-logical-schedule*)
-        (let ((lschedule (quil::make-lscheduler)))
+      (when (and protoquil
+                 print-logical-schedule)
+        (let ((lschedule (quil::make-lscheduler))
+              (metadata metadata))
```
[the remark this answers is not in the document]
appleby: The `(metadata metadata)` re-binding, I mean.

app/src/entry-point.lisp
```diff
@@ -52,17 +38,13 @@
     (("gate-blacklist") :type string :optional t :documentation "when calculating statistics, ignore these (comma-separated) gates")
     (("gate-whitelist") :type string :optional t :documentation "when calculating statistics, consider only these (comma-separated) gates")
     (("without-pretty-printing") :type boolean :optional t :documentation "turns off pretty-printing features")
-    #-forest-sdk
```
ecpeterson: 👍

app/src/entry-point.lisp
```diff
@@ -52,17 +38,13 @@
...
-    (("verbose") :type boolean :optional t :documentation "verbose compiler trace output")
-    #-forest-sdk
-    (("json-serialize" #\j) :type boolean :optional t :documentation "serialize output as a JSON object")
-    #-forest-sdk
-    (("print-logical-schedule" #\s) :type boolean :optional t :documentation "include logically parallelized schedule in JSON output; requires -P")
```
ecpeterson: cf. #95

app/src/entry-point.lisp
```diff
@@ -52,17 +38,13 @@
...
+    (("verbose") :type boolean :optional t :documentation "verbose compiler trace output") 
+    (("print-logical-schedule" #\s) :type boolean :optional t :documentation "include logically parallelized schedule in output; requires -P")
     (("isa") :type string :optional t :initial-value "8Q" :documentation "set ISA to one of \"8Q\", \"20Q\", \"16QMUX\", \"bristlecone\", \"ibmqx5\", or path to QPU description file")
     (("enable-state-prep-reductions") :type boolean :optional t :documentation "assume that the program starts in the ground state")
     (("protoquil" #\P) :type boolean :optional t :documentation "restrict input/output to ProtoQuil")
     (("help" #\h) :type boolean :optional t :documentation "print this help information and exit")
-    (("server-mode-http" #\S) :type boolean :optional t :documentation "run as a web server *and* an RPCQ server. ignores --port and uses 6000 and 5555 respectively")
+    (("server-mode-http" #\S) :type boolean :optional t :documentation "run as a web server *and* an RPCQ server. ignores --port and uses 6000 and 5555 respectively. This is deprecated and will eventually be removed.")
```
ecpeterson: I've been wondering if this is the right deprecation path. Should we keep `-S`, toss `-R`, and disable the web server on `-S`? Hm.

app/src/entry-point.lisp
```diff
@@ -395,150 +361,210 @@ HTTP server for good.
...
-
-(defun run-CLI-mode (isa-descriptor)
-  (let* ((program-text (slurp-lines))
-         (program (quil::parse-quil program-text)))
-    (process-program program isa-descriptor)))
-
-(defun process-program (program chip-specification)
+         
+         (let* ((program-text (slurp-lines))
+                (program (parse-quil program-text)))
```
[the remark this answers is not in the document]
ecpeterson: I think I would too, if they aren't being used for anything but the first argument to `process-program`.

app/src/entry-point.lisp
```diff
@@ -395,150 +361,210 @@ HTTP server for good.
...
+      (when protoquil
+        (setf (parsed-program-executable-code processed-program)
               (coerce
-               (loop :for instr :across (quil::parsed-program-executable-code processed-program)
+               (loop :for instr :across (parsed-program-executable-code processed-program)
                      :for j :from 0
                      ;; if there's a comment, consider extracting from it the final_rewiring
                      ;; entry for use in the statistics dictionary.
-                     :when (quil::comment instr)
+                     :when (comment instr) :do
```
ecpeterson: I'd prefer the `:do` on the next line, which will probably undo the big block of re-indenting that's in the diff.

app/src/entry-point.lisp
ecpeterson: I was cautious about how costly it was when I was first writing this, but compared to compilation itself it isn't costly at all, and we should just do it all the time.

ecpeterson: (We'll want to continue to consume all the associated CLI flags, issue a deprecation warning when they're present, and otherwise ignore them.)

app/src/entry-point.lisp
```diff
@@ -395,150 +361,210 @@ HTTP server for good.
...
-                     *compute-fidelity*
-                     *compute-unused-qubits*))
+      (when (and protoquil
+                 (or compute-gate-depth
+                     compute-gate-volume
+                     compute-runtime
+                     compute-fidelity
+                     compute-unused-qubits))
         ;; calculate some statistics based on logical scheduling
         (let ((lschedule (quil::make-lscheduler)))
```
ecpeterson: If you're deleting `quil:`s, can you get rid of these too? Or do they need to stay because of the `::`?

app/src/entry-point.lisp
notmgsk: I would be happy to have both `-S` and `-R` invoke the RPCQ server. Whether we should nix `-R` altogether is what I'm unsure of.

notmgsk:
BTW, @appleby, @ecpeterson, @stylewarning: This absolutely breaks the HTTP server, since I have refactored/removed stuff without making appropriate changes to the HTTP server. That was 80/20 lazy/intentional. If we are not OK with killing the HTTP server entirely with this PR, then I'll do the extra work. Speak now or forever hold your HTTP requests.

app/src/entry-point.lisp
notmgsk: The latter. If you would rather, I can bring back the `quil:`.

notmgsk: Otherwise we can export more stuff from `cl-quil`.

app/src/entry-point.lisp
notmgsk: hmph

app/src/entry-point.lisp
notmgsk: It is not costly. I did the numbers.

notmgsk: I don't know if we want to deprecate the flags. I suspect we don't want to print all the stats by default? I'm thinking either:
  * keep the individual stats flags, but have their meaning now be "print this thing"; or
  * keep and deprecate the individual flags, and introduce a `--print-statistics` flag.
Which do you prefer?

appleby: > BTW, @appleby, @ecpeterson, @stylewarning: This absolutely breaks the HTTP server, since I have refactored/removed stuff without making appropriate changes to the HTTP server. That was 80/20 lazy/intentional. If we are not OK with killing the HTTP server entirely with this PR, then I'll do the extra work. Speak now or forever hold your HTTP requests.

How long has the HTTP server been deprecated?  I defer to @stylewarning and @ecpeterson here, since I don't have a good feel for how many people/customers might still be using the HTTP interface.

ecpeterson: 5-6 months. Anyone using pyQuil is by now definitely using the RPCQ endpoint, and I couldn't guess at the fraction of users sending requests through custom software.

app/src/entry-point.lisp
stylewarning: I don't want to export internal data structures yet. I would keep `quil::` for rly internal stuff

stylewarning:
If we remove the HTTP server, we should delete it.

app/src/entry-point.lisp
```diff
@@ -37,44 +25,165 @@
                                       :maximum-priority ':info
                                       :log-writer (cl-syslog:null-log-writer)))
 
-(defparameter *statistics-dictionary* (make-hash-table :test #'equal))
-
 (defparameter *option-spec*
```
stylewarning: I think there should be these variables

```
*option-spec* : currently active and supported options
*deprecated-option-spec*: active but deprecated options
*ignored-option-spec*: inactive (ignored) and deprecated options
*retired-option-spec*: invalid options
```

When options are processed, only process `(append *option-spec* *deprecated-option-spec*)`.

app/src/entry-point.lisp
```diff
@@ -37,44 +25,165 @@
...
+    (("print-statistics"
+      :type boolean
+      :optional t
+      :documentation "print program statistics.  Requires -P."))
+
+    (("compute-matrix-reps" #\m)
+     :type boolean
+     :optional t
+     :documentation "prints matrix representations for comparison  Requires -P.  This is deprecated and will eventually be removed.  See --print-statistics.")
+    (("help" #\h)
```
stylewarning: space above this

app/src/entry-point.lisp
```diff
@@ -37,44 +25,165 @@
...
+     :initial-value nil
+     :documentation "Check for a new SDK version at launch.")
+
+    (("proxy")
+     :type string
+     :optional t
+     :initial-value nil
+     :documentation "Proxy to use when checking for an SDK update.")
+
+    ;; TODO Below are the deprecated command line options.
```
stylewarning: see note above about separate variable

app/src/entry-point.lisp
```diff
@@ -251,10 +354,10 @@ end of the name."
                           (without-pretty-printing nil)
                           (print-logical-schedule nil)
                           (verbose nil)
-                          (json-serialize nil)
```
stylewarning: Ok, so we are going to have a lot of non-functional deprecated options. I recommend making this function

```
(defun process-options (&rest all-args &key <non-ignored options> &allow-other-keys)
   ...
   (when <all-args contains *ignored-option-spec* then warn>)
```

app/src/entry-point.lisp
```diff
@@ -269,6 +372,10 @@ end of the name."
                           (quiet nil)
                           (check-sdk-version nil)
                           (proxy nil))
+  ;; Deprecated options.
+  (declare (ignore compute-gate-depth compute-gate-volume compute-runtime
```
stylewarning: this can be removed in light of the above

app/src/entry-point.lisp
```diff
@@ -351,14 +449,14 @@ Version ~A is available from https://www.rigetti.com/forest~%"
       (reload-foreign-libraries)
 
       (cond
-	;; web server mode requested.
-	;; currently provides both web and RPCQ server as a transition.
-	(server-mode-http
+        ;; web server mode requested.
+        ;; currently provides both web and RPCQ server as a transition.
+        (server-mode-http
```
stylewarning: we can prob delete, but @ecpeterson shall chime in on this matter

app/src/entry-point.lisp
```diff
@@ -351,14 +449,14 @@ Version ~A is available from https://www.rigetti.com/forest~%"
...
          ;; launch the polling loop
-	 (unless quiet
+         (unless quiet
            (show-banner))
 
-	 (unless quiet
+         (unless quiet
```
stylewarning: this will be able to be deleted

app/src/entry-point.lisp
```diff
@@ -369,176 +467,189 @@ HTTP server for good.
...
+                     (quil::lscheduler-walk-graph
+                      lschedule
+                      :bump-value (lambda (instr value)
+                                    (setf collect
+                                          (quil::resource-union collect
+                                                                (quil::instruction-resources instr)))
+                                    value))
+                     collect))
+                 (unused-qubits
+                   (loop :for i :below (length (quil::vnth 0 (quil::chip-specification-objects chip-specification)))
```
stylewarning: i think there's a helper function for (length (vnth 0 ...)) in cl-quil; check the respective file

app/src/entry-point.lisp
```diff
@@ -369,176 +467,189 @@ HTTP server for good.
...
-          (print-matrix-with-comment-hashes original-matrix *human-readable-stream*)
-          (format *human-readable-stream* "~%#Matrix read off from compiled code~%")
-          (print-matrix-with-comment-hashes same-same-but-different *human-readable-stream*)
-          (format *human-readable-stream* "~%")
-          (format *human-readable-stream* "#Matrices are~a equal~%" (if (quil::matrix-equals-dwim original-matrix same-same-but-different) "" " not"))
-          (finish-output *human-readable-stream*)))
-      (publish-json-statistics))))
+          (loop :for instr :across (parsed-program-executable-code processed-program)
+                :when (and (typep instr 'gate-application)
+                           (<= 2 (length (application-arguments instr))))
```
stylewarning: there's an `arity` func, looks like it should be exported

app/src/printers.lisp
```diff
@@ -9,94 +9,59 @@
...
-  (yason:encode *statistics-dictionary* *json-stream*)
-  *statistics-dictionary*)
-
-
 ;; custom encoder for rewiring objects
 (defmethod yason:encode ((object quil::rewiring) &optional (stream *standard-output*))
   (yason:encode (quil::rewiring-l2p object) stream))
+
+(defun print-statistics (statistics &optional (stream *human-readable-stream*))
+  (a:when-let ((depth (gethash "gate_depth" statistics)))
```
stylewarning: each of these can be massively simplified, use `~@[...~]`:

```
CL-USER> (format t "maybe: ~@[~D~]~%" 5)
maybe: 5
CL-USER> (format t "maybe: ~@[~D~]~%" nil)
maybe: 
```

`~@[...~]` only prints the contents if the thing is non-nil boom gottem

app/src/printers.lisp
```diff
@@ -9,94 +9,59 @@
...
+    (format stream
```
stylewarning: example
```
(format stream "~@[# Compiled gate depth: ~D~%~]" (gethash "gate-depth" statistics))
```

app/src/printers.lisp
```diff
@@ -9,94 +9,59 @@
...
+            unused-qubits))
+  (a:when-let ((fidelity (gethash "program_fidelity" statistics)))
+    (format stream
+            "# Estimated compiled program fidelity: ~5d~%"
+            fidelity))
+  (a:when-let ((topological-swaps (gethash "topological_swaps" statistics)))
+    (format stream
+            "# SWAPs incurred by topological considerations: ~d~%"
+            topological-swaps))
+  (a:when-let ((duration (gethash "program_duration" statistics)))
```
stylewarning: maybe keep the when-let for this one tho

app/src/entry-point.lisp
notmgsk: @ecpeterson @stylewarning @appleby friendly ping for opinion

app/src/entry-point.lisp
appleby: `(quil::chip-spec-n-qubits chip-specification)`

...saved you a grep

app/src/entry-point.lisp
appleby: My $0.02: If we are deleting the http server, then I like the idea of "stealing" `-S` for the RPCQ server. That way the option for "run in server mode" is the same for both `quilc` and `qvm`.  I would vote to deprecate (but continue to accept) `-R` for some period of time before removing it completely, though.

app/src/entry-point.lisp
notmgsk: Do you mean `*deprecated-option-spec*` in the `when`? The lib that handles command line args will be unaware of `*ignored-option-spec*` (following your advice for `(append *option-spec* *deprecated-option-spec*))` and throw an error if any of `*ignored-option-spec*` are given. Should I include that in the `append`?

app/src/entry-point.lisp
notmgsk: Just to be clear, @appleby, you're saying we should keep both `-R` and `-S` and have them do the same thing?

appleby: Yes, but only temporarily and only assuming we do decide to delete the HTTP server.  Just print a deprecation message if the user supplies `-R` saying: this will go away soon, use `-S` instead.

notmgsk: gneat

app/src/entry-point.lisp
```diff
@@ -24,7 +24,7 @@
                                       :facility ':local0
                                       :maximum-priority ':info
                                       :log-writer (cl-syslog:null-log-writer)))
-
+(caar (first *option-spec*))
```
appleby: Whoopsie?

app/src/entry-point.lisp
```diff
@@ -171,26 +173,38 @@
      :type boolean
      :optional t
      :documentation "prints approximate compiled circuit fidelity  Requires -P.  This is deprecated and will eventually be removed.  See --print-statistics.")
+    
     (("compute-2Q-gate-depth" #\2)
      :type boolean
      :optional t
-     :documentation "prints compiled circuit multiqubit gate depth; ignores white/blacklists, requires -P.  This is deprecated and will eventually be removed.  See --print-statistics.")
+     :documentation "prints compiled circuit multiqubit gate depth; ignores white/blacklists, requires -P.  This is deprecated and will eventually be removed.  See --print-statistics."
+     :deprecated-by "print-statistics")
```
appleby: Nice. Does it make sense to add this `:deprecated-by` to all the deprecated options?

app/src/entry-point.lisp
```diff
@@ -171,26 +173,38 @@
...
     (("show-topological-overhead" #\t)
      :type boolean
      :optional t
-     :documentation "prints the number of SWAPs incurred for topological reasons.  This is deprecated and will eventually be removed.  See --print-statistics.")
-    ))
+     :documentation "prints the number of SWAPs incurred for topological reasons.  This is deprecated and will eventually be removed.  See --print-statistics."))
+  "Supported and deprecated options.")
+
+(defparameter *ignored-option-spec*
+  1
```
appleby: Does `'()` not work here?

app/src/entry-point.lisp
notmgsk: This was a "backup stuff while I swap laptops" commit. Ignore :)

appleby: lol. I thought the commit message was "ignore" because `*ignored-option-spec*`

app/src/entry-point.lisp
notmgsk: Can't find it. If you can't either, should I define it?

app/src/entry-point.lisp
notmgsk: Can it? These kwargs are still not referenced anywhere

app/src/entry-point.lisp
```diff
@@ -351,194 +471,178 @@ Version ~A is available from https://www.rigetti.com/forest~%"
...
-
-(defun process-program (program chip-specification)
-  (let* ((original-matrix
-           (when (and *protoquil* *compute-matrix-reps*)
-             (quil::parsed-program-to-logical-matrix program)))
+         (let* ((program-text (slurp-lines))
+                (program (parse-quil program-text))
+                (original-matrix (when (and protoquil compute-matrix-reps)
+                                   (parsed-program-to-logical-matrix program))))
+           ;; :print-logical-schedule print-logical-schedule
```
ecpeterson: what's this doing here

app/src/entry-point.lisp
```diff
@@ -351,194 +471,178 @@ Version ~A is available from https://www.rigetti.com/forest~%"
...
+               (process-program program (lookup-isa-descriptor-for-name isa)
+                                :protoquil protoquil
+                                :gate-whitelist gate-whitelist
+                                :gate-blacklist gate-blacklist)
+             (print-program processed-program *quil-stream*)
+             (when print-statistics
+               (print-statistics statistics *quil-stream*))
+             (when (and protoquil compute-matrix-reps)
+               (let* ((processed-program-matrix (parsed-program-to-logical-matrix processed-program :compress-qubits t))
+                      (same-same-but-different (quil::scale-out-matrix-phases processed-program-matrix
```
ecpeterson: I'd encourage you to not even give this a name :P

app/src/entry-point.lisp
```diff
@@ -351,194 +471,178 @@ Version ~A is available from https://www.rigetti.com/forest~%"
...
     ;; do the compilation
     (multiple-value-bind (processed-program topological-swaps)
-        (quil::compiler-hook program chip-specification :protoquil *protoquil*)
+        (compiler-hook program chip-specification :protoquil protoquil)
 
       ;; if we're supposed to output protoQuil, we need to strip the final HALT
       ;; instructions from the output
-      (when *protoquil*
-        (setf (quil::parsed-program-executable-code processed-program)
+      (when protoquil
```
ecpeterson: I know this isn't your fault, but re-reading this gives me some heebie-jeebies to see this whole blob inlined into `process-program`. I'd feel better if it were factored out and given some human-readable name. I might open an Issue task to that effect.

app/src/rpc-server.lisp
```diff
@@ -33,42 +33,48 @@
               (2 (incf runtime coeff-twoq)))
           :finally (return runtime))))
 
+(defun statistics-to-metadata (statistics)
+  (check-type statistics hash-table)
+  (make-instance 'rpcq::|NativeQuilMetadata|
```
ecpeterson: Why not use `|NativeQuilMetadata|` everywhere, instead of this intermediate hash-table? All the fields are optional, right, so it shouldn't matter if you build it up piecewise...?

app/src/web-server.lisp
```diff
@@ -1,361 +0,0 @@
-;;;; server.lisp
```
ecpeterson: 🎉

app/src/rpc-server.lisp
notmgsk: I think I did that in one iteration of this PR, and something turned me off it. But now I don't remember what it was, so I'll give it a shot again.

notmgsk: Oh -- the slots aren't `setf`able. I suppose I can just collect all values before making the object.

notmgsk: OK I'm back to thinking that using a `|NativeQuilMetadata|` isn't as simple as it looks.

app/src/entry-point.lisp
notmgsk: OK. In the interest of closing this PR, I'll open a task for that.

notmgsk:
I would appreciate it if at least one of you could compile and build `quilc`, and then run the pyquil test suite using the newly built `quilc -R`.

app/src/rpc-server.lisp
ecpeterson: Ah: good answer. We can, of course, make them `setf`able by modifying the message spec—but that should indeed be a task for another day.

notmgsk:
OK OK. After https://github.com/rigetti/rpcq/pull/85 is merged mgsk will be happy.

BC in pyquil will be maintained. This is achieved by allowing three values for the `:protoquil` flag:
  * `nil` (`None`): defer to server
  * `:false` (`False`): do not restrict to protoquil
  * `t` (`True`): do restrict to protoquil

appleby:
> I would appreciate it if at least one of you could compile and build `quilc`, and then run the pyquil test suite using the newly built `quilc -R`.

I will give it a try

app/tests/rpcq-tests.lisp
```diff
@@ -51,7 +54,62 @@
...
+                                          :|quil| quil
+                                          :|target_device| target-device)))
+      (flet ((parse-response (protoquil)
+               (quil:parse-quil (rpcq::|NativeQuilResponse-quil| (rpcq:rpc-call client "quil-to-native-quil" server-payload :protoquil protoquil)))))
+        ;; :protoquil nil means defer to server, i.e. this should produce protoquil
+        (is (quil:protoquil-program-p (parse-response nil)))
+        ;; :protoquil ':false means no protoquil, override server's -P
+        (is (not (quil:protoquil-program-p (parse-response ':false))))
+        ;; :protoquil t means yes protoquil, regardless of what the server says
+        (is (quil:protoquil-program-p (parse-response t))))))
```
appleby: Would it make sense to also do these same tests with `*protoquil*` set to `nil` (flipping the first test to make sure it doesn't return protoquil)?

app/src/globals.lisp
```diff
@@ -0,0 +1,18 @@
+(in-package #:quilc)
+
+(defparameter *program-name* "quilc")
+(defparameter *print-logical-schedule* nil)
```
appleby: Straggler or still used?

app/src/entry-point.lisp
```diff
@@ -7,87 +7,182 @@
...
+     :type boolean
+     :optional t
+     :documentation "prints matrix representations for comparison  Requires -P.  This is deprecated and will eventually be removed.")
+    
+    (("help" #\h)
+     :type boolean
+     :optional t
+     :documentation "print this help information and exit")
+
+    (("server-mode-rpc" #\R #\S)
```
appleby: Are we planning to accept `-R` and `-S` as synonyms indefinitely, or should one be moved into `*deprecated-option-spec*`?

notmgsk: I can't make my mind up which is the better option. `-S` seems appropriate for `S`erver, but we've had that option deprecated for a while and so maybe it's weird to undeprecate it and deprecate the thing the were telling folks to use.

ecpeterson: I'm in the exact same boat.

app/tests/rpcq-tests.lisp
notmgsk: Yes. Good idea.

app/src/entry-point.lisp
appleby: I guess there is no rush. We could always leave both as valid for now, and pick one to deprecate (or not) in the future.

notmgsk: Let's put a pin in it.

doc/man/quilc.1
```diff
@@ -38,34 +34,40 @@ either 8Q, 20Q, 16QMUX, bristlecone, ibmqx5, or a path to a QPU
 description file.
 .IP "-P, --protoquil"
 Prescribe that the input and output must be ProtoQuil, which is Quil
-that is comprised of gates and measurements, with no control flow.
+that is, in order, comprised of memory declarations, gates, and measurements, with no control flow.
+.IP "--print-statistics"
```
appleby: Should this have some explanatory text with it?  Or are you meant to refer to the below deprecated options' description?

notmgsk: Always with the good spots.

appleby:
Latest round of changes LGTM

notmgsk:
This is no longer blocked, and -- unless there are any remaining objections -- mergeable.

PR #290, opened 2019-06-14
Add cleanall make target

notmgsk:


Makefile
```diff
@@ -160,6 +160,15 @@ clean:
...
+clean-quicklisp:
+	@echo "Cleaning up old projects in Quicklisp"
+	$(QUICKLISP) \
+             --eval '(ql-dist:clean (ql-dist:dist "quicklisp"))'
+
 clean-cache:
 	@echo "Deleting $(LISP_CACHE)"
 	rm -rf "$(LISP_CACHE)"
+
+cleanall: clean clean-cache clean-quicklisp
```
jmbr: Minor suggestions:

  1. I'd like to see these targets marked with `.PHONY: clean-quicklisp cleanall`

  2. I'd replace the name `cleanall` by either `clean-all` (for consistency) or by `clobber`. Since `clobber` is a typical `make` target (see https://www.gnu.org/software/make/manual/make.html#index-clobber-_0028standard-target_0029), I think it makes sense to use it here.

notmgsk: 1. Do you mean just adding the `.PHONY` above `cleanall`?
2. I was also going for consistency -- at least with QVM who uses `cleanall`. But I would prefer `clean-all`.

jmbr: Adding `.PHONY: clean-all` in a line by itself would do.

jmbr: Also, if there's already a `cleanall` target in QVM it's probably better to leave the name as is rather than than introduce an inconsistency between both packages. Alternatively, we could change it simultaneously in both.

Makefile
```diff
@@ -154,12 +154,22 @@ test-ccl:
...
+	@echo "All cleaned and reindexed."
```
appleby: I think the "reindexed" here is referring to the `ql:register-local-projects` which is done in the QVM Makefile's `clean-cache` target, but which is not done in the above `clean-cache` target for some reason.

Not sure if there is a reason quilc's `clean-cache` doesn't reindex, but probably makes sense to do it here, too, if only to reduce the diff with the QVM Makefile.

PR #291, opened 2019-06-16
Build on CCL

notmgsk:
Majority of the diff is just getting rid of warnings. There are two changes, one controversial, one not.

* not-controversial: CCL does not like `#\OCR_FORK`, so use something it does like.
* controversial: CCL does not like read-time algebraic data types. I haven't explored the finer details of this issue, but evaluating stuff like `(named-operator ...)` at run-time rather than read-time with `#.(named-operator ...)` fixes the issue.

- [ ] Seems like fiasco is [broken on CCL](https://github.com/joaotavora/fiasco/issues/28).
- [x] RPCQ gives a [very strange error](https://github.com/rigetti/rpcq/issues/72) on CCL.

Closes #291.

### DIFF
--- a/src/parser.lisp
+++ b/src/parser.lisp
@@ -100,7 +100,7 @@
    (return (tok ':DAGGER)))
   ((eager #.(string #-ccl #\HELM_SYMBOL #+ccl #\u+2388))
    (return (tok ':CONTROLLED)))
-  ((eager #.(string #\OCR_FORK))
+  ((eager #.(string #-ccl #\OCR_FORK #+ccl #\u+2442))
    (return (tok ':FORKED)))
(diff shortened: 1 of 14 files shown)


src/parser.lisp
```diff
@@ -100,7 +100,7 @@
    (return (tok ':DAGGER)))
   ((eager #.(string #-ccl #\HELM_SYMBOL #+ccl #\u+2388))
    (return (tok ':CONTROLLED)))
-  ((eager #.(string #\OCR_FORK))
+  ((eager #.(string #-ccl #\OCR_FORK #+ccl #\u+2442))
```
PuercoPop: :thought_balloon: SBCL (and ECL) both understand the `#\u+2442` so it may be better to use that notation for all implementations instead to avoid using feature expressions.

appleby: Counter-point: the `#\OCR_FORK` (and `#\HELM_SYMBOL` / `#\DAGGER`, above) are more self-documenting, even though it's fairly clear from context based on what `tok` gets returned.

stylewarning:
I'm fine evaluating `named-operator`

stylewarning:
Let's close since so much has changed.

PR #295, opened 2019-06-21
Provide estimated runtime in compiled protoquil metadata

notmgsk:
~Blocked by #288.~

The main changes here are:

* `quil-to-native-quil` constructs a hash-table directly for the response metadata, rather than a `rpcq::|NativeQuilMetadata|` object. This allows us to extend the metadata we return without having to change RPCQ's message definitions.
* The `program_duration` value is converted to a float. Previously it was a rational and would cause an error because we have no en/decoding routines for that type. This is an open issue in RPCQ: https://github.com/rigetti/rpcq/issues/23.
* Compute an estimation for the compiled native quil program. An *estimate*.
* Implement `protoquil-program-p` and `not-protoquil` condition.


app/src/rpc-server.lisp
```diff
@@ -15,6 +15,29 @@
...
+  (when (and (typep parsed-protoquil-program 'quil:parsed-program)
+             (quil:protoquil-program-p parsed-protoquil-program))
+    ;; These opaque numbers come from an analysis of the runtimes of a
+    ;; large number of randomly generated programs targeting a 16Q
+    ;; lattice. Those programs were a random mixture of 1- and 2Q
+    ;; gates, followed by MEASUREs on all 16 qubits.
+    (let ((coeff-oneq 1.6291)
+          (coeff-twoq 1.701)
+          (coeff-const 181.89)
+          (num-oneq (count-if (lambda (instr)
```
ecpeterson: I’d prefer this as a `(loop … :count … :into …)`. You could do both binds and the final computation all at once.

app/src/rpc-server.lisp
```diff
@@ -27,16 +50,22 @@
...
-                                  :|gate_volume| (gethash "gate_volume" dictionary)
-                                  :|multiqubit_gate_depth| (gethash "multiqubit_gate_depth" dictionary)
-                                  :|program_duration| (ensure-optional-real (gethash "program_duration" dictionary))
-                                  :|program_fidelity| (ensure-optional-real (gethash "program_fidelity" dictionary))
-                                  :|topological_swaps| (gethash "topological_swaps" dictionary))))
+         (processed-program (gethash "processed_program" dictionary))
+         ;; TODO If "qpu_runtime_estimation" can be folded into
+         ;; "program_duration" then we can replace this whole thing
+         ;; with DICTIONARY.
+         (metadata (a:alist-hash-table
```
ecpeterson: Did we really mean to change the type of this?

src/compilation-methods.lisp
```diff
@@ -90,6 +90,22 @@
 
 ;;; Core public-facing routine for a full compilation pass.
 
+(define-condition not-protoquil (simple-error)
```
ecpeterson: nice

app/src/rpc-server.lisp
ecpeterson: It’s easy to imagine pyQuil not complaining about this, btw. It’s not particularly fussy about types, and you’re providing the same dictionary-type accessors it uses as sugar on the real object.

notmgsk: I opted to do this because (1) I didn't want to overwrite the already-present "program_duration" key, and (2) I didn't want to have to extend the definition of `rpcq::|NativeQuilMetadata|`. If you think either (and which) of those would be a better choice, LMK.

ecpeterson: Let's extend the structure definition in the RPCQ message spec. There's an `:optional` qualifier that you can use to maintain backwards compatibility; it'll take just a few seconds.

src/compilation-methods.lisp
notmgsk: 😎

app/src/rpc-server.lisp
```diff
@@ -15,6 +15,24 @@
...
+    ;; large number of randomly generated programs targeting a 16Q
+    ;; lattice. Those programs were a random mixture of 1- and 2Q
+    ;; gates, followed by MEASUREs on all 16 qubits.
+    (loop :with coeff-oneq := 1.6291
+          :with coeff-twoq := 1.701
+          :with runtime := 181.89
+          :for instr :across (quil:parsed-program-executable-code parsed-protoquil-program)
+          :when (typep instr 'quil:application) :do
+            (let ((nargs (length (quil:application-arguments instr))))
+              (when (= nargs 1) (incf runtime coeff-oneq))
```
stylewarning:
```
(case (length ...)
  (1 (incf ...))
  (2 ...))
```

app/src/rpc-server.lisp
```diff
@@ -15,6 +15,24 @@
...
 
+(defun runtime-estimation (parsed-protoquil-program)
+  "Estimated QPU runtime of PARSED-PROTOQUIL-PROGRAM. Likely to be an over-estimate for small depth programs, where runtime is dominated by network latency and compilation, etc. Take these results with a grain of salt."
+  (when (and (typep parsed-protoquil-program 'quil:parsed-program)
+             (quil:protoquil-program-p parsed-protoquil-program))
+    ;; These opaque numbers come from an analysis of the runtimes of a
+    ;; large number of randomly generated programs targeting a 16Q
+    ;; lattice. Those programs were a random mixture of 1- and 2Q
+    ;; gates, followed by MEASUREs on all 16 qubits.
+    (loop :with coeff-oneq := 1.6291
```
stylewarning: nice magic numbers

app/src/rpc-server.lisp
```diff
@@ -27,16 +45,21 @@
...
                                   :|gate_depth| (gethash "gate_depth" dictionary)
                                   :|gate_volume| (gethash "gate_volume" dictionary)
                                   :|multiqubit_gate_depth| (gethash "multiqubit_gate_depth" dictionary)
                                   :|program_duration| (ensure-optional-real (gethash "program_duration" dictionary))
                                   :|program_fidelity| (ensure-optional-real (gethash "program_fidelity" dictionary))
-                                  :|topological_swaps| (gethash "topological_swaps" dictionary))))
+                                  :|topological_swaps| (gethash "topological_swaps" dictionary)
+                                  :|qpu_runtime_estimation| (runtime-estimation
+                                                             (quil:parse-quil processed-program)))))
+    (print (runtime-estimation (quil:parse-quil processed-program)))
```
stylewarning: delete?

app/src/rpc-server.lisp
```diff
@@ -155,6 +175,7 @@
 
 (declaim (special *program-name*))
 (defun start-rpc-server (&key
+                           (protocol "tcp")
```
stylewarning: can we check that `protocol` is from a valid set?

app/src/rpc-server.lisp
notmgsk: unavoidable. I note their source above

stylewarning: i know

app/src/rpc-server.lisp
notmgsk: Maybe. Does a valid set exist?

app/src/rpc-server.lisp
notmgsk: i know you know

app/src/rpc-server.lisp
```diff
@@ -15,6 +15,24 @@
...
+    ;; These opaque numbers come from an analysis of the runtimes of a
+    ;; large number of randomly generated programs targeting a 16Q
+    ;; lattice. Those programs were a random mixture of 1- and 2Q
+    ;; gates, followed by MEASUREs on all 16 qubits.
+    (loop :with coeff-oneq := 1.6291
+          :with coeff-twoq := 1.701
+          :with runtime := 181.89
+          :for instr :across (quil:parsed-program-executable-code parsed-protoquil-program)
+          :when (typep instr 'quil:application) :do
+            (case (nargs (length (quil:application-arguments instr)))
```
stylewarning: not right

tests/compiler-hook-tests.lisp
```diff
@@ -424,10 +424,10 @@ PRAGMA END_PRESERVE_BLOCK
 MEASURE 1
 ")))
     ;; TODO Maybe there should be a (define-condition invalid-protoquil ...) ?
```
stylewarning: invalid old news

src/compilation-methods.lisp
```diff
@@ -96,14 +96,18 @@
...
+               (format s "Misplaced or illegal instruction in ProtoQuil program:~%  ")
+               (cond ((zerop j)
+                      (write-string "(BEGINNING OF PROGRAM)" s))
+                     (t
+                      (print-instruction (aref p (1- j)) s)))
+               (format s "~%>>>")
+               (print-instruction (aref p j) s)
+               (format s "~%  ")
+               (cond ((= (length p) (1+ j))
+                      (write-string "END OF PROGRAM"))
```
stylewarning: write to s

stylewarning: also parentheses

PR #301, opened 2019-06-24
fixing tests to adhere to previous PR for fix/clifford_ket

### LINKED ISSUE #299: pauli-matrix-p and clifford ordering have changed
Since merging #174 at least two tests are failing.

1. The ordering of Cliffords has changed in `serialize-clifford-sequence`. This breaks `test-serialize-clifford-sequence`.
2. Something has changed in `pauli-matrix-p` and this has broken `test-pauli-matrix-p`.

AShiTheCoder:
Eliminated errors from `phase-to-string` and `string-to-phase` so that the test suite could continue and reject invalid pauli matrices on its own, and changed the qubit ordering convention in `test-clifford-from-quil` to be consistent with the clifford/paulis' new computational basis argument convention.

Fixes #299

### DIFF
--- a/src/clifford/benchmarking-procedures.lisp
+++ b/src/clifford/benchmarking-procedures.lisp
@@ -130,7 +130,7 @@
         ((quil::double~ phase #C(0 1)) "i")
         ((quil::double~ phase #C(0 -1)) "-i")
         ;; This phase is invalid, but don't error so the test suite can continue.
-        (t nil)))
+        (t (error "Invalid phase number: ~A~%" phase))))
 
 (defun string-to-phase (phase-str)
   "Returns the phase corresponding to a string form of a fourth root of unity, given by PHASE-STR, NIL otherwise."
@@ -139,7 +139,7 @@
         ((string= phase-str "i") #C(0 1))
         ((string= phase-str "-i") #C(0 -1))
         ;; This phase string is invalid, but don't error so the test suite can continue.
-        (t nil)))
+        (t (error "Invalid phase string: ~A~%" phase-str))))
 
 (defun pauli-matrix-p (p)
   "Returns two strings. The first is a string representation of the phase of P, assuming P is a Pauli matrix, which will be a fourth root of unity. The second return value is a string representation of P as a tensor product of single qubit Pauli operators. If P is not a Pauli operator, then the second value will be NIL."
@@ -159,7 +159,7 @@
       ((not (valid-pauli-dim m n)) NIL)
       ((= m n 2)
        (if (not (null pauli))
-           (values (phase-to-string phase) pauli)
+           (handler-bind ((error #'(lambda (x) (declare (ignore x)) nil))) (values (phase-to-string phase) pauli))
            NIL))
       ((or (factor-I p) (factor-Z p))
        (multiple-value-bind (coeff next-pauli)
--- a/src/clifford/benchmarking-procedures.lisp
+++ b/src/clifford/benchmarking-procedures.lisp
@@ -129,7 +129,8 @@
         ((quil::double~ phase -1) "-")
         ((quil::double~ phase #C(0 1)) "i")
         ((quil::double~ phase #C(0 -1)) "-i")
-        (t (error "Invalid phase number: ~A~%" phase))))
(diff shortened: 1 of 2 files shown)


tests/benchmarking-procedures-tests.lisp
```diff
@@ -143,9 +143,9 @@
 	(XZ (cl-quil.clifford:pauli-from-string "XZ"))
 	(-YY (cl-quil.clifford:pauli-from-string "-YY"))
 	(XX (cl-quil.clifford:pauli-from-string "XX"))
-	(CNOT01H0-quil (format nil "~{~a~%~}" (list "CNOT 0 1" "H 0")))
+	(CNOT01H0-quil (format nil "~{~a~%~}" (list "CNOT 1 0" "H 0")))
 	(CNOT01H0 (cl-quil.clifford:clifford-from-quil CNOT01H0-quil))
-	(H0CNOT01-quil (format nil "~{~a~%~}" (list "H 0" "CNOT 0 1")))
+	(H0CNOT01-quil (format nil "~{~a~%~}" (list "H 0" "CNOT 1 0")))
```
notmgsk: nit-pick: do you think the variable name should be changed also

AShiTheCoder: fine

src/clifford/benchmarking-procedures.lisp
```diff
@@ -129,15 +129,17 @@
         ((quil::double~ phase -1) "-")
         ((quil::double~ phase #C(0 1)) "i")
         ((quil::double~ phase #C(0 -1)) "-i")
-        (t (error "Invalid phase number: ~A~%" phase))))
```
notmgsk: @stylewarning is this bad form? it seems like we're being weird here just to keep the tests happy, but it should be the other way round maybe

src/clifford/benchmarking-procedures.lisp
```diff
@@ -129,15 +129,17 @@
...
+        (t nil)))
 
 (defun string-to-phase (phase-str)
   "Returns the phase corresponding to a string form of a fourth root of unity, given by PHASE-STR, NIL otherwise."
   (cond ((string= phase-str "") 1)
         ((string= phase-str "-") -1)
         ((string= phase-str "i") #C(0 1))
         ((string= phase-str "-i") #C(0 -1))
-        (t (error "Invalid phase string: ~A~%" phase-str))))
+        ;; This phase string is invalid, but don't error so the test suite can continue.
```
notmgsk: likewise @stylewarning

notmgsk:
I see now that `serialize-clifford` and its ilk are not referenced anywhere, and so this should not be a BC-breaking change, and so I'm less concerned now.

src/clifford/benchmarking-procedures.lisp
stylewarning: I don't think this is good.

src/clifford/benchmarking-procedures.lisp
stylewarning: No bueno.

src/clifford/benchmarking-procedures.lisp
```diff
@@ -159,7 +159,7 @@
       ((not (valid-pauli-dim m n)) NIL)
       ((= m n 2)
        (if (not (null pauli))
-           (values (phase-to-string phase) pauli)
+           (handler-bind ((error #'(lambda (x) (declare (ignore x)) nil))) (values (phase-to-string phase) pauli))
```
stylewarning: use handler-case

src/clifford/benchmarking-procedures.lisp
```diff
@@ -157,7 +157,7 @@
       ((not (valid-pauli-dim m n)) NIL)
       ((= m n 2)
        (if (not (null pauli))
-           (values (phase-to-string phase) pauli)
+           (handler-case (values (phase-to-string phase) pauli) (error () nil))
```
notmgsk: This also feels drastic. But the body of `phase-to-string` is sufficiently simple that it shouldn't be confusing.

PR #304, opened 2019-06-25
clang++-7 -> g++

notmgsk:
`g++` seems like a more reasonable default


jmbr:
Just for the record, the name `c++` is standard among *nix systems to invoke the system's C++ compiler (likewise for `cc` and the C compiler).

appleby:
> Just for the record, the name `c++` is standard among *nix systems to invoke the system's C++ compiler (likewise for `cc` and the C compiler).

Good point.

PR #305, opened 2019-06-26
Build libtweedledum with CMake

jmbr:
This changes the way quilc is built so that invoking `make tweedledum` will build the dynamic library using a suitable compiler on your system. You only need to do this *once* when hacking on the quilc code.

Alternatively, running

```make LIBSBCL=/path/to/sbcl/src/runtime/libsbcl.a tweedledum```

will additionally create a new binary called `src/contrib/tweedledum/rigetti-sbcl` that incorporates tweedledum statically. This binary could be used to compile quilc without the need to place the shared tweedledum in any particular directory.

Notes:

1. I have only tested this in Mac OS X so far but I expect it to work on Linux as well.
2. Perhaps one remaining task would be to invoke `cmake` from cl-quil.asd for people who solely want to use Quicklisp to build quilc.

### CODE REVIEW

**Comment by jmbr** [comment]
Thoughts? @appleby @notmgsk @stylewarning

File: src/contrib/tweedledum/tweedledum.lisp [hunk 1]
```diff
@@ -81,3 +80,21 @@ GIVE-UP-COMPILATION if INSTR is not a permutation gate."
...
   (uiop:symbol-call ':cl-quil-tests
                     '#:run-cl-quil-tests))
+
+(defun tweedledum-present-p ()
+  "Determine if tweedledum is callable."
+  ;; On a non-SBCL system it will fall back to the default behavior for dynamic libraries.
+  (let ((present-p t))
+    #+sbcl
+    (handler-case (synthesis-dbs '(0))
+      (sb-kernel::undefined-alien-function-error (condition)
```
appleby: I feel like `sb-alien` *ought* to provide some way to check this without having to try calling the function and letting it fail, but if that's not the case, so be it.

Would it work to instead check for `sb-alien:undefined-alien-error` here (which is exported)?

src/contrib/tweedledum/CMakeLists.txt
```diff
@@ -0,0 +1,78 @@
...
+  POSITION_INDEPENDENT_CODE True)
+target_compile_definitions(tweedledum-object PRIVATE FMT_HEADER_ONLY)
+target_compile_features(tweedledum-object PRIVATE cxx_std_17)
+target_include_directories(tweedledum-object
+  PRIVATE ${TWEEDLEDUM_DIR}/tweedledum/include
+  PRIVATE ${TWEEDLEDUM_DIR}/tweedledum/libs/fmt
+  PRIVATE ${TWEEDLEDUM_DIR}/tweedledum/libs/easy
+  PRIVATE ${TWEEDLEDUM_DIR}/tweedledum/libs/glucose
+  PRIVATE ${TWEEDLEDUM_DIR}/tweedledum/libs/kitty
+  PRIVATE ${TWEEDLEDUM_DIR}/tweedledum/include)
```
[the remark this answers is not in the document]
jmbr: It is intentional in this case but thank you for pointing it out.

src/contrib/tweedledum/CMakeLists.txt
```diff
@@ -0,0 +1,78 @@
...
+  add_executable(rigetti-sbcl empty.cpp)
+  target_link_options(rigetti-sbcl
+    PRIVATE
+    # The following linker flags come from sbcl/src/runtime/sbcl.mk:
+    # LINKER:-macosx_version_min,10.6
+    # LINKER:-arch,x86_64
+    # LINKER:-dynamic
+    # LINKER:-twolevel_namespace
+    # LINKER:-bind_at_load
+    LINKER:-pagezero_size,0x100000
```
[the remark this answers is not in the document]
jmbr: I know what the ld manual page says about it: it is going to create a page without read access at address zero with the specified size. If a pointer to NULL is dereferenced, then this page will be accessed and the program will immediately crash with a memory violation.

In more practical terms, if you omit this option you'll get a sbcl binary that jumps right into LDB complaining that the image has been corrupted.

src/contrib/tweedledum/tweedledum.lisp
```diff
@@ -81,3 +80,21 @@ GIVE-UP-COMPILATION if INSTR is not a permutation gate."
...
+        (declare (ignore condition))
+        (setf present-p nil)))
+    present-p))
+
+;; TODO Some error handling here
+(unless *tweedledum-libs-loaded*
+  (unless (tweedledum-present-p)
```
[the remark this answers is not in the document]
jmbr: Yes, definitely.

src/contrib/tweedledum/tweedledum.lisp
jmbr: Will check this out. Thanks.

src/contrib/tweedledum/CMakeLists.txt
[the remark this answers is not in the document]
appleby: > If you omit this pagezero_size option, you'll get an sbcl binary that jumps right into LDB complaining that the image has been corrupted.

Maybe just copy/paste the above portion (or whatever else you want to add) into a comment in the source file?  Some future rigetti-lisp-hacker from the year 2131 :space_invader:  will now doubt thank us.

src/contrib/tweedledum/CMakeLists.txt
```diff
@@ -0,0 +1,83 @@
...
+
+  add_library(tweedledum-static STATIC $<TARGET_OBJECTS:tweedledum-object>)
+  set_target_properties(tweedledum-static PROPERTIES OUTPUT_NAME ${TWEEDLEDUM_NAME})
+  set(LIBTWEEDLEDUM_STATIC_NAME
+    "${CMAKE_STATIC_LIBRARY_PREFIX}${TWEEDLEDUM_NAME}${CMAKE_STATIC_LIBRARY_SUFFIX}")
+
+  file(WRITE empty.cpp "// This file exists to placate CMake's need \
+for a source file being associated to a target.
+")
+  add_executable(rigetti-sbcl empty.cpp)
```
appleby: Maybe also add `rigetti-sbcl` and the cmake `build` directory to `.gitignore`

jmbr: Yes, and empty.cpp as well.

jmbr: On a second thought, I think only empty.cpp belongs in .gitignore. The rest of the files you mention will vanish after `make clean-tweedledum`.

appleby: The point of `.gitignore` is "I do not want to version control these files", which I think still applies to any build artifacts, regardless of what `make clean` deletes. I don't want to have to run `make clean-tweedledum` to make them disappear from my (ma)git status output :-).

In fact, `empty.cpp` should probably also be removed by `make clean-tweedledum`, since it's also a build artifact.

appleby:
> Other pie-in-the-sky wishlist items:
> 
> * Doesn't need to happen in this PR, but I think we should just pick one of GNU make vs cmake and use it everywhere...

Having now had to time to relax with a nice cup of tea, my stance on this has softened. I still think it would better if we picked one of GNU make vs CMake and used it everywhere, but I guess there are more important problems to solve.

jmbr:
I would also like to have an all-encompassing build system. However, that's not how CMake works. You can think of CMake as a (slicker) combination of GNU autoconf and automake: it handles automatic configuration (that is, stuff like figuring out the location and flags of the compiler and linker or finding out the extension of libraries on your system) and then produces a build file in a variety of formats such as GNU make files, Ninja files, and even Visual Studio. My point is that CMake is not a full replacement for Make, they do different things.

Quicklisp is trickier. CMake creates a so-called compilation database that you may have seen in build/compile_options.json. This contains the exact command (and arguments) used to generate object files. It is mostly straight-forward to grab the content of that file and run the same command from the ASD file. The compilation database unfortunately does not include the commands necessary for *linking* the resulting objects. I spent some time figuring out the way to introspect from within the CMakeLists.txt file and produce a config.lisp file with the exact commands and flags but only got half-way there (and I'm not even sure that it's possible to get all the way there).

appleby:
> I would also like to have an all-encompassing build system. However, that's not how CMake works. You can think of CMake as a (slicker) combination of GNU autoconf and automake: it handles automatic configuration (that is, stuff like figuring out the location and flags of the compiler and linker or finding out the extension of libraries on your system) and then produces a build file in a variety of formats such as GNU make files, Ninja files, and even Visual Studio. My point is that CMake is not a full replacement for Make, they do different things.

I get what you're saying, but don't lots of projects (admittedly I don't know of any primarily *common lisp* projects, so maybe it doesn't make sense here) have their build configuration specified entirely in CMake and then just let CMake generate the other build files? I guess for any non-trivial build maybe you end up having to maintain a Makefile.am (or whatever the CMake equivalent is), but my wistful hope here (and this is where I definitely don't understand CMake well enough to know if it would make a difference) is that the result would be tighter integration between the top-level build and the tweedledum build. For instance, the toplevel makefile currently treats `tweedledum` as a `.PHONY` target and just hands dependency tracking off to the cmake build, which it doesn't know much about. This also maybe ties in to my other vague desire to have the `rigetti-sbcl` binary wind up in the toplevel directory for visibility purposes, which I assume (for no good reason) might be easier to pull of if the whole build could be handled by a single top-level CMake.

On the other hand, I don't want to push *too* hard for converting everything to CMake, because then I will actually have to read the CMake documentation :-)

> Quicklisp is trickier. CMake creates a so-called compilation database that you may have seen in build/compile_options.json. This contains the exact command (and arguments) used to generate object files. It is mostly straight-forward to grab the content of that file and run the same command from the ASD file. The compilation database unfortunately does not include the commands necessary for _linking_ the resulting objects. I spent some time figuring out the way to introspect from within the CMakeLists.txt file and produce a config.lisp file with the exact commands and flags but only got half-way there (and I'm not even sure that it's possible to get all the way there).

I definitely prefer the approach you're describing here, but if it can't be made to work, would it be acceptable to just have the ASDF build shell out to `cmake` to do the heavy lifting? This obviously means an ASDF-build-time dependency on cmake, which is unfortunate, but if the only alternative is to require an out-of-band manual `cmake`, maybe it's not so bad.

jmbr:
Oh, I thought you meant something else. I wholeheartedly agree with what you said above and I'm 100% on-board with having CMake orchestrate everything. The reason why I didn't do it that way this time around was because I wanted to submit a PR that introduced a gradual change that would be more palatable. I'm glad we concur.

Yeah, launching CMake from ASDF is totally doable too. My misgiving about it is that it would be ugly but if others are OK with it, it's fine.

appleby:
> Oh, I thought you meant something else. I wholeheartedly agree with what you said above and I'm 100% on-board with having CMake orchestrate everything. The reason why I didn't do it that way this time around was because I wanted to submit a PR that introduced a gradual change that would be more palatable. I'm glad we concur.

I guess if I'm being really honest, my desire to standardize on one or the other of make vs cmake stems from two OCD impulses that ultimately don't really matter:

1. The desire to have a "single" build system that's used everywhere, even if that "single" system is cmake and winds up generating Makefiles. This is not really important, but for some reason soothes my soul.

2. The desire to iron out the minor differences between the various Makefiles for quilc, qvm, and rpcq. This is unrelated and low priority, but somehow rewriting everything in CMake seemed like a good excuse to tackle this at the same time :-).

The tighter integration between the toplevel build and tweedledum are also real (minor) concerns, but if I'm being honest, the above two points are probably what's really motivating me here.

I should add though that moving to CMake everywhere is definitely a change that would require buy-in from everyone. I am OK having to learn a new build system, but others might feel differently (or maybe I am the only caveman that doesn't already know how to drive CMake!).

Also, I agree with your decision to limit the changes in this PR to just introducing CMake for the tweedledum build.

> Yeah, launching CMake from ASDF is totally doable too. My misgiving about it is that it would be ugly but if others are OK with it, it's fine.

It's definitely ugly, but if the only other option is "no working ASDF build", I prefer "ugly" :-).

appleby:
> > Yeah, launching CMake from ASDF is totally doable too. My misgiving about it is that it would be ugly but if others are OK with it, it's fine.
> 
> It's definitely ugly, but if the only other option is "no working ASDF build", I prefer "ugly" :-).

I guess the other option is to just have the ASDF continue to work as it did before. Is that a workable option?

jmbr:
> > > Yeah, launching CMake from ASDF is totally doable too. My misgiving about it is that it would be ugly but if others are OK with it, it's fine.
> > 
> > 
> > It's definitely ugly, but if the only other option is "no working ASDF build", I prefer "ugly" :-).
> 
> I guess the other option is to just have the ASDF continue to work as it did before. Is that a workable option?

Yes, of course. Although it violates the Don't-Repeat-Yourself principle, which means that it could potentially become a source of errors down the line (e.g., if subtle flag changes in the CMake side that does not get propagated to the ASDF side and this results in some mismatch and a hard to debug problem).

jmbr:
> I guess if I'm being really honest, my desire to standardize on one or the other of make vs cmake stems from two OCD impulses that ultimately don't really matter:
> 
> 1. The desire to have a "single" build system that's used everywhere, even if that "single" system is cmake and winds up generating Makefiles. This is not really important, but for some reason soothes my soul.
> 2. The desire to iron out the minor differences between the various Makefiles for quilc, qvm, and rpcq. This is unrelated and low priority, but somehow rewriting everything in CMake seemed like a good excuse to tackle this at the same time :-).
> 
> The tighter integration between the toplevel build and tweedledum are also real (minor) concerns, but if I'm being honest, the above two points are probably what's really motivating me here.

I could not possibly agree more with you.
 
> I should add though that moving to CMake everywhere is definitely a change that would require buy-in from everyone. I am OK having to learn a new build system, but others might feel differently (or maybe I am the only caveman that doesn't already know how to drive CMake!).
> 
> Also, I agree with your decision to limit the changes in this PR to just introducing CMake for the tweedledum build.
> 
> > Yeah, launching CMake from ASDF is totally doable too. My misgiving about it is that it would be ugly but if others are OK with it, it's fine.
> 
> It's definitely ugly, but if the only other option is "no working ASDF build", I prefer "ugly" :-).

OK but see my other answer below.

appleby:
> > > > Yeah, launching CMake from ASDF is totally doable too. My misgiving about it is that it would be ugly but if others are OK with it, it's fine.
> > > 
> > > 
> > > It's definitely ugly, but if the only other option is "no working ASDF build", I prefer "ugly" :-).
> > 
> > 
> > I guess the other option is to just have the ASDF continue to work as it did before. Is that a workable option?
> 
> Yes, of course. Although it violates the Don't-Repeat-Yourself principle, which means that it could potentially become a source of errors down the line (e.g., if subtle flag changes in the CMake side that does not get propagated to the ASDF side and this results in some mismatch and a hard to debug problem).

This is a very good point.  I don't know enough about what's required for quicklisp packaging to know which would be preferable from quicklisp's point of view: having an external cmake dependency or having the ASDF build do something slightly different.

I'm also throwing half-baked suggestions over the fence because I'm still on vacation and have nothing better to do than harass you :-).  I'll be less annoying in code review starting monday once I have my own PRs to work on again.

Basically, I trust you to do whatever you think seems like the sanest option here.

notmgsk:
Is this still WIP or is it ready for The Big Time? I'm coming around to the idea of incorporating tweedledum directly into the binary.

jmbr:
> Is this still WIP or is it ready for The Big Time? I'm coming around to the idea of incorporating tweedledum directly into the binary.

The changes @appleby are easy to incorporate so, for the purpose of the review, I think you can consider it ready for the big time.

.gitignore
```diff
@@ -10,3 +10,8 @@ doc/*.aux
 doc/*.log
 doc/*.pdf
 doc/*.toc
+empty.cpp
+src/contrib/tweedledum/empty.cpp
+src/contrib/tweedledum/build
+src/contrib/tweedledum/rigetti-sbcl
+src/contrib/tweedledum/libtweedledum.{so,dylib}
```
notmgsk: For completeness, add `src/contrib/tweedledum.{so,dylib}`

src/contrib/tweedledum/tweedledum.lisp
notmgsk: @jmbr did you resolve this?

src/contrib/tweedledum/tweedledum.lisp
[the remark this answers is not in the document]
notmgsk: @jmbr likewise can we resolve this comment?

src/contrib/tweedledum/tweedledum.lisp
```diff
@@ -82,20 +87,16 @@ GIVE-UP-COMPILATION if INSTR is not a permutation gate."
   (uiop:symbol-call ':cl-quil-tests
                     '#:run-cl-quil-tests))
 
+#+sbcl
 (defun tweedledum-present-p ()
   "Determine if tweedledum is callable."
   ;; On a non-SBCL system it will fall back to the default behavior for dynamic libraries.
```
notmgsk: Is comment still true?

src/contrib/tweedledum/tweedledum.lisp
```diff
@@ -82,20 +87,16 @@ GIVE-UP-COMPILATION if INSTR is not a permutation gate."
...
-        (declare (ignore condition))
-        (setf present-p nil)))
-    present-p))
+  (handler-case (synthesis-dbs '(0))
+    (sb-kernel::undefined-alien-function-error (condition)
+      (declare (ignore condition))
+      (return-from tweedledum-present-p nil)))
+  t)
+#-sbcl
+(error "Don't know what to do on non-SBCL systems.")
```
notmgsk: Why are we restricting loading tweedledum to SBCL? I don't recall trying to work with tweedledum on other implementations, but I don't see why it wouldn't work.

I suppose what I am saying is: I don't want this to be SBCL-specific.

src/contrib/tweedledum/tweedledum.lisp
```diff
@@ -82,20 +87,16 @@ GIVE-UP-COMPILATION if INSTR is not a permutation gate."
...
   "Determine if tweedledum is callable."
   ;; On a non-SBCL system it will fall back to the default behavior for dynamic libraries.
-  (let ((present-p t))
-    #+sbcl
-    (handler-case (synthesis-dbs '(0))
-      (sb-kernel::undefined-alien-function-error (condition)
-        (declare (ignore condition))
-        (setf present-p nil)))
-    present-p))
+  (handler-case (synthesis-dbs '(0))
```
notmgsk: This seems like a complicated solution to what I assume was also solved by managing `*tweedledum-libs-loaded*`. What's the benefit of this change?

src/contrib/tweedledum/tweedledum.lisp
```diff
@@ -9,7 +9,7 @@
 (in-package #:cl-quil.tweedledum)
 
 (cffi:define-foreign-library
-    (libtweedledum :search-path "/usr/local/lib/rigetti/")
+    (libtweedledum :search-path "/usr/local/lib/")
```
notmgsk: Why is this path changed?

notmgsk:
Rebased onto master.

jmbr:
We seem to have moved away from libtweedledum. Is this PR still relevant/desirable?

notmgsk:
I hope to revive it at some point, but we can close this for the time being.

PR #307, opened 2019-06-28
Fixing various leftover qubit ordering issues in quilc.

AShiTheCoder:
Previously, I changed how cliffords and paulis were printed internally, as well as the order in which they took arguments; for example, IIX became an X on qubit 0 instead of qubit 2. However, there were many internal inconsistencies left over, which I've fixed in this PR:

- `embed` for both cliffords and paulis weren't taking arguments in the correct computational basis ordering (i.e. `(embed '(X Y) 3 '(0 1))` yielded `IXY` instead of `IYX`, the latter of which is what you should get if you matched the pauli symbol list and qubit argument list element-wise)
- `tensor-mul` for paulis was changed to be consistent with ordering changes
- Tests were changed to reflect these new ordering conventions
- `clifford-from-quil`, which was not working correctly before, is now working as expected
- RPCQ and web server calls to `embed` and `pauli-from-string` now pass in qubits in the correct computational basis order

I'm still a little bit nervous about making this many changes, so please review them carefully/criticize them appropriately if necessary.


src/clifford/benchmarking-procedures.lisp
```diff
@@ -162,11 +162,11 @@
       ((or (factor-I p) (factor-Z p))
        (multiple-value-bind (coeff next-pauli)
            (pauli-matrix-p (magicl::slice p 0 (/ m 2) 0 (/ n 2)))
-         (values coeff (concatenate-or-nil next-pauli pauli))))
+         (values coeff (concatenate-or-nil pauli next-pauli))))
       ((or (factor-Y p) (factor-X p))
        (multiple-value-bind (coeff next-pauli) (pauli-matrix-p
                                                   (magicl::slice p (/ m 2) m 0 (/ n 2)))
-         (values (phase-to-string (* (string-to-phase coeff) (if (factor-Y p) #C(0 -1) 1))) (concatenate-or-nil next-pauli pauli))))
+         (values (phase-to-string (* (string-to-phase coeff) (if (factor-Y p) #C(0 -1) 1))) (concatenate-or-nil pauli next-pauli))))
```
notmgsk: Newlines will make this more readable.

ecpeterson:
Closes #309

PR #308, opened 2019-06-28
Configure docker env var for CXX

notmgsk:
The new default `g++` is not available in the docker builds, or at least it is too old. The docker builds do have `clang++-7` however, so configure them to use that.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,6 +2,8 @@ FROM rigetti/rpcq
 
 ARG build_target
 
+ENV CXX="clang++-7"
+
 # install build dependencies
 COPY Makefile /src/quilc/Makefile
 WORKDIR /src/quilc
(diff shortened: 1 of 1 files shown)


notmgsk:
> This LGTM. Does `c++` on these docker images point to the out-of-date `g++` or the snazzy new `clang++-7`? If the latter, it might make sense to take @jmbr's recommendation from the other recent PR into account and just specifiy `c++` here.

I will investigate.

> Also, would it suffice to pass `CXX` as an argument when calling `make` to build the quilc executable, or does this need to be defined at runtime as well?

Just build time, AFAICT. Changed.

jmbr:
It might also be useful to give https://github.com/rigetti/quilc/pull/305 a look because eventually the results of that PR could impact this (e.g., if we go ahead and migrate the whole build system to CMake).

Issue #284, opened 2019-06-12
TEST-MEASURE-SEMANTICS: Expression (EVERY #'DOUBLE= (AMPLITUDES QVM) (AMPLITUDES QVM-COMP)) evaluated to false


appleby:
TEST-MEASURE-SEMANTICS fails for me when running `make test`. This is testing on master and after a `make clean clean-cache`.

```
TEST-MEASURE-SEMANTICS
Unhandled FIASCO::FAILED-ASSERTION in thread #<SB-THREAD:THREAD "main thread" RUNNING
                                                {10005205B3}>:
  Test assertion failed when running TEST-MEASURE-SEMANTICS:

Expression (EVERY #'DOUBLE= (AMPLITUDES QVM)
                  (AMPLITUDES QVM-COMP)) evaluated to false
0: #'DOUBLE= => #<FUNCTION CL-QUIL::DOUBLE=>
1: (AMPLITUDES QVM) => #(#C(1.0000000000000004d0 0.0d0) #C(0.0d0 0.0d0)
                         #C(0.0d0 0.0d0) #C(0.0d0 0.0d0) #C(0.0d0 0.0d0)
                         #C(0.0d0 0.0d0) #C(0.0d0 0.0d0) #C(0.0d0 0.0d0)
                         #C(0.0d0 0.0d0) #C(0.0d0 0.0d0) #C(0.0d0 0.0d0)
                         #C(0.0d0 0.0d0) #C(0.0d0 0.0d0) #C(0.0d0 0.0d0)
                         #C(0.0d0 0.0d0) #C(0.0d0 0.0d0) #C(0.0d0 0.0d0)
                         #C(0.0d0 0.0d0) #C(0.0d0 0.0d0) #C(0.0d0 0.0d0)
                         #C(0.0d0 0.0d0) #C(0.0d0 0.0d0) #C(0.0d0 0.0d0)
                         #C(0.0d0 0.0d0) #C(0.0d0 0.0d0) #C(0.0d0 0.0d0)
                         #C(0.0d0 0.0d0) #C(0.0d0 0.0d0) #C(0.0d0 0.0d0)
                         #C(0.0d0 0.0d0) #C(0.0d0 0.0d0) #C(0.0d0 0.0d0)
                         #C(0.0d0 0.0d0) #C(0.0d0 0.0d0) #C(0.0d0 0.0d0)
                         #C(0.0d0 0.0d0) #C(0.0d0 0.0d0) #C(0.0d0 0.0d0)
[...]

PR #311, opened 2019-07-01
Bump version to 1.9.1

notmgsk:
This fixes

  * boondoggle package build failure
  * fix clifford qubit ordering issues


notmgsk:
> Build failure looks like the same out-of-date QVM issue I encountered in #284.

Yep, this is an on-going problem that ought to fixed soon (@karalekas his this on his plate). The actual deploy stage does not depend on tests passing, however. Docker images will be built when a "vX.Y.Z" tag is pushed.

Issue #313, opened 2019-07-02
PRINT-INSTRUCTION-GENERIC fails to print gate applications with complex-valued parameters


appleby:
Minimal reproduction case:

```
echo "RX(i) 0" | ./quilc
! ! ! Error: There is no applicable method for the generic function
               #<STANDARD-GENERIC-FUNCTION CL-QUIL::PRINT-INSTRUCTION-GENERIC (28)>
             when called with arguments
               (#C(0.0d0 1.0d0) #<SB-IMPL::STRING-OUTPUT-STREAM {100501CED3}>).
See also:
  The ANSI Standard, Section 7.6.6
```

The following files from the parser tests also tickle the bug:

```
tests/good-test-files/good-complex-complex-number.quil
tests/good-test-files/good-complex-params.quil
tests/good-test-files/good-simple-params.quil
```

For example:

```
$ ./quilc < tests/good-test-files/good-simple-params.quil
! ! ! Error: There is no applicable method for the generic function
               #<STANDARD-GENERIC-FUNCTION CL-QUIL::PRINT-INSTRUCTION-GENERIC (28)>
             when called with arguments
               (#C(-0.5d0 1.0d0) #<SB-IMPL::STRING-OUTPUT-STREAM {10054CC323}>).
See also:
  The ANSI Standard, Section 7.6.6
```

Note that these Quil programs can be parsed and printed just fine, as long as they don't pass through `compiler-hook`.

```
CL-USER> (quil::print-parsed-program (quil:parse-quil "RX(i) 0"))
RX(1.0i) 0
NIL
CL-USER> (quil::print-parsed-program
	  (quil:compiler-hook (quil:parse-quil "RX(i) 0")
			      (quil::build-nq-linear-chip 2)))
; Evaluation aborted on #<SB-PCL::NO-APPLICABLE-METHOD-ERROR {100212D953}>.

There is no applicable method for the generic function
  #<STANDARD-GENERIC-FUNCTION CL-QUIL::PRINT-INSTRUCTION-GENERIC (28)>
when called with arguments
  (#C(0.0d0 1.0d0)
   #<SB-IMPL::STRING-OUTPUT-STREAM {100212D163}>).
   [Condition of type SB-PCL::NO-APPLICABLE-METHOD-ERROR]
See also:
  Common Lisp Hyperspec, 7.6.6 [:section]

Restarts:
 0: [RETRY] Retry calling the generic function.
[...]

PR #316, opened 2019-07-03
Fix printing of parameters in MATRIX-GATE-DEFINITION entries

### LINKED ISSUE #272: PRINT-INSTRUCTION-GENERIC fails to print MATRIX-GATE-DEFINITIONs whose entries contain bare parameters
The following Quil program from `tests/good-test-files/good-defgate-with-param.quil` fails to print.

```
DEFGATE BASIC(%alpha):
    1, 0
    0, %alpha

BASIC(1.0) 1
```

Here is the backtrace from the quilc binary produced by the `quilc-unsafe` make target.

```
./quilc < tests/good-test-files/good-defgate-with-param.quil
Unhandled SB-KERNEL:CASE-FAILURE in thread #<SB-THREAD:THREAD "main thread" RUNNING
                                              {10006185B3}>:
  #:|alpha| fell through ETYPECASE expression. Wanted one of (NUMBER LIST).

0: (SB-DEBUG::DEBUGGER-DISABLED-HOOK #<SB-KERNEL:CASE-FAILURE expected-type: (OR NUMBER LIST) datum: #:|alpha|> #<unused argument> :QUIT T)
1: (SB-DEBUG::RUN-HOOK *INVOKE-DEBUGGER-HOOK* #<SB-KERNEL:CASE-FAILURE expected-type: (OR NUMBER LIST) datum: #:|alpha|>)
2: (INVOKE-DEBUGGER #<SB-KERNEL:CASE-FAILURE expected-type: (OR NUMBER LIST) datum: #:|alpha|>)
3: (ERROR SB-KERNEL:CASE-FAILURE :NAME ETYPECASE :DATUM #:|alpha| :EXPECTED-TYPE (OR NUMBER LIST) :POSSIBILITIES (NUMBER LIST))
4: ((:METHOD CL-QUIL::PRINT-INSTRUCTION-GENERIC (CL-QUIL::MATRIX-GATE-DEFINITION STREAM)) #<CL-QUIL:PARAMETERIZED-GATE-DEFINITION {1004D5B103}> #<SB-IMPL::STRING-OUTPUT-STREAM {1006090D63}>) [fast-method]
5: (CL-QUIL::PRINT-PARSED-PROGRAM #<CL-QUIL:PARSED-PROGRAM {100607D353}> #<SB-IMPL::STRING-OUTPUT-STREAM {1006090D63}>)
6: (QUILC::PRINT-PROGRAM #<CL-QUIL:PARSED-PROGRAM {100607D353}> #<SYNONYM-STREAM :SYMBOL SB-SYS:*STDOUT* {1000005EA3}>)
7: (QUILC::PROCESS-PROGRAM #<CL-QUIL:PARSED-PROGRAM {1004E00223}> #<CL-QUIL::CHIP-SPECIFICATION of 8:8 objects>)
8: (QUILC::PROCESS-OPTIONS :PREFER-GATE-LADDERS NIL :COMPUTE-GATE-DEPTH NIL :COMPUTE-GATE-VOLUME NIL :COMPUTE-RUNTIME NIL :COMPUTE-FIDELITY NIL :COMPUTE-MATRIX-REPS NIL :COMPUTE-2Q-GATE-DEPTH NIL :COMPUTE-UNUSED-QUBITS NIL :SHOW-TOPOLOGICAL-OVERHEAD NIL :GATE-BLACKLIST NIL :GATE-WHITELIST NIL :WITHOUT-PRETTY-PRINTING NIL :PRINT-LOGICAL-SCHEDULE NIL :VERBOSE NIL :JSON-SERIALIZE NIL :ISA "8Q" :ENABLE-STATE-PREP-REDUCTIONS NIL :PROTOQUIL NIL :VERSION NIL :CHECK-LIBRARIES NIL :BENCHMARK NIL :SERVER-MODE-HTTP NIL :SERVER-MODE-RPC NIL :HOST "*" :PORT 5555 :TIME-LIMIT 0 :HELP NIL :LOG-LEVEL "info" :QUIET NIL :CHECK-SDK-VERSION NIL :PROXY NIL)
9: (QUILC::%ENTRY-POINT NIL)
10: ((LAMBDA NIL :IN MAKE-TOPLEVEL-FUNCTION))
11: ((FLET SB-UNIX::BODY :IN SAVE-LISP-AND-DIE))
12: ((FLET "WITHOUT-INTERRUPTS-BODY-14" :IN SAVE-LISP-AND-DIE))
13: ((LABELS SB-IMPL::RESTART-LISP :IN SAVE-LISP-AND-DIE))

```

The failure occurs in `print-instruction-generic` when attempting to print the `matrix-gate-definition`, which expects all `gate-definition-entries` to be of type `number` or `list`, and chokes on the bare `#:|alpha|` symbol.

```lisp
(:method ((gate matrix-gate-definition) (stream stream))
    (let ((gate-size (isqrt (length (gate-definition-entries gate)))))
      (format stream "DEFGATE ~a~@[(~{%~a~^, ~})~]:~%"
              (gate-definition-name gate)
              (if (typep gate 'static-gate-definition)
                  nil
                  (gate-definition-parameters gate)))
      (dotimes (i gate-size)
        (format stream "    ~{~a~^, ~}~%"
                (mapcar (lambda (z)
                          (with-output-to-string (s)
                            (etypecase z
                              (number
                               (format-complex z s))
                              (list
                               (print-instruction (make-delayed-expression nil nil z) s)))))
                        (subseq (gate-definition-entries gate)
                                (* i gate-size)
                                (* (1+ i) gate-size)))))
      (terpri stream)))
```

Another example that fails similarly can be found in `tests/good-test-files/good-general-defgate.quil`:

```
DEFGATE GEN-GATE(%a,%b,%c,%d):
    %a, %b
    %c, %d

GEN-GATE(0,1,2,3) 0
```

```
./quilc < tests/good-test-files/good-general-defgate.quil
! ! ! Error: #:|a| fell through ETYPECASE expression.
             Wanted one of (NUMBER LIST).
```

### LINKED ISSUE #273: Maybe: Plunder golden inputs for TEST-PRINT-PARSED-PROGRAM-GOLDEN-FILES
[As discussed](https://github.com/rigetti/quilc/pull/257#discussion_r288610497) in #257, it might be useful to have a function to steal golden inputs from other Quil test files for use in `test-print-parsed-program-golden-files`.

Two potential sources of pilfered golden inputs are `tests/good-test-files/` and `tests/compiler-hook-test-files/`. Of the two, `good-test-files` appears to have the juicier inputs. Possibly there are other sources that could be plundered as well.

It's debatable whether this is a good idea or overkill, and probably depends on how often new test cases are added to the above directories. Note, however, that plundering golden inputs has already yielded one potential bug (#272).

## Additional considerations

Any such function would probably need to maintain a blacklist of input files that should be skipped.

One reason to skip input files is that their printed form contains floats with lots of digits in the printed representation, which may or may not be portable and in any case seems sketchy to depend on the default formatting for `~F`.

Also consider that INCLUDES would need to have their paths re-written since the newly produced golden file will not live in the same directory as the source file that is being plundered.

For example, the file `tests/good-test-files/good-include.quil` contains the following Quil program:

```
# include a file

INCLUDE "bell.quil"

BELL
```

In order to work with the existing `test-print-parsed-program-golden-files` test, the INCLUDE path would need to be re-written to `INCLUDE "good-test-files/bell.quil"`, and also the test would need to be run with `*allow-unresolved-applications*` bound to `t` (I think).

Alternatively, any file containing an INCLUDE could be added to the blacklist.

## POC bash script

Here is a proof-of-concept bash script. If we decide to implement the gold-plundering mechanism described in this issue, then the script would be converted to a lisp function and moved into `tests/printer-tests.lisp` alongside the function `update-print-parsed-program-golden-files`. Presumably, the cutesy pirate references would also be toned down!

This script is written with the assumption that it lives in the directory `tests/printer-test-files/`.

```bash
#!/usr/bin/bash

BURIED_TREASURE="../good-test-files/*.quil ../compiler-hook-test-files/*.quil"
TREASURE_CHEST=plundered-gold.quil

for gold in $BURIED_TREASURE; do
    echo "# Input"
    echo "# Plundered from $gold"
    cat "$gold"
    echo
    echo "# Output"
    echo
done >> "$TREASURE_CHEST"

echo "Arr. All the gold has been plundered from $BURIED_TREASURE."
echo "Don't forget to run CL-QUIL-TESTS::UPDATE-PRINT-PARSED-PROGRAM-GOLDEN-FILES in your lisp REPL."
```

appleby:
**Fix printing of parameters in MATRIX-GATE-DEFINITION entries**

If the entry is a symbol, it must be parameter. Pass it along to the PRINT-INSTRUCTION method specialized on DELAYED-EXPRESSION, like we already do for LISTs that represent Quil functions.

Fixes #272

**Plunder golden tests from good-test-files and compiler-hook-test-files**

Do a one-time import of golden test cases for TEST-PRINT-PARSED-PROGRAM-GOLDEN-FILES from tests/{good,compiler-hook}-test-files.

Some of these plundered tests require regex support in order to verify the printed output. Those tests have been sequestered in their own gold-regex subdirectory, and the all the existing tests plus new plundered tests that do not require regex support have been moved in to a gold-standard subdir. The reason for separating the tests like this is to avoid hoop-jumping when automatically updating golden file output sections in UPDATE-PRINT-PARSED-PROGRAM-GOLDEN-FILES, which now only updates the gold-standard files since I don't see an easy way to automate updating the output sections of golden files that require arbitrary regexes.

Closes #273

**Remove unnecessary let-binding from PARSE-ARITHMETIC-TOKENS**
**Move CL-PPCRE dependency out of CL-QUIL and into CL-QUIL-TESTS**


cl-quil.asd
```diff
@@ -29,7 +29,6 @@
                #:trivial-garbage        ; weak hash tables
                #:cl-permutation
                #:queues.priority-queue
-               #:cl-ppcre               ; CHP file parsing
```
appleby: As far as I can tell, `cl-ppcre` is no longer used directly by `cl-quil`.  Easy to revert if I'm missing something.

src/ast.lisp
```diff
@@ -1293,7 +1293,7 @@ For example,
                             (etypecase z
                               (number
                                (format-complex z s))
-                              (list
+                              ((or list symbol)
                                (print-instruction (make-delayed-expression nil nil z) s)))))
```
appleby: Not sure if this is the *right* fix, but it is *a* fix.  Open to other suggestions.

Note that the `print-instruction-generic` method that is specialized on `delayed-expression` already knows how to print all three of these cases, so we could just punt everything to `print-instruction-generic`.  I didn't do that because, currently, `print-instruction-generic` on a `delayed-expression` adds (gratuitous imho) parentheses around numeric literals, so for example:

```
DEFGATE FOO(%a):
    %a*2, 0
    0, 0
```

Prints as

```
DEFGATE FOO(%a):
    (%a*(2.0)), 0.0
    0.0, 0.0
```

If we removed those extra parens around the `(2.0)` from `print-instruction-generic`, then all three types in this `etypecase` could be handled by a single `(format nil "~/instr-fmt or whatever it's called/" (make-delayed-expression ...))`

src/parser.lisp
```diff
@@ -953,8 +953,7 @@ INPUT-STRING that triggered the condition."
 "
   (labels ((simplify (entry)
              (parse-arithmetic-tokens entry :eval t)))
-    (let ((simple-entries (mapcar #'simplify entries)))
-      simple-entries)))
+    (mapcar #'simplify entries)))
```
appleby: Totally unrealted, but couldn't help meself.

tests/printer-test-files/gold-regex/plundered-regex-gold.quil
```diff
@@ -0,0 +1,457 @@
...
+PRAGMA BLOCK
+CNOT 0 2
+RZ\(0\.5\) 2
+CNOT 0 2
+PRAGMA END_BLOCK
+PRAGMA END_COMMUTING_BLOCKS
+CZ 1 2
+
+# Input
+# Plundered from ../../compiler-hook-test-files/johannes-CPHASE.quil
```
appleby: Some of these test cases (from `compiler-hook-test-files` especially) are of dubious value as printer tests. On the other hand, they are fairly cheap and you never know when printing complex numbers might break on you (see #313).

If anyone objects to any of these, just say the word and I'll remove them.

tests/printer-test-files/gold-standard/plundered-gold.quil
```diff
@@ -0,0 +1,1228 @@
...
+RESET
+HALT
+
+# Input
+# Disable fixed-point check
+#
+# The fixed-point check is disabled for this test because something in
+# the PARSE-QUIL -> PRINT-PARSED-PROGRAM cycle causes sequences of
+# CONTROLLED DAGGER or DAGGER CONTROLLED to permute, which makes
+# fixed-point check break. For example:
```
appleby: As far as I understand it, the flip-flopping of CONTROLLED and DAGGER modifiers described here does not affect correctness, but I am not 100% certain, so let me know if that is not the case.

The re-ordering is slightly confusing though, so I was planning to file a follow-on issue.  If it's not too onerous, I'll fix it so that parse -> print is idempotent w.r.t. these modifiers, at which I'll point re-enable the fixed-point check here.

ecpeterson: You're right, this doesn't affect semantic correctness. You're also right that CL-QUIL shouldn't do this, or it should at least pick a preferred ordering and stick with it.

appleby: #317

src/ast.lisp
stylewarning: what about printing negative numbers? does that make it gratuitous?

cl-quil.asd
stylewarning: it was used in cl-quil.clifford but looks like it's gone now

src/ast.lisp
appleby: Oof. Owie. Ouch. Good point. I guess it would be a problem for the `^` operator.

tests/printer-tests.lisp
```diff
@@ -7,7 +7,16 @@
 (defparameter *printer-test-files-directory*
   (asdf:system-relative-pathname
    ':cl-quil-tests
-   "tests/printer-test-files/"))
+   "tests/printer-test-files/")
+  "Path to directory containing printer test files. Note that unlike most other test file directories used in this test-suite, all the actual input files are contained in subdirectories of this one. See also *PRINTER-TEST-FILES-GOLD-STANDARD-DIRECTORY* and *PRINTER-TEST-FILES-GOLD-REGEX-DIRECTORY*.")
+
+(defparameter *printer-test-files-gold-standard-directory*
+  (merge-pathnames #P"gold-standard/" *printer-test-files-directory*)
+  "Path to directory containing golden files who's expected output sections can be compared with STRING=.")
```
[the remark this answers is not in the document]
appleby: Grammar is not my strong suit.  I typed "whose", doubted myself, looked it up on google, came to the conclusion it should be "who's" and changed it...

appleby:
Latest commit now with more guder speling

tests/printer-test-files/gold-standard/plundered-gold.quil
appleby: Reversing of gate modifiers is "fixed" in the latest commit, but I wasn't sure if this was a parser bug or a printer bug, so I just picked one to get the code review conversation going.

Prior to this change, gate modifiers were applied left-to-right, like so:

```
CL-USER> (let ((pp (quil::parse-quil "CONTROLLED DAGGER FORKED Y 0 1 2")))
	   (quil::print-parsed-program pp)
	   (quil::application-operator
	    (aref (quil::parsed-program-executable-code pp) 0)))
FORKED DAGGER CONTROLLED Y 0 1 2
#.(CL-QUIL::FORKED-OPERATOR
  #.(CL-QUIL:DAGGER-OPERATOR
    #.(CL-QUIL:CONTROLLED-OPERATOR
      #.(CL-QUIL:NAMED-OPERATOR "Y"))))
```

After this change, operators are applied right-to-left, like so:

```
CL-USER> (let ((pp (quil::parse-quil "CONTROLLED DAGGER FORKED Y 0 1 2")))
	   (quil::print-parsed-program pp)
	   (quil::application-operator
	    (aref (quil::parsed-program-executable-code pp) 0)))
CONTROLLED DAGGER FORKED Y 0 1 2
#.(CL-QUIL:CONTROLLED-OPERATOR
  #.(CL-QUIL:DAGGER-OPERATOR
    #.(CL-QUIL::FORKED-OPERATOR
	  #.(CL-QUIL:NAMED-OPERATOR "Y"))))
```

But maybe left-to-right is the correct order, and I should just change the printer to reverse the order when printing.

@notmgsk @stylewarning @ecpeterson
QUILMASTERS PLEASE ADVISE!

src/parser.lisp
```diff
@@ -566,7 +566,7 @@ INPUT-STRING that triggered the condition."
             :finally (progn
                        (setf (application-operator app)
                              (apply-modifiers-to-operator
-                              processed-modifiers
+                              (reverse processed-modifiers)
```
appleby: If the consing here is hateful, I could always modify the loop to push modifiers onto the head of the list instead of `:collect`ing, or `nreverse`, or turn the recursion in `apply-modifiers-to-operator` inside-out, but that would destroy tail-recursion and make everyone sad.

stylewarning: not worried, also waters says reverse is OK

src/ast.lisp
stylewarning: my bones

tests/printer-test-files/gold-standard/plundered-gold.quil
notmgsk: I would have to check to be sure, but I think this is no bueno. In your first example, qubit 0 is consumed by the `FORKED` and qubit 1 by the `CONTROLLED`. In your second example, q0 is consumed by the `CONTROLLED` and q1 by the `FORKED`. This would require changes to how modified instructions are applied to qubits.

notmgsk: In the case of `DAGGER CONTROLLED ...` it doesn't matter so much because `DAGGER` does not consume one of the application arguments.

src/parser.lisp
appleby: Good enough for Waters, good enough for me.

tests/printer-test-files/gold-standard/plundered-gold.quil
ecpeterson: I agree with Mark.

appleby: Yeah, I was wondering whether qubit order mattered, or whether something in the way the operators are built sorts it out (or if it just didn't matter). Admittedly, I didn't even know FORKED was a thing until today, so I was mostly thinking about combinations of CONTROLLED and DAGGER...

appleby: For what it's worth, this change didn't fail any tests in either quilc or qvm, but maybe there are just no testcases that cover it.

stylewarning:
looks like I merged too early; let's fix whatever needs to be fixed.

appleby:
> looks like I merged too early; let's fix whatever needs to be fixed.

Should I just open another PR reverting the last commit and have discussion about DAGGER/CONTROLLED/FORKED modifiers continue there instead?

tests/printer-test-files/gold-standard/plundered-gold.quil
notmgsk: @applyby
```
(setf a (parsed-program-to-logical-matrix (parse-quil "CONTROLLED FORKED Y 0 1 2")))
(setf b (parsed-program-to-logical-matrix (parse-quil "FORKED CONTROLLED Y 0 1 2")))
(matrix-equality a b)
NIL
```

notmgsk: And for completeness:
```
(setf c (parsed-program-to-logical-matrix (parse-quil "CONTROLLED FORKED Y 1 0 2")))
(matrix-equality b c)
T
```

notmgsk:
@appleby yes, do

tests/printer-test-files/gold-standard/plundered-gold.quil
appleby: Doh. Guess I will add such a test to the test suite :-).

To be fair, I did exactly that with a bunch of combinations of DAGGER and CONTROLLED trying to find a case where order mattered. I also tried testing combinations with FORKED once I "discovered" it, but at that point had switched to testing by piping stuff to QVM, and I guess QVM doesn't support FORKED yet?

appleby:
> @appleby yes, do

#319

PR #318, opened 2019-07-08
Add support for pretty-printing small integer multiples of pi

### LINKED ISSUE #314: pi/2 is pretty printed but not 2*pi
Is there some reason to not pretty-print integer multiples of pi?  If not, fix it.

This originally came up in a code review here https://github.com/rigetti/quilc/pull/257#discussion_r288606203, where it was suggested that `*print-fractional-radians*` would be a good place to start investigating this.

appleby:
Add support for pretty-printing `n * pi` for `-10 <= n <= 10`. The number 10 was chosen arbitrarily and is based on the existing idea of fractional radian printing where we just check a handful of "sane" fractions. If we want to pretty-print `74628294764 * pi`, I guess I could switch to checking to see if the remainder of `(floor r pi)` is zero (or ~pi, apparently; see my comments in #314).

This sequence of commits is a choose-your-own-adventure game. Each successive commit represents a further refactoring on the original implementation of `FORMAT-REAL`, with my own personal favorite appearing as the final commit. If you feel I've "jumped the shark" at any point, it's easy to revert to a previous version. Otherwise, I will squash them before merging.

Note that the final "reasonably rational" implementation is a tad slower (despite consing less) than the original `numer` / `denom` nested-loops implementation (on SBCL) due to SBCL trying hard to convert the `RATIO`s to `FLOAT`s without rounding error in `SB-KERNEL:FLOAT-RATIO` (according to `SB-SPROF`). If we are concerned about the performance penalty, most of it can be clawed back by passing a `:key` argument to `FIND` in `REASONABLY-RATIONAL` that just does `(/ (float (numerator x)) (denominator x))` to do the `RATIO -> FLOAT` conversion. Since we are only testing via `double~`, the rounding error that might be introduced by the division doesn't appear to matter (it does if I change the test to `double=` though). Likewise, a declare/declaim inline for `REASONABLY-RATIONAL` might help.

Take these benchmarks with a grain of salt. Run on my laptop with who-knows-what running in the background.

```
CL-USER> (defparameter *test-floats* 
                       (concatenate 'vector (copy-seq quil::**reasonable-rationals**) 
                                            (loop :repeat 1000 :collect (random 1000000.0d0))))
*TEST-FLOATS*

;; nested numer / denom loop implementation
CL-USER> (let ((s (make-broadcast-stream)))
	   (gc :full t)
	   (time (loop :repeat 100 :do 
                   (loop :for r :across *test-floats* :do 
                     (quil::format-real-prev r s)))))
Evaluation took:
  1.319 seconds of real time
  0.967264 seconds of total run time (0.890748 user, 0.076516 system)
  [ Run times consist of 0.036 seconds GC time, and 0.932 seconds non-GC time. ]
  73.31% CPU
  3,332,709,211 processor cycles
  720,223,792 bytes consed

;; "reasonably rational" implementation (final commit)
CL-USER> (let ((s (make-broadcast-stream)))
	   (gc :full t)
	   (time (loop :repeat 100 :do 
                   (loop :for r :across *test-floats* :do 
                     (quil::format-real r s)))))
Evaluation took:
  1.624 seconds of real time
  1.206542 seconds of total run time (1.149760 user, 0.056782 system)
  [ Run times consist of 0.022 seconds GC time, and 1.185 seconds non-GC time. ]
  74.32% CPU
  4,103,977,161 processor cycles
  430,687,456 bytes consed

;; "reasonably rational" + float division instead of SB-KERNEL:FLOAT-RATIO
CL-USER> (let ((s (make-broadcast-stream)))
	   (gc :full t)
	   (time (loop :repeat 100 :do 
                   (loop :for r :across *test-floats* :do 
                     (quil::format-real-float-div r s)))))
Evaluation took:
  0.897 seconds of real time
  0.662924 seconds of total run time (0.615438 user, 0.047486 system)
  [ Run times consist of 0.013 seconds GC time, and 0.650 seconds non-GC time. ]
  73.91% CPU
  2,266,545,527 processor cycles
  239,539,504 bytes consed
```

Fixes #314


tests/printer-tests.lisp
```diff
@@ -128,3 +128,250 @@ TEST(0.5) 0 1
...
+                        (1234.1234 . "1234.1234")
+                        (2e10 . "20000000000.0"))))
+    (%check-format #'string= "~/quil::real-fmt/" simple-cases :print-fractional-radians t)
+    (%check-format #'string= "~/quil::real-fmt/" simple-cases :print-fractional-radians nil))
+
+  ;; Test *PRINT-FRACTIONAL-RADIANS* = NIL.
+  (%check-format #'uiop:string-prefix-p
+                 "~/quil::real-fmt/"
+                 `((,(- pi) . "-3.1415926")
+                   (,pi . "3.141592653")
```
appleby: Whoops. Meant to truncate these to 4 decimal places.

src/ast.lisp
```diff
@@ -1046,34 +1046,40 @@ N.B., The fractions of pi will be printed up to a certain precision!")
...
+
+(defun reasonably-rational (r &key (test #'double~))
+  "Return the first RATIONAL in **REASONABLE-RATIONALS** that is DOUBLE~ to R.
+
+If no such RATIONAL exists, return NIL.
+
+R can be any NUMBER, but probably REAL.
+
+When R is a FLOAT, you can think of REASONABLY-RATIONAL as an imprecise version of CL:RATIONALIZE that is limited to checking RATIONALs in **REASONABLE-RATIONALS**."
+  (find r **reasonable-rationals** :test test))
```
appleby: The optimization I mentioned in the PR description would be to add a `:key (lambda (x) (/ (float (numerator x)) (denominator x)))` here, in which case the `:test` argument should probably be removed  to require `double~`.

notmgsk:
Nothing to add just yet, other than the pun in here is wonderful:
> Add support for pretty-printing n * pi for -10 <= n <= 10. The number 10 was chosen arbitrarily and is based on the existing idea of fractional radian printing where we just check a handful of "sane" fractions. If we want to pretty-print 74628294764 * pi, I guess I could switch to checking to see if the remainder of (floor r pi) is zero (or ~pi, apparently; see my comments in #314).

appleby:
> Nothing to add just yet, other than the pun in here is wonderful:
> 
> > Add support for pretty-printing n * pi for -10 <= n <= 10. The number 10 was chosen arbitrarily and is based on the existing idea of fractional radian printing where we just check a handful of "sane" fractions. If we want to pretty-print 74628294764 * pi, I guess I could switch to checking to see if the remainder of (floor r pi) is zero (or ~pi, apparently; see my comments in #314).

As much as I would love to take credit, if there is a pun in the quoted paragraph, it was unintentional (and I am oblivious) :(

notmgsk:
The issue #314 seems appropriately numbered.

appleby:
> The issue #314 seems appropriately numbered.

`o m g`

notmgsk:
We may have peaked.

appleby:
> We may have peaked.

Can't file any more pi-related issues till we hit issue # 3141

tests/printer-tests.lisp
appleby: Done.

tests/printer-tests.lisp
```diff
@@ -128,3 +128,250 @@ TEST(0.5) 0 1
...
+  (let ((simple-cases '((0 . "0.0")
+                        (0.0 . "0.0")
+                        (0.0d0 . "0.0")
+                        (#C(0 0) . "0.0")
+                        (#C(0.0 0.0) . "0.0")
+                        (#C(0.0d0 0.0d0) . "0.0")
+                        (42.42 . "42.42")
+                        (42.42d0 . "42.42")
+                        (#C(4.2 0) . "4.2"))))
+    (%check-format #'string= "~/quil::complex-fmt/" simple-cases :print-polar-form t)
```
stylewarning: is complex-fmt exported? i think so; prefer `:`

tests/printer-tests.lisp
```diff
@@ -128,3 +128,250 @@ TEST(0.5) 0 1
...
+                           ;; within +/- +DOUBLE-COMPARISON-THRESHOLD-LOOSE+.
+                           (destructuring-bind (input . expected-output) testcase
+                             (list `(,(+ (- input quil::+double-comparison-threshold-loose+)
+                                         double-float-epsilon)
+                                     . ,expected-output)
+                                   `(,input . ,expected-output)
+                                   `(,(- (+ input quil::+double-comparison-threshold-loose+)
+                                         double-float-negative-epsilon)
+                                     . ,expected-output))))
+                         `((,(* -10 quil::pi) . "-10*pi")
```
stylewarning: oof, can't you generate these?

src/ast.lisp
```diff
@@ -1046,34 +1046,40 @@ N.B., The fractions of pi will be printed up to a certain precision!")
 (defvar *print-polar-form* nil
   "When true, FORMAT-COMPLEX prints out complex numbers in polar form with syntax AMPLITUDE∠PHASE.")
 
+(global-vars:define-global-parameter **reasonable-rationals**
+    (nconc (a:iota 10 :start 1)
```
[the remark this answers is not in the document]
appleby: I actually did that on my first pass, then switched to a list and found that it (surprisingly) doesn't seem to have a noticeable affect on runtime.  Maybe because `(upgraded-array-element-type 'rational) => T`?

tests/printer-tests.lisp
appleby: 'tis not

tests/printer-tests.lisp
appleby: I did generate them and just copy-pasted here.

The problem with generating them at run-time is you now have the problem of generating the pretty-printed representation, which is precisely the thing I'm trying to test.

tests/printer-tests.lisp
appleby: I lied. It is.

tests/printer-tests.lisp
appleby: Or did you mean getting rid of the `mapcan` that adds the +/- `+double-comparison-threshold-loose+` and just copy/pasting all those as well?

src/ast.lisp
```diff
@@ -1046,34 +1046,40 @@ N.B., The fractions of pi will be printed up to a certain precision!")
...
-              (cond
-                ((= numer 1)
-                 (format stream "~:[~;-~]pi/~d" (minusp r) denom))
-                ((zerop numer)
-                 (format stream "~F" r))
-                (t (format stream "~:[~;-~]~d*pi/~d" (minusp r) numer denom)))
-              (return-from format-real))))
-        ;; If we cannot find a nice fraction of pi, just print the
-        ;; real number
+  (if (or (not *print-fractional-radians*) (double~ (abs r) 0))
```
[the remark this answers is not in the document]
appleby: Done

tests/printer-tests.lisp
appleby: Done

tests/printer-test-files/gold-standard/plundered-gold.quil
```diff
@@ -192,7 +192,7 @@ DEFCIRCUIT EULER(%alpha, %beta, %gamma) q:
 
 
 RX(pi) 0
-RY(6.283185307179586) 0
+RY(2*pi) 0
```
appleby: Isn't she a beaut'?

tests/printer-tests.lisp
```diff
@@ -157,3 +157,250 @@ TEST(0.5) 0 1
...
+                        (2e10 . "20000000000.0"))))
+    (%check-format #'string= "~/quil:real-fmt/" simple-cases :print-fractional-radians t)
+    (%check-format #'string= "~/quil:real-fmt/" simple-cases :print-fractional-radians nil))
+
+  ;; Test *PRINT-FRACTIONAL-RADIANS* = NIL.
+  (%check-format #'uiop:string-prefix-p
+                 "~/quil:real-fmt/"
+                 `((,(- pi) . "-3.1415")
+                   (,pi . "3.1415")
+                   (,(* (- 1) (/ pi 2)) . "-1.5707")
```
appleby: Whoops. Forgot to `quil::pi` these.

src/ast.lisp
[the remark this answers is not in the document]
appleby: Still want?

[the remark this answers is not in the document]
stylewarning: it's definitely `t`, and i don't expect a huge difference, except we can go crazy with length-100000 vectors of rationals and be able to binary search it in proper logarithmic time. Up to you.

tests/printer-tests.lisp
stylewarning: i mean format the pretty representation yourself

```
(cons (* n quil::pi) (format nil "~D*pi" n))
```

but maybe it's not worth it with all of the special cases. I'll be happy if you write a non-test function to generate these cases somewhere in a #+#:ignore comment.

src/ast.lisp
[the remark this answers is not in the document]
appleby: Binary searchability is a good point I wasn't considering.  I suspect the list of rationals will stay small, so I think I'll leave it as-is for now. We can always swap it for a vector if it ever grows large enough to pass the tipping point for binary search.

tests/printer-tests.lisp
appleby: I could always write some simpler "obviously correct" (lol) re-implementation of the pretty-printing bits of `format-real`.  I will give that a try and see how it looks.  If not, I will add the function to auto-generate this list. I also had the idea to extend the tests with some test cases to make sure, e.g., `p/q * pi +/- (threshold + epsilon)` does not trigger pretty-printing.

I will send these changes in a separate PR.

Issue #317, opened 2019-07-05
Fix reversal of gate-application modifiers


appleby:
Something in the `PARSE-QUIL` -> `PRINT-PARSED-PROGRAM` cycle causes sequences of `CONTROLLED DAGGER` or `DAGGER CONTROLLED` to permute. For example:

```
CONTROLLED DAGGER Y 0 1
-> parse -> print ->
DAGGER CONTROLLED Y 0 1
-> parse -> print ->
CONTROLLED DAGGER Y 0 1
```

This does not affect correctness, but is mildly annoying and does affect the "fixed-point" check of `TEST-PRINT-PARSED-PROGRAM-GOLDEN-FILES`.

We should either print these modifiers in the same order they appear in the input Quil program, or else pick some canonical ordering so they are always printed the same way.

Once implemented, be sure to re-enable the fixed-point check for the test in `tests/printer-test-files/gold-standard/plundered-gold.quil` which is reference in the following code review discussion thread:

https://github.com/rigetti/quilc/pull/316#discussion_r300177587

appleby:
This issue is (temporarily) undead. #319

appleby:
I updated the issue title since this applies to FORKED modifiers as well. For example

	./quilc <<EOF
    > DEFCIRCUIT MODIFIERS q0 q1 q2:
    >	FORKED CONTROLLED DAGGER Y q0 q1 q2
    > HALT
    > EOF
    DEFCIRCUIT MODIFIERS q0 q1 q2:
	    DAGGER CONTROLLED FORKED Y q0 q1 q2

    HALT

appleby:
This was fixed in #329

PR #319, opened 2019-07-09
Revert changes to gate modifier parsing

appleby:
This a manual revert of the last commit from #316.

For context: https://github.com/rigetti/quilc/pull/316#discussion_r301745558


notmgsk:
@appleby does the fix for the original issue remain in master after this PR? Is there anything more to do with this PR?

appleby:
> @appleby does the fix for the original issue remain in master after this PR? Is there anything more to do with this PR?

No, this merely reverts my breakage from #316. I would say go ahead and merge this to unbreak things, then I will send another PR to fix #317.

PR #327, opened 2019-07-12
Print operator modifiers in reverse order

### LINKED ISSUE #317: Fix reversal of gate-application modifiers

appleby:
After my [failed attempt](https://github.com/rigetti/quilc/pull/316#discussion_r301745558) to reverse modifier order in the parser, this PR takes the opposite approach of leaving the parser alone, but reversing the order in which operator modifiers are printed in `PRINT-OPERATOR-DESCRIPTION`.

Note that a program like `CONTROLLED FORKED DAGGER Y 0 1 2` still results in an operator that looks like:

```
#.(CL-QUIL:DAGGER-OPERATOR
  #.(CL-QUIL::FORKED-OPERATOR
    #.(CL-QUIL:CONTROLLED-OPERATOR
	  #.(CL-QUIL:NAMED-OPERATOR "Y"))))
```

but the `CONTROLLED FORKED DAGGER` modifiers are now printed in the same order as they appeared in the original Quil source.

quilc / qvm / pyquil tests all passed locally.

Reviewers, let me know if reversing the printed order like this is insufficient.

**Commit message**

Prior to this change, gate-application modifiers got printed in the reverse order to how they appeared in the Quil text that was parsed. For example

BEFORE

```
    ./quilc <<EOF
    > DEFCIRCUIT MODIFIERS q0 q1 q2:
    >	FORKED CONTROLLED DAGGER Y q0 q1 q2
    > HALT
    > EOF
    DEFCIRCUIT MODIFIERS q0 q1 q2:
	DAGGER CONTROLLED FORKED Y q0 q1 q2

    HALT
```

AFTER

```
    ./quilc <<EOF
    > DEFCIRCUIT MODIFIERS q0 q1 q2:
    >	FORKED CONTROLLED DAGGER Y q0 q1 q2
    > HALT
    > EOF
    DEFCIRCUIT MODIFIERS q0 q1 q2:
	FORKED CONTROLLED DAGGER Y q0 q1 q2

    HALT
```

Fixes #317


src/ast.lisp
```diff
@@ -901,14 +901,20 @@ For example, `DAGGER DAGGER H 0` should produce `H 0`."
...
-    ((named-operator name) (write-string name stream))
-    ((controlled-operator o) (write-string "CONTROLLED " stream)
-                             (print-operator-description o stream))
-    ((dagger-operator o) (write-string "DAGGER " stream)
-                         (print-operator-description o stream))
-    ((forked-operator o) (write-string "FORKED " stream)
-                         (print-operator-description o stream))))
+  (labels ((print-reversed-modifiers (od)
+             (adt:match operator-description od
+               ((named-operator _))
```
appleby: I could capture the name here by `setf`ing a closed-over binding and save a call to `operator-description-root-name`, below.  I figure nobody is going to add 10,000 modifiers to their operator, so I went with the non-mutatey version.

notmgsk:
I feel very strongly that this should be fixed at the level of the parser.

appleby:
> I feel very strongly that this should be fixed at the level of the parser.

Fixing it in the parser was my first inclination as well, so I can certainly revisit that. I'm just not sure what else in the compiler will break if I change the parse order and permute the qubit args. But I can try it and see. I guess I'm still a little gun-shy about making such a change since the last time I tried it I "silently" broke the parser.

Why do you feel strongly it should be fixed in the parser? Is the current approach broken in some way? As far as I can tell quilc does the right thing with the modifiers, at least w.r.t. `parsed-program-to-logical` matrix it appears to produce the expected matrix. But I could certainly be missing something and might be allowing the compiler's *current* behavior to influence my opinion of how it *should* behave. For example, even thought the modifiers are parsed "in reverse", when `gate-matrix` is called, it processes them non-tail recursively, so the leftmost modifier in quil-source order (right-most in parsed representation) still gets applied first. One advantage of this approach is that handling of the `FORKED` modifier param splitting falls out of the recursion naturally, since `FORKED` can just pass half the params in each of the two recursive calls it makes to `gate-matrix`. For example:

```
QUIL> (defparameter *pp* (parse-quil "CONTROLLED FORKED Y 2 1 0"))
*PP*
QUIL> (aref (parsed-program-executable-code *pp*) 0)
#<FORKED CONTROLLED Y 2 1 0>
QUIL> (application-operator *)
#.(FORKED-OPERATOR #.(CONTROLLED-OPERATOR #.(NAMED-OPERATOR "Y")))
QUIL> (trace gate-matrix :methods nil)
(GATE-MATRIX)
QUIL> (parsed-program-to-logical-matrix *pp*)
  0: (CL-QUIL:GATE-MATRIX #<FORKED CONTROLLED Y 2 1 0>)
    1: (CL-QUIL:GATE-MATRIX #<CL-QUIL::FORKED-GATE {100200AB33}>)
      2: (CL-QUIL:GATE-MATRIX #<CL-QUIL:CONTROLLED-GATE {1001F8A123}>)
        3: (CL-QUIL:GATE-MATRIX #<CL-QUIL:SIMPLE-GATE {1001EFD5F3}>)
        3: GATE-MATRIX returned
             #<MAGICL:MATRIX [Z] 2x2:
                0.000 + 0.000j    -0.000 - 1.000j
                0.000 + 1.000j     0.000 + 0.000j>
...snip...
```

So when the recursion unwinds, you effectively get `(FORKED (CONTROLLED Y)))`.

But maybe handling of `parsed-program-to-logical-matrix` is not the whole story and there are other reasons to prefer fixing it in the parser that I am missing.

Or are you saying you expect the modifiers to apply left-to-right rather than right-to-left, and instead produce `(CONTROLLED (FORKED Y))`?

notmgsk:
Yeah, maybe you're right. I think my dislike is something to do with the mismatch between internal and external representations, and maybe that means that we could do better. But maybe not.

appleby:
> Yeah, maybe you're right. I think my dislike is something to do with the mismatch between internal and external representations, and maybe that means that we could do better. But maybe not.

I basically agree, hence why my first attempt was to fix it in the parser. Once you pointed out that I broke everything, I decided to retreat and go with the smallest diff that AFAICT fixes the bug while being least likely to introduce new bugs.

appleby:
Per an offline slack conversation, it turns out the original approach of reversing modifier order in the parser as implemented #316 was correct. I'm going to close this and include the fix in #329.

PR #329, opened 2019-07-15
Back to the Reversal Part II

appleby:
~~This is change is motivated by my recent [faux-pas](https://github.com/rigetti/quilc/pull/316#discussion_r300177587) in #316 (reverted in #319) where I reversed the order in which gate modifiers were parsed into an `OPERATOR-DESCRIPTION` without also permuting the order of qubit arguments. That change broke the parser, but did it in such a way that that no existing tests failed as a result.~~

Great Scott! After some offline discussion and little help from my friends, it seems that the [original fix](https://github.com/rigetti/quilc/pull/316#discussion_r300177587) from #316 was, indeed, correct.  The latest commits in this PR now reverse the reversal of reversing the gate modifiers that was reverted in #319.  Shout out to the Three Amigos who helped guide me back to the One True Fix.

~~This is still not ready for Prime Time. Remaining TODOs~~

- [x] Drop the golden file tests
- [x] Add some `compiler-hook` tests
- [x] ~~Track down rewiring bug~~ Not a bug: PEBKAC
- [ ] ~~Maybe whittle down the test cases to some minimum required set~~ Meh. Tests only take ~1s to run, even after adding compilation.  It is a pain to verify all these by eyeballing them though, so if reviewers complain, I can revisit this.

Changes included in this PR:

1. Reverse the order of `processed-modifiers` in `parse-modified-application` before calling `apply-modifiers-to-operator`.

2. Add some tests to ensure that `parsed-program-to-logical-matrix` is working as expected, especially w.r.t. gate modifiers and qubit consumption order.

3. Add some test cases to the compiler-hook-test-files that include a random assortment of gate modifier combinations.


tests/printer-tests.lisp
```diff
@@ -82,12 +82,53 @@ to compare output sections.
...
+  "Ensure that appleby didn't break everything again."
+
+  ;; Why hello. Has this test failed due to innocuous changes to the printed matrix representation
+  ;; of the parsed program? Then you should consider running UPDATE-SANITY-CHECK-GOLDEN-FILES,
+  ;; above, to update the output sections of the golden files. If you do so, however, be sure to
+  ;; examine the diffs of the old vs new output *carefully* to ensure all the changes are intended
+  ;; or expected. Golden files are precious, and their sanctity must be preserved. Thank you.
+
+  (let ((golden-files (uiop:directory-files *printer-test-files-gold-sanity-directory* #P"*.quil")))
+    (format t "~&  Gold-n-sanity tests:")
```
appleby: Is it insane to compare matrices via `STRING=`? Of course it is. But now that I have my golden-file hammer, everything looks like a golden nail.

In fairness, testing this way made it extremely convenient while I was iterating on test cases, since I could just call `UPDATE-SANITY-CHECK-GOLDEN-FILES`, and then "easily" eyeball the resulting output to make sure it looked correct.

If someone yells at me hard enough, I'll move these out of the golden files and into a `matrix-equality` test. I'll sulk about it some, though.

tests/printer-test-files/gold-sanity/matrix-insanity.quil
```diff
@@ -0,0 +1,272 @@
...
+CONTROLLED X 1 0
+# Output
+#<MAGICL:MATRIX [Z] 4x4:
+   1.000 + 0.000j     0.000 + 0.000j     0.000 + 0.000j     0.000 + 0.000j
+   0.000 + 0.000j     1.000 + 0.000j     0.000 + 0.000j     0.000 + 0.000j
+   0.000 + 0.000j     0.000 + 0.000j     0.000 + 0.000j     1.000 + 0.000j
+   0.000 + 0.000j     0.000 + 0.000j     1.000 + 0.000j     0.000 + 0.000j>
+# Input
+# Test CONTROLLED on non-permutation gate
+CONTROLLED Y 0 1
```
appleby: I have attempted to verify these output matrices by eye and by checking their action on basis states, and believe them to be correct.  But if someone who has the superpower of staring at an arbitrary matrix and verifying at-a-glance what it does (*cough* @ecpeterson *cough*) wants to double-check my work here, that would be appreciated.

tests/printer-tests.lisp
notmgsk: My concern is with the precision of the printed matrices. We have made the aesthetically convenient choice to print only three decimal places. This would be bad if the compared matrices differed beyond this meagre precision.

notmgsk: but really `whocares.jpg` about precision

appleby: Valid concern and one of the reasons I tried to stick to matrices with "simple" entries like `0, 1, +/- i` in most places.  The lone exceptions are the `RX(pi/2)` incantations when I want to test `DAGGER` on something non-hermitian.

Also, the intention of these tests is more as a basic sanity check rather than checking numerical accuracy to 15 sig figs, but making these do a real `matrix-equality` (or `operator~` or whatever it's called) tests would mean we get that "for free".

notmgsk: Also LAPACK is weird and sometimes outputs `- 0j` rather than `+ 0j`. I'm not convinced that would show up where tests are sufficiently simple and do not probe LAPACK's quirks, but there are so many weird things that *could* pop up that this method of testing makes me lean towards the more reliable method.

appleby: Fine. I *guess* I'll do the right thing. I'm just being lazy to copy/paste these into a `.lisp` file and format them nicely.

I will leave this version up for a bit first though, to give reviewers time to point out errors, since it really is easier (I think) to eyeball the output and make sure the resulting matrices look correct.

I might be overly optimistic in thinking anyone will actually take the time to do that, but hope springs eternal.

appleby:
Pushed a commit that does `parsed-program-to-logical-matrix` sanity testing more sanely, rather than via string comparisons. Note that the limited precision of the printed reps was something of a feature, since the tests are only really meant to check that the entries appear in the right places and the `FORKED`, `CONTROLLED`, `DAGGER` do the expected thing and consume qubits in the expected order.

I also switch from `RX(theta)` to `PHASE(theta)` as my non-hermitian `DAGGER` test since it means all entries for the test matrices are drawn exclusively from {0, 1, i, -i}.

Current diff has both versions. Let me know which one you prefer. I was planning to mark the PR as a draft, but seems that's only possible at PR creation, so I modified the title to make it clear it shouldn't be merged in it's current state.

ecpeterson:
We talked about this offline, and we decided that there's a genuine bug where modifiers are mis-ordered at parse time, which invalidates some of the test data here. Once that bug is squashed and the reference test data here is regenerated, this looks OK to me. I _do_ like being able to see the matrix printouts—it's how we resolved a bug!—but I agree with Mark that I'd prefer not to use string equality to test the matrix pairs. The matrix entries are printed to 10^-3 and are supposed to agree to within ~10^-10, but -0 and 0 are close even by that metric.

appleby:
Fine. I will kill the golden file tests (`*sniffle*`). Actually, it turned out that the formatted lisp lists are maybe even more readable than the golden file outputs once I switched from `RX` to `PHASE` for the `DAGGER` tests. I might also expand the tests with a handful of multi-`FORKED` gates, just for good measure.

I will update this PR with the original fix of reversing modifiers at parse time and also update the test expectations.

appleby:
I think this is ready for review now.

~500:1 ratio of line-of-test-diff to lines-of-fix diff.  Have fun eyeballing the expected output matrices in `test-logical-matrix-sanity`. I know I did :-).

If anyone can think of additional tests to add, I am all ears.

tests/logical-matrix-sanity-tests.lisp
```diff
@@ -0,0 +1,423 @@
...
+                            expected)))
+            ;; FIASCO:IS always evaluates it's format arguments, even if the test assertion
+            ;; succeeds.  Formatting via MATRIX-MISMATCH-FMT will only compute the MATRIX-MISMATCH
+            ;; when/if the associated test assertion actually fails.
+            (is (quil::matrix-equality actual expected)
+                "Checking input: ~S~%~/cl-quil-tests::matrix-mismatch-fmt/"
+                input
+                (list actual expected))
+            (is (quil::operator= compiled expected)
+                "Checking compiled input: ~S~%~/cl-quil-tests::matrix-mismatch-fmt/"
```
appleby: This test is quite similar to what `test-compiler-hook` does, but limited to the default architecture, rather than checking against all of them.  Doing the compilation here slows down this test significantly, so if we don't feel like the compiled version is adding value over and above what `test-compiler-hook` already does, I can drop it, possibly adding a subset of the below test cases to compiler-hook-test-files and running them through the full complement of supported architectures.

PR #331, opened 2019-07-22
Use swap fidelity instead of gate time as a distance metric

karalekas:
Will add more information once this is closer to being ready for review.


src/addresser/logical-schedule.lisp
```diff
@@ -596,7 +596,11 @@ mapping instructions to their tags. "
                                                                   (equalp x (gate-binding-operator y)))))))
                           (unless gate-record
                             (warn-and-skip instr))
-                          (setf fidelity (gate-record-fidelity gate-record))))))
+                          (dohash ((key val) specs-hash)
```
ecpeterson: I think this chunk can be dropped. The `findhash` call above accomplishes this now.

PR #210, opened 2019-04-29
Annotate compiler functions

ecpeterson:
The goal of this (sequence of) PR(s) is to uniformize the different kinds of compilation subroutines appearing in CL-QUIL, to annotate them with information about e.g. which architecture they target, and to replace the hard-coded material in `chip-specification.lisp` with a mechanism that extracts a subset of compilers that target a given architecture. The hope is that this will also feed into our plans for "fine-grained compilation": this will both make it easier for users to define custom compilers and to control the list of compiler techniques that CL-QUIL uses as it proceeds.

What follows are some musings about the road between here and there. Keeping in mind our main goal of automating the decision of what compiler to use when, shat are the salient features of a compiler?

* A compiler consumes a string of instructions (and perhaps some context), then emits a string of instructions.
* A compiler matches on a particular form of input: instruction name, qubit count, ..., all the way up to arbitrary predicates based on the matrix form of an instruction.
* A compiler provides guarantees about the emitted sequence, e.g., what gate set it belongs to, or the ordering of instructions, or the number of instructions (either total or of a particular type).

What "breeds" of compilers do we currently have?

* "Translators": one in, many out; match on operator name and perhaps a #'double= condition on the parameter value.
* Chip-sensitive "translators": one in, many out; uses context like chip topology to decide what constitutes legal output.
* "Rewriting rules": many in, many out; matches on operator name and perhaps on complicated conditions on parameter values; usually stays within a prescribed gate set for both in and out.
* Context-sensitive "rewriting rules": all of the above, except there are also complicated conditions involving the context, e.g., whether the present wavefunction of the quantum computer is an eigenstate for the instruction.
* Generic compilers: one in, many out; some don't match on anything except qubit count (optimal-2q, qs-compile); some match on name (ucr-compile); some match on matrix structure (ucr-recognize).
* "Approximation templates": one in (actually, a canonical coordinate (which maybe we could replace with a pseudoinstruction?)), many out; uses chip information to make a decision about when we might return not-quite-the-right gate.

What are some useful tools can you might use to make guarantees about successful termination? Here it seems like there are only a few major features of instructions, in what I'd consider to be their decreasing order of severity: their qubit count, their operator name, and their parameter values. Target gate sets typically impose strong conditions on each of these (e.g.: ≤2Q gates; of the 2Q gates only CZs; of the 1Q gates either RX or RZ; of the RX gates only RX(k pi/2)). CL-QUIL resolves these constraints in the order given:

[...]

PR #339, opened 2019-07-24
Add diagonal matrix compilation routine from tweedledum

notmgsk:


src/contrib/tweedledum/tweedledum.lisp
```diff
@@ -69,6 +69,43 @@ GIVE-UP-COMPILATION if INSTR is not a permutation gate."
...
+  ;; instruction. Like with permutation gates, diagonality should be
+  ;; available as part of the structure definition. Maybe that already
+  ;; exists.
+  (let ((m (quil:gate-matrix instr)))
+    (unless (diagonal-p m)
+      (quil::give-up-compilation))
+    ;; This synthesis routine works on the unitary
+    ;;   U = diag(1, e^{-i t_1}, ..., e^{-i t_{2^n - 1}})
+    ;; and takes as input the angles t_i; hence (rest ...) below.
+    (let ((angles (rest (mapcar #'phase (magicl:matrix-diagonal m)))))
```
jmbr: Suggestion: use `angles (mapcar #'phase (rest (magicl:matrix-diagonal m)))` instead (no need to evaluate `phase` on the constant 1 and then drop it, we can just drop it from the get go).

jmbr:
This PR gives me a wonderful opportunity to raise https://github.com/rigetti/quilc/pull/305 back from the dead and bring it to your (@notmgsk and possibly @stylewarning 's) attention.

notmgsk:
> This PR gives me a wonderful opportunity to raise #305 back from the dead and bring it to your (@notmgsk and possibly @stylewarning 's) attention.

ok ok ok :)

src/contrib/tweedledum/tweedledum.c
```diff
@@ -14,6 +15,20 @@
...
+  extern char* tweedledum_synthesis_diagonal(double* angles, uint32_t size) {
+    using namespace tweedledum;
+    std::vector<double> vangles(angles, angles + size);
+    std::cout << vangles.size() << std::endl;
+    auto network = diagonal_synth<gg_network<mcmt_gate>>(vangles);
+
+    std::stringstream ss;
+    write_quil(network, ss);
+    const std::string resultstr = ss.str();
+    char* str = (char*) malloc( (resultstr.length()+1) * sizeof(char) );
```
jmbr: Lines 27-29 can be replaced by `return strdup(resultstr.c_str());`

src/contrib/tweedledum/tweedledum.c
```diff
@@ -14,6 +15,20 @@
 
 extern "C" {
 
+  extern char* tweedledum_synthesis_diagonal(double* angles, uint32_t size) {
+    using namespace tweedledum;
+    std::vector<double> vangles(angles, angles + size);
+    std::cout << vangles.size() << std::endl;
```
jmbr: Do we really want to output this? It seems to be for debugging purposes only.

notmgsk: Gone :)

src/contrib/tweedledum/tweedledum.c
notmgsk: I remember thinking there must be a simpler way. Ty :)

ecpeterson:
Recording / broadcasting some internal discussion:

The ideas around this PR—but not the PR itself—have two warts.

1. The implementation of this decomposition routine inside of `tweedledum` counts the position inside of the diagonal entries with an `int32`, and it detects the most significant nonzero bit in that position with a single-precision floating point logarithm. Together, these limit the number of qubits to which the routine applies. The routine itself doesn't actually look that complicated, and so there's some discussion of complaining to tweedledum's maintainers, filing a fix in their repository, rewriting it in Lisp to make use of its more flexible integer type, ... .

2. The gates to which this routine applies belong to the middle class in (uniformly controlled RZs == multiply-FORKED RZs) <= diagonal gates <= dense unitary operators. This routine also gives much better output than that of the dense decomposition `qs-compile.lisp`, but worse output than that of the FORKED compiler `ucr-compile.lisp`. Accordingly, the author of the diagonal compilation routine has to know to inject this compilation routine at a particular point in `*global-compilers*` — but a lot of the work in #210 was about automating these kinds of decisions in the case of a fixed qubit count, and it would be great to be able to extend that work to make this injection decision as well.

PR #340, opened 2019-07-25
Publish test-compiler-hook stats

ecpeterson:
Working on a compilation subroutine and you aren't sure how it's affecting output quality for example programs? Try running `cl-quil-tests::test-compiler-hook` with `print-stats` turned on!

### DIFF
--- a/tests/compiler-hook-tests.lisp
+++ b/tests/compiler-hook-tests.lisp
@@ -99,11 +99,15 @@ JUMP @a")))
          (proc-prog
            (quil::compiler-hook (quil::transform 'quil::compress-qubits
                                                  (cl-quil::read-quil-file file))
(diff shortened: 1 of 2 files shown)


tests/compiler-hook-tests.lisp
```diff
@@ -114,7 +117,9 @@ some test programs."
         (format *debug-io* "      Testing file ~a:" (pathname-name file))
         (dolist (architecture (list ':cz ':iswap ':cphase ':piswap ':cnot))
           (format *debug-io* " ~a" architecture)
-          (compare-compiled file architecture))
+          (let ((stats (compare-compiled file architecture)))
+            (when print-stats
```
notmgsk: `when` outside the `let`

ecpeterson:
If we want to add more statz later, we certainly can.

PR #342, opened 2019-07-26
Fix bugs related to unpacking fidelity values

ecpeterson:
(1) Turn some list-walkers into sequence-walkers: quilc's web interface deserializes lists into lists, and the RPCQ interface deserializes lists into vectors. I made some bad assumptions in #210's `chip-reader.lisp` about which sequence type would be coming down the wire. The first commit corrects that.

(2) `loads-specs-layer` was confusing `gate-record`s and `compiler-binding`s, by using `copy-binding` on a `gate-record` and expecting it to update `:fidelity` key field. The second commit straightens out this gibberish.


ecpeterson:
FWIW: I can round-trip jobs from pyQuil with this PR in place.

src/chip-reader.lisp
```diff
@@ -116,8 +116,8 @@
                              :qubit (intern-if-string (gethash "qubit" gate-datum))))
       (t
        (make-gate-binding :operator (named-operator (gethash "operator" gate-datum))
-                          :parameters (mapcar #'intern-if-string (gethash "parameters" gate-datum))
-                          :arguments (mapcar #'intern-if-string (gethash "arguments" gate-datum)))))))
+                          :parameters (map 'list #'intern-if-string (gethash "parameters" gate-datum))
```
notmgsk: Whenever I see `(map 'list ...)` I immediately think "hey isn't that what `maplist` is for" and then I remember that lisp is weird.

ecpeterson:
Let's not merge this until I figure out the relationship to Mark's other bug.

PR #349, opened 2019-07-29
Rework the Docker inheritance and CI pipeline

karalekas:
Goals
------
- [x] Allow quilc to change quicklisp dependency version without an rpcq release
- [x] Start explicitly specifying the dependent image versions instead of using `latest`
- [x] Allow dependent image labels to be overridden during `docker build`
- [x] Have the quilc image depend on the latest qvm image (needed for unit tests)
- [x] Run the CI tests from within freshly-built quilc Docker images, rather than using the "build within a Docker image" out-of-the-box feature of GitLab CI (because you can only specify one image to build FROM, as opposed to the flexibility of multi-stage builds if you use your own Dockerfile)
- [x] Update the README to explain how to upgrade the Dockerfile

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -10,22 +10,25 @@ test-cl-quil:
   stage: test
   tags:
     - github
+    - dockerd
   only:
     - branches
-  image: rigetti/rpcq
+  image: docker:stable
   script:
-    - make dump-version-info
-    - make install-test-deps
-    - make test-cl-quil RIGETTI_LISP_LIBRARY_HOME=/src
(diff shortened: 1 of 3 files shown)


ecpeterson:
Perhaps a premature approval, but so far things look fine

karalekas:
For this to truly address all the problems that were raised, I'm going to need to spend some more time on this. Will mark as non-draft when it's ready 2 go.

karalekas:
The next commit is where it starts 2 get tricky.

karalekas:
This should be good to go now.

PR #350, opened 2019-07-30
Undelete deprecated option

notmgsk:


app/src/entry-point.lisp
```diff
@@ -168,7 +168,12 @@
   '((("print-logical-schedule" #\s)
      :type boolean
      :optional t
-     :documentation "include logically parallelized schedule in output.  Requires -P.  This is inactive and will eventually be removed."))
+     :documentation "include logically parallelized schedule in output.  Requires -P.  This is inactive and will eventually be removed.")
+    
+    (("json-serialize" #\j)
+     :type boolean
+     :optional t
+     :documentation "serialize output as a JSON object"))
```
appleby: Maybe a "This is inactive and will eventually be removed" note in the doc string, like for `print-logical-schedule` above.

PR #355, opened 2019-07-30
eliminate identities via rewriting rule

stylewarning:
Better than Euler-compiling `I`...

### DIFF
--- a/src/compressor/rewriting-rules.lisp
+++ b/src/compressor/rewriting-rules.lisp
@@ -7,6 +7,11 @@
 
 ;;; rewriting rules in general
 
+(define-compiler eliminate-identities
+    ((x ("I" () _)))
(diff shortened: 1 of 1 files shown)


src/compressor/rewriting-rules.lisp
```diff
@@ -31,7 +36,8 @@
 
 (defun global-rewriting-rules ()
   "Rewriting rules that may be useful in manipulating instruction sequences that do not have an underlying notion of 'native hardware'.  In particular, this includes canonicalizing the ordering of commuting instructions."
-  (list #'sort-CZs
+  (list #'eliminate-identities
```
ecpeterson: FWIW, this isn't necessary: this compiler will get picked up by all hardware objects.

PR #356, opened 2019-07-30
Hotfix: Euler infinite loop

ecpeterson:
The recent "annotated compilers" PR (#210) introduced an infinite loop on Linux. A summary of the changes in this PR:

1. In one place, program gate parameters were being compared against a reference set of gate parameters by `equalp` rather than `double=`. This was causing the bug to only show on certain Linux installations and never on macOS (but it should have been showing everywhere).
2. The decision about which compiler to include into a rewriting ruleset was guarded by two conditions: the compiler should emit gates that belong to the target gateset (without any further intervening compilers), and the compiler's output should have estimated fidelity at least as good as its input. The former was checked in two ways: quilc first tried to reason about gateset inclusion (e.g.: if `RZ(_)` is permitted, then the rule `RZ(phi) RZ(theta) |-> RZ(phi+theta)` should be allowed) and second it resorted to case-checking (e.g.: if `RX(k pi/2)` is permitted for -2 ≤ k ≤ 2, then the rule `RX(phi) RX(theta) |-> RX(phi + theta mod 2pi)` should be allowed, even though it broadly consumes gates of the form `RX(_)`). This is not tight enough to exclude Euler ZXZ decomposition: for instance, the gate RX(pi/2) can be decomposed as RZ(0) RX(pi/2) RZ(0), which has native output of better-or-equal fidelity to its input. We included two fixes for this error:
2a. In actuality, instructions like RZ are not perfect (i.e., of fidelity 1d0) because of the limitations of digital electronics: waveforms, phase shifts, and angle values in the input program are all stored with finite granularity, and so one should only expect RZ to be represented with about 2^-(# bits) of fidelity. We introduced a constant, `+near-perfect-fidelity+`, to account for such situations.
2b. We never want to permit "unguarded" compilers, which match on any input gate, to enter a set of rewriting rules. Rewriting rules are always allowed to take action, and the rewriting process rewound and begun again, whenever an instruction sequence is in their domain of definition. An unguarded rewriting rule, then, will _always_ trigger an infinite loop.
3. Version bump to 1.10.1, since we'll certainly want to re-release once this problem is fixed.

Both 2a and 2b have some defects. The value purported as "constant" in 2a is not actually so: it's a function of the characteristics of the control electronics. We could use `1 - +double-comparison-threshold-strict+` to at least reflect other choices that we've made internally to quilc, but even that has a note of kludgeiness to it. In 2b (and in the surrounding code), we make the somewhat bold assumption that _any_ predicate (including a disjunction of your favorite predicates) is _always_ falsifiable, which may not be true generally and _certainly_ may not be true when restricting attention to gates drawn from a particular set. So, 2b may not be aggressive enough to avoid infinite loops.

Either of 2a and 2b alone is enough to nip this particular problem in the bud.


src/define-compiler.lisp
```diff
@@ -695,33 +695,43 @@ N.B.: The word \"shortest\" here is a bit fuzzy.  In practice it typically means
...
+                                 (symbolp (gate-binding-operator b))
+                                 (endp (gate-binding-options b))))))
+                 (compiler-bindings compiler))
+          ;; have cleared these checks, we support two kinds of matches:
+          ;; (1) a "blind" match, where we can tell just by considering bindings
+          ;;     that the compiler will always have reasonable input and output
+          ;; (2) an "exhaustive" match, where we instantiate different test cases
+          ;;     to pump through the compiler, convincing ourselves that in all
+          ;;     applicable situations the compiler gives good output.
+          (or
```
ecpeterson: This clause is all the same, just re-indented by the preceding `and`.

src/define-compiler.lisp
```diff
@@ -695,33 +695,43 @@ N.B.: The word \"shortest\" here is a bit fuzzy.  In practice it typically means
...
-          ;; if that falls through, try the exhaustive match
-          (let ((matches (loop :for binding :in (compiler-bindings compiler)
-                               :collect (gates-that-match-binding binding))))
-            (unless (every #'identity matches)
-              (return-from consider-compiler nil))
-            (every #'identity (apply #'a:map-product (a:curry #'consider-input-test-case compiler) matches))))))
+         ;; certain things we will never tolerate:
+         ;; (1) unguarded matches will always result in infinite loops.
+         (and
+          (every (lambda (b)
```
notmgsk: notany

ecpeterson:
I think it would be cheap to maintain a total count of the instruction list and to warn if it ever did something like triple in length. I agree, though: not in the PR.

notmgsk:
> I think it would be cheap to maintain a total count of the instruction list and to warn if it ever did something like triple in length. I agree, though: not in this PR.

Isn't that exactly what `H 0` is going to do?

ecpeterson:
`H 0` isn't native & so won't ever appear as input to the rewriter.

PR #362, opened 2019-07-30
accept misordered ISA keys

ecpeterson:
The pyQuil tests have been QPU descriptors with `specs` layers containing misordered ISA keys all along. This PR extends quilc's tolerance toward this, though it still scolds with a warning.

### DIFF
--- a/src/chip-reader.lisp
+++ b/src/chip-reader.lisp
@@ -73,10 +73,10 @@
   (:report "Invalid QPU description file: missing required ISA layer or sub-layer."))
 
 (defun expand-key-to-integer-list (key)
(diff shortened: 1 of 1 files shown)


ecpeterson:
oh, oops, my local copy has a `warn` when sort acts nontrivially. that’s ok, serves me right for not saving my sass.

src/chip-reader.lisp
```diff
@@ -234,6 +233,7 @@
 
 (defun load-specs-layer (chip-spec specs-hash)
   "Loads the \"specs\" layer into a chip-specification object."
+  (declare (optimize (debug 3)))
```
appleby: Friendly reminder: `C-u C-c C-c`

:-)

PR #365, opened 2019-07-31
Use proper parameter arithmetic in rewriters

notmgsk:
Pending test fixtures


src/compressor/rewriting-rules.lisp
```diff
@@ -197,7 +197,7 @@
     ((x ("PISWAP" (theta) p q)
         :where (and (typep theta 'double-float)
                     (double= theta (* pi (floor theta pi))))))
-  (list (build-gate "PISWAP" `(,(- theta pi)) p q)
+  (list (build-gate "PISWAP" `(,(param-+ theta pi)) p q)
```
appleby: I assume switching from - to + here doesn't matter because theta + pi is equivalent to theta - pi?

ecpeterson: Good eye. We do need a `+` here: the `PISWAP` parameter is only 4pi-periodic.

tests/translator-tests.lisp
```diff
@@ -90,76 +101,126 @@
...
+                                                  (gethash param parameter-assignments))
+                                                 (t
+                                                  (error 'test-case-too-complicated
+                                                         :reason "I don't know how to generate this parameter."))))
+                                             (quil::gate-binding-parameters binding))))
+                           (qubits (mapcar (lambda (qubit)
+                                             (cond
+                                               ((numberp qubit)
+                                                qubit)
+                                               ((eql qubit (intern "_"))
```
stylewarning: no need to intern; just use `'_`

tests/translator-tests.lisp
```diff
@@ -90,76 +101,126 @@
...
+                                   ((numberp qubit)
+                                    (push qubit output-qubits))
+                                   ((null qubit)
+                                    (push (random-with-exceptions qubit-bound
+                                                                  (append output-qubits occupied-qubits))
+                                          output-qubits)))))))
+                      (cond
+                        ((symbolp (quil::gate-binding-operator binding))
+                         (apply #'quil::anon-gate "ANONYMOUS-INPUT"
+                                (quil::random-special-unitary (ash 1 (length qubits)))
```
stylewarning: prefer `(expt 2 (length qubits))`

stylewarning:
(I added nitpicks because it's not ready-4-merge, according to @notmgsk)

PR #369, opened 2019-08-01
Parallelize tests

notmgsk:
## Without parallelization:
```
Evaluation took:
  939.605 seconds of real time
  938.227703 seconds of total run time (920.563960 user, 17.663743 system)
  [ Run times consist of 47.040 seconds GC time, and 891.188 seconds non-GC time. ]
  99.85% CPU
  92,010 forms interpreted
  2,679 lambdas converted
  2,548,142,861,497 processor cycles
  617,767,259,472 bytes consed
```

## With parallelization:
```
Evaluation took:
  261.228 seconds of real time
  644.712250 seconds of total run time (627.920342 user, 16.791908 system)
  [ Run times consist of 46.721 seconds GC time, and 597.992 seconds non-GC time. ]
  246.80% CPU
  35 forms interpreted
  99 lambdas converted
  708,435,090,211 processor cycles
  324,243,353,296 bytes consed
```

Note: tweedledum is used in neither of these as I haven't rebased onto master in a while.

Some TODO items:
- [x] Figure out the number of available threads, and configure the lparallel kernel to use these. A sub task is to investigate if there is an optimal number of threads.
- [x] Clean up the output of the test suite. My expectation is that most of the text (e.g. "RUNNING TEST BLAHBLAH ..... [OK]") produced by a single test would be replaced by a single dot, to show progress. Otherwise you'll see a bunch of garbled text as threads compete for stdout.


appleby:
You stole my idea! j/k great minds think alike.  Now that I have this beefy 16-core laptop, I like this idea even more!

notmgsk:
### without parallel, without tweedledum
```
Evaluation took:
  341.940 seconds of real time
  343.317939 seconds of total run time (335.434383 user, 7.883556 system)
  [ Run times consist of 24.926 seconds GC time, and 318.392 seconds non-GC time. ]
  100.40% CPU
  93,992 forms interpreted
  2,572 lambdas converted
  927,327,078,390 processor cycles
  237,929,154,560 bytes consed
```

### without parallel, with tweedledum
```
Evaluation took:
  187.137 seconds of real time
  187.230604 seconds of total run time (183.131409 user, 4.099195 system)
  [ Run times consist of 14.214 seconds GC time, and 173.017 seconds non-GC time. ]
  100.05% CPU
  94,404 forms interpreted
  60 lambdas converted
  507,511,488,400 processor cycles
  121,453,557,872 bytes consed
```
  
### with parallel, without tweedledum
```
Evaluation took:
  199.927 seconds of real time
  288.363019 seconds of total run time (278.110210 user, 10.252809 system)
  [ Run times consist of 32.193 seconds GC time, and 256.171 seconds non-GC time. ]
  144.23% CPU
  48 forms interpreted
  99 lambdas converted
  542,197,138,558 processor cycles
  196,295,109,648 bytes consed
```

### with parallel, with tweedledum
```
Evaluation took:
  73.254 seconds of real time
  193.507896 seconds of total run time (185.209192 user, 8.298704 system)
  [ Run times consist of 18.481 seconds GC time, and 175.027 seconds non-GC time. ]
  264.16% CPU
  48 forms interpreted
  198,663,128,394 processor cycles
  102,801,500,896 bytes consed
```

notmgsk:
OK. This is probably ready for review.

It adds a couple things that invoke a parallelized test suite:
* A `:parallel` keyword to `cl-quil-tests::run-cl-quil-tests` that is the number of threads to use. If `nil` fall back to the standard test suite (i.e. so very slow).
* A `test-fast` target to the `Makefile` which will use tweedledum *and* some threading. You can configure the number of threads with `THREADS=1000 make test-fast` and it defaults to two.

With four threads and tweedledum, tests take approx 70s.

Makefile
```diff
@@ -13,6 +13,7 @@ QUICKLISP_BOOTSTRAP_URL=https://beta.quicklisp.org/quicklisp.lisp
 UNAME_S=$(shell uname -s)
 ZMQ_REPO=https://download.opensuse.org/repositories/network:/messaging:/zeromq:/release-stable/xUbuntu_16.04/
 PREFIX ?= /usr/local
+THREADS ?= 2
```
appleby: Any guestimate here on what is the "sweet spot" for minimizing test time? The exact value probably depends on the specifics of your laptop etc., but I think it makes sense to aim for conservative-but-close-to-optimal-assuming-relatively-recent-developer-laptop here rather than just conservative, since you only get threading if you specifically request it.

tests/compiler-hook-tests.lisp
```diff
@@ -114,21 +109,28 @@ JUMP @a")))
...
-        (format *debug-io* "      Testing file ~a:" (pathname-name file))
-        (dolist (architecture (list ':cz ':iswap ':cphase ':piswap ':cnot))
-          (format *debug-io* " ~a" architecture)
-          (let ((stats (compare-compiled file architecture)))
-            (when print-stats
-              (format *debug-io* "~a" stats))))
-        (terpri *debug-io*)))))
+(eval-when (:compile-toplevel :load-toplevel :execute)
+  
+  (defmacro define-compiler-hook-tests ()
```
appleby: Maybe just `macrolet` unless you think this will be useful elsewhere.

tests/compiler-hook-tests.lisp
```diff
@@ -114,21 +109,28 @@ JUMP @a")))
...
+                                                 (concatenate 'string "TEST-COMPILER-HOOK-"
+                                                              (string-upcase (pathname-name test-file)))
+                                                 (concatenate 'string "TEST-COMPILER-HOOK-"
+                                                              (string-upcase (pathname-name test-file))
+                                                              "-WITH-STATE-PREP-COMPRESSION"))
+                             :append
+                             (loop :for arch :in (list ':cz ':iswap ':cphase ':piswap ':cnot)
+                                   :for name := (concatenate 'string test-name "-" (string arch))
+                                   :collect
+                                   `(deftest ,(intern name) (&key print-stats)
```
appleby: Nice. IIUC, this will make it even more obvious which test failed since every file now gets it's own `deftest`.

Did you mean to bind `quil::*enable-state-prep-compression*` here?

Also looks like `print-stats` is ignored.

tests/compiler-hook-tests.lisp
```diff
@@ -114,21 +109,28 @@ JUMP @a")))
...
-      (format *debug-io* "~&    With *ENABLE-STATE-PREP-COMPRESSION* ~a~%" quil::*enable-state-prep-compression*)
-      (dolist (file (uiop:directory-files *compiler-hook-test-file-directory* #P"*.quil"))
-        (format *debug-io* "      Testing file ~a:" (pathname-name file))
-        (dolist (architecture (list ':cz ':iswap ':cphase ':piswap ':cnot))
-          (format *debug-io* " ~a" architecture)
-          (let ((stats (compare-compiled file architecture)))
-            (when print-stats
-              (format *debug-io* "~a" stats))))
-        (terpri *debug-io*)))))
+(eval-when (:compile-toplevel :load-toplevel :execute)
```
appleby: `eval-when` still needed here? asking for a friend.

tests/compiler-hook-tests.lisp
appleby: > That, but also because `test-compiler-hook` was the bottleneck

What I meant to say was that in addition to enabling parallel execution, this seems like a small win even for the single-threaded case.

notmgsk: Ah, then yes :)

Makefile
notmgsk: For me (with tweedledum):
* 1 thread: ∞s
* 2 threads: 120s
* 4 threads: 95s
* 8 threads: 75s
* 16 threads: 77s

I think 4 cores is a reasonable assumption these days, so I'll bump the default thread count to 4.

notmgsk:
> do you get readable output/backtraces when one or more tests fail from a make test-fast run?

Yes! :D

> does interactive debugging in slime repl work, specifically the fiasco "retest" restart

Also yes! The other threads will continue testing while a broken test presents (and acts upon) the normal restarts.

tests/compiler-hook-tests.lisp
```diff
@@ -114,21 +109,32 @@ JUMP @a")))
...
+                                :append
+                                (loop :for arch :in (list ':cz ':iswap ':cphase ':piswap ':cnot)
+                                      :for name := (concatenate 'string test-name "-" (string arch))
+                                      :collect
+                                      `(deftest ,(intern name) (&key print-stats)
+                                         (finish-output *debug-io*)
+                                         (let* ((cl-quil::*enable-state-prep-compression* ,state-prep)
+                                                (stats (compare-compiled ,test-file ,arch)))
+                                           (when print-stats
+                                             (format *debug-io* "~a" stats)))
```
appleby: I noticed in the parser tests you swapped formatting on `*debug-io*` for `fiasco::*test-run-standard-output*`. Do you want `fiasco::*test-run-standard-output*` here as well? Ditto in the `finish-output` above and the `terpri` below.

tests/compilation-tests.lisp
```diff
@@ -91,40 +91,40 @@
 
 (deftest test-cnot-rewiring-in-cnot-architecture ()
   "Test that all CNOTs on each pair of qubits compile into a single CNOT rewired appropriately."
-  (format t "~&    [Test output: ~%")
+  (format fiasco::*test-run-standard-output* "[Test output: ~%")
```
stylewarning: Is this the right thing to do? I thought `*standard-output*` got rebound

tests/compiler-hook-tests.lisp
```diff
@@ -114,21 +109,32 @@ JUMP @a")))
...
-          (let ((stats (compare-compiled file architecture)))
-            (when print-stats
-              (format *debug-io* "~a" stats))))
-        (terpri *debug-io*)))))
+(macrolet
+    ((define-compiler-hook-tests ()
+       `(progn
+          ,@(loop :for test-file :in (uiop:directory-files (asdf:system-relative-pathname
+                                                            ':cl-quil-tests
+                                                            "tests/compiler-hook-test-files/") #P "*.quil")
```
stylewarning:
```suggestion
                                                            "tests/compiler-hook-test-files/") #P"*.quil")
```

tests/compiler-hook-tests.lisp
```diff
@@ -114,21 +109,32 @@ JUMP @a")))
...
+                                                    (concatenate 'string "TEST-COMPILER-HOOK-"
+                                                                 (string-upcase (pathname-name test-file))
+                                                                 "-WITH-STATE-PREP-COMPRESSION"))
+                                :for state-prep :in (list nil t)
+                                :append
+                                (loop :for arch :in (list ':cz ':iswap ':cphase ':piswap ':cnot)
+                                      :for name := (concatenate 'string test-name "-" (string arch))
+                                      :collect
+                                      `(deftest ,(intern name) (&key print-stats)
+                                         (finish-output fiasco::*test-run-standard-output*)
```
stylewarning: factor the body of this test out into toplevel function

```
(defun compiler-hook-test (compression file arch)
  (let ((*enable... compression))
   ...))
```

tests/compiler-hook-tests.lisp
```diff
@@ -114,21 +109,32 @@ JUMP @a")))
...
+                                         (let* ((cl-quil::*enable-state-prep-compression* ,state-prep)
+                                                (stats (compare-compiled ,test-file ,arch)))
+                                           (when print-stats
+                                             (format fiasco::*test-run-standard-output* "~a" stats)))
+                                         (terpri fiasco::*test-run-standard-output*))))))))
+
+  (define-compiler-hook-tests))
```
stylewarning: this macrolet business is slightly unsettling; a bunch of tests being generated dynamically depending on the filesystem.

Maybe there should be some way to do in-test parallelism. Maybe a `fiasco:async-is` which submits a task to the lparallel kernel?

tests/compiler-hook-tests.lisp
```diff
@@ -104,31 +99,41 @@ JUMP @a")))
...
-        (format *debug-io* "      Testing file ~a:" (pathname-name file))
-        (dolist (architecture (list ':cz ':iswap ':cphase ':piswap ':cnot))
-          (format *debug-io* " ~a" architecture)
-          (let ((stats (compare-compiled file architecture)))
-            (when print-stats
-              (format *debug-io* "~a" stats))))
-        (terpri *debug-io*)))))
+  "Test whether the compiler hook preserves semantic equivalence for some test programs."
+  (finish-output fiasco::*test-run-standard-output*)
+  (lparallel:pmap nil
```
stylewarning: these nested maps are unsightly; I suggest parallelizing yourself:

```
(defun test-compiler-hook ()
  (let ((num-tasks 0))
    (flet ((create-task (state-prep file arch)
            (incf num-tasks)
            (lambda ()
              (let ((*state-prep* state-prep))
                (compare-compiled file architecture)))))
      (let ((ch (lparallel:make-channel)))
        (dolist (state-prep ...)
          (dolist (file ...)
            (dolist (arch ...)
              (lparallel:submit-task ch (create-task state-prep file arch)))))
        (loop :repeat num-tasks :do (lparallel:receive-result ch))))))
```

tests/compiler-hook-tests.lisp
```diff
@@ -170,12 +175,12 @@ RX(pi) 2
               (shuffle-list l k))))
 
 (deftest test-compiler-hook-random-4Q ()
-  (finish-output *debug-io*)
+  (finish-output fiasco::*test-run-standard-output*)
```
stylewarning: Was there an explanation why we have to use fiasco internal variables anytime we want to print?

tests/suite.lisp
```diff
@@ -4,13 +4,43 @@
 
 (in-package #:cl-quil-tests)
 
-(defun run-cl-quil-tests (&key (verbose nil) (headless nil))
-  "Run all CL-QUIL tests. If VERBOSE is T, print out lots of test info. If HEADLESS is T, disable interactive debugging and quit on completion."
+(defun run-cl-quil-tests (&key (verbose nil) (headless nil) (parallel nil))
+  "Run all CL-QUIL tests. If VERBOSE is T, print out lots of test info. If HEADLESS is T, disable interactive debugging and quit on completion. If PARALLEL is non-nil, it is the number of threads to use."
   ;; Bug in Fiasco commit fe89c0e924c22c667cc11c6fc6e79419fc7c1a8b
```
stylewarning: (check-type parallel (or null (integer 1)))

tests/compiler-hook-tests.lisp
notmgsk: 🤷‍♂ I haven't the energy to investigate.

ecpeterson: I have done no investigation, but it doesn't seem crazy to me: my memory from working on the RPCQ server is that the standard streams get NIL'd out when working inside a thread, and so you have to either pass a stream along or preserve a special variable to send output anywhere. I'm surprised that `fiasco::*test-run-standard-output*` is surviving specialization to a thread, but I'm not surprised that `*debug-io*` isn't.

notmgsk: I think `fiasco::*test-run-standard-output*` is configured by the `deftest` macro.

notmgsk:
@stylewarning Please re-re-review.

tests/compilation-tests.lisp
```diff
@@ -91,40 +91,40 @@
 
 (deftest test-cnot-rewiring-in-cnot-architecture ()
   "Test that all CNOTs on each pair of qubits compile into a single CNOT rewired appropriately."
-  (format t "~&    [Test output: ~%")
+  (format *standard-output* "[Test output: ~%")
```
stylewarning: Is it not possible to use `t`? `t` is a designator for `*terminal-io*`.

It doesn't sit well with me that unidiomatic code can't be written in tests; I want to preserve the ability to write normal code, especially if the functions themselves are writing to some other standard stream.

tests/compiler-hook-tests.lisp
```diff
@@ -101,34 +96,46 @@ JUMP @a")))
...
+             (bt:with-lock-held (tasks-lock)
+               (incf num-tasks))
+             (lambda ()
+               (let ((quil::*enable-state-prep-compression* state-prep))
+                 (format *standard-output*
+                         "    Testing file ~a (~a, with~:[out~;~] state prep)~%"
+                         (pathname-name file) arch state-prep)
+                 (compare-compiled file arch)))))
+      (let ((ch (lparallel:make-channel)))
+        (dolist (state-prep (list nil t))
```
stylewarning: use constant list `'(nil t)`

tests/compiler-hook-tests.lisp
```diff
@@ -101,34 +96,46 @@ JUMP @a")))
...
+          (dolist (file (uiop:directory-files *compiler-hook-test-file-directory* #P"*.quil"))
+            (dolist (arch (list ':cz ':iswap ':cphase ':piswap ':cnot))
```
stylewarning: use constant list `'(:cz :iswap ...)`

tests/compiler-hook-tests.lisp
```diff
@@ -101,34 +96,46 @@ JUMP @a")))
...
+              (lparallel:submit-task ch (create-task state-prep file arch)))))
+        (loop :repeat num-tasks :do
+          (let ((fiasco::*print-test-run-progress* nil))
```
stylewarning: what's the reasoning for this binding? (again, it goes against the expectation of writing idiomatic code)

tests/compiler-hook-tests.lisp
```diff
@@ -170,29 +177,42 @@ RX(pi) 2
...
+                 (format *standard-output*
+                         "    Testing architecture ~a, with~:[out~;~] state prep~%"
+                         arch state-prep)
+                 (let* ((num-qubits 4)
+                        (v (quil::random-special-unitary (expt 2 num-qubits)))
+                        (args (shuffle-list (a:iota num-qubits :start (1- num-qubits) :step -1)))
+                        (parsed-prog
+                          (make-instance
+                           'quil::parsed-program
+                           :executable-code (make-array 1
```
stylewarning: `(vector (make-instance...))`

tests/compiler-hook-tests.lisp
```diff
@@ -170,29 +177,42 @@ RX(pi) 2
...
+                                                                          :operator (named-operator "RANDO-GATE")
+                                                                          :gate v
+                                                                          :arguments (mapcar #'qubit args)))))
+                        (processed-program
+                          (quil::compiler-hook parsed-prog (quil::build-nQ-linear-chip num-qubits
+                                                                                       :architecture arch))))
+                   (quil::matrix-equals-dwim (quil::kq-gate-on-lines v num-qubits args)
+                                             (quil::parsed-program-to-logical-matrix processed-program)))))))
+      (let ((ch (lparallel:make-channel)))
+        (dolist (state-prep (list nil t))
```
stylewarning: constant lists

tests/parser-tests.lisp
```diff
@@ -17,15 +17,15 @@
 (deftest test-parsing-good-test-files ()
   "Test whether all valid test files parse."
   (dolist (file (uiop:directory-files *good-test-file-directory* #P"*.quil"))
-    (format *debug-io* "~&    Testing good file ~A~%" (pathname-name file))
+    (format fiasco::*test-run-standard-output* "~&    Testing good file ~A~%" (pathname-name file))
```
stylewarning: these should be changed

tests/suite.lisp
```diff
@@ -4,13 +4,49 @@
 
 (in-package #:cl-quil-tests)
 
-(defun run-cl-quil-tests (&key (verbose nil) (headless nil))
-  "Run all CL-QUIL tests. If VERBOSE is T, print out lots of test info. If HEADLESS is T, disable interactive debugging and quit on completion."
+(defun run-cl-quil-tests (&key (verbose nil) (headless nil) (parallel nil))
+  "Run all CL-QUIL tests. If VERBOSE is T, print out lots of test info. If HEADLESS is T, disable interactive debugging and quit on completion. If PARALLEL is non-nil, it is the number of threads to use."
+  ;; Since each test is run in its own thread *and* each test may try
```
stylewarning: I don't think this is very intuitive and should be ameliorated. 

- Rename `parallel` to `num-threads`, defaulting to `1`
- `(check-type num-theads (integer 1))`
- bind for both branches `(lparallel:*kernel* (lparallel:make-kernel (max 2 num-threads)))`
- replace `parallel` branch with the test `(< 1 num-threads)`

tests/suite.lisp
```diff
@@ -4,13 +4,49 @@
...
+  ;; to spin up subthreads, the minimum number of threads required is
+  ;; two. If we allow just a single thread, then any test which tries
+  ;; to run a subthread will block because there are no more threads
+  ;; available.
+  (check-type parallel (or null (integer 2)))
```
stylewarning: see above note

tests/suite.lisp
```diff
@@ -4,13 +4,49 @@
...
   (setf fiasco::*test-run-standard-output* (make-broadcast-stream
                                             *standard-output*))
-  (let ((quil::*compress-carefully* t))
+  (let ((quil::*compress-carefully* t)
+        ;; Some tests make use of lparallel directly. If we don't set
+        ;; a kernel here, then the (non-parallel) test suite will
+        ;; pause and request that we create a kernel. To avoid that,
+        ;; we create a single thread kernel that ought to behave as if
+        ;; there was no parallelisation.
+        (lparallel:*kernel* (lparallel:make-kernel 2)))
```
stylewarning: see above note; this can be bound correctly once

tests/suite.lisp
```diff
@@ -4,13 +4,49 @@
...
     (cond
+      (parallel
```
stylewarning: see above note on changing this test

tests/suite.lisp
```diff
@@ -4,13 +4,49 @@
...
+       (let ((lparallel:*kernel* (lparallel:make-kernel parallel))
```
stylewarning: see above note; this line can be eliminated

tests/suite.lisp
```diff
@@ -4,13 +4,49 @@
...
+             (fiasco::*debug-on-unexpected-error* nil)
```
stylewarning: all this internal business is scary and prone to breakage. Can you write a comment here about it?

tests/suite.lisp
stylewarning: this test should also check that `headless` is `t` I think

tests/translator-tests.lisp
```diff
@@ -250,7 +250,7 @@
 
 (deftest test-global-instruction-expansion ()
   "Tests the expansion of gate applications on non-adjacent qubits by comparing the matrix of the gate before and afte expansion."
-  (finish-output *debug-io*)
+  (finish-output fiasco::*test-run-standard-output*)
```
stylewarning: more weirdness

tests/compiler-hook-tests.lisp
notmgsk: Attempt to curtail fiasco's output, which I expected would be a garbled mess for parallel tests. Anyway, it doesn't work well so I'll remove it.

tests/suite.lisp
notmgsk: Why?

stylewarning: I think I was thinking that parallelization should only happen when headless, only because of the ordering of results, possibly terrible error handling experience (i.e., none), etc.

tests/compiler-hook-tests.lisp
```diff
@@ -101,34 +96,45 @@ JUMP @a")))
...
+             (bt:with-lock-held (tasks-lock)
+               (incf num-tasks))
+             (lambda ()
+               (let ((quil::*enable-state-prep-compression* state-prep))
+                 (format *standard-output*
+                         "    Testing file ~a (~a, with~:[out~;~] state prep)~%"
+                         (pathname-name file) arch state-prep)
+                 (compare-compiled file arch)))))
+      (let ((ch (lparallel:make-channel)))
+        (dolist (state-prep (list nil t))
```
stylewarning: this is still a list that can be `'(nil t)`

PR #372, opened 2019-08-02
358/remove-preprocessors

cbosborn:
- Added compilation to random-quil producers step.
- Moved compilation to processors step.
- Removed preprocessors step entirely.
- Added examples.lisp with the updated example from the README.
- Update example script in README to match the script in examples.lisp.


boondoggle/README.md
```diff
@@ -10,34 +10,32 @@ Here is a Lisp snippet illustrating the usage of this library:
...
+      (progn
+        (quil::print-parsed-program (produced-program))
+        (apply-process (processors i) (produced-program))))
+    (qvm-results ((j consumers) (i processors))
+      (progn
+        (quil::print-parsed-program (compiled-program i))
+        (consume-quil (consumers j) (compiled-program i))))
+    (l1-distance ((l consumers) (k processors)
+                  (j consumers) (i processors))
+      (apply-process (post-process) (qvm-results j i) (qvm-results l k)))))
```
ecpeterson: Is this still emitting values close to `2`? If so: if we reduce `quil::build-8q-chip` to `quil::build-nQ-linear-chip 2`, does the L1 number go down? If not: something is wrong somewhere.

boondoggle/src/processors.lisp
```diff
@@ -33,32 +33,12 @@
 (defmethod apply-process ((processor processor-identity) &rest data)
   (first data))
 
-(defclass processor-measures (processor)
-  ((chip-specification :initform nil
-                       :initarg :chip-specification
-                       :reader processor-measures-chip-specification)))
-
 (defun measure-at-close-instrs (chip-specification)
```
ecpeterson: Did you mean to kill this too, since it's now been replicated in `boondoggle/src/producers.lisp`?

boondoggle/src/examples.lisp
```diff
@@ -0,0 +1,40 @@
+;;;; examples.lisp
```
ecpeterson: Another option is to define a test suite and define this as a test rather than an example.

boondoggle/src/processors.lisp
cbosborn: That's correct, forgot to push.

boondoggle/README.md
```diff
@@ -10,34 +10,32 @@ Here is a Lisp snippet illustrating the usage of this library:
 
 ```
 (let ((*debug-noise-stream* *standard-output*)
-      (chip-spec ...))
+       (chip-spec (quil:build-8q-chip)))
   (pipeline
-   ;; variables / data / definitions
+     ;; variables / data / definitions
```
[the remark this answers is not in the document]
cbosborn: Strange, for some reason I don't see this in my editor.

boondoggle/src/examples.lisp
```diff
@@ -0,0 +1,40 @@
...
+                                              :chip-specification chip-spec
+                                              :respect-topology t))
+    
+      (processors    (i)       (list (make-instance 'processor-identity)
+                                     (make-instance 'processor-quilc
+                                                    :executable-path "<path-to-quilc-executable>")))
+      (consumers     (j)       (list (make-instance 'consumer-local-qvm
+                                                    :trials 1000)))
+      (post-process  ()        (make-instance 'processor-L1-distance)))
+     ;;
```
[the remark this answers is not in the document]
cbosborn: Documentation is a wip, sorry :)

boondoggle/src/examples.lisp
```diff
@@ -0,0 +1,40 @@
...
+     (produced-program ()
+                       (produce-quil-program (producer)))
+     (compiled-program ((i processors))
+                       (progn
```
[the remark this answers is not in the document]
cbosborn: This indentation style is my own doing. I'll look up the accepted way to do it

boondoggle/README.md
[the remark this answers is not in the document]
notmgsk: Yeah me neither.

boondoggle/src/examples.lisp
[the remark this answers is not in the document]
notmgsk: I think it's less to do with your emacs settings, and more to do with how the macro is implemented. Slime does some magic to figure out how to indent macro bodies, but it may be unsuccessful here because that macro expands into a bunch of nested `flet`s and `let`s. IMO the macro could be rewritten to output a single `labels` which slime would understand better (maybe). Anyway, I think that's probably outside the scope of this PR.

boondoggle/src/producers.lisp
```diff
@@ -156,4 +162,14 @@ PARAMETER-BOUNDS is a list of maximum random values for the gate parameters."
     (setf (quil::parsed-program-executable-code parsed-program)
           (make-array (length instruction-list)
                       :initial-contents instruction-list))
+    ;;; Add measure instructions
+    (push (quil::make-memory-descriptor
+                :name "ro"
+                :type quil::quil-bit
+                :length (quil::chip-spec-n-qubits chip-specification))
```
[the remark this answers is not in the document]
cbosborn: I got an error suggesting this one needed `::`.

boondoggle/README.md
cbosborn: With `quil::build-8q-chip` I am getting ~ `((((0.0d0 1.754d0)) ((1.754d0 0.0d0))))`. With `quil::build-nQ-linear-chip 2` I'm getting ~ `((((0.0d0 0.3d0)) ((0.3d0 0.0d0))))`.

boondoggle/tests/.#suite.lisp
```diff
@@ -0,0 +1 @@
+ChrisOsborn@MacBook-Pro.local.16255
```
notmgsk: Delete

boondoggle-tests.asd
```diff
@@ -0,0 +1,21 @@
+;;;; boondoggle-tests.asd
+;;;;
+;;;; Author: Chris Osborn
+
+(asdf:defsystem #:boondoggle-tests
+  :description "Boondoggle tests."
+  :author "Chris Osborn <chris@rigetti.com>"
+  :license "Apache License 2.0 (See LICENSE.txt)"
+  :depends-on (#:quilc
```
notmgsk: Does this really depend on quilc, or is it cl-quil?

boondoggle/src/consumers.lisp
```diff
@@ -38,7 +38,10 @@
                        instr))))))
 
 (defmethod consume-quil ((consumer consumer-local-qvm) parsed-program)
-  "Runs a Quil program on the local QVM."
+  "This function runs a Quil program on the QVM, and returns a histogram of counts of each resulting bitstring, ordered lexicographically, and normalized against total number of trials.
+
+For a program constructing a bell state with a large sampling count, the returned value would be (0.5d0 0.0d0 0.0d0 0.5d0).
+"
```
notmgsk: Delete newline

boondoggle/src/processors.lisp
```diff
@@ -33,34 +33,10 @@
...
-                       (measure-at-close-instrs
-                        (processor-measures-chip-specification processor))))
-    data))
-
 (defclass processor-L1-distance (processor)
   ())
 
 (defmethod apply-process ((processor processor-L1-distance) &rest data)
+  "Calculate the L1 distance between two arguments. &rest expects only 2 position arguments, each of which is expected to be a normalized histogram count of bitstring outputs from the QVM for a given quil program.
+"
```
notmgsk: Delete newline

boondoggle/src/producers.lisp
```diff
@@ -65,6 +65,14 @@ PARAMETER-BOUNDS is a list of maximum random values for the gate parameters."
 (defclass producer-from-directory (producer)
   ((directory-path :initform nil)))
 
+(defun measure-at-close-instrs (chip-specification)
+  "Apply MEASURE instructions on all qubits in a chip-specification. Assumes qubits are ordered from 0 to n with no skipped indices.
+"
```
notmgsk: Delete newline

boondoggle/src/producers.lisp
```diff
@@ -156,4 +164,15 @@ PARAMETER-BOUNDS is a list of maximum random values for the gate parameters."
...
           (make-array (length instruction-list)
                       :initial-contents instruction-list))
+    ;; Add readout memory definitions to the program
+    (push (quil::make-memory-descriptor
+                :name "ro"
+                :type quil::quil-bit
+                :length (quil::chip-spec-n-qubits chip-specification))
+          (quil:parsed-program-memory-definitions parsed-program))
+    ;; Add measure instructions to the end of the program
+    (setf (quil::parsed-program-executable-code parsed-program)
```
notmgsk: This could be cleaned up with `alexandria:appendf`

boondoggle/tests/examples.lisp
```diff
@@ -0,0 +1,51 @@
...
+  (let ((*debug-noise-stream* *standard-output*)
+        (chip-spec (quil::build-8q-chip)))
+    (let ((l1-results (boondoggle::pipeline
+                        ((producer      ()        (make-instance 'boondoggle::producer-random
+                                                                 :program-volume-limit 20
+                                                                 :chip-specification chip-spec
+                                                                 :respect-topology t))      
+                         (processors    (i)       (list (make-instance 'boondoggle::processor-identity)
+                                                        (make-instance 'boondoggle::processor-quilc
+                                                                       :executable-path "~/code/lisp/quilc/quilc")))
```
notmgsk: Perhaps default this path to the relative `#P"../quilc"`.

notmgsk: Actually, this is going to be awkward if these tests are expected to be run in CI/CD. Maybe `boondoggle::processor-quilc` should be rewritten to use `quilc::process-program` rather than the binary directly.

notmgsk: Likewise, `consumer-local-qvm`.

boondoggle/src/producers.lisp
cbosborn: @notmgsk after trying this out I was pointed to the fact that these are `vector`s, and appendf operates on `list`s.

boondoggle/tests/examples.lisp
cbosborn: @notmgsk These are legitimate Qs. How about I capture them as issues for when we choose to incorporate boondoggle into CI/CD, and address then?

cbosborn: @notmgsk Solving for your original suggestion with `#P"..."` I've implemented the following:

```
:executable-path (namestring #P"./quilc")
```

I found that the kwarg requires a string input, but I believe you're right that `#P` is needed to resolve relative directories. I'm using `./quilc` because `boondoggle-tests.asd` resides in the `quilc/` folder, not the `boondoggle/` folder. I should probably move them.

cbosborn:
Ok this is no longer wip.

boondoggle/src/processors.lisp
```diff
@@ -33,34 +33,9 @@
...
-                       (measure-at-close-instrs
-                        (processor-measures-chip-specification processor))))
-    data))
-
 (defclass processor-L1-distance (processor)
   ())
 
 (defmethod apply-process ((processor processor-L1-distance) &rest data)
+  "Calculate the L1 distance between two arguments. &rest expects only 2 position arguments, each of which is expected to be a normalized histogram count of bitstring outputs from the QVM for a given quil program."
   (reduce #'+ (mapcar #'abs (mapcar #'- (first data) (second data)))))
```
jmbr: I know you haven't written this line but it's asking to be rewritten as `(reduce #'+ (mapcar (a:compose #'abs #'-) ...))`

boondoggle/README.md
```diff
@@ -10,34 +10,32 @@ Here is a Lisp snippet illustrating the usage of this library:
...
-    (processors    (j) (list (make-instance 'processor-measures
-                                            :chip-specification chip-spec)
-                             (make-instance 'processor-identity)))
-    (consumers     (j) (list (make-instance 'consumer-local-qvm
-                                            :trials 10)))
+  
+    (processors    (i)       (list (make-instance 'processor-identity)
+                                   (make-instance 'processor-quilc
+                                                  :executable-path "<path-to-quilc-executable>")))
+    (consumers     (j)       (list (make-instance 'consumer-local-qvm
```
ecpeterson: Mentioned in DM: you can drop the `list` here, then also the `j`, then some of the obnoxious indexing / parens further below.

boondoggle/README.md
ecpeterson: And those numbers go down when trials go up?

boondoggle/src/processors.lisp
ecpeterson: love 2 cons

boondoggle/tests/tests.lisp
```diff
@@ -0,0 +1,50 @@
...
+                                     (progn
+                                       (quil::print-parsed-program (compiled-program i))
+                                       (boondoggle::consume-quil (consumers j) (compiled-program i))))
+                        (l1-distance ((l consumers) (k processors) (j consumers) (i processors))
+                                     (boondoggle::apply-process (post-process) (qvm-results j i) (qvm-results l k))))))
+      (destructuring-bind ((((a b)) ((c d)))) l1-results ; l1 results are of the form e.g. ((((0.0d0 0.5d0)) ((0.5d0 0.0d0)))) 
+        (fiasco:is (= b c)) ; Off-diagonal elements must equal
+        (fiasco:is (= a d)) ; Diagonal elements must equal
+        (fiasco:is (= a 0.0)) ; Diagonal elements must be zero
+        (fiasco:is (<= b 2.0)) ; Off-diagonal elements must be less than 2.0
```
ecpeterson: I was asking the questions in the README thread in an effort to make this number _much_ smaller than `2.0`. In fact, if you use a weak `<=`, the largest value the mathematically-correct implementation of `L1-...` could produce is `2.0`, and the assertion that it's strictly less than `2.0` (indeed: _much_ less than `2.0`!) should be taken to mean "the compiler is working".

boondoggle/src/processors.lisp
jmbr: Following @stylewarning 's meme:

Who -> `(reduce #'+ (mapcar #'abs (mapcar #'- (first data) (second data))))`
Whom -> `(reduce #'+ (mapcar (a:compose #'abs #'-) (first data) (second data)))`
Whomst -> `(loop :for x :in (first data) :for y :in (second data) :summing (abs (- x y)))`
Whomst'd -> `(series:collect-sum (series:mapping (((x y) (series:scan-multiple 'list (first data) (second data)))) (abs (- x y))))`

appleby: Haha. A+ meme. I was going to give a shout-out to `SERIES` here. Beat me to it :)

cbosborn: @jmbr I don't get it yet but 😂 anyway

jmbr: @cbosborn They all do the same but the first two variants are more memory-hungry (they create lists with intermediate results before `reduce` is called). The fourth variant uses an iteration package named series that's existed for a long time but didn't make it into the ANSI CL standard (despite having some niceties like lazy comprehensions and what not). I think the preferred approach in our code base is the third option (i.e., use `loop`).

boondoggle/tests/tests.lisp
cbosborn: @ecpeterson Ok, after a detour into this correctness question I've Implemented a two-sample chi-squared test. See the test docstring for further explanation. In doing this I had to change the output from `consume-quil` to the unnormalized bitstring histograms. Let me know if you have any questions.

ecpeterson:
There's more to do generally—boondoggle dates from Dec 2017 and so its code has quite a stale taste—but I don't think those things necessarily belong in this PR.

Issue #91, opened 2019-02-08
Ensure `PROCESS-INCLUDES` works fine with memory declarations


ecpeterson:
There's some parsed-program appending in this transform, and declared memory might not be saved from the include.

PR #374, opened 2019-08-03
Include memory definitions from included quil

kilimanjaro:
This addresses #91. When a memory region is declared across multiple files, this signals a `quil-parse-error`.

### DIFF
--- a/src/analysis/process-includes.lisp
+++ b/src/analysis/process-includes.lisp
@@ -30,16 +30,24 @@
     (unless (uiop:file-exists-p file)
       (error "Could not include ~S because it does not exist." file))
     (let ((incl-pp (parse-quil-into-raw-program (a:read-file-into-string file))))
-      (setf (parsed-program-gate-definitions pp)
-            (append (parsed-program-gate-definitions pp)
-                    (parsed-program-gate-definitions incl-pp)))
-      (setf (parsed-program-circuit-definitions pp)
-            (append (parsed-program-circuit-definitions pp)
-                    (parsed-program-circuit-definitions incl-pp)))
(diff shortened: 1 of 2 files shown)


kilimanjaro:
> My only concern here is how this handles the case where files A and B both define the same memory region, possibly with incompatible types. If you try to redefine a memory region in a single Quil file, the parser will complain (even if the definitions are identical).

True, probably the appropriate thing is to raise a parse error if there any any conflicts.

src/analysis/process-includes.lisp
```diff
@@ -30,12 +30,24 @@
     (unless (uiop:file-exists-p file)
       (error "Could not include ~S because it does not exist." file))
     (let ((incl-pp (parse-quil-into-raw-program (a:read-file-into-string file))))
+      (let ((collisions (intersection (parsed-program-memory-definitions pp)
```
notmgsk: `a:when-let`

src/analysis/process-includes.lisp
```diff
@@ -30,12 +30,24 @@
...
+but this region has already been declared elsewhere."
+                            (memory-descriptor-name (first collisions))
+                            file)))
       (setf (parsed-program-gate-definitions pp)
             (append (parsed-program-gate-definitions pp)
                     (parsed-program-gate-definitions incl-pp)))
       (setf (parsed-program-circuit-definitions pp)
             (append (parsed-program-circuit-definitions pp)
                     (parsed-program-circuit-definitions incl-pp)))
+      (setf (parsed-program-memory-definitions pp)
```
notmgsk: I wonder how @stylewarning feels about `alexandria:appendf`

Issue #361, opened 2019-07-30
Stochastic failure to compile 'disconnected' program on connected topology. 


kylegulshen:
This came up while running quantum volume, e.g. 
```python
qc = get_qc('Aspen-4-8Q-A', as_qvm=True, noisy=False)
measure_quantum_volume(qc, num_circuits=200, depths=[4])
```
I've been able to isolate the problem to the point where for `Program` prog = 

```
DEFGATE LYR0_RAND0:
    0.3453183205304793-0.12696801053233828i, 0.5265386376910978+0.1931838984531052i, -0.29052305165907266-0.594441852976343i, -0.1506008541194645-0.29937609929246506i
    -0.1593953114488653-0.20866235427010307i, -0.16168835357073963+0.2520089033062991i, 0.239834418759946-0.2438490569066952i, -0.7460675648562454+0.4096362084408356i
    -0.12856619973598493+0.7223359877459085i, 0.4807485314801305-0.24994946297635628i, 0.20519011099777268+0.10677911378343408i, -0.33834032003758086-0.011341762443828618i
    -0.04332809717096084+0.5054816193435523i, -0.5205466694277492+0.1766995012896463i, -0.61250636255294-0.12877703191585974i, -0.18249325797051483-0.12396815063332628i

DEFGATE LYR0_RAND1:
    -0.21163731076953685+0.1556329927465134i, -0.3001566064245327-0.33802345390852884i, 0.7431016288284251-0.37714645363637356i, -0.1772723072697441+0.027734947336669865i
    0.20484396673693783+0.5578246542068888i, 0.5182948848663171+0.13152502746494504i, 0.23098272365238026-0.16896663770943743i, 0.3885456697969944-0.3578710705261025i
    0.05563181510118506-0.18850346757941397i, -0.11536779759544692-0.5264188304392133i, -0.09087636334777575-0.0776098534213045i, 0.7836539637683549+0.20627576785173588i
    -0.6294273438803477-0.378221347023483i, 0.4673940664514512-0.028879206744236774i, -0.14087341305332418-0.4353511883395492i, -0.009238897502739374-0.17893372117702133i

DEFGATE LYR1_RAND0:
    -0.3161540949620567+0.33398130454144603i, 0.5730670791443696-0.1988884256830049i, -0.5087925775965276+0.27789976292939966i, -0.24656615939351895+0.153777622475405i
[...]

PR #375, opened 2019-08-05
Relabel rewiring comments in COMPRESS-QUBITS

appleby:
The `COMPRESS-QUBITS` transform wasn't updating rewiring comments, which causes `PARSED-PROGRAM-TO-LOGICAL-MATRIX` to generate the wrong matrix if `COMPRESS-QUBITS` wasn't a no-op.

This was noticed while working on the fix for #361, which changes the default initial rewiring type for some programs from `':NAIVE` to `':PARTIAL` or `':GREEDY`. Making that change caused the tests in `QUILC-TESTS::TEST-FAITHFULNESS` to fail due to this `COMPRESS-QUBITS` bug.

~~This is still a WIP. Posting for early feedback to see if the basic approach seems sane.~~

Summary of changes in this PR:

1. Call RELABEL-REWIRING on rewiring comments in COMPRESS-QUBITS (see above)
2. Sprankle some abstraction over rewiring comments
3. Tests

TODOs:

- [x] Replace `READ-FROM-STRING` in `PARSE-ENTERING/EXITING-REWIRING` with something akin to `MAKE-REWIRING-FROM-STRING`
- [x] Refactor existing rewiring-related code to use new abstractions
- [x] Replace `COMPILER-HOOK` tests with more focused `COMPRESS-QUBITS` tests


app/src/entry-point.lisp
```diff
@@ -490,7 +490,7 @@ Version ~A is available from https://www.rigetti.com/forest~%"
          (let* ((program-text (slurp-lines))
                 (program (parse-quil program-text))
                 (original-matrix (when (and protoquil compute-matrix-reps)
-                                   (parsed-program-to-logical-matrix program))))
+                                   (parsed-program-to-logical-matrix program :compress-qubits t))))
```
appleby: The motivation for this change is that when the user requests `---compute-matrix-reps`, we end up comparing `ORIGINAL-MATRIX` to the matrix of the compiled program, and when calling `PARSED-PROGRAM-TO-LOGICAL-MATRIX` on the compiled program, below, we pass `:COMPRESS-QUBITS T`.  If we don't also compress qubits here, there is no guarantee the resulting matrices will be the same size. There are several ways to resolve this:

1. Compress qubits here (current diff represents this choice)
2. Don't compress qubits for `PROCESSED-PROGRAM-MATRIX`, below
3. Call `MATRIX-RESCALE` on the two matrices to ensure they are the same size before passing them to `SCALE-OUT-MATRIX-PHASES` and/or `MATRIX-EQULS-DWIM`.

The downside of (1) is that `ORIGINAL-MATRIX` is meant to be the "known good" base case, so by passing it through `COMPRESS-QUBITS` there is more chance for bugs to invalidate the test. For example, if `COMPRESS-QUBITS` causes `P-P-T-L-M` to generate the same incorrect matrix for both `ORIGINAL-MATRIX` and `PROCESSED-PROGRAM-MATRIX`

The downside to (2) is that it's one less test of `COMPRESS-QUBITS`. For example, if the existing code hadn't been calling `COMPRESS-QUBITS` here, I would have never noticed the bug.

The downside to (3) is I haven't tested it yet and don't know if it actually works :))

tests/compiler-hook-tests.lisp
```diff
@@ -118,9 +120,11 @@ some test programs."
         (format *debug-io* "      Testing file ~a:" (pathname-name file))
         (dolist (architecture (list ':cz ':iswap ':cphase ':piswap ':cnot))
           (format *debug-io* " ~a" architecture)
-          (let ((stats (compare-compiled file architecture)))
-            (when print-stats
-              (format *debug-io* "~a" stats))))
+          (dolist (rewiring-type '(nil :naive :partial :greedy))
```
appleby: This makes the already-slow `COMPILER-HOOK-TEST` waaaay slower, which is why my intention is to remove this before merging and introduce some smaller, more focused `COMPRESS-QUBITS` tests.

notmgsk:
Without actually reviewing: you have my thanks for introducing some first-class representation of rewiring comments. I think the stuff in `process-program` could be cleaned up with this (and likely other places that scan for rewiring comments).

tests/compiler-hook-tests.lisp
appleby: Thumbs about making COMPILER-HOOK tests slower, or thumbs down about removing this before merging?

src/ast.lisp
```diff
@@ -158,6 +158,89 @@ The keys are typically INSTRUCTION instances and associated values are STRINGs."
...
+  "Parse an exiting REWIRING from REWIRING-STRING."
+  (%parse-single-rewiring rewiring-string +exiting-rewiring-prefix+))
+
+(defun parse-entering/exiting-rewiring (rewiring-string)
+  "Parse entering and exiting REWIRINGs from REWIRING-STRING.
+
+Return (VALUES ENTERING-REWIRING EXITING-REWIRING)."
+  ;; TODO:(appleby) Stolen from PARSED-PROGRAM-TO-LOGICAL-MATRIX. Replace this with something akin to MAKE-REWIRING-FROM-STRING.
+  (let ((*read-eval* nil))
+    (destructuring-bind (entering-vector . exiting-vector)
```
notmgsk: I could probably run-it-and-see, but I'm tired: how does this handle a malformed entering/exiting form? e.g. is the error raised by `(parse-entering/exiting-rewiring "Entering/exiting rewiring: ((0 1 2 3) . ())")` intelligible to an end-user?

tests/compiler-hook-tests.lisp
notmgsk: Maybe both! But mostly the slower tests.

src/ast.lisp
appleby: Currently will blow up more-or-less inscrutably. One of my TODOs above is to rewrite this function to not use `read-from-string` and instead parse it explicitly, a-la `make-rewiring-from-string`.  In that function, I will include a decent error message.

Actually, one question I had (maybe for @ecpeterson) is whether the length of the entering and exiting rewirings always has to match, or would it be possible to have something like

```
(#(0 1 2 3) . #(1 0))
```

appleby: ^ For context, I mention this because I noticed there is a `trim-rewiring` function that selects a minimum prefix of rewiring, discarding any tail that is an identity mapping.

notmgsk: Ah, it would make sense that redundant info can be elided.

src/addresser/rewiring.lisp
```diff
@@ -271,3 +271,41 @@ BODY as an implicit PROGN."
...
+           :END-ANCHOR)
+         str))
+    (assert matchedp
+            nil
+            "Malformed rewiring pair string: ~@
+             input ~A is not of the form (#(...) . #(...))."
+            str)
+    (let ((first-rewiring-string (aref matches 0))
+          (second-rewiring-string (aref matches 1)))
+      (assert (= (length (aref matches 0)) (length (aref matches 1)))
```
appleby: Unclear if the assertion is actually valid and/or required. I opted to err on the side of caution and require the entering/exiting rewirings to be the same length.

tests/compiler-hook-tests.lisp
appleby: These extra compiler hook tests have been removed now. More thorough tests is still a TODO, but I've at least added a new `app/tests/faithfulness-test-files/CNOT-partial-rewiring.quil`.

appleby:
Marking this as "Do Not Review" for now.

The changes in this PR appear to fix the original bug (namely running the "faithfulness" tests in `quilc-tests` with something other than `:naive` as the initial rewiring type). I've added a new test case to the `faithfulness-test-files` that demonstrates this.

However, in the process of writing additional tests, I came across some additional corner cases that need more investigation.

src/addresser/rewiring.lisp
appleby: Also note that if the entering and exiting rewiring must be the same length, then we could possibly do away with the above regex and speed things up (at the expense of being less forgiving in what we parse) by just splitting the string in half and peeling off outer parens and passing each half to `make-rewiring-from-string`.

app/src/entry-point.lisp
appleby: Everything I said above is wrong wrong wrong. Turns out the person who wrote this originally knew what they were doing and I was confused (go figure).

Basically, I was expecting that passing `:compress-qubits t` to `p-p-t-l-m` would result in a matrix of the same size regardless of whether the `parsed-program` you passed has been compiled or not. Wrong!

Long story short: `p-p-t-l-m` has to take the rewirings into account, which means you wind up with a matrix whose size depends on `(max largest-compressed-qubit-index (length (trim-rewiring rewiring))`.  If you run the program through `compress-qubits` first, then the trimmed rewiring might be as long as the highest qubit index in your original source program.

Uncompiled parsed-program's don't have rewirings attached, so `compress-qubits` will produce a matrix of the minimal size based on how many distinct qubit indices appeared in your original parsed-program.

tl;dr

This was correct as it was before: don't `compress-qubits` for the `original-matrix`, but do `compress-qubits` for the `processed-program-matrix`.

appleby:
> Marking this as "Do Not Review" for now.
> 
> The changes in this PR appear to fix the original bug (namely running the "faithfulness" tests in `quilc-tests` with something other than `:naive` as the initial rewiring type). I've added a new test case to the `faithfulness-test-files` that demonstrates this.
> 
> However, in the process of writing additional tests, I came across some additional corner cases that need more investigation.

These "additional corner cases" turned out to be a bug my understanding of how `:compress-qubits t` interacts with `parsed-program-to-logical-matrix` when operating on a `parsed-program` with rewirings attached. See my comment [here](https://github.com/rigetti/quilc/pull/375#discussion_r310744650) for more.

I have one more refactoring TODO to tackle, then I will mark this as ready for review.

src/addresser/rewiring.lisp
appleby: I just noticed this calculation in `entry-point.lisp` which seems to indicate that entering and exiting rewirings are at least expected to be the same length.

```lisp
(setf (gethash "final_rewiring" statistics)
      (read-from-string
       (subseq (comment instr) length (- (length comment)
                                         (/ (- (length comment) length) 2)
                                         2))))
```

If folks prefer, I could do away with the regex here and do some simpler/faster "parsing by string offset" here as well.

app/src/entry-point.lisp
```diff
@@ -523,58 +523,11 @@ Returns a values tuple (PROCESSED-PROGRAM, STATISTICS), where PROCESSED-PROGRAM
...
-                              (setf (comment (aref (parsed-program-executable-code processed-program)
-                                                   (1- j)))
-                                    (comment instr))))
-                     :unless (typep instr 'halt)
-                       :collect instr)
-               'vector))
-
-        (setf (gethash "topological_swaps" statistics)
-              topological-swaps)
+              (%strip-halts-respecting-rewirings processed-program))
```
appleby: N.B. we now do potentially two passes over `parsed-program-executable-code` to extract the `final_rewiring` and twiddle the `halt`s, whereas we used to do it in a single pass.

However, `extract-final-exit-rewiring-vector` is smart enough to loop over the instructions in reverse, and I assume that the common case is that `processed-program` terminates in with `halt` that has an exit rewiring attached.

app/src/entry-point.lisp
```diff
@@ -638,3 +591,38 @@ Returns a values tuple (PROCESSED-PROGRAM, STATISTICS), where PROCESSED-PROGRAM
                          (quil::lscheduler-calculate-depth lschedule)))))
 
       (values processed-program statistics))))
+
+(defun %strip-halts-respecting-rewirings (processed-program)
```
appleby: Please review the changes to this function with extra care. Part of the reason I split it out into a separate function was so that I could write tests to verify my refactoring didn't break things, but the tests are not exhaustive.

I recommend looking at the diff for https://github.com/rigetti/quilc/pull/375/commits/c57e884584b6426668ccb074f21fdc785dc29de3 to get an easier-to-eyeball apples-to-oranges comparison of the before and after.

Note in particular that I am now checking to make sure rewirings attached to a HALT and it's antecedent are "compatible", and will trigger an assertion if not.

app/tests/misc-tests.lisp
```diff
@@ -43,3 +43,121 @@
...
+             (let ((halt-instr (quil::nth-instr index pp)))
+               (assert (quil::haltp halt-instr))
+               (setf (quil:comment halt-instr)
+                     (quil::make-rewiring-comment :exiting exiting-vector)))
+
+             (when (and (plusp index) (not (null entering-vector)))
+               (setf (quil:comment (quil::nth-instr (1- index) pp))
+                     (quil::make-rewiring-comment :entering entering-vector))))))
+
+    ;; Multiple HALTs in a row are not handled correctly (but should "never" happen).
```
appleby: I am basing this assumption on the fact the the original `halt` stripping code did not handle this case.

app/src/entry-point.lisp
```diff
@@ -638,3 +591,38 @@ Returns a values tuple (PROCESSED-PROGRAM, STATISTICS), where PROCESSED-PROGRAM
...
+      :for instr :across (parsed-program-executable-code processed-program)
+      :for prev-instr := (and (plusp j) (quil::nth-instr (1- j) processed-program))
+      :when (and (plusp j) (typep instr 'halt) (comment instr)) :do
+        (multiple-value-bind (this-entering this-exiting) (quil::instruction-rewirings instr)
+          (multiple-value-bind (prev-entering prev-exiting) (quil::instruction-rewirings prev-instr)
+            (assert-rewirings-compatible ':ENTERING instr prev-instr this-entering prev-entering)
+            (assert-rewirings-compatible ':EXITING instr prev-instr this-exiting prev-exiting)
+            ;; Because of the (COMMENT INSTR) in the above :WHEN loop clause, we know that at least
+            ;; one of THIS-ENTERING and THIS-EXITING is non-NIL. Furthermore, due to the
+            ;; ASSERT-REWIRINGS-COMPATIBLE calls we know that either at most one the this/prev
```
appleby:
```suggestion
            ;; ASSERT-REWIRINGS-COMPATIBLE calls we know that either at least one the this/prev
```

src/ast.lisp
```diff
@@ -1313,6 +1431,80 @@ For example,
...
+  (check-type parsed-program parsed-program)
+  (let* ((code (parsed-program-executable-code parsed-program))
+         (length (length code)))
+    (%bounds-check-bidirectional-index index length)
+    (funcall continuation code (%normalize-bidirectional-index index length))))
+
+(defun nth-instr (index parsed-program)
+  "Return the INSTRUCTION at position INDEX in PARSED-PROGRAM's EXECUTABLE-PROGRAM vector.
+
+INDEX can be a \"bidirectional\" index. That is any integer value in the range [-length, length), where length is (LENGTH (PARSED-PROGRAM-EXECUTABLE-CODE PARSED-PROGRAM)).
```
appleby: Another, perhaps "lispier", choice is to only allow non-negative INDEX, but accept a `:from-end` keyword option.

app/src/entry-point.lisp
ecpeterson: indeed, the only case: within the `protoquil` guard, there can't be any JUMPs, LABELs, ... .

src/addresser/rewiring.lisp
```diff
@@ -271,3 +271,41 @@ BODY as an implicit PROGN."
                      (error "Malformed rewiring string: unexpected token ~a" token))))
                 tokens)))
     (make-rewiring-from-l2p integer-vec)))
+
+(defun make-rewiring-pair-from-string (str)
```
ecpeterson: <3

src/ast.lisp
```diff
@@ -158,6 +158,120 @@ The keys are typically INSTRUCTION instances and associated values are STRINGs."
...
+If REWIRING-STRING does not have a valid rewiring comment prefix, signal an error."
+  (cond ((comment-entering-rewiring-p rewiring-string)
+         ':ENTERING)
+        ((comment-exiting-rewiring-p rewiring-string)
+         ':EXITING)
+        ((comment-entering/exiting-rewiring-p rewiring-string)
+         ':ENTERING/EXITING)
+        (t (error "Invalid rewiring comment: ~S" rewiring-string))))
+
+(defun make-rewiring-comment (&key entering exiting)
```
ecpeterson: Alternatively (to much of this file), `rewiring-comment` could actually be a struct that stores the tag and the two vectors directly, and then you teach the printer how to turn it into the appropriate string. I don't think we often make the assumption that comments are literally strings—just that they're printable objects.

src/ast.lisp
```diff
@@ -350,6 +464,10 @@ numbers. It must start with a string.")
   #+#:appleby-sufficiently-classy
```
ecpeterson: :)

src/ast.lisp
ecpeterson: This alternative would be my preference, I think, but I also wouldn't stand on it.

src/cfg.lisp
```diff
@@ -525,16 +525,15 @@ Return the following values:
...
-               (format nil "Entering/exiting rewiring: (~a . ~a)"
-                       (rewiring-l2p (basic-block-in-rewiring blk))
-                       (rewiring-l2p (basic-block-out-rewiring blk)))))
+               (make-rewiring-comment :entering (basic-block-in-rewiring blk)
+                                      :exiting (basic-block-out-rewiring blk))))
         (t
          (when (basic-block-in-rewiring blk)
            (setf (comment (aref code-section 0))
-                 (format nil "Entering rewiring: ~a" (rewiring-l2p (basic-block-in-rewiring blk)))))
+                 (make-rewiring-comment :entering (basic-block-in-rewiring blk))))
```
ecpeterson: 👍

app/src/entry-point.lisp
ecpeterson: I suspect this assumption would greatly simplify the private method below.

appleby: Yes, that would absolutely simplify things if we can be sure there is at most one HALT stuck on the end.  I realize now that maybe the original code to strip the HALT was just piggy-backing on the pre-existing loop to extract the "final_rewiring", but I assumed it was because there might be multiple HALTs.

src/ast.lisp
appleby: I think I agree. I will make this change. No sense trying to turn lisp into Python 🐍

src/ast.lisp
appleby: This is a good idea but unless you feel it's a blocker, I'll add it to the backlog, mostly just so I can move on to Actual Sprint Commitments (tm).

I did consider using some first-class datatype for comments, but went with a minimal gloss over the existing strings/vectors instead.

src/ast.lisp
appleby: This is done. Changed it take a `:from-end` keyword in the latest commit.

src/ast.lisp
appleby: 🎩 
🍎

app/src/entry-point.lisp
appleby: I take it back. I implemented this change, and it actually made the `%strip-halts...` function slightly longer, not shorter, due to the fact that many of the edge cases fall out of the loop conditions naturally, whereas they have to be explicitly tested if not looping (is the code empty?, does it have a single instruction?, is the final instruction a HALT?, etc).

It's also not really any faster to avoid the loop, since `parsed-program-executable-code` is a `simple-vector`, so you wind up having to copy instructions anyway just to peel off the last one.  What this refactoring *would* do is make the associated tests much simpler to assume it only has to test cases with a single HALT at the end...

But I'm inclined to leave it as is, since it the refactoring doesn't make the code much shorter/faster, and the current version handles multiple HALTs, in case that is ever needed.

src/ast.lisp
appleby: Filed #381

ecpeterson: I don't think this is a blocker. `'pragma` was around for ~2 years before someone thought to subclass it.

src/ast.lisp
```diff
@@ -158,6 +158,120 @@ The keys are typically INSTRUCTION instances and associated values are STRINGs."
   (check-type comment-string string)
   (setf (gethash x **comments**) comment-string))
 
+(a:define-constant +entering-rewiring-prefix+
```
stylewarning: I like this refactor, but this in/out rewiring stuff is absurdly complicated for something seemingly so simple.

appleby: I think @ecpeterson's idea [here](https://github.com/rigetti/quilc/pull/375#discussion_r312027757) to make rewiring comments a struct with a `print-object` method, rather than always converting to/from strings and vectors would simplify things a fair bit, which is now #381.

stylewarning: Well still, the fact that comment data affects semantics in various ways is troubling.

PR #377, opened 2019-08-06
quilt parsing

kilimanjaro:
Opening up for review. This extends `cl-quil` with support for `quilt` instructions.

Since this is such a big PR, I'll briefly comment on the changes.

- Add AST objects for new quilt instructions and primitives (e.g. frames, waveforms)
- Extend `parser.lisp` to support quilt (this includes a few minor quality of life improvements as well)
- Refactor the high level behavior of `parse-quil`. Here we process includes for a list of "raw" quil instructions, then parse to an "unresolved program", and then resolve objects. 
- When converting raw quil to an unresolved program, we check and handle ambiguous definitions.
- Object resolution (previously application resolution) now includes things like frames and waveforms, which must have their definitions resolved.
- `expand-calibrations` is a new pass which uses much of the circuit expansion machinery (which has been moved to `expansion.lisp`). The behavior here is to produce simple quilt with all gate applications, measurements, etc resolved.
- The `type-check` pass has been extended to verify memory references in `CAPTURE` and `RAW-CAPTURE`
- An additional pass, `fill-delays`, can turn a quilt program with "empty" time intervals and return a program with an entire block of time accounted for (via additional `DELAY` instructions). Some of the machinery in here is written to also be useful for the pulse tracing qvm.

Anticipated remaining work:
-  the primary language feature that is left unaddressed here is how to support affine transformations of waveforms (or the equivalent, so that one can `CAPTURE` on a "projected" iq value). This will be a separate PR.
- quilc app may need some additional work on error reporting (e.g. a nice message if unexpected quilt instructions show up)


src/parser.lisp
```diff
@@ -1390,6 +1320,353 @@ result of BODY, and the (possibly null) list of remaining lines.
...
+When ALLOW-EXPRESSIONS is set, we allow for general arithmetic expressions in a parameter context."
+  (unless (eql ':LEFT-PAREN (token-type (first params-args)))
+    (return-from parse-parameters (values nil params-args)))
+
+  ;; Remove :LEFT-PAREN
+  (pop params-args)
+
+  ;; Parse out the parameters enclosed.
+  (multiple-value-bind (found-params rest-line)
+      (take-while-from-end (lambda (x) (eql ':RIGHT-PAREN (token-type x)))
```
appleby: I know you didn't write it, but `take-until-from-end` seems like a less confusing name for this function.

src/parser.lisp
```diff
@@ -1390,6 +1320,353 @@ result of BODY, and the (possibly null) list of remaining lines.
...
+  (when (null tok-lines)
+    (quil-parse-error "EOF reached when waveform definition expected"))
+
+  ;; Get the parameter and body lines
+  (let (name)
+    (destructuring-bind (parameter-line &rest body-lines) tok-lines
+      (destructuring-bind (op . params-args) parameter-line
+        ;; Check that we are dealing with a DEFWAVEFORM.
+        (unless (eql ':DEFWAVEFORM (token-type op))
+          (quil-parse-error "DEFWAVEFORM expected. Got ~S"
```
appleby: `disappointing-token-error`?

src/ast.lisp
```diff
@@ -381,6 +483,101 @@ as the reset is formally equivalent to measuring the qubit and then conditionall
...
+       (defclass ,name (simple-frame-mutation)
+         ()
+         (:documentation ,docstring))
+
+       (defmethod mnemonic ((inst ,name)) (values ',mnemonic ',name)))))
+
+(defmethod arguments ((instr simple-frame-mutation))
+  (with-slots (qubits frame value)
+      instr
+    (coerce (list frame value) 'vector)))
```
appleby: `(vector frame value)`?

src/ast.lisp
```diff
@@ -381,6 +483,101 @@ as the reset is formally equivalent to measuring the qubit and then conditionall
...
+
+(defmacro define-simple-frame-mutation (name mnemonic &body body)
+  (check-type mnemonic string)
+  (check-type body list)
```
appleby: There's a bit of a mismatch and some overlap between the `check-type` here (only `mnemonic`) and in `expand-simple-frame-mutation-definition` (`name` and `mnemonic`). Could probably just check in `expand-simple-frame-mutation-definition`.

Also, no real need to `(check-type body list)`.

src/ast.lisp
```diff
@@ -381,6 +483,101 @@ as the reset is formally equivalent to measuring the qubit and then conditionall
...
+(eval-when (:compile-toplevel :load-toplevel :execute)
+  (defun expand-simple-frame-mutation-definition (name mnemonic docstring)
+    (check-type name symbol)
+    (check-type mnemonic string)
+    `(progn
+       (defclass ,name (simple-frame-mutation)
+         ()
+         (:documentation ,docstring))
+
+       (defmethod mnemonic ((inst ,name)) (values ',mnemonic ',name)))))
```
appleby: No need to quote if `mnemonic` is a `string`

src/ast.lisp
```diff
@@ -1201,6 +1409,38 @@ For example,
...
 
+  (:method  ((instr simple-frame-mutation) (stream stream))
+    (format stream "~A~{ ~A~}"
+            (mnemonic instr)
+            (map 'list #'print-instruction-to-string (arguments instr))))
+
+  (:method ((instr pulse) (stream stream))
+    (format stream "PULSE ~A ~A"
+            (print-instruction-generic (pulse-frame instr) nil)
+            (print-instruction-generic (pulse-waveform instr) nil)))
```
appleby:
```lisp
(format stream "PULSE ~/cl-quil::instruction-fmt/ ~/cl-quil::instruction-fmt/"
        (pulse-frame instr)
        (pulse-waveform instr))
```

ditto elsewhere

src/ast.lisp
```diff
@@ -1289,7 +1529,55 @@ For example,
...
+                          (number
+                           (format-complex z s))
+                          ((or list symbol)
+                           (print-instruction (make-delayed-expression nil nil z) s)))))
+                    (waveform-definition-entries thing))))
+
+  (:method ((defn gate-calibration-definition) (stream stream))
+    (format stream "DEFCAL ~a" (calibration-definition-name defn))
+    (flet ((print-thing (thing)
+             (print-instruction thing nil)))
```
appleby: `print-thing` can be replaced with `print-instruction-to-string`. Or maybe `~/cl-quil::instruction-fmt/`.

src/ast.lisp
kilimanjaro: For this sort of thing I have a slight preference for keeping the format string simple and keeping the work in ordinary lisp-land.

appleby: Fair enough. I'm not bothered personally. I'm just pattern matching on other recent quilc review comments where there has been a push to switch everything to `~/cl-quil::instruction-fmt/` but piecemeal as they come up in review, rather than in one fell swoop.

src/analysis/defence.lisp
```diff
@@ -0,0 +1,80 @@
...
+
+
+(defun expand-fences-to-delays (parsed-program &optional sample-rate)
+  "Expand FENCE instructions to corresponding DELAY instructions on each relevant
+qubit line. For custom waveform definitions, SAMPLE-RATE (in Hz) must be provided."
+  (let ((qubit-clocks (make-hash-table)) ; indexed by qubit indices i.e. numbers
+        new-instrs)
+    (loop :for instr :across (parsed-program-executable-code parsed-program)
+          :for qubits := (quilt-instruction-qubits instr)
+          :when qubits
```
appleby: Did you want `typecase` rather than `etypecase` in `quilt-instruction-qubits`?

src/analysis/expansion.lisp
```diff
@@ -1,12 +1,21 @@
...
   '(member nil :DEFCIRCUIT :DEFCAL))
 
 (defparameter *expansion-context* nil
   "The context for expansion, mainly used for error messages.")
 
+(defun expansion-error (format-control &rest format-args)
+  "Signal a QUIL-PARSE-ERROR, incorporating information about the expansion context."
+  (quil-parse-error "While expanding ~A: ~A"
+                    (or *expansion-context* "circuit or calibration")
+                    (apply #'format nil format-control format-args)))
```
appleby: I usually forget about it as well, but `~?` is a [fun little `format` snack](http://www.lispworks.com/documentation/HyperSpec/Body/22_cgf.htm):

```
(quil-parse-error "While expanding ~A: ~?" 
                  (or *expansion-context* "circuit or calibration")
		  format-control
		  format-arguments)
```

kilimanjaro: Nifty!

src/analysis/expand-calibrations.lisp
```diff
@@ -0,0 +1,181 @@
...
+(defgeneric apply-calibration (instr gate-cals measure-cals measure-discard-cals)
+  (:documentation "If INSTR has an associated calibration, return a list of instructions instantiated from the body of the calibration definition. Otherwise, return NIL.")
+
+  (:method ((instr gate-application) gate-cals measure-cals measure-discard-cals)
+    (declare (ignore measure-cals measure-discard-cals))
+    (let ((op (application-operator instr)))
+      (unless (plain-operator-p op)
+        (return-from apply-calibration))
+
+      (a:if-let ((defn (find-if (lambda (defn) (calibration-matches-p defn instr))
```
appleby: `a:when-let` ? Ditto in the other methods just below.

src/analysis/expand-calibrations.lisp
```diff
@@ -0,0 +1,181 @@
...
+  (:method ((instr measure-discard) gate-cals measure-cals measure-discard-cals)
+    (declare (ignore gate-cals measure-cals))
+    (a:if-let ((defn (find-if (lambda (defn) (calibration-matches-p defn instr))
+                              measure-discard-cals)))
+      (instantiate-definition defn
+                              nil
+                              (list (measurement-qubit instr)))
+      nil))
+
+  (:method ((instr t) gate-cals measure-cals measure-discard-cals)
```
appleby: nitpick: no need for `t` specializer

src/analysis/expand-calibrations.lisp
```diff
@@ -0,0 +1,181 @@
...
+      ;; Recursively expand the new instructions
+      (a:mappend (lambda (instr)
+                   (recursively-expand-instruction instr
+                                                   gate-cals
+                                                   measure-cals
+                                                   measure-discard-cals))
+                 expanded)
+      ;; Otherwise, handle the base case.
+      (if (and *require-applicable-calibration*
+               (null expanded)
```
appleby: `expanded` guaranteed to be `null` here, no?

src/analysis/expand-calibrations.lisp
```diff
@@ -0,0 +1,181 @@
...
+
+(defun expand-calibrations (parsed-program)
+  "Expand all gate applications and measurements in PARSED-PROGRAM for which there is a corresponding calibration definition."
+  (multiple-value-bind (gate-cals
+                        measure-cals
+                        measure-discard-cals)
+      (compute-calibration-tables parsed-program)
+    (let ((*expansion-context* ':DEFCAL)
+          (*expansion-depth* 0))
+      (let ((expanded
```
appleby: Ditto here re: smushed `let*`

src/analysis/expand-calibrations.lisp
kilimanjaro: `when` would do the job, but I suppose I opted for `if` because in most usage, `when` and `unless` are called for side-effecting behavior, and I want to be explicit here that I care about the `nil` return value.

src/analysis/expand-calibrations.lisp
kilimanjaro: The aesthetic reason for having two `let` statements is that the outer one is setting up special variables for the computation to happen within. Switching to `let*` works but with the demarcation between setup and computation less clear (although one less indent is nice)

src/analysis/expand-circuits.lisp
```diff
@@ -7,323 +7,20 @@
...
-                   :if (listp instantiated)
-                     :append instantiated
-                   :else
-                     :collect instantiated)))
-      (setf (parsed-program-executable-code parsed-program)
-            (coerce flat-instrs 'vector))
-      parsed-program)))
+    (let ((*expansion-context* ':DEFCIRCUIT)
+          (*expansion-depth* 0))
+      (let ((flat-instrs
```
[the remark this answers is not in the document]
kilimanjaro: keeping w/ the above comment, i'm fine with either approach (will let the heavier voices weigh in on this)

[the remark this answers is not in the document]
appleby: either if fine by me as well. I just thought it might have slipped by unnoticed from some intermediate refactoring pass.

src/analysis/expand-calibrations.lisp
appleby: totally reasonable

stylewarning: not only reasonable, but preferable!

appleby: 4 out of 5 @stylewarnings approve

src/analysis/expansion.lisp
```diff
@@ -0,0 +1,447 @@
...
+An instruction is unitary if it is of type APPLICATION, whether that be INSTR itself or, if INSTR is a CIRCUIT-APPLICATION, the body of the circuit. With the caveat that UNRESOLVED-APPLICATIONS must be explicitly allowed by setting *ALLOW-UNRESOLVED-APPLICATIONS* to T."
+  (etypecase instr
+    (gate-application
+     t)
+    (circuit-application
+     (every #'unitary-instruction-p
+            (circuit-definition-body (circuit-application-definition instr))))
+    (unresolved-application
+     (unless *allow-unresolved-applications*
+       (error "Unable to determine if the unresolved application is unitary:~%    ~A"
```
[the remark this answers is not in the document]
notmgsk: You didn't hear this from me, but he called the formatter ugly.

src/analysis/expand-calibrations.lisp
```diff
@@ -0,0 +1,181 @@
...
+(defun compute-calibration-tables (parsed-program)
+  "Extract the calibration definitions from PARSED-PROGRAM. Returns three values:
+
+1. a hash table of gate calibrations (keyed by name),
+
+2. a list of measure calibrations,
+
+3. a list of measure discard calibrations."
+  (let ((gate-calibrations
+          (make-hash-table :test 'equalp)) ; EQUALP b/c this table is keyed by OPERATOR-DESCRIPTIONs
```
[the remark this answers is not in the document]
kilimanjaro: whipping one up here, but the general feature is also a candidate for a `cl-adt` pr

src/analysis/expand-calibrations.lisp
```diff
@@ -0,0 +1,181 @@
...
+                             gate-calibrations)))
+        (measure-calibration-definition
+         (push defn measure-calibrations))
+        (measure-discard-calibration-definition
+         (push defn measure-discard-calibrations))))
+    (values gate-calibrations
+            measure-calibrations
+            measure-discard-calibrations)))
+
+(defgeneric apply-calibration (instr gate-cals measure-cals measure-discard-cals)
```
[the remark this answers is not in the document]
kilimanjaro: how about `instantiate-applicable-calibration`

src/analysis/fill-delays.lisp
```diff
@@ -0,0 +1,205 @@
...
+
+(defun frame-intersects-p (frame qubits)
+  "Does the FRAME involve any of the specified QUBITS?"
+  (intersection qubits
+                (frame-qubits frame)
+                :test #'equalp))
+
+(defun frame-on-p (frame qubits)
+  "Does FRAME involve exactly the specified QUBITS?"
+  (equalp qubits (frame-qubits frame)))
```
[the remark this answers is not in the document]
kilimanjaro: yes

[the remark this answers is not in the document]
kilimanjaro: changing docstring to clarify

src/analysis/expansion.lisp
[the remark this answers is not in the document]
kilimanjaro: How should I use the formatter?

src/analysis/expansion.lisp
```diff
@@ -0,0 +1,447 @@
...
+
+(deftype expansion-context ()
+  '(member nil :DEFCIRCUIT :DEFCAL))
+
+(defparameter *expansion-context* nil
+  "The context for expansion, mainly used for error messages.")
+
+(defun expansion-error (format-control &rest format-args)
+  "Signal a QUIL-PARSE-ERROR, incorporating information about the expansion context."
+  (quil-parse-error "While expanding ~A: ~?"
```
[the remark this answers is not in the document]
kilimanjaro: adding a new `quil-expansion-error` condition

src/ast.lisp
kilimanjaro: I'm warming up to formatters the more I think about it, although their use imo makes things less readable in some contexts. For example, in `print-instruction-generic` I think the nicest thing is that the printing of an instruction be controlled by the equivalent of a single format string that looks like a templated form of the actual instruction (e.g. with `"PULSE ~A ~A"`). Using `~/cl-quil:instruction-fmt/` is compatible with that, except if you use it a few times the line gets quite wide and it's not so easy on the eyes. Perhaps this is less annoying to others though.

kilimanjaro: One option mentioned by @stylewarning is to use http://edicl.github.io/cl-interpol/ instead. Not sure if it's worth it in this context; the basic tradeoff is that we can easily call `print-instruction-to-string`, but these string allocations might impose too severe of a runtime cost on either qvm or quilc (though I'm doubtful in both cases).

src/waveform.lisp
```diff
@@ -0,0 +1,127 @@
...
+   (t0 :quilt-name "t0"
+       :type float
+       :documentation "Center time coordinate of the shape in seconds. Defaults
+       to mid-point of pulse.")
+   (anh :quilt-name "anh"
+        :type float
+        :documentation "Anharmonicity of the qubit, f01-f12 in (Hz)")
+   (alpha :quilt-name "alpha"
+          :type float
+          :documentation "Dimensionless DRAG parameter"))
```
stylewarning: love dimensionless `<thing>` parameters

src/waveform.lisp
```diff
@@ -0,0 +1,127 @@
...
+
+;;; These are produced from the similar definitions in RPCQ.
+
+(define-standard-waveform gaussian-waveform "gaussian"
+  ((fwhm :quilt-name "fwhm"
+         :type float
+         :documentation "Full Width Half Max shape parameter, in seconds.")
+   (t0 :quilt-name "t0"
+       :type float
+       :documentation "Center time coordinate of the shape in seconds. Defaults
```
stylewarning: newlines in doc strings are evil

src/waveform.lisp
```diff
@@ -0,0 +1,127 @@
...
+
+    `(progn
+       (defclass ,class-name (standard-waveform)
+         ,slot-specs
+         ,@(when documentation
+            `((:documentation ,documentation))))
+
+       (setf (gethash ,quilt-name *quilt-to-waveform-class*) ',class-name)
+
+       (defmethod quilt-waveform-parameter-alist ((class-name (eql ',class-name)))
```
stylewarning: I think I'd prefer this to be stashed away in a hash table so that this could be a function, but this is pretty simple, even if it's using a tank to drive a nail

src/ast.lisp
appleby: `~/cl-quil:instruction-fmt/` is a bit of mouthful and I agree with your point that single-letter built-in format controls are more readable in cases like this one.

As for cl-interpol, I'm a fan of edi weitz packages in general, but mildly opposed to adding another dependency to cl-quil, especially for something as fringe as string interpolation.

src/analysis/expansion.lisp
```diff
@@ -0,0 +1,452 @@
...
+         (instantiate-definition (circuit-application-definition instr)
+                                 params
+                                 args))
+        (t
+         (error "Unable to instantiate the modifiers in the complex instruction ~S."
+                (print-instruction instr nil))))))
+
+  (:method ((instr application) param-value arg-value)
+    (let ((remake nil))
+      (let ((params (mapcar (transform-if (constantly t) (substitute-parameter param-value))
```
appleby:
```suggestion
      (let ((params (mapcar (substitute-parameter param-value)
```

I might be missing something here

src/analysis/expansion.lisp
```diff
@@ -0,0 +1,452 @@
...
+                            (application-parameters instr)))
+            (args (mapcar (transform-if #'is-formal arg-value)
+                          (application-arguments instr))))
+        (setf remake t)
```
appleby: `remake` no longer needed?

tests/quilt-tests.lisp
```diff
@@ -0,0 +1,330 @@
...
+                (signature "DEFWAVEFORM foo 1.0:~%    1.0, 1.0")))
+    (is (not (equalp (signature "DEFWAVEFORM foo 1.0:~%    1.0, 1.0")
+                     (signature "DEFWAVEFORM Foo 1.0:~%    1.0, 1.0"))))
+    (is (equalp (signature "DEFCAL RX(%theta) q:~%    NOP")
+                (signature "DEFCAL RX(%theta) q:~%    NOP")))
+    (is (equalp (signature "DEFCAL RX(0) 0:~%    NOP")
+                (signature "DEFCAL RX(0) 0:~%    NOP")))
+    ;; argument names don't matter
+    (is (equalp (signature "DEFCAL RX(0) s:~%    NOP")
+                (signature "DEFCAL RX(0) s:~%    NOP")))
```
appleby: Are the argument names here meant to differ?

tests/printer-test-files/gold-standard/plundered-gold.quil
```diff
@@ -984,8 +984,8 @@ CONTROLLED Y 0 1
 CONTROLLED CONTROLLED Y 0 1 2
 CONTROLLED CONTROLLED CONTROLLED Y 0 1 2 3
 DAGGER Y 0
-Y 0
-DAGGER Y 0
+DAGGER DAGGER Y 0
+DAGGER DAGGER DAGGER Y 0
```
appleby:
```

                                 _____  _____
                                <     `/     |
                                 >          (
                                |   _     _  |
                                |  |_) | |_) |
                                |  | \ | |   |
                                |            |
                 ______.______%_|            |__________  _____
               _/                                       \|     |
              |             INVOLUTIVE-DAGGER-OPERATOR         <
              |_____.-._________              ____/|___________|
                                |            |
                                | + 07/10/19 |
                                |            |
                                |            |
                                |   _        <
                                |__/         |
                                 / `--.      |
                               %|            |%
                           |/.%%|          -< @%%%
                           `\%`@|     v      |@@%@%%    - mfj
                         .%%%@@@|%    |    % @@@%%@%%%%
                    _.%%%%%%@@@@@@%%_/%\_%@@%%@@@@@@@%%%%%%
```

src/analysis/expansion.lisp
kilimanjaro: Good catch. Actually, I'm keeping it in (but in a way that actually uses it...).

src/analysis/expansion.lisp
```diff
@@ -0,0 +1,452 @@
...
+(define-condition quil-expansion-error (simple-error)
+  ()
+  (:documentation "Representation of an error arising in Quil circuit or calibration expansion."))
+
+(defun quil-expansion-error (format-control &rest format-args)
+  "Signal a QUIL-EXPANSION-ERROR, incorporating information about the expansion context."
+  (error 'quil-expansion-error :format-control (format nil "While expanding ~A: ~A"
+                                                       (or *expansion-context* "circuit or calibration")
+                                                       format-control)
+                               :format-arguments format-args))
```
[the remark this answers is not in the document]
kilimanjaro: I was being unreasonably lazy w/ my use of `format`, so I prefer yours.  Re: `*expansion-context*` I'm just going to replace it with `(and (boundp '*expansion-context*) *expansion-context*)` here. I think for the purpose of reporting an error we should be lenient about it (e.g. we might want to call `quil-expansion-error` in a place where there is no expectation that the context has been set).

tests/quilt-tests.lisp
kilimanjaro: It was a test! I was testing you! Now that we've settled that I'm... uh... just going to quietly change this.

tests/printer-test-files/gold-standard/plundered-gold.quil
kilimanjaro: If anything I am _increasing_ the number of annihilated `involutive-dagger-operator`!

tests/quilt-tests.lisp
appleby: One day a master painter and an apprentice were painting a house for a very persnickety customer. The type that would stand over their shoulder while they worked and remark: "you missed a spot", etc.

They were nearing completion of the job and the apprentice was just about to paint the final column. "Leave that unpainted," said the master. "Why?", asked the apprentice. The master did not reply.

Later that day, the customer sent word to the painters: you forgot to paint the final column! How could you make such an obvious error! "We're very sorry," said the master, "I'll send someone right away to finish the job." So he sent the apprentice. When the apprentice returned, he asked the master: "why did you tell me to leave that column unpainted? It's obvious that this customer, who is exacting, would notice and demand that we finish!" The master replied: "the persnickety customer will always find fault with the job, no matter what. If we hadn't left something obvious for them to find, who knows what difficult thing they would have made us undo!"

And with that, the apprentice was enlightened.

notmgsk: "Well they refused to pay us for breach of contract," snapped the apprentice.

"Damn, dude. I was super baked," spake master.

appleby: :gong:

src/analysis/fill-delays.lisp
```diff
@@ -0,0 +1,206 @@
...
+     (delay-on-qubits
+      (loop :for defn :in (parsed-program-frame-definitions parsed-program)
+            :for frame := (frame-definition-frame defn)
+            :when (equalp (frame-qubits frame)
+                          (delay-qubits instr))
+              :collect frame))
+     (fence
+      (loop :for defn :in (parsed-program-frame-definitions parsed-program)
+            :for frame := (frame-definition-frame defn)
+            :when (intersection (frame-qubits frame)
```
appleby: `frame-intersects-p` ?

src/analysis/fill-delays.lisp
```diff
@@ -0,0 +1,206 @@
...
+     (capture (capture-frame instr))
+     (raw-capture (raw-capture-frame instr))
+     (delay-on-frames (delay-frames instr))
+     (simple-frame-mutation (frame-mutation-target-frame instr))
+     (swap-phase (swap-phase-left-frame instr) (swap-phase-right-frame instr))
+     ;; delay affects all frames on precisely the delay qubits
+     (delay-on-qubits
+      (loop :for defn :in (parsed-program-frame-definitions parsed-program)
+            :for frame := (frame-definition-frame defn)
+            :when (equalp (frame-qubits frame)
```
appleby: `frame-on-p` ?

src/analysis/fill-delays.lisp
```diff
@@ -0,0 +1,206 @@
...
+                                    (frame-intersects-p f (fence-qubits instr)))
+                                  (tracked-frames clocks)))
+           (latest (latest-time clocks frames)))
+      (loop :for f :in frames
+            :for lag := (- latest (local-time clocks f))
+            ;; handle implicit delays
+            :when (plusp lag)
+                  :do (incf (local-time clocks f) lag)
+            :when (plusp lag)
+              :collect (make-instance 'delay-on-frames
```
appleby: `:and :collect`

src/analysis/fill-delays.lisp
```diff
@@ -0,0 +1,206 @@
...
+
+(defgeneric emit-delays (instr clocks)
+  (:documentation "Update the local times on CLOCKS according to the specified INSTR, returning a list of implicit DELAYs.")
+
+  (:method ((instr delay-on-frames) clocks)
+    (dolist (f (delay-frames instr))
+      (incf (local-time clocks f)
+            (delay-duration instr)))
+    ;; no implicit delays
+    nil)
```
appleby: kids these days and their dangling-nil lifestyles... where will it lead us?

src/analysis/fill-delays.lisp
```diff
@@ -0,0 +1,206 @@
...
+
+  (:method ((instr swap-phase) clocks)
+    (with-slots (left-frame right-frame) instr
+      (let ((latest (max (local-time clocks left-frame)
+                         (local-time clocks right-frame))))
+        (loop :for f :in (list left-frame right-frame)
+              :for lag := (- latest (local-time clocks f))
+              :when (plusp lag)
+                :do (incf (local-time clocks f) lag)
+                :and :collect (make-instance 'delay-on-frames )))))
```
appleby:
```suggestion
                :and :collect (make-instance 'delay-on-frames 👻)))))
```

spooky 👻 args

src/analysis/fill-delays.lisp
```diff
@@ -0,0 +1,206 @@
...
+  (typecase instr
+    (pulse (pulse-frame instr))
+    (capture (capture-frame instr))
+    (raw-capture (raw-capture-frame instr))))
+
+(defun frame-intersects-p (frame qubits)
+  "Does the FRAME involve any of the specified QUBITS?"
+  (intersection qubits
+                (frame-qubits frame)
+                :test #'eql
```
stylewarning: `#'=`

src/analysis/fill-delays.lisp
```diff
@@ -0,0 +1,206 @@
...
+                  :when (plusp lag)
+                    :do (incf (local-time clocks f) lag)
+                    :and :collect (make-instance 'delay-on-frames
+                                                 :frames (list f)
+                                                 :duration (constant lag))))))))
+
+(defun fill-delays (parsed-program &key (omit-fences t) (synchronize-at-end t))
+  "Introduce any implicit DELAY instructions in the quilt program PARSED-PROGRAM.
+
+If OMIT-FENCE is T, then FENCE instructions will be removed from the resulting program.
```
stylewarning: please note in the doc string that these are defaults: `If OMIT-FENCE is T (default), ...`

src/analysis/fill-delays.lisp
```diff
@@ -0,0 +1,206 @@
...
+               (quil-parse-error "Cannot resolve timing information for non-quilt instruction ~A" instr))
+             ;; Synchronization is not needed with DELAYs
+             (unless (and omit-fences (typep instr 'fence))
+               (push instr new-instrs))
+             (dolist (delay (emit-delays instr frame-clocks))
+               (push delay new-instrs))))
+      (loop :for instr :across (parsed-program-executable-code parsed-program)
+            :do (process-instr instr))
+      (when synchronize-at-end
+        (let ((all-qubits (reduce #'union
```
stylewarning: `(reduce f x :key g)` instead of `(reduce f (mapcar g ...))`

src/analysis/type-safety.lisp
```diff
@@ -330,6 +344,47 @@
...
+  (:method ((instr capture) memory-regions)
+    (let* ((mref (capture-memory-ref instr))
+           (mdesc (find-descriptor-for-mref mref memory-regions)))
+      (enforce-mref-bounds mref mdesc)
+      (adt:match quil-type (memory-descriptor-type mdesc)
+        (quil-real
+         (if (> 2 (mref-available-length mref mdesc))
+             (quil-type-error "CAPTURE instruction target ~A must be a REAL ~
+                               vector of length no less than 2."
+                              (print-instruction-to-string mref))
```
stylewarning: prefer the format string

src/analysis/type-safety.lisp
```diff
@@ -330,6 +344,47 @@
...
+             t))
+        (_
+         (quil-type-error "CAPTURE instruction target must be of type ~
+                           REAL, but got ~A of type ~A"
```
stylewarning: same here

src/analysis/type-safety.lisp
```diff
@@ -330,6 +344,47 @@
...
+           (mdesc (find-descriptor-for-mref mref memory-regions))
+           (frame-defn (frame-name-resolution
+                        (raw-capture-frame instr))))
+      (enforce-mref-bounds mref mdesc)
+      (adt:match quil-type (memory-descriptor-type mdesc)
+        (quil-real
+         (a:if-let ((samples (raw-capture-num-real-samples instr)))
+           (if (> samples (mref-available-length mref mdesc))
+               (quil-type-error "RAW-CAPTURE instruction target ~A must be a REAL ~
+                                 vector of length no less than ~A."
```
stylewarning: same here

src/ast.lisp
```diff
@@ -127,13 +127,59 @@ EXPRESSION should be an arithetic (Lisp) form which refers to LAMBDA-PARAMS."
...
           de))))
 
-
 (defun map-de-params (f de)
   "Create a new DELAYED-EXPRESSION from DE, applying F to all of the parameters of DE."
   (let ((c (copy-structure de)))
     (setf (delayed-expression-params c) (mapcar f (delayed-expression-params de)))
     c))
 
+(defstruct (frame (:constructor frame (qubits name)))
```
stylewarning: can we organize this file so that the Quilt extensions are clearly delineated? I'm not saying there has to be just one giant Quilt section, and I'd even be happy with comments above the structs and functions that are Quilt specific, e.g.,

```
;;; Quilt
(defstruct (frame ...

;;; Quilt
(defun frame=
...
```

I want "pure quil" and "quilt" to not get too wrapped up in one another.

src/ast.lisp
```diff
@@ -127,13 +127,59 @@ EXPRESSION should be an arithetic (Lisp) form which refers to LAMBDA-PARAMS."
...
+  "A reference to a QuilT rotating frame, relative to which control or readout waveforms may be defined."
+  (name nil :type string)
+  (qubits nil :type list)
+  ;; Will later be resolved
+  (name-resolution nil :type (or null frame-definition)))
+
+(defun frame= (a b)
+  (and (string= (frame-name a) (frame-name b))
+       (equalp (frame-qubits a) (frame-qubits b))))
```
stylewarning: if you add `qubit=` you'll need to change this

src/ast.lisp
```diff
@@ -127,13 +127,59 @@ EXPRESSION should be an arithetic (Lisp) form which refers to LAMBDA-PARAMS."
...
-
 (defun map-de-params (f de)
   "Create a new DELAYED-EXPRESSION from DE, applying F to all of the parameters of DE."
   (let ((c (copy-structure de)))
     (setf (delayed-expression-params c) (mapcar f (delayed-expression-params de)))
     c))
 
+(defstruct (frame (:constructor frame (qubits name)))
+  "A reference to a QuilT rotating frame, relative to which control or readout waveforms may be defined."
+  (name nil :type string)
```
stylewarning: can the first two be `read-only`

src/ast.lisp
```diff
@@ -127,13 +127,59 @@ EXPRESSION should be an arithetic (Lisp) form which refers to LAMBDA-PARAMS."
...
+  ;; Will later be resolved
+  (name-resolution nil :type (or null
+                                 standard-waveform
+                                 waveform-definition)))
+
+(defun waveform-ref (name &rest plist)
+  "Construct a waveform reference with keyword-value pairs given by ARGS."
+  (assert (evenp (length plist)))
+  (%waveform-ref name
+                 (loop :for (name val) :on plist :by #'cddr :while val
```
stylewarning: put while val on newline

src/ast.lisp
```diff
@@ -127,13 +127,59 @@ EXPRESSION should be an arithetic (Lisp) form which refers to LAMBDA-PARAMS."
...
+  ;; A list of (name val) lists.
+  (parameters nil :read-only t :type list)
+  ;; Will later be resolved
+  (name-resolution nil :type (or null
+                                 standard-waveform
+                                 waveform-definition)))
+
+(defun waveform-ref (name &rest plist)
+  "Construct a waveform reference with keyword-value pairs given by ARGS."
+  (assert (evenp (length plist)))
```
stylewarning: write a nice error message

src/ast.lisp
```diff
@@ -127,13 +127,59 @@ EXPRESSION should be an arithetic (Lisp) form which refers to LAMBDA-PARAMS."
...
+                       :collect (list name val))))
+
+
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;; QuilT Waveforms ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+
+(defclass standard-waveform ()
+  ((duration :initarg :duration
+             :reader waveform-duration
+             :type float
+             :documentation "Duration of the waveform, in seconds."))
```
stylewarning: is `seconds` an appropriate unit? I'd rather have some arbitrary units and `+seconds-per-arbitrary-unit+` so we can have Mega Precision

src/ast.lisp
```diff
@@ -413,6 +461,97 @@ as a permutation."
                  :body body
                  :context context))
 
+;;; Frame Definitions (QuilT)
```
stylewarning:
```suggestion
;;; Frame Definitions (Quilt)
```

src/ast.lisp
```diff
@@ -127,13 +127,59 @@ EXPRESSION should be an arithetic (Lisp) form which refers to LAMBDA-PARAMS."
...
+  (name nil :read-only t :type string)
+  ;; A list of (name val) lists.
+  (parameters nil :read-only t :type list)
+  ;; Will later be resolved
+  (name-resolution nil :type (or null
+                                 standard-waveform
+                                 waveform-definition)))
+
+(defun waveform-ref (name &rest plist)
+  "Construct a waveform reference with keyword-value pairs given by ARGS."
```
stylewarning:
```suggestion
  "Construct a waveform reference with keyword-value pairs given by PLIST."
```

src/ast.lisp
```diff
@@ -127,13 +127,59 @@ EXPRESSION should be an arithetic (Lisp) form which refers to LAMBDA-PARAMS."
...
+       (equalp (frame-qubits a) (frame-qubits b))))
+
+(defun frame-hash (f)
+  #+sbcl
+  (sb-int:mix (sxhash (frame-name f)) (sxhash (frame-qubits f)))
+  #-sbcl
+  (logxor (sxhash (frame-name f)) (sxhash (frame-qubits f))))
+
+(defstruct (waveform-ref (:constructor %waveform-ref (name parameters)))
+  "An reference to a (possibly parametric) QuilT waveform."
```
stylewarning:
```suggestion
  "An reference to a (possibly parametric) Quilt waveform."
```

src/ast.lisp
```diff
@@ -127,13 +127,59 @@ EXPRESSION should be an arithetic (Lisp) form which refers to LAMBDA-PARAMS."
...
 
-
 (defun map-de-params (f de)
   "Create a new DELAYED-EXPRESSION from DE, applying F to all of the parameters of DE."
   (let ((c (copy-structure de)))
     (setf (delayed-expression-params c) (mapcar f (delayed-expression-params de)))
     c))
 
+(defstruct (frame (:constructor frame (qubits name)))
+  "A reference to a QuilT rotating frame, relative to which control or readout waveforms may be defined."
```
stylewarning:
```suggestion
  "A reference to a Quilt rotating frame, relative to which control or readout waveforms may be defined."
```

stylewarning: change all QuilT to Quilt

src/ast.lisp
```diff
@@ -413,6 +461,97 @@ as a permutation."
                  :body body
                  :context context))
 
+;;; Frame Definitions (QuilT)
+
+(defclass frame-definition ()
```
stylewarning: add class and slot documentation

src/ast.lisp
```diff
@@ -413,6 +461,97 @@ as a permutation."
...
+                      :initform nil
+                      :reader frame-definition-initial-frequency)
+   (context :initarg :context
+            :type lexical-context
+            :accessor lexical-context)))
+
+;;; Waveform Definitions (QuilT)
+
+(defclass waveform-definition ()
+  ((name :initarg :name
```
stylewarning: add documentation of slots

src/ast.lisp
```diff
@@ -413,6 +461,97 @@ as a permutation."
...
+(defclass parameterized-waveform-definition (waveform-definition)
+  ((parameters :initarg :parameters
+               :reader waveform-definition-parameters
+               :documentation "A list of symbol parameter names."))
+  (:documentation "A waveform definition that has named parameters."))
+
+(defun make-waveform-definition (name parameters entries sample-rate &key context)
+  (check-type name string)
+  (check-type parameters symbol-list)
+  (if parameters
```
stylewarning: `(not (endp parameters))`

src/ast.lisp
```diff
@@ -519,6 +657,112 @@ as the reset is formally equivalent to measuring the qubit and then conditionall
...
+(defclass simple-frame-mutation (instruction)
+  ((frame :initarg :frame
+          :accessor frame-mutation-target-frame)
+   (value :initarg :value
+          :accessor frame-mutation-value))
+  (:documentation "An instruction representing the mutation of a frame attribute.")
+  (:metaclass abstract-class))
+
+(eval-when (:compile-toplevel :load-toplevel :execute)
+  (defun expand-simple-frame-mutation-definition (name mnemonic docstring)
```
stylewarning: s/expand/define/

src/ast.lisp
```diff
@@ -519,6 +657,112 @@ as the reset is formally equivalent to measuring the qubit and then conditionall
...
+(defclass swap-phase (instruction)
+  ((left-frame :initarg :left-frame
+               :accessor swap-phase-left-frame)
+   (right-frame :initarg :right-frame
+                :accessor swap-phase-right-frame))
+  (:documentation "An instruction representing a phase swap between two frames."))
+
+;;; QuilT Operators
+
+(defclass pulse (instruction)
```
stylewarning: document these more?

src/ast.lisp
```diff
@@ -1006,6 +1249,35 @@ For example, `DAGGER DAGGER H 0` should produce `H 0`."
     ((named-operator name)   name)
     (_ (error "The application doesn't have a canonical name."))))
 
+(defun operator-description-equalp (od1 od2)
```
stylewarning: I like using `=` as the suffix

src/ast.lisp
```diff
@@ -1326,6 +1598,20 @@ For example,
                   (print-instruction expr stream)))))
       (print-delayed-expression (delayed-expression-expression thing) stream)))
 
+  (:method ((thing frame) (stream stream))
+    (format stream "~{~A ~}\"~A\""
+            (mapcar #'print-instruction-to-string (frame-qubits thing))
```
stylewarning: you can put this in the iterator above

src/ast.lisp
```diff
@@ -1326,6 +1598,20 @@ For example,
...
+            (frame-name thing)))
+
+  (:method ((thing waveform-ref) (stream stream))
+    (format stream "~A~@[(~{~A~^, ~})~]"
+            (waveform-ref-name thing)
+            (mapcar (lambda (name-and-value)
+                      (format nil "~A: ~A"
+                              (param-name (first name-and-value))
+                              (print-instruction-to-string (second name-and-value))))
```
stylewarning: same

src/ast.lisp
```diff
@@ -1334,14 +1620,66 @@ For example,
     (format stream "RESET"))
 
   (:method ((instr reset-qubit) stream)
-    (format stream "RESET ~a" (print-instruction (reset-qubit-target instr) nil)))
+    (format stream "RESET ~a" (print-instruction-to-string (reset-qubit-target instr))))
```
stylewarning: :/

src/ast.lisp
```diff
@@ -1334,14 +1620,66 @@ For example,
...
   (:method ((instr wait) (stream stream))
     (format stream "WAIT"))
 
   (:method ((instr no-operation) (stream stream))
     (format stream "NOP"))
 
+  (:method  ((instr simple-frame-mutation) (stream stream))
+    (format stream "~A~{ ~A~}"
+            (mnemonic instr)
+            (map 'list #'print-instruction-to-string (arguments instr))))
```
stylewarning: :/ :/ :/

src/ast.lisp
```diff
@@ -1363,50 +1701,65 @@ For example,
...
+            (print-instruction-to-string (measurement-qubit instr))))
 
   (:method ((instr application) (stream stream))
     (print-operator-description (application-operator instr) stream)
     (format stream "~@[(~{~a~^, ~})~]~{ ~a~}"
-            (mapcar (lambda (thing) (print-instruction thing nil))
+            (mapcar #'print-instruction-to-string
                     (application-parameters instr))
-            (mapcar (lambda (thing) (print-instruction thing nil))
+            (mapcar #'print-instruction-to-string
```
stylewarning: all of these TO-STRING methods are making my eyes bleed

src/ast.lisp
```diff
@@ -1444,13 +1873,22 @@ For example,
                      :accessor parsed-program-gate-definitions)
    (circuit-definitions :initarg :circuit-definitions
                         :accessor parsed-program-circuit-definitions)
+   (waveform-definitions :initarg :waveform-definitions
```
stylewarning: lets document this class better

src/analysis/fill-delays.lisp
kilimanjaro: Trying to really drive the point home, this thing returns nil!

src/analysis/fill-delays.lisp
kilimanjaro: Oof, not sure how that one happened.

src/ast.lisp
kilimanjaro: This is mostly taking its cue from RPCQ, which tracks pulse length in seconds and in the lisp code declares the type to be `float`. I don't have the impression anyone expects to be working with numbers smaller than roughly 10^{-10} without some radical breakthrough in control electronics.

src/analysis/fill-delays.lisp
```diff
@@ -0,0 +1,206 @@
+(in-package :cl-quil)
+
+(define-transform fill-delays (fill-delays)
+  "This transform fills empty time on quilt frames with explicit DELAY instructions in a greedy fashion."
+  expand-calibrations
+  resolve-waveform-references)
+
+;;; Syntactic conveniences
+
+(deftype simple-quilt-instruction ()
```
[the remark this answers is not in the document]
kilimanjaro: Added a comment

src/analysis/fill-delays.lisp
```diff
@@ -0,0 +1,206 @@
...
+    (waveform-ref-name-resolution wf-ref)))
+
+(defun waveform-active-duration (wf-or-wf-defn)
+  "Get the active duration of the waveform or waveform definition, in seconds.
+If WF-OR-WF-DEFN is a waveform definition, SAMPLE-RATE (Hz) must be non-null. "
+  (etypecase wf-or-wf-defn
+    (standard-waveform (constant-value (waveform-duration wf-or-wf-defn)))
+    (waveform-definition
+     (/ (length (waveform-definition-entries wf-or-wf-defn))
+        (constant-value (waveform-definition-sample-rate wf-or-wf-defn))))))
```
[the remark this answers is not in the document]
kilimanjaro: Adding a check for this in `parse-sample-rate`

src/ast.lisp
kilimanjaro: Should this also apply to `expand-classical-instruction-definitions`?

src/analysis/expansion.lisp
```diff
@@ -0,0 +1,465 @@
...
+    (gate-calibration-definition (values (calibration-definition-parameters defn)
+                                         (calibration-definition-arguments defn)
+                                         (calibration-definition-body defn)))
+    (measure-calibration-definition (values nil
+                                            (list (measurement-calibration-qubit defn)
+                                                  (measure-calibration-address defn))
+                                            (calibration-definition-body defn)))
+    (measure-discard-calibration-definition (values nil
+                                                    (list (measurement-calibration-qubit defn))1
+                                                    (calibration-definition-body defn)))))
```
notmgsk: What is this `1` doing?

notmgsk: Other than being spooky

src/analysis/expansion.lisp
```diff
@@ -0,0 +1,465 @@
...
+            (circuit-definition-body (circuit-application-definition instr))))
+    (unresolved-application
+     (unless *allow-unresolved-applications*
+       (error "Unable to determine if the unresolved application is unitary:~%    ~/cl-quil:instruction-fmt/"
+              instr))
+     t)))
+
+(defun transform-if (test transform)
+  "Given a unary function TRANSFORM, return a new function which either applies TRANSFORM or does not,
+depending on whether TEST passes."
```
notmgsk: line break

src/analysis/expansion.lisp
```diff
@@ -0,0 +1,465 @@
...
+                        (application-arguments instr))))
+      (cond
+        ((simple-dagger-operator-p (application-operator instr))
+         (let ((instrs (instantiate-definition (circuit-application-definition instr)
+                                               params
+                                               args)))
+           (dolist (instr instrs)
+             (unless (unitary-instruction-p instr)
+               (error "DAGGER cannot be applied to the impure instruction ~S"
+                      (print-instruction instr nil)))
```
notmgsk: Some `print-instruction`s lying around

src/analysis/expansion.lisp
```diff
@@ -0,0 +1,465 @@
...
+             (setf (application-operator instr)
+                   (involutive-dagger-operator (application-operator instr))))
+           ;; The Hermitian transpose reverses the order of operator
+           ;; applications
+           (setf instrs (reverse instrs))))
```
notmgsk: `nreverse`

src/analysis/fill-delays.lisp
```diff
@@ -0,0 +1,206 @@
...
+
+(defun quilt-instruction-duration (instr)
+  "Get the duration of the specified quilt instruction INSTR."
+  (typecase instr
+    ((or pulse capture)
+     (waveform-active-duration (resolved-waveform instr)))
+    (delay
+      (constant-value (delay-duration instr)))
```
notmgsk: extra space

src/analysis/fill-delays.lisp
```diff
@@ -0,0 +1,206 @@
...
+
+(defun quilt-instruction-frames (instr parsed-program)
+  (a:ensure-list
+   (typecase instr
+     (pulse (pulse-frame instr))
+     (capture (capture-frame instr))
+     (raw-capture (raw-capture-frame instr))
+     (delay-on-frames (delay-frames instr))
+     (simple-frame-mutation (frame-mutation-target-frame instr))
+     (swap-phase (swap-phase-left-frame instr) (swap-phase-right-frame instr))
```
notmgsk: Should this return a list of both functions' results? If I understand correctly, it will only return the result of `swap-phase-right-frame`.

src/analysis/fill-delays.lisp
```diff
@@ -0,0 +1,206 @@
...
+  (let ((frame-clocks (make-frame-table))
+        new-instrs)
+
+    ;; initialize clocks
+    (dolist (defn (parsed-program-frame-definitions parsed-program))
+      (setf (local-time frame-clocks (frame-definition-frame defn)) 0.0d0))
+
+    (flet ((process-instr (instr)
+             (unless (typep instr 'simple-quilt-instruction)
+               (quil-parse-error "Cannot resolve timing information for non-quilt instruction ~A" instr))
```
notmgsk: period

src/analysis/resolve-objects.lisp
```diff
@@ -0,0 +1,215 @@
...
+                        (waveform-ref-name waveform-ref)
+                        expected-parameters))
+    t))
+
+(defun resolve-standard-waveform (waveform-ref waveform-class)
+  (let ((param-map (quilt-waveform-parameter-alist waveform-class)))
+    (validate-waveform-parameters waveform-ref
+                                  (mapcar #'first param-map))
+    (let ((obj (make-instance waveform-class)))
+      (loop :for (param  val) :in (waveform-ref-parameters waveform-ref)
```
notmgsk: space

src/analysis/resolve-objects.lisp
```diff
@@ -0,0 +1,215 @@
...
+  frame)
+
+;;; TODO: Factor out this gate arity computation to something nicer.
+(defgeneric resolve-instruction (instr parsed-program)
+  (:method ((instr unresolved-application) parsed-program)
+    (macrolet
+        ((assert-and-print-instruction (test-form &optional places datum &rest arguments)
+           `(assert ,test-form
+                    ,places
+                    (format nil "Error in resolving ~s: ~a"
```
notmgsk: period

src/analysis/resolve-objects.lisp
```diff
@@ -0,0 +1,215 @@
...
+      (resolve-frame (raw-capture-frame instr)
+                     (parsed-program-frame-definitions parsed-program)))
+    instr)
+
+  (:method ((instr t) parsed-program)
+    (declare (ignore parsed-program))
+    instr))
+
+(defun resolve-objects (unresolved-program)
+  "Perform all object resolution within the list RAW-QUIL, returning a PARSED-PROGRAM."
```
notmgsk: RAW-QUIL?

src/analysis/type-safety.lisp
```diff
@@ -30,11 +30,15 @@
...
+     (memory-ref-position mref)))
+
 (defun enforce-mref-bounds (mref memory-descriptor)
   (when (and (typep mref 'memory-ref)
-             (<= (memory-descriptor-length memory-descriptor)
-                 (memory-ref-position mref)))
-    (quil-type-error "Memory ref \"~A\" is exceeds region size ~A"
+             (not (plusp (mref-available-length mref
+                                                memory-descriptor))))
+    (quil-type-error "Memory ref \"~A\" exceeds region size ~A"
```
notmgsk: period

src/analysis/type-safety.lisp
```diff
@@ -330,6 +344,47 @@
...
+         (a:if-let ((samples (raw-capture-num-real-samples instr)))
+           (if (> samples (mref-available-length mref mdesc))
+               (quil-type-error "RAW-CAPTURE instruction target ~A must be a REAL ~
+                                 vector of length no less than ~A."
+                                (print-instruction-to-string mref)
+                                samples))
+           (warn "RAW-CAPTURE on frame ~A with unknown sample rate."
+                 (frame-definition-frame frame-defn))))
+        (_
+         (quil-type-error "RAW-CAPTURE instruction target must be of type ~
```
notmgsk: period

src/ast.lisp
```diff
@@ -349,12 +397,10 @@ If no exit rewiring is found, return NIL."
...
 
 (defun permutation-from-gate-entries (entries)
-  "Create the permutation (list of natural numbers) that represents
-the input matrix ENTRIES. Return nil if ENTRIES cannot be represented
-as a permutation."
+  "Create the permutation (list of natural numbers) that represents the input matrix ENTRIES. Return nil if ENTRIES cannot be represented as a permutation."
   (let* ((n (isqrt (length entries)))
          (perm (make-list n)))
-    (dotimes (i n perm)
+     (dotimes (i n perm)
```
notmgsk: no

src/cl-quil.lisp
```diff
@@ -18,23 +71,83 @@
...
+    (etypecase instr
+      (gate-definition (make-condition 'ambiguous-gate-or-circuit-definition :conflicts combined))
+      (circuit-definition (make-condition 'ambiguous-gate-or-circuit-definition :conflicts combined))
+      (waveform-definition (make-condition 'ambiguous-waveform-definition :conflicts combined))
+      (calibration-definition (make-condition 'ambiguous-calibration-definition :conflicts combined))
+      (frame-definition (make-condition 'ambiguous-frame-definition :conflicts combined))
+      (memory-descriptor (make-condition 'ambiguous-memory-declaration :conflicts combined)))))
+
+(defun raw-quil-to-unresolved-program (code)
+  "This constructes a PARSED-PROGRAM object from the given quil CODE, without any resolution of applications.
```
notmgsk: constructEs

src/parser.lisp
```diff
@@ -312,6 +314,16 @@ the immediately preceding line."
 (defvar *formal-arguments-allowed* nil
   "Dynamic variable to control whether formal parameters and arguments are allowed in the current parsing context.")
 
+(defparameter *parse-context* nil
+  "A Quil keyword (e.g. :DEFGATE) indicating the context in which a portion of a Quil instruction is being parsed.")
+
+(defun disappointing-token-error (found-token expected-msg)
+  (quil-parse-error "Expected ~A~@[ in ~A~], but observed a token ~A with value ~A"
```
notmgsk: period

tests/bad-test-files/bad-quilt-defframe.quil
```diff
@@ -0,0 +1,3 @@
+DEFFRAME 0 1 "xy":
+    SAMPLE-RATE: 1.0
+    SAMPLE-RATE: 2.0
```
notmgsk: For the sake of not having tests emit warnings, could you add newlines following any final indented line.

src/analysis/expansion.lisp
appleby: Sneak level 💯

src/analysis/expand-calibrations.lisp
```diff
@@ -0,0 +1,173 @@
+(in-package #:cl-quil)
+
+;;; Calibration Expansion
+;;;
+;;; This looks a lot like circuit expansion, and under the hood a lot of the
+;;; mechanics are the same. A few differences are worth pointing out:
+;;;
+;;;   - The task of finding a calibration that matches a measurement or gate application
```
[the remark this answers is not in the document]
kilimanjaro: Mostly because calibrations are rather low level and hardware, and so a single gate could reasonable have multiple calibrations. For example, some qubits in our chip are tuneable, and others are not, so perhaps the calibration varies from qubit to qubit. Likewise perhaps for parametric gates you wish to have a calibration for "special" values of the parameter, which differs from calibrations for the generic value of the parameter, etc. So the quilt spec proposes a pattern matching syntax, and identifies which calibration is applicable in a specific context. It also encourages having a simple deterministic rule for tiebreaking (which vanilla Quil would prefer not to have).

src/analysis/fill-delays.lisp
kilimanjaro: oof

tests/bad-test-files/bad-quilt-defframe.quil
kilimanjaro: After talking to @stylewarning I'm just nixing the warning, so EOF counts as an implicit dedent (this is compatible with a plausible use case, where up-to-date calibrations or frame definitions are just be tacked on to the end of some base quilt program). FWIW the tokenizer was gobbling up terminal newlines anyways, so outside of changing that one would have needed an explicit `HALT` or something.

src/ast.lisp
kilimanjaro: I added this + a blurb at the top.

src/ast.lisp
```diff
@@ -1326,6 +1712,23 @@ For example,
...
+    (format stream "~{~/quil:instruction-fmt/ ~}\"~A\""
+            (frame-qubits thing)
+            (frame-name thing)))
+
+  ;;; Quilt
+  (:method ((thing waveform-ref) (stream stream))
+    (format stream "~A" (waveform-ref-name thing))
+    (when (not (endp (waveform-ref-parameter-alist thing)))
+      (format stream "(")
+      (loop :for (param . value) :in param-alist
```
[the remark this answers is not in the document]
appleby: Parser appears to call the `%waveform-ref` constructor directly passing an alist of cons pairs (not proper lists) built in `parse-waveform-parameter-and-value`. So probably the alist construction in `waveform-ref` above should change.

src/ast.lisp
```diff
@@ -1438,20 +1995,55 @@ For example,
...
+                       :documentation "The memory definitions introduced by DECLARE.")
    (executable-program :initarg :executable-code
-                       :accessor parsed-program-executable-code))
+                       :accessor parsed-program-executable-code
+                       :type (vector instruction)
+                       :documentation "A vector of executable Quil instructions."))
   (:default-initargs
    :gate-definitions nil
    :circuit-definitions nil
+   :waveform-definitions nil
```
appleby: nitpick: `'()` etc

src/ast.lisp
```diff
@@ -1505,45 +2097,51 @@ Examples:
...
+    (dolist (gate-defn gate-definitions)
+      (print-instruction gate-defn s))
+    (unless (endp gate-definitions)
+      (terpri s))
+
+    ;; write out circuits
+    (dolist (circuit-defn circuit-definitions)
+      (print-instruction circuit-defn s))
+    (unless (endp circuit-definitions)
+      (terpri s))
```
appleby:
```
(flet ((print-definitions (definitions) ...)
  (print-definitions memory-definitions)
  (print-definitions frame-definitions)
  ...
  (print-definitions circuit-definitions)
``` 

imho

Or maybe `print-instruction-sequence` and learn to live with the extra newlines (boo!).

src/analysis/resolve-objects.lisp
```diff
@@ -0,0 +1,217 @@
...
+      (quil-parse-error "Unable to resolve formal ~A outside of definition body." formal-qubit)
+      (a:if-let ((defn (find frame frame-definitions
+                             :key #'frame-definition-frame
+                             :test #'frame=)))
+        (setf (frame-name-resolution frame) defn)
+        (quil-parse-error "No frame definition found for referenced frame ~A."
+                          (print-instruction-to-string frame)))))
+  frame)
+
+;;; TODO: Factor out this gate arity computation to something nicer.
```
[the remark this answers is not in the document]
kilimanjaro: I'm not sure who wrote that TODO, or what they had in mind, but I will leave it for them,

src/ast.lisp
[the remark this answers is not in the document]
kilimanjaro: Yeah, this is just some code path that hasn't been worked since the changes I made this afternoon. Good eye.

kilimanjaro:
Closing until I have addressed feedback, made changes, and tested them adequately.

src/ast.lisp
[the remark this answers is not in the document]
appleby: Just realized this the perfect opportunity to give my GOLDEN FILES sales pitch (*slicks back hair with greasy comb*).

If you chuck the various Quilt examples from your parser tests into a golden file under `tests/printer-test-files`, then `test-print-parsed-program-golden-files` will make sure the examples can be both parsed and printed (and that the printed output is stable, tho you can disable that with a magic comment if need be).  The golden tests don't replace parsers tests, since they don't verify anything about the structure of what's parsed, only that it can be parsed and printed without error and that the printed output matches the expected output. But it's a ~~low-cost~~ medium-cost way to add printer tests if you already have a bunch of parser tests lying around.

To find out more, log on to `tests/utilities.lisp` and scroll down to the mini-novella of a comment near the end, specifically the section `ADDING NEW INPUT/OUTPUT pairs`.  If you don't need regex support, you can stick the new test cases in a file under `tests/printer-test-files/gold-standard`, then call `cl-quil-tests::update-print-parsed-program-golden-files` to automatically generate the output sections.

You'll probably have lot's of questions, like why didn't appleby capitalize "pairs" in "ADDING NEW INPUT/OUTPUT pairs"? Totally valid question. 🤷‍♂ Feel free to ping me if you wanna take this baby for a test drive.

Alternatively, feel free to ghost me if it all seems like more trouble than it's worth. It is the seasonally appropriate reaction, after all.

src/ast.lisp
```diff
@@ -127,13 +160,64 @@ EXPRESSION should be an arithetic (Lisp) form which refers to LAMBDA-PARAMS."
...
+
+;;; Quilt
+(defun waveform-ref (name &rest plist)
+  "Construct a waveform reference with keyword-value pairs given by PLIST."
+  (unless (evenp (length plist))
+    (error "WAVEFORM-REF takes an odd number of arguments: first a name, and then alternating keyword-value pairs."))
+  (%waveform-ref name
+                 (loop :for (name val) :on plist :by #'cddr
+                       :while val
+                       :collect (list name val))))
```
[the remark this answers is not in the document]
kilimanjaro: I'm nixing this whole function, since constructing from a plist is not a common operation in practice.

src/ast.lisp
kilimanjaro: Reasonable suggestion. I'm opting for something more like the former, although with a bit more careful control of newline printing.

PR #383, opened 2019-08-09
support Allegro CL

stylewarning:


stylewarning:
do-not-merge yet, still not working

src/package.lisp
```diff
@@ -2,8 +2,12 @@
 ;;;;
 ;;;; Author: Robert Smith
 
+#+(or allegro)
```
kilimanjaro: Can you add a one-line comment about why Allegro gets special treatment / what's the deal with Allegro

appleby: I actually had to check as well. Turns out the `:local-nicknames` option isn't standard. Who knew?

PR #391, opened 2019-08-20
Fixup: clamp fidelity overflow

### LINKED ISSUE #378: Fidelity estimates exceeding 100% crash quilc
The fidelity estimates stored in Forest DB (e.g., Aspen-4's qubit 17's f1QRB at the time of this writing) can exceed 100%. This causes quilc to crash.

An easy fix for this is to clamp interpreted fidelities to `+near-perfect-fidelity+`. It is probably better to understand why these values appear at all before proceeding.

ecpeterson:
As reported in #378, the changes in #210 cause quilc to throw a fit resulting in OOM when fed fidelity values > 1. This patch catches these values and clamps them to `+near-perfect-fidelity+`.

I included a commit where such a clamp is accompanied by a warning. I don't know if this is a good idea or if it's obnoxious, since it'll warn the _user_ but it's typically _Forest's_ fault.

Closes #378.


src/chip-reader.lisp
```diff
@@ -124,12 +124,23 @@
...
+(defun deforestify-fidelity (fidelity)
+  "The fidelity values recorded in Forest DB are subject to various quirks. This \"parser\" un-quirks them."
+  (cond 
+    ((double= fidelity +forest-error-sentinel+)
+     (warn "Chip specification contains a fidelity field populated by the Forest error sentinel. Replacing it with the default ~a." +totally-awful-fidelity+)
+     +totally-awful-fidelity+)
+    ((> fidelity +near-perfect-fidelity+)
+     (warn "Chip specification contained fidelity ~a > 1. Truncating to ~a." fidelity +near-perfect-fidelity+)
+     +near-perfect-fidelity+)
+    (t
```
appleby: Maybe add a `(< fidelity 0)` check as well, just in case?

tests/chip-specification-tests.lisp
```diff
@@ -34,8 +34,7 @@
                      "CNOT 0 15"
                      "CNOT 0 16"
                      "CNOT 1 2")
-                   ;; TODO:(appleby) switch this to use Aspen-4-10Q-A.qpu once gh-378 is resolved.
-                   (cl-quil-tests::%read-test-chipspec "Aspen-4-10Q-A-doctored-for-gh-378.qpu"))))
```
appleby: 👍 ty

src/chip-reader.lisp
stylewarning: `minusp`

ecpeterson: I just added a `minusp` branch that throws an _error_, since that doesn't generically correspond to some known goofy Forest DB behavior.

PR #394, opened 2019-08-21
Fix governed queue creation and other portability fixes

stylewarning:
just ANSI things


stylewarning:
oops accidentally hauled in other junk into this PR

PR #395, opened 2019-08-26
Use swap fidelity instead of gate time as a distance metric

karalekas:
And benchmark the peformance.

Replaces stale PR #331

–––––––

Some things to inspect before merge, inherited from #449:

* `chip-specification` sometimes carries extra data used by a particular addressing method to make more intelligent decisions, like the `time-bound` key inside of various `hardware-object`'s `misc-data` slots. It would probably be better to make this part of (the subclasses of) `addresser-state`, since which addresser cares about what is really specific to the style of addressing being used. (But it's also conceivable that the compressor will want to see that same information and optimize along the same `cost-function`...?) (Also, I don't think the `cost-bound` field gets populated at all? correctly? for the fidelity addresser!)
* I did a bunch of monkey-patching around Corwin's contributions to the addresser (viz., the various bits of SWAP-searching), which involved passing different "versions" of the parent `cost-function` around with different argument specializations and so on. It's gross, and it sort-of breaks the abstraction of addresser-state.
* We should very thoroughly test the case of running programs near the saturation limit on _small_ chips, as this is the most common use case today.
* Seriously consider adding terms to the fidelity cost function which represent (T1-type) decay.


benchmarking/rewiring-analysis.lisp
```diff
@@ -296,3 +297,14 @@
     :a*-depth  (:partial :a*           :a*           1d0 :depth)
     :a*-sum    (:partial :a*           :a*           1d0 :sum)
     ))
+
+(defvar *cost-fn-weight-style-assn*
+    (make-assignments
+        ((*random-state* #+sbcl (sb-ext:seed-random-state 1)
```
stylewarning: i'd prefer to move these into separate function so we know it needs porting

benchmarking/rewiring-analysis.lisp
```diff
@@ -296,3 +297,14 @@
...
+                         #+ecl  (make-random-state 1)
+                         #-(or sbcl ecl) (error "don't know how to seed random state"))
+         (quil::*compressor-passes* 1))
+        (quil::*cost-fn-weight-style*)
+      :duration   (:duration)
+      :fidelity   (:fidelity)
```
stylewarning: close parens

src/addresser/logical-schedule.lisp
```diff
@@ -634,7 +635,7 @@ mapping instructions to their tags. "
     (multiple-value-bind (max-value value-hash)
         (lscheduler-walk-graph lschedule
                                :base-value 1d0
-                               :bump-value #'fidelity-bumper
+                               :bump-value (lambda (instr value) (let ((fid (fidelity-bumper instr value))) (format t "~/cl-quil:instruction-fmt/ -> ~a~%" instr fid) fid))
```
stylewarning: did you intend to keep the print there

src/addresser/cost-function.lisp
```diff
@@ -40,9 +53,9 @@
     (dotimes (link-index (length (vnth 1 (chip-specification-objects chip-spec))))
       (let ( ; eventually this should look up "SWAP" in this list, but for now
              ; this is guaranteed to be the only 2Q permutation anyway.
-            (weight (permutation-record-duration
-                     (vnth 0 (hardware-object-permutation-gates
-                              (chip-spec-nth-link chip-spec link-index)))))
+            (weight (ecase *cost-fn-weight-style*
+                      (:duration (permutation-record-duration (vnth 0 (hardware-object-permutation-gates (chip-spec-nth-link chip-spec link-index)))))
+                      (:fidelity (- (log (swap-fidelity chip-spec link-index))))))
```
stylewarning: can we document this formula for the future reader

karalekas:
Just pushed additional work-in-progress work, still need to address the partial review comments above.

benchmarking/rewiring-analysis.lisp
```diff
@@ -63,7 +63,7 @@
 (defparameter *rewiring-explicit-test-chips*
   '((:0006q-2ring *chip-6q-2ring*)
     (:0008q *chip-0008q*)
-    (:0016q *chip-0016q*)
+    #+ignore(:0016q *chip-0016q*)
```
stylewarning: this shouldn't make it into the final merge

src/addresser/cost-function.lisp
```diff
@@ -55,7 +55,7 @@
              ; this is guaranteed to be the only 2Q permutation anyway.
             (weight (ecase *cost-fn-weight-style*
                       (:duration (permutation-record-duration (vnth 0 (hardware-object-permutation-gates (chip-spec-nth-link chip-spec link-index)))))
-                      (:fidelity (- (log (swap-fidelity chip-spec link-index))))))
+                      (:fidelity (* (permutation-record-duration (vnth 0 (hardware-object-permutation-gates (chip-spec-nth-link chip-spec link-index)))) (- (log (swap-fidelity chip-spec link-index)))))))
```
stylewarning: put each factor on a newline

src/addresser/logical-schedule.lisp
```diff
@@ -599,6 +599,7 @@ mapping instructions to their tags. "
      (hash-table-count (lscheduler-earlier-instrs lschedule))))
 
 (defun lscheduler-calculate-fidelity (lschedule chip-spec)
+  (declare (optimize (debug 3)))
```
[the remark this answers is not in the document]
stylewarning: delete unless needed

src/chip-specification.lisp
```diff
@@ -293,7 +293,7 @@ used to specify CHIP-SPEC."
                                   (hardware-object-gate-information obj))
                          (make-gate-record :duration 150
                                            :fidelity fidelity)))))
-      (dolist (data `((:cz     "CZ"     ()  (_ _) 0.89d0)
+      (dolist (data `((:cz     "CZ"     ()  (_ _) 0.98d0)
```
stylewarning: can we write a comment where these numbers come from, or why their choices are reasonable?

src/chip-specification.lisp
```diff
@@ -346,10 +346,10 @@ used to specify CHIP-SPEC."
       (when (member ':RZ type)
         (stash-gate-record "RZ" '(_) (list q) 1/100 +near-perfect-fidelity+))
       (when (member ':X/2 type)
-        (stash-gate-record "RX" '(#.(/ pi 2))  `(,q) 9 0.98d0)
```
stylewarning: likewise here

src/addresser/fidelity-addresser.lisp
```diff
@@ -0,0 +1,169 @@
...
+;;; itself as a heuristic for producing "efficient" programs, where "efficient"
+;;; is primarily taken to mean "short". In this routine, we make the necessary
+;;; modifications to take "efficient" to mean "high-fidelity".
+;;;
+;;; See DO-FIDELITY-ADDRESSING below for the main entry point.
+
+(defclass fidelity-addresser-state (addresser-state)
+  ())
+
+(defstruct fidelity-cost
```
stylewarning:
```
(defstruct (fidelity-cost (:constructor fidelity-cost (value)))
  (value nil :type real))
```
add `:read-only t` if it's read-only too.

This will change the constructor to just be `(fidelity-cost 10)`

src/addresser/fidelity-addresser.lisp
```diff
@@ -0,0 +1,169 @@
...
+        (gate-weights-cost 0d0))
+    
+    ;; calculate log-infidelity coming from INSTR, using recombination:
+    ;;   + calculate the cut point
+    ;;   + calculate the infidelity of instructions since the cut point
+    ;;   + use only the additional infidelity penalty
+    (when (and instr (typep instr 'gate-application))
+      (let* ((chip-spec (addresser-state-chip-specification state))
+             (instruction-expansion
+               (let ((*compress-carefully* nil))
```
stylewarning: why?

src/addresser/fidelity-addresser.lisp
```diff
@@ -0,0 +1,169 @@
...
+                 (expand-to-native-instructions (list instr) chip-spec))))
+        (setf instr-cost (- (log (calculate-instructions-fidelity instruction-expansion chip-spec))))
+        
+        ;; then, see if there's a non-naive cost available
+        (a:when-let* ((hardware-object (lookup-hardware-object chip-spec instr))
```
stylewarning: are you sure _all_ of these should participate in `when`? It sounds like it should be
```
(let* (...)
  ;; assert some things are non-null
  (when <actual thing that can be null>
    ...))
```

src/addresser/addresser-common.lisp
```diff
@@ -0,0 +1,788 @@
...
+   "Select a permutation and schedule it for execution. The permutation is selected to lower the
+cost-function associated to the current lschedule.")
+  (:method (state rewirings-tried)
+    (with-slots (lschedule initial-l2p working-l2p chip-sched chip-spec qq-distances)
+        state
+      (format *compiler-noise-stream*
+              "SELECT-AND-EMBED-A-PERMUTATION: entering SWAP selection phase.~%")
+      (let ((gates-in-waiting (assign-weights-to-gates state)))
+        (ecase *addresser-swap-search-type*
+          #+ignore  ; TODO: test this
```
stylewarning: TODO

src/addresser/fidelity-addresser.lisp
```diff
@@ -0,0 +1,169 @@
...
+  (double= (fidelity-cost-value val1) (fidelity-cost-value val2)))
+
+(defmethod build-worst-cost ((state fidelity-addresser-state))
+  (make-fidelity-cost :value most-positive-fixnum))
+
+(defmethod assign-weights-to-gates ((state fidelity-addresser-state) &optional (1q-descaling 1/10))
+  (multiple-value-bind (max-value value-hash)
+      (lscheduler-walk-graph (addresser-state-logical-schedule state)
+                             :base-value 0
+                             :bump-value (lambda (instr value)
```
stylewarning: abstract the lambda into an flet

src/addresser/fidelity-addresser.lisp
```diff
@@ -0,0 +1,169 @@
...
+                                             (t
+                                              value)))
+                             :test-values #'max)
+    (declare (ignore max-value))
+    value-hash))
+
+(defun initial-fidelity-addresser-working-state (chip-spec initial-rewiring)
+  (let* ((*cost-fn-weight-style* ':fidelity)
+         (state (initial-addresser-working-state chip-spec initial-rewiring)))
+    (change-class state 'fidelity-addresser-state)
```
stylewarning: colour me dubious

src/addresser/fidelity-addresser.lisp
```diff
@@ -0,0 +1,169 @@
...
+    state))
+
+(defun do-greedy-fidelity-addressing (instrs
+                               chip-spec
```
stylewarning: indentation

src/addresser/outgoing-schedule.lisp
```diff
@@ -104,6 +104,12 @@ BEFORE-INST to make use of RESOURCE."
                  (not (resource-subsetp (instruction-resources instr) resource)))
         :maximize (chip-schedule-end-time schedule instr)))
 
+(defun chip-schedule-from-carving-point (schedule resource point)
+  (loop :for instr :being :the :hash-keys :of (chip-schedule-times schedule)
```
stylewarning: document

src/addresser/temporal-addresser.lisp
```diff
@@ -4,107 +4,173 @@
...
+(defclass temporal-addresser-state (addresser-state)
+  ((1q-queues :accessor temporal-addresser-state-1q-queues
+              :initarg :1q-queues
+              :documentation "The family of queues where not-yet-scheduled 1Q instructions live, while we get them out of the way to process 2Q instructions.")))
+
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;; Cost function ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+
+(defstruct temporal-cost
```
stylewarning: document, provide read-only and types qw n33e3e

src/addresser/temporal-addresser.lisp
```diff
@@ -4,107 +4,173 @@
...
-    (if (= q0 qubit) q1 q0)))
-
-;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;; Chip Schedule Management ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+;;; This file contains the subroutines specific to the operation of the version
+;;; of the addresser that uses overall program runtime to optimize output.
+
+(defclass temporal-addresser-state (addresser-state)
+  ((1q-queues :accessor temporal-addresser-state-1q-queues
+              :initarg :1q-queues
+              :documentation "The family of queues where not-yet-scheduled 1Q instructions live, while we get them out of the way to process 2Q instructions.")))
```
stylewarning: document class

src/addresser/temporal-addresser.lisp
```diff
@@ -4,107 +4,173 @@
...
+;;;; CZ 3 7
+;;;; on a chip with a linear circuit topology, and (for whatever reason) assuming
+;;;; qubits 0 and 7 to be fixed, a naive sum of the SWAP distances will result
+;;;; in a standstill. the 0-3 distance is 2 SWAPs and the 3-7 distance is 3 SWAPs,
+;;;; and trading 3 for 2 or 3 for 4 preserves the total number of 1+4 = 3+2 = 5
+;;;; SWAPs in all. to break this kind of deadlock, we make very distant gates
+;;;; suffer a decay penalty, so that gates that are already nearer are preferred
+;;;; and executed first.
+
+(defparameter *cost-fn-tier-decay* 0.5
```
stylewarning: defvar for configuration (if it's so)

src/addresser/temporal-addresser.lisp
```diff
@@ -523,550 +302,118 @@ cost-function associated to the current lschedule."
...
+     ;; dequeue and set the dirty bit  
+     (lscheduler-dequeue-instruction (addresser-state-logical-schedule state) instr)
+     (push instr (aref (temporal-addresser-state-1q-queues state)
+                       (qubit-index (first (application-arguments instr)))))
+     (values t nil nil))
+    (t
+     (call-next-method))))
+
+;; flush the queues before adding the 2Q instruction
+(defmethod append-instr-to-chip-schedule :before
```
stylewarning: why do we want to do this in a before method?

src/compilation-methods.lisp
```diff
@@ -248,8 +248,9 @@ Returns a value list: (processed-program, of type parsed-program
 
          (touch-unpreserved-block (blk registrant)
            ;; actually process this block
-           (multiple-value-bind (initial-l2p chip-schedule final-l2p)
-               (do-greedy-temporal-addressing
+           (multiple-value-bind (chip-schedule initial-l2p final-l2p)
+               ;; TODO: here, and below, we need to find a way to not hard-code this method call
```
stylewarning: what does this mean?

tests/compiler-hook-tests.lisp
```diff
@@ -477,7 +477,7 @@ MEASURE 1
 (deftest test-clever-CCNOT-depth-reduction ()
   "Test that the ':GREEDY-QUBIT swap selection strategy brings CZ depth down to optimal for CCNOT."
   (let ((p (quil::compiler-hook (quil::parse-quil "
-PRAGMA INITIAL_REWIRING \"GREEDY\"
```
stylewarning: why?

tests/initial-rewiring-tests.lisp
```diff
@@ -117,6 +117,7 @@ HALT
     ;; It's questionable whether asserting an error is the right thing to do here, but it at least
     ;; documents the current behavior. And if we ever improve NAIVE rewiring so that this doesn't
     ;; error, this test will fail and remind us to update it to NOT-SIGNALS.
+    #+ignore
```
stylewarning: oof

src/addresser/addresser-common.lisp
```diff
@@ -0,0 +1,788 @@
...
+;;; The potential heuristics used to search for solutions to the addressing
+;;; problem are manifold, and different ones are called for in different
+;;; situations. This file contains the guts common to each such method, and
+;;; different methods will invoke these guts after binding particular behaviors
+;;; to various hooks.
+;;;
+;;; See DO-GREEDY-TEMPORAL-ADDRESSING below for the main entry point.
+
+
+(defparameter *addresser-swap-lookahead-depth* 2
```
stylewarning: defvar for config vars

src/addresser/addresser-common.lisp
```diff
@@ -0,0 +1,788 @@
...
+
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;; Addresser State Updates ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+
+;;; The main entry point for temporal addressing is DO-GREEDY-TEMPORAL-ADDRESSING
+;;; below. However, this manages a fair amount of state as it navigates the
+;;; addressing process. The struct below bundles this together for the sake of
+;;; convenience. At any given point of the execution there is only one of these
+;;; objects, with values being mutated along the way.
+
+(defclass addresser-state ()
```
stylewarning: can we add types to the slots, or at least document them in the strings?

src/addresser/addresser-common.lisp
```diff
@@ -0,0 +1,788 @@
...
+  (let* ((hardware-object (chip-spec-nth-link chip-spec link-index))
+         (permutation-record (vnth 0 (hardware-object-permutation-gates
+                                      hardware-object)))
+         (swap (apply #'build-gate
+                      (permutation-record-operator permutation-record)
+                      '()
+                      (coerce (chip-spec-qubits-on-link chip-spec link-index) 'list))))
+    (calculate-instructions-fidelity (expand-to-native-instructions (list swap) chip-spec) chip-spec)))
+
+(defparameter *cost-fn-weight-style* ':duration)
```
stylewarning: defvar

src/addresser/addresser-common.lisp
```diff
@@ -0,0 +1,788 @@
...
+                   :working-l2p (copy-rewiring initial-l2p)
+                   :lschedule (make-lscheduler)
+                   :qq-distances (precompute-qubit-qubit-distances chip-spec)
+                   :qubit-cc (a:extremum (chip-spec-live-qubit-cc chip-spec)
+                                         #'>
+                                         :key #'length)
+                   :chip-sched (make-chip-schedule chip-spec)
+                   :chip-spec chip-spec)))
+
+;; TODO: why isn't this part of the state?
```
stylewarning: defvar

src/addresser/addresser-common.lisp
```diff
@@ -0,0 +1,788 @@
...
+(defparameter *addresser-use-free-swaps* nil
+  "Does the addresser treat the initial rewiring as something that can be changed?")
+
+(defgeneric assign-weights-to-gates (state)
+  (:documentation "Generic method for assigning weights to gates, for consumption by COST-FUNCTION."))
+
+(defgeneric cost-function (state &key gate-weights instr)
```
stylewarning: it seems this sometimes returns NIL. Is that true?

Should these be keyword arguments or should they be real args so we can dispatch on `instr`?

This function is super odd.

src/addresser/addresser-common.lisp
```diff
@@ -0,0 +1,788 @@
...
+                                     :instr instr))))
+             (push (copy-rewiring working-l2p) rewirings-tried)
+             (embed-swap (select-cost-lowering-swap working-l2p chip-spec #'cost-function rewirings-tried)
+                         initial-l2p
+                         working-l2p
+                         chip-spec
+                         chip-sched
+                         :use-free-swaps *addresser-use-free-swaps*)
+             rewirings-tried))
+          ;; TODO: reenable this
```
stylewarning: TODO

benchmarking/rewiring-analysis.lisp
```diff
@@ -296,3 +297,14 @@
...
+(defvar *cost-fn-weight-style-assn*
+    (make-assignments
+        ((*random-state* #+sbcl (sb-ext:seed-random-state 1)
+                         #+ecl  (make-random-state 1)
+                         #-(or sbcl ecl) (error "don't know how to seed random state"))
+         (quil::*compressor-passes* 1))
+        (quil::*cost-fn-weight-style*)
+      :duration   (:duration)
+      :fidelity   (:fidelity)
+      ))
```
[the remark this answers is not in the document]
ecpeterson: when in rome

src/addresser/fidelity-addresser.lisp
ecpeterson: I _think_ (but I also forget, since I failed to leave an explanatory comment!) that `*compress-carefully*` interacts poorly with `application-thread-invocation`s, since those expand into instruction sequences but don't themselves have associated matrices to check against.

src/addresser/fidelity-addresser.lisp
ecpeterson: Definitely this is a case of not wanting to push the indentation level further and further up. The first two really belong to the `when`, the rest to a `let*`. I'll break it apart.

src/addresser/fidelity-addresser.lisp
ecpeterson: I'm almost definitely abusing CLOS here. What I want is a (fairly flat) class tree w/ a corresponding tree of custom constructors which all behave like `:after` specializations. This is one way to achieve that, but it's probably not the right one. What's the right one?

src/addresser/temporal-addresser.lisp
ecpeterson: idk "qw n33e3e"

src/addresser/temporal-addresser.lisp
ecpeterson: Not quite: it's a tunable parameter, and tuning it can get the addresser out of deadlocks, but it's very unclear _how_ to successfully tune it, and so we just randomize it across different addresser loop iterations to keep out of trouble.

src/addresser/temporal-addresser.lisp
ecpeterson: The `1q-queues` are populated with single-qubit instructions that we deferred scheduling so that we could inspect 2Q instructions further down the line. If we're calling `append-instr-to-chip-schedule`, then we're actually trying to schedule a (2Q) instruction, which means we can no longer defer any intersecting 1Q instructions—which have to be handled `:before` we handle the requested 2Q instruction.

src/addresser/addresser-common.lisp
ecpeterson: I'm definitely open to reconsidering this function, but I don't think it's possible for it to return NIL, and I do think the docstring makes that clear.

src/addresser/fidelity-addresser.lisp
stylewarning: Why not define `after` methods on all of the classes?

src/chip-specification.lisp
```diff
@@ -369,6 +360,8 @@ used to specify CHIP-SPEC."
           (declare (ignore chip-spec))
           (when (optimal-2q-target-meets-requirements arch ':cz)
             'cnot-to-native-cnots))
+        (constantly 'cz-to-native-czs)
+        #+ignore
```
stylewarning: why is this ignored

tests/compilation-tests.lisp
```diff
@@ -165,4 +165,7 @@ CNOT 0 2"))
     (is (every (link-nativep chip) 2q-code))
     ;; NOTE: Decomposing into fewer 2q gates is more of a regression
     ;; test on quality of compilation, and not on correctness.
-    (is (>= 6 (length 2q-code)))))
+
+    ;; NOTE: This test used to be upper-bounded 6. Changes to the
```
stylewarning: sounds like an `XXX TODO:` not a note; maybe it should warn if it's not < 7?

ecpeterson: seconded

tests/initial-rewiring-tests.lisp
ecpeterson: seconded

ecpeterson: I updated this to `not-signals` instead, but that only makes me think that the bad test case has changed a bit & so this test isn't testing anything anymore. 🤷‍♂

appleby: fwiw I'm fine with removing it if it's not a useful test

ecpeterson: I don't think that the test is bad—it captures a real failure of the naive addresser!—but it's fragile, in the sense that changes to the addresser stand a high chance of changing where the edge cases are, and this test hard-codes an edge case in.

appleby: I'd vote to remove then. imo fragile test == bad test :-)

appleby: I had a feeling it might be fragile when I added it, but wasn't sure... so I won't shed any tears to see it go!

benchmarking/qpu-rewiring/Aspen-32Q-Y.qpu
```diff
@@ -0,0 +1,509 @@
+{
+  "name": "Aspen-4-16Q-A",
```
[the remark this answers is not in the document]
ecpeterson: it won't hurt to update it, but the `name` field is mostly meant for over-the-wire ISA payloads, so it's allowed to be independent of the file name (which might not exist at all)

src/compilation-methods.lisp
```diff
@@ -147,6 +147,8 @@
 ;; Forward declaration from compressor.lisp
 (declaim (special *compressor-passes*))
 
+(defparameter *addresser-state-constructor* 'initial-fidelity-addresser-working-state)
```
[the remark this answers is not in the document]
ecpeterson: I actually am inclined to make this the default for now, though it would be smart to expose some kind of toggle through the application layer. I'm inclined to promote that to a fresh Issue rather than deal with it now.

src/addresser/fidelity-addresser.lisp
ecpeterson: OK, I feel good about this now. Let me know what you think.

ecpeterson:
Reminder that much of this PR is refactor smoke and mirrors: `embed-swap.lisp`, `addresser-common.lisp`, and `temporal-addresser.lisp` are 99% copy/pasted from the old `temporal-addresser.lisp`, with some `defun`s changed to `defmethod`s. Don't be put off by the 3.7k line count.

src/addresser/path-heuristic.lisp
```diff
@@ -77,35 +77,34 @@ rewiring."
...
+    :for gate :being :the :hash-keys :of gates-in-waiting
+    :for tier := (gethash gate gates-in-waiting)
+    :for tier-weight := (expt 0.5d0 tier)
 
-    :for tier :in gates-in-waiting
-    :and tier-index :below 3
-    :and tier-weight := 1d0 :then (* tier-weight 0.5d0)
+    :do (when (and (typep gate 'gate-application)
+                   (= 2 (length (application-arguments gate))))
+          (destructuring-bind (q0 q1) (mapcar #'qubit-index (application-arguments gate))
```
appleby: another chance for `application-qubit-indices` to shine!

honestly, `(mapcar #'quibit-index ...)` is pretty common in quilc sources (probably predating the above function), and not that much longer / harder to read. feel free to leave it.

there is even a duplicate definition called `application-argument-indicies` in `compilation-tests.lisp`  :-)

notmgsk:
> Reminder that much of this PR is refactor smoke and mirrors: `embed-swap.lisp`, `addresser-common.lisp`, and `temporal-addresser.lisp` are 99% copy/pasted from the old `temporal-addresser.lisp`, with some `defun`s changed to `defmethod`s. Don't be put off by the 3.7k line count.

Figuring out where that 1% lies is a PITA.

ecpeterson:
I agree. If we’d had the foresight to do this refactor before doing the
fidelity stuff, we surely would have. Alas: we are merely programmers.

ecpeterson:
@appleby: Good point / idea. Rather than putting in extra lööps to re-run all the existing tests over both addresser styles, I put in a more meager test which exercises the different styles on a single, simple-enough test input.

src/addresser/addresser-common.lisp
```diff
@@ -0,0 +1,763 @@
...
+        ;; for each right vertex...
+        (dotimes (j vertex-count)
+          (let ((ij (aref dist i j))
+                (ik (aref dist i k))
+                (kj (aref dist k j)))
+            ;; is our current best path i->j worse than the concatenation of the
+            ;; current best paths i->k and k->j?
+            (when (> ij (+ ik kj))
+              ;; then replace the old best path with these new ones.
+              (setf (aref dist i j) (+ ik kj))
```
notmgsk: You could totally collapse these `setf`s and impress @jmbr.

src/addresser/addresser-common.lisp
```diff
@@ -0,0 +1,763 @@
...
+         (swap (apply #'build-gate
+                      (permutation-record-operator permutation-record)
+                      '()
+                      (coerce (vnth 0 (hardware-object-cxns hardware-object)) 'list))))
+    (calculate-instructions-fidelity (expand-to-native-instructions (list swap) chip-spec) chip-spec)))
+
+;; nearly ripped straight out of the Wikipedia article for Floyd-Warshall
+(defun precompute-qubit-qubit-distances (chip-spec initial-map)
+  "Implements Floyd-Warshall to compute the minimum weighted distance between any pair of qubits on a CHIP-SPECification, weighted by swap duration."
+  (let* ((vertex-count (length (vnth 0 (chip-specification-objects chip-spec))))
```
notmgsk: `chip-spec-qubits`

src/addresser/addresser-common.lisp
```diff
@@ -0,0 +1,763 @@
...
+(defun next-best-instruction (state instrs)
+  "Find the instruction from the given INSTRS which has the best COST-FUNCTION valuatiuon. Returns three values: the instruction, the index of the hardware object it lies on, and the optional qubit assignments if the selected instruction was only partially assigned."
+  (with-slots (chip-spec chip-sched lschedule) state
+    (let ((best-cost (build-worst-cost state))
+          instr
+          hardware-index)
+      
+      (loop
+        :for (candidate-hardware-index candidate-instr) :in instrs
+        :for physical-qubits := (coerce (vnth 0
```
notmgsk: 🤢

ecpeterson: I agree, but it is what it is.

tests/initial-rewiring-tests.lisp
```diff
@@ -117,7 +117,7 @@ HALT
     ;; It's questionable whether asserting an error is the right thing to do here, but it at least
     ;; documents the current behavior. And if we ever improve NAIVE rewiring so that this doesn't
     ;; error, this test will fail and remind us to update it to NOT-SIGNALS.
```
stylewarning: add follow-on comment, but keep the paper trail!

src/compilation-methods.lisp
```diff
@@ -147,6 +147,8 @@
 ;; Forward declaration from compressor.lisp
 (declaim (special *compressor-passes*))
 
+(defparameter *default-addresser-state-class* 'fidelity-addresser-state)
```
stylewarning: probably want to `DEFVAR` since it's configuration

src/addresser/temporal-addresser.lisp
```diff
@@ -523,548 +308,156 @@ cost-function associated to the current lschedule."
...
+                      (setf (gethash hw (temporal-addresser-state-cost-bounds instance))
+                            (temporal-cost-heuristic-value
+                             (cost-function instance :gate-weights (assign-weights-to-gates instance)))))
+        :finally (setf (addresser-state-initial-l2p instance) old-initial-l2p
+                       (addresser-state-working-l2p instance) old-working-l2p
+                       (addresser-state-logical-schedule instance) old-lscheduler
+                       (addresser-state-chip-schedule instance) old-chip-sched)))
+
+;; also, we randomize the cost function weights during select-and-embed-a-permutation
+(defmethod select-and-embed-a-permutation ((state temporal-addresser-state) rewiring-tried)
```
stylewarning: it probably makes more sense to make this an `:around` method

src/addresser/outgoing-schedule.lisp
```diff
@@ -104,6 +104,18 @@ BEFORE-INST to make use of RESOURCE."
                  (not (resource-subsetp (instruction-resources instr) resource)))
         :maximize (chip-schedule-end-time schedule instr)))
 
+(defun chip-schedule-from-carving-point (schedule resource point)
+  "Returns the (time-ordered) subsequence of instructions from SCHEDULE which intersect RESOURCE and which start on or after the time POINT.
+
+If POINT is generated by CHIP-SCHEDULE-RESOURCE-CARVING-POINT, then the resulting subsequence has a further \"information light-cone\" property: SCHEDULE will have no other instructions which can receive information from RESOURCE from after POINT."
+  (sort
```
stylewarning: add a comment that this might be optimized by computing the `GETHASH`'s once a la
```
(loop :for instr :being the :hash-keys :of ...
        :using (hash-value v)
  ...
      :collect (cons v instr))
```
then you can sort on `:key #'car` and do a `mapcar #'cdr` on the whole affair.

What is this called again? Schwartzian transform or something? It's some perl hack IIRC.

[ No need to actually optimize it though, but it might be low hanging fruit in the future. ]

PR #400, opened 2019-08-31
define-compiler local macro INST

### LINKED ISSUE #335: Repurpose `INST` from `DEFINE-VOP` for `DEFINE-COMPILER`
Bytecode generation in SBCL looks something like this:
```
(define-vop (fast-negate/signed signed-unop)
  (:translate %negate)
  (:generator 2
    (move res x)
    (inst neg res)))
```
and (after #210) compiled circuit fabrication in quilc looks something like this:
```
(define-compiler sort-RX-after-Z
    ((x ("RX" (theta) q))
     (y ("RZ" (pi)  q)))
  (list (build-gate "RZ" `(,pi)       q)
        (build-gate "RX" `(,(- theta)) q)))
```
quilc could benefit from adopting the macro `inst`, which stows an instruction to be eventually output at the termination of the bytecode generation routine, rather than relying on the compiler author to manually manipulate the _list_ of instructions being returned.

@stylewarning brought this up here, where he also provided an example implementation: https://github.com/rigetti/quilc/pull/210#issuecomment-511951521 . Overall, I think this is a great idea, but I think that the exact implementation that we adopt should...

1. ... hew closer to the VOP version than is presented in the example. One instruction at a time is probably fine; we can try to provide an `inst*` if it isn't.
2. ... be terser than in the example. In the VOP version, the `&rest` arguments are used to define the new instruction, and the symbol `neg` is used to designate a particular operation. If we used the example as-is, we'd have to write `(inst (build-gate ...) (build-gate ...) ...)`, which I find distasteful.

Before I get around to working on this issue, I'd like to solicit other advice about how to make this look its prettiest.

ecpeterson:
Introduces the local macros `INST`, `INST*` and `FINISH-COMPILER` within a `DEFINE-COMPILER` body. Makes `DEFINE-COMPILER` look & feel a bit more `DEFINE-VOP`-y, and obviates the compiler author from having to manually track the construction of the instruction sequence.

Things to do:

- [x] `approx.lisp` hasn't yet been converted.
- [x] Deal with topology nativizers: they need manual tracking of instructions and of the expected output table.
- [x] `define-canonical-circuit-approximation`'s use of `predicate` seems broken, unrelated to this PR. Investigate and file an issue.
- [x] I think `define-searching-approximate-template` now discards the fidelity value, which I'd prefer to retain.

Closes #335.


src/build-gate.lisp
```diff
@@ -64,11 +64,14 @@ EXAMPLE: The Quil line \"CPHASE(pi) 2 3\" corresponds to the S-expression (build
                  :gate matrix
                  :arguments (mapcar #'%capture-arg qubits)))
 
+(defun repeatedly-fork (op n)
```
notmgsk: There really ought to be an `iterate` function in `alexandria`.

src/define-compiler.lisp
```diff
@@ -768,65 +768,49 @@ N.B.: This routine is somewhat fragile, and highly creative compiler authors wil
...
-             `(build-gate ,name ,param-list
-                          ,@(loop :for j :below (gate-definition-qubits-needed (lookup-standard-gate name))
-                                  :collect '_)))))
-         (otherwise
-          (make-occurrence-table)))))
+    (inst
+     (case (length body)
+       ;; are we a raw gate?
+       (2
+	(a:plist-hash-table (list (make-gate-binding :operator '_
```
appleby: more funky indentation

src/define-compiler.lisp
```diff
@@ -768,65 +768,49 @@ N.B.: This routine is somewhat fragile, and highly creative compiler authors wil
...
+						     :parameters '_
+						     :arguments nil)
+				  1)
+			    :test #'equalp))
+       ;; we must be a gate description.
+       (otherwise
+	(destructuring-bind (head name param-list &rest qubit-list) body
```
appleby: ditto

src/compilers/euler-compile.lisp
```diff
@@ -29,16 +29,16 @@
...
-                                                    (phase (magicl:ref ,v0 0 0)))))
-                               q)
-                   (build-gate ,inner-gate `(,(* ,inner-prefactor 2 (first ,angles)))
-                               q)
-                   (build-gate ,outer-gate `(,(* ,outer-prefactor
-                                                 (- (phase (magicl:ref ,u1 0 0))
-                                                    (phase (magicl:ref ,u0 0 0)))))
-                               q))))))))
+             (inst ,outer-gate `(,(* ,outer-prefactor
+				     (- (phase (magicl:ref ,v1 0 0))
```
notmgsk: tabs

src/compilers/ucr-explode.lisp
```diff
@@ -110,33 +110,32 @@
               (subseq parameters
                       (/ (length parameters) 2)))
             (averages
-              (mapcar (lambda (x y) (constant
-                                     (/ (+ (constant-value x)
-                                           (constant-value y))
-                                        2)))
-                      high-order-params
-                      low-order-params))
+	      (mapcar (lambda (x y) (constant
```
notmgsk: tabs

src/compilers/ucr-explode.lisp
```diff
@@ -216,52 +211,50 @@
...
                  (string= "RY" roll-type))
-            (build-gates `(("RY"    ,averages    ,target)
-                           ("Z"     ()           ,control)
-                           ("Z"     ()           ,target)
-                           ("RZ"    (,pi/2)      ,target)
-                           ("ISWAP" ()           ,target ,control)
-                           ("RY"    ,differences ,control)
-                           ("ISWAP" ()           ,target ,control)
-                           ("RZ"    (,-pi/2)     ,target))))
+	    (inst "RY"    averages    target)
```
notmgsk: tab

src/define-compiler.lisp
```diff
@@ -857,6 +841,87 @@ N.B.: This routine is somewhat fragile, and highly creative compiler authors wil
...
+
+(defmacro finish-compiler ()
+  "Finish compiling, retuning all instructed emitted with INST.
+
+FINISH-COMPILER is a local macro usable within a compiler body."
+  (error "FINISH-COMPILER can only be used in the body of a compiler."))
+
+(defmacro with-inst (&body body)
+  "Define INST, INST*, and FINISH-COMPILER handlers extending over BODY."
+  (let ((list (gensym "LIST"))
```
appleby: `a:with-gensyms` ?

ecpeterson: Ah! This `let` was in Robert's original comment, and I imagined that he hadn't used `with-gensyms` because he wanted all these things to have recognizable names—but `with-gensyms` supplies an intelligent argument to `gensym` based on the symbol name. Great!

src/compilers/ucr-explode.lisp
```diff
@@ -201,11 +199,7 @@
                                         2)))
                       high-order-params
                       low-order-params)))
-       (labels ((ucr (params args)
-                  (apply #'build-gate op (alexandria:ensure-list params) (alexandria:ensure-list args)))
-                (build-gates (gate-data)
-                  (mapcar (lambda (arg-list) (apply #'build-gate arg-list)) gate-data))
-                (but-last (ell)
+       (labels ((but-last (ell)
```
appleby: `cl:butlast` ?

src/compilers/state-prep.lisp
```diff
@@ -191,7 +191,9 @@
               (setf source-wf (nondestructively-apply-instrs-to-wf prefix-circuit
                                                                    source-wf
                                                                    qubit-complex))
-              (setf abs-so-source (abs (segre-obstruction-2Q source-wf))))))))
+              (setf abs-so-source (abs (segre-obstruction-2Q source-wf)))
+              (dolist (instr prefix-circuit)
+                (inst instr)))))))
```
appleby: File this under Very Small Wins, but this `dolist` idiom appears here and there. Maybe a tiny macro that expands to the loop form would be a small creature comfort.

src/compilers/approx.lisp
```diff
@@ -523,58 +526,58 @@ Additionally, if PREDICATE evaluates to false and *ENABLE-APPROXIMATE-COMPILATIO
                        (give-up-compilation))
                       ;; otherwise, this solution will do.
                       (t
-                       (values (circuit-template ,template-values ,q1 ,q0) (- 1 ,goodness)))))))
```
appleby: Just curious: what happened to `(- 1 ,goodness)` here?

src/define-compiler.lisp
```diff
@@ -633,7 +637,7 @@ N.B.: The word \"shortest\" here is a bit fuzzy.  In practice it typically means
                  (special-cost (occurrence-table-cost (first special-path) target-gateset)))
             ;; did we in fact beat out the generic machinery?
             (when (and (not (path-has-a-loop-p special-path (compiler-bindings compiler)))
-                       (>= special-cost generic-cost))
+                       (> special-cost generic-cost))
```
appleby: I should probably go read the source closely, but I'm lazy and it's easier to ask :): why do we want to maximize cost?

src/define-compiler.lisp
```diff
@@ -768,20 +772,26 @@ N.B.: This routine is somewhat fragile, and highly creative compiler authors wil
...
              (param-list (cond
-                           ((endp param-list)
-                            nil)
-                           ((and (typep param-list 'list)
-                                 (= 2 (length param-list))
-                                 (typep (second param-list) 'list))
-                            (loop :for item :in (second param-list)
+                           ((typep param-list 'list)
+                            (loop :for item :in (case (first param-list)
+                                                  ((quote #+sbcl sb-int:quasiquote)
```
appleby: fancy

src/define-compiler.lisp
```diff
@@ -1043,6 +1101,19 @@ N.B.: This routine is somewhat fragile, and highly creative compiler authors wil
     (expand-bindings bindings nil)))
 
 (defun enact-compiler-options (options body)
+  "Produces code used to enact options passed to DEFINE-COMPILER.
+
+FULL-CONTEXT: Names a symbol to which the active COMPILATION-CONTEXT will be bound.
```
appleby: Was momentarily confused about where these params come from since they don't explicitly appear in the lambda list.  Maybe something like "OPTIONS is a list of zero or more of the following keyword arguments"

src/define-compiler.lisp
ecpeterson: Dangerous, really: I'm not happy that there's a portability concern here.

ecpeterson: perhaps there's a `trivial-syntax-tree`

appleby: Closest thing I could find is

https://gitlab.common-lisp.net/frideau/fare-quasiquote

I *think* I lean towards not pulling in a dependency for this, but I could be convinced otherwise.

Another option is to add feature expressions for any platforms we expect to / dream of supporting, then add a final clause like 

```
#+(not (or sbcl ccl ecl ...whatever else...))
(cerror "Who cares?" "Don't know how to pattern match quasiquote on ~A" (lisp-implementation-type))
```

src/compilers/approx.lisp
```diff
@@ -681,9 +684,16 @@ Additionally, if PREDICATE evaluates to false and *ENABLE-APPROXIMATE-COMPILATIO
...
-                                       a d b q1 q0))))
+	  (let ((canonical-coords (get-canonical-coords-from-diagonal d)))
+	    (destructuring-bind (b0 b1 center-circuit a0 a1)
+		(sandwich-with-local-gates (list (build-gate "CAN" canonical-coords q1 q0))
+					   a d b q1 q0)
+              (declare (ignore center-circuit))
+	      (inst "B0"  (gate-matrix b0) q0)
+	      (inst "B1"  (gate-matrix b1) q1)
+	      (inst "CAN" canonical-coords q1 q0)
+              (inst "A0"  (gate-matrix a0) q0)
```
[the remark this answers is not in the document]
ecpeterson: ugh where do these tabs keep coming from

src/compilers/ucr-explode.lisp
ecpeterson: ‼️

src/compilers/approx.lisp
ecpeterson: It's being handled one macro level up: https://github.com/rigetti/quilc/pull/400/files/185bf02ceb2b58d0fa279be58bd3c4cf3dc31897#diff-448a30484fa4acd3e0877b063434edb6R485 .

appleby: Aha! Ty.

src/compilers/ucr-explode.lisp
appleby: `cl:kitchensink` amiright?

src/compilers/state-prep.lisp
ecpeterson: I have mixed feelings about this. This idiom appears in the following places:

1. Inside of macro code, when percolating one `with-inst` up to another.
2. In compilers that I was too lazy to properly port over to `inst`, like the translators. They could probably be written in a way where they call out to `inst` as they evaluate their recursion.
3. In this situation, which is kind of similar to 1: the circuit is being formed and checked earlier, and it needs to get "percolated up" to another `with-inst` (although here there's no explicit parent `with-inst`).

I think I'd rather fix 2 along the lines indicated here ^, and then this loop becomes sufficiently non-idiomatic that it wouldn't warrant a special name.

src/define-compiler.lisp
ecpeterson: It's a bad name: "cost" here means "estimated fidelity".

src/compilers/state-prep.lisp
appleby: Fair enough

src/define-compiler.lisp
ecpeterson: CC: @stylewarning , @rigettizach .

stylewarning: I would defer to @rigettizach . I'm not confident enough about using fare-quasiquote; but I do feel danger in using SBCL's deal

src/define-compiler.lisp
```diff
@@ -857,6 +839,82 @@ N.B.: This routine is somewhat fragile, and highly creative compiler authors wil
...
 
+;;; here are the forms directly related to DEFINE-COMPILER, whether its
+;;; destructuring or utilities available with its body.
+
+;; within the body of a DEFINE-COMPILER, we publish instructions to output
+;; using a DEFINE-VOP-like INST macro, and we quit a compiler early using
+;; FINISH-COMPILER.
+
+;; first, prohibit the use of these items outside of a compiler body
+(defun inst (&rest xs)
```
stylewarning: Do we have a use-case for defining functions which themselves `inst`, but _those_ functions can only be called during the dynamic extent of the compiler?

If so, we should do something like
```
;; dynamically bound during compilers
(defvar *inst-function*)

(defun inst (...)
  (apply *inst-function* ...))
```

or so. This would allow nicer abstraction where you can write something like
```
(defun inst-xyx (alpha beta gamma)
  (inst "RX" ...)
  ...)
```

and it will still only be valid in the context.

src/define-compiler.lisp
```diff
@@ -857,6 +839,82 @@ N.B.: This routine is somewhat fragile, and highly creative compiler authors wil
...
+
+(defmacro with-inst (&body body)
+  "Define INST, INST*, and FINISH-COMPILER handlers extending over BODY."
+  (a:with-gensyms (list tail compiler-context x xs retval-p)
+    `(block ,compiler-context
+       (let* ((,list (cons nil nil))
+              (,tail ,list))
+         (declare (dynamic-extent ,list)
+                  (type cons ,list ,tail))
+         (labels ((inst (&rest ,xs)
```
stylewarning: If you went with the above idea, you'd still do this, but you'd

```
(let ((*inst-function #'inst))
  ....)
```

with the appropriate gensymming of `inst`

src/define-compiler.lisp
ecpeterson: Possibly. There's some `inst` percolation in the changes in `approx.lisp` where a macro with an implicit `inst` wants to pass its results back out to a parent macro with an implicit `inst`. I think in the "searching" situation I want it to stay as it is, but in the other situation maybe there's a case to be made for this.

src/define-compiler.lisp
ecpeterson: I'm not sold yet, but I know what you mean here.

src/define-compiler.lisp
ecpeterson: This is the situation I was thinking of https://github.com/rigetti/quilc/pull/400/files#diff-448a30484fa4acd3e0877b063434edb6R472 , and I don't see how it'd be made easier with this `inst` extent business. I'll try to think of some other situations too, but my suspicion is that this isn't functionality we need now.

src/define-compiler.lisp
rigettizach: This is, I think, equally fragile, but at least it's not sbcl-specific: ```#.(first (quote `(t)))```. This returns `SB-INT:QUASIQUOTE` on SBCL, `SI:QUASIQUOTE` on ecl, `SYSTEM::BACKQUOTE` on CLISP, and `QUOTE` on Clozure CL.

ecpeterson: It bothers me a little that there's no opportunity for this to loudly fail (vs Mike's suggestion), but I do appreciate that it's at least somewhat flexible. Any objection to installing both this 'generic' expression and a Mike-type feature-guard?

rigettizach: It makes sense to do both.

appleby: Probably obvious, but I just noticed this is happening in a `loop`, so if you do go with a loudly-complaining feature guard, either make it a non-continuable error, or else move it out of the loop body :-).

ecpeterson: I made it a read-time error, since that's when the problem really arises anyhow.

src/define-compiler.lisp
```diff
@@ -1046,17 +1046,24 @@ FINISH-COMPILER is a local macro usable within a compiler body."
...
-                                  :gensym-name "PARAM"
-                                  :seq-accessor 'application-parameters
-                                  :gate-name (compiler-binding-name binding)
-                                  :ele-accessor 'constant-value
-                                  :ele-type 'constant
-                                  :test 'double=)))
+             (cond
+               ((wildcard-pattern-p (gate-binding-parameters binding))
+                rest)
+               ((and (gate-binding-parameters binding) ; no NILs please
```
stylewarning: this reminded me of "amanda please"

src/define-compiler.lisp
```diff
@@ -1046,17 +1046,24 @@ FINISH-COMPILER is a local macro usable within a compiler body."
...
+                         (mapcar #'constant-value (application-parameters ,(compiler-binding-name binding)))))
+                   ,rest))
+               (t
+                (expand-sequence (gate-binding-parameters binding)
+                                 env
+                                 rest
+                                 :gensym-name "PARAM"
+                                 :seq-accessor 'application-parameters
+                                 :gate-name (compiler-binding-name binding)
+                                 :ele-accessor 'constant-value
```
stylewarning: any chance these could be renamed with `elt-type` and `elt-accessor`?

ecpeterson: sure. zach named them, but i can un-name them.

ecpeterson: ok, ready 4 ur re-approval

src/define-compiler.lisp
ecpeterson: 🙃

PR #322, opened 2019-07-11
Install libtweedledum as part of the docker image

jmbr:

PR #409, opened 2019-09-06
Fix startup spam

notmgsk:
Moves the tweedledum load call into `build-app.lisp`.

### DIFF
--- a/app/src/entry-point.lisp
+++ b/app/src/entry-point.lisp
@@ -292,8 +292,6 @@
   (disable-debugger))
 
 (defun entry-point (argv)
-  #-win32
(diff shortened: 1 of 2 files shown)


appleby:
I'm trying to remember why the load was moved to `entry-point` in the first place. Was this required for the barebones package? Or is no longer needed since jmbr's fix in https://github.com/rigetti/quilc/pull/322?

notmgsk:
> I'm trying to remember why the load was moved to `entry-point` in the first place. Was this required for the barebones package? Or is no longer needed since jmbr's fix in #322?

It's still needed. I don't recall the reason why it was in entry-point, however. Looking forward to getting #305 merged, so we can dump tweedledum into the image rather than distribute it as a shared library.

PR #411, opened 2019-09-06
institute package constants pi/2, -pi/2, 2pi, 4pi, -pi

### LINKED ISSUE #275: Create package constants pi/2 and -pi/2
These constants show up all over the code and are often written as, eg, `#.(/ pi 2)`. We should save ourselves the trouble / ugliness and define package-level constants.

(Suggested by Robert here: https://github.com/rigetti/quilc/pull/237#discussion_r290486777 .)

ecpeterson:
This PR and #400 touch a lot of the same code, so I based this one off of it. Merge it first.

Closes #275.


src/compilers/ucr-explode.lisp
```diff
@@ -177,8 +177,6 @@
             (control (first arguments))
             (target (first (last arguments)))
             (rest (rest arguments))
-            (pi/2 (constant (/ pi 2)))
-            (-pi/2 (constant (/ pi -2)))
```
appleby: I assume the answer is no, but these don't need to be ensconced in `constant` anymore?

ecpeterson: They do: `constant` here is a constructor for the struct type `constant` that boxes a constant value in a Quil AST.

appleby: That's what I was confused about. I think I see now: `%capture-param` handles the boxing in `build-gate`.

appleby: Oops, I'm of course looking at the old version of the code without the `INST` changes... anyway: something else is doing the boxing :)

ecpeterson: Oh! I didn't respond to the right question at all. That's right: something else (viz., `build-inst`) is doing the boxing.

stylewarning:
"block constants" @ecpeterson  <3

src/utilities.lisp
```diff
@@ -203,6 +203,11 @@ contains the bits of INTEGER. See http://www.cliki.net/ROTATE-BYTE"
     #-(or ccl ecl sbcl allegro) (error "double-float-positive-infinity not available."))
 
 (a:define-constant pi #.(coerce cl:pi 'double-float))
```
stylewarning: defconstant is OK, and I don't think you need `#.`

ecpeterson: oh, idk, i just copied what was already there. now changed to comport

PR #419, opened 2019-09-17
Refactor expand-to-native-instructions

notmgsk:


src/compressor/compressor.lisp
```diff
@@ -346,61 +346,39 @@ other's."
...
+        :while instruction :do
+          (multiple-value-bind (native-instructions already-native)
+              (expand-instruction-to-native-instructions instruction chip)
+            (cond
+              (already-native
+               (push instruction processed))
+              (t
+               ;; TODO This feels inefficient to calculate on every iteration.
+               ;; TODO What if NATIVE-INSTRUCTIONS is NIL?
+               (check-matrix-consistency (list instruction) native-instructions)
```
notmgsk: You will notice that I am using `instruction` rather than `instructions` to calculate the minimal relabeling. I *think* this is OK (tests pass) but it's possible that I am missing some subtlety.

src/compressor/compressor.lisp
```diff
@@ -346,61 +346,36 @@ other's."
...
-                                     (ilog2 (magicl:matrix-rows mat))))
-                     (kroned-mat (kron-matrix-up mat kron-size))
-                     (kroned-ref-mat (kron-matrix-up ref-mat kron-size)))
-                (assert
-                 (matrix-equality
-                  kroned-ref-mat
-                  (scale-out-matrix-phases kroned-mat kroned-ref-mat)))))
-            ;; expand the set of instructions which need to be processed.
-            (setf instructions (append translation-results (rest instructions))))))))))
+  (loop :with processed := nil
```
ecpeterson: 😎

notmgsk: hopefully not entirely gratuitous :)

src/compressor/compressor.lisp
```diff
@@ -346,61 +346,42 @@ other's."
       ;; when we make it to this point, no rewrite rules apply, so quit.
       (peephole-rewriter-nodes->instrs (peephole-rewriter-node-next head)))))
 
+(defgeneric expand-instruction-to-native-instructions (instr chip)
+  (:documentation "Expand INSTR into a list of instructions native for CHIP.
+
+If INSTR is already native for CHIP, then the return value is (VALUES (LIST INSTR) T). Otherwise, it is (VALUES NATIVE-INSTRUCTIONS NIL) where NATIVE-INSTRUCTIONS is a list of native instructions.")
```
stylewarning: I think this API is a little odd to say the least. Why the second value?

src/compressor/compressor.lisp
```diff
@@ -346,61 +346,42 @@ other's."
...
+    ;; assume it to be a classical instruction which should be treated
+    ;; as native.
+    (values (list instr) t))
+
+  (:method ((instr no-operation) chip)
+    ;; A NO-OPERATION is not native, and should be popped from the
+    ;; instruct list (i.e. does not expand into any instructions).
+    (values nil nil))
+
+  (:method ((instr application) chip)
```
stylewarning: gate-application?

src/compressor/compressor.lisp
```diff
@@ -346,61 +346,42 @@ other's."
       ;; when we make it to this point, no rewrite rules apply, so quit.
       (peephole-rewriter-nodes->instrs (peephole-rewriter-node-next head)))))
 
+(defgeneric expand-instruction-to-native-instructions (instr chip)
+  (:documentation "Expand INSTR into a list of instructions native for CHIP.
+
+If INSTR is already native for CHIP, then the return value is (VALUES (LIST INSTR) T). Otherwise, it is (VALUES NATIVE-INSTRUCTIONS NIL) where NATIVE-INSTRUCTIONS is a list of native instructions.")
+  ;; Note: (list instr) is used below to satisfy one's predilection
+  ;; for symmetry.
+  (:method (instr chip)
```
stylewarning: this seems a little dangerous; what about pragmas? includes? random Common LIsp objects?

src/compressor/compressor.lisp
```diff
@@ -346,61 +346,42 @@ other's."
...
-                (assert
-                 (matrix-equality
-                  kroned-ref-mat
-                  (scale-out-matrix-phases kroned-mat kroned-ref-mat)))))
-            ;; expand the set of instructions which need to be processed.
-            (setf instructions (append translation-results (rest instructions))))))))))
+  (loop :with processed := nil
+        :with instructions := instructions
+        :for instruction := (pop instructions)
+        :while instruction :do
```
stylewarning: I prefer `:until (null instruction)`

src/compressor/compressor.lisp
notmgsk: Yeah you're right. Way over-engineered. I can remove the `values` business, and any conditional logic in the calling function, and just prepend whatever `expand-instructions-to-native-instructions` returns onto the instructions pipeline.

src/compressor/compressor.lisp
ecpeterson: I also think this should be OK: `native-instructions` is only produced from `instruction`, so this local check ought to make just as much sense as the global check as if `instructions` were entirely equal to `(list instruction)`. I think you're fine.

src/matrix-operations.lisp
```diff
@@ -364,3 +364,21 @@ as needed so that they are the same size."
   (loop :for j :below (min (magicl:matrix-rows m)
                            (magicl:matrix-cols m))
         :collect (magicl:ref m j j)))
+
+;; TODO I have the feeling we have too many matrix comparison routines.
```
ecpeterson: There are definitely a handful: variants that take instructions, instruction sequences, raw matrices; variants that compare up to double=, double~; variants that compare across all matrix entries, across all matrix entries after scaling out phases, across just the first column after scaling out phases. It'd be nice to be able to organize these, but so far they've just resisted organization.

src/compressor/compressor.lisp
notmgsk: Well there's no hope of expanding pragmas, includes, control flow, etc. into instructions. I don't think they deserve to be removed from the instruction stream though.

src/matrix-operations.lisp
notmgsk: True, true. I've renamed the function to be a little more descriptive.

notmgsk:
It now occurs to me that `expand-instruction-to-native-instructions` is a very misleading name. Don't merge until I clean it up.

notmgsk:
Requesting a final review, since i changed some stuff. Apologies for the bother.

src/compressor/compressor.lisp
```diff
@@ -346,61 +346,28 @@ other's."
...
-                     (kroned-mat (kron-matrix-up mat kron-size))
-                     (kroned-ref-mat (kron-matrix-up ref-mat kron-size)))
-                (assert
-                 (matrix-equality
-                  kroned-ref-mat
-                  (scale-out-matrix-phases kroned-mat kroned-ref-mat)))))
-            ;; expand the set of instructions which need to be processed.
-            (setf instructions (append translation-results (rest instructions))))))))))
+  "Expand INSTRUCTIONS into a list of instructions that are native for CHIP. Makes no attempt to perform rewiring or simplication."
+  (alexandria:mappend (a:rcurry #'expand-instruction-to-native-instructions chip) instructions))
```
appleby: `(alexandria:mappend (a:rcurry ...`

weird flex but ok.  variety being the spice of life and all (`alexandria:` vs `a:`).

notmgsk: @rigettizach made me feel guilty about that. And this is probably my subconscious trying to make amends.

notmgsk:
> I'm not actually sure what changed, but looks ok to me :-)

I should've made a list, but here we are.

src/compressor/compressor.lisp
appleby: > > I'm not actually sure what changed, but looks ok to me :-)
> 
> I should've made a list, but here we are.

The dark side of the Force-push, if you will.

notmgsk: > > > I'm not actually sure what changed, but looks ok to me :-)
> > 
> > 
> > I should've made a list, but here we are.
> 
> The dark side of the Force-push, if you will.

I won't.

appleby: > > > > I'm not actually sure what changed, but looks ok to me :-)
> > > 
> > > 
> > > I should've made a list, but here we are.
> > 
> > 
> > The dark side of the Force-push, if you will.
> 
> I won't.

understandable.

src/compressor/compressor.lisp
ecpeterson: I still think you could specialize on the type `instruction` without suffering anything major.

PR #420, opened 2019-09-17
Muffle start up warnings, and search more paths for libtweedledum.so

notmgsk:
I previously (#409) tried to clear up the quilc start-up warning spam that was introduced in #276, but that introduced a second issue (in the barebones distribution, the library would not be found).

So this PR covers two things:
* Muffle the warnings that happen at quilc startup. I'm still not sure what is causing these, but they're AFAICT harmless warnings that we can ignore.
* Add more paths to the list of paths that CFFI searches for libraries (this removes the requirement of modifying `LD_LIBRARY_PATH` to pick up the library in the current dir: e.g. `LD_LIBRARY_PATH=./ quilc`).

### DIFF
--- a/app/src/entry-point.lisp
+++ b/app/src/entry-point.lisp
@@ -293,7 +293,8 @@
 
 (defun entry-point (argv)
   #-win32
-  (uiop:symbol-call ':cl-quil.tweedledum '#:load-tweedledum)
+  (handler-bind ((warning #'muffle-warning))
+    (uiop:symbol-call ':cl-quil.tweedledum '#:load-tweedledum))
   (handler-case
       (%entry-point argv)
     (interactive-interrupt (c)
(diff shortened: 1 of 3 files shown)


app/src/entry-point.lisp
```diff
@@ -292,6 +292,9 @@
   (disable-debugger))
 
 (defun entry-point (argv)
+  #-win32
+  (handler-bind ((warning #'muffle-warning))
+    (uiop:symbol-call ':cl-quil.tweedledum '#:load-tweedledum))
```
jmbr: I removed the `handler-bind` form to get a chance to see the warnings but no warnings appeared. Could you please post an example of those?

notmgsk: @jmbr They appear only from the SDK packaged quilc. Find that here: https://downloads.rigetti.com/qcs-sdk/forest-sdk.dmg

PR #426, opened 2019-09-18
[elf-library project] Produce dylibs if on macos

notmgsk:


elf-library/Makefile
```diff
@@ -9,8 +9,14 @@ run_shared:
...
 
-quil_example.out: example.c cl-quil--all-systems.so
-	$(CC) `ecl-config --cflags` -o $@ example.c cl-quil--all-systems.so \
+ifeq ($(shell uname),Darwin)
+LIBEXTENSION := dylib
+else
+LIBEXTENSION := so
+endif
+
+quil_example.out: example.c cl-quil--all-systems.${LIBEXTENSION}
```
notmgsk: Is this an acceptable method of dispatching based on the OS? (i.e. using `${VARIABLE}` in the dependencies)

PR #431, opened 2019-09-20
Exact 2Q state preparation

ecpeterson:
A long while back, Oscar Perdomo provided us with some formulas to perform exact state-to-state transformations between two-qubit systems. This PR implements his techniques.

TODO:
* [x] The case where the source is unentangled and the target is entangled is handled by reversing the state prep instruction, compiling that, and then reversing the output. _Very_ confusingly, this seems to result in a loss of precision. Addressing this is probably a requirement before merging.
* [x] Write to Oscar to announce the inclusion of these methods.
* [x] Change `#+#:ignore` to whatever the informative attic feature is.

### DIFF
--- a/src/compilers/state-prep.lisp
+++ b/src/compilers/state-prep.lisp
@@ -96,7 +96,7 @@
      (* (aref vector 1) (aref vector 2))))
 
 (defun canonicalize-wf (vector)
-  "Calculates a special-orthogonal MATRIX that moves a unit-length VECTOR in C^4 into the form V = (a+bi c 0 0).  Returns (LIST MATRIX V)."
+  "Calculates a special-orthogonal MATRIX that moves a unit-length VECTOR in C^4 into the form V = (a+bi c 0 0).  Returns (VALUES MATRIX V)."
   (let ((matrix (magicl:diag 4 4 (list 1d0 1d0 1d0 1d0)))
         (v (copy-seq vector)))
     ;; start by moving all of the imaginary components into the 0th position.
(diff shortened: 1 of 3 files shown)


jmbr:
Are the notes available somewhere?

src/compilers/state-prep.lisp
```diff
@@ -229,6 +232,109 @@
...
+(define-compiler state-prep-2Q-compiler
+    ((instr (_ _ q1 q0)
+            :where (typep instr 'state-prep-application)))
+  "Exact, optimal compiler for STATE-PREP-APPLICATION instances that target a pair of qubits."
+  (flet ((orthogonal-vector (v)
+           (make-row-major-matrix 2 1 (list (- (conjugate (magicl:ref v 1 0))) (conjugate (magicl:ref v 0 0)))))
+         (normalize-vector (v)
+           (let ((norm (sqrt (+ (expt (abs (magicl:ref v 0 0)) 2)
+                                (expt (abs (magicl:ref v 1 0)) 2)))))
+             (print norm) (terpri)
```
appleby: 👀

src/compilers/state-prep.lisp
```diff
@@ -145,11 +146,236 @@
...
+                                                        0 0 -1 0
+                                                        0 0  0 1))))
+          (setf matrix (m* m matrix))
+          (setf v (nondestructively-apply-matrix-to-vector matrix vector)))))
     (list matrix v)))
 
+(defun state-prep-within-local-equivalence-class (source-wf target-wf)
+  "Produces a circuit that carries a two-qubit wavefunction SOURCE-WF to a two-qubit target wavefunction TARGET-WF under the hypothesis that SOURCE-WF and TARGET-WF have the same Segre obstruction.
 
+Returns a pair (LIST C0 C1) of 2x2 matrices with (C0 (x) C1) SOURCE-WF = TARGET-WF."
```
stylewarning: can you state why you return a pair and not values?

src/compilers/state-prep.lisp
```diff
@@ -145,11 +146,236 @@
...
+                    (d (normalize-vector (m* A b-vector)))
+                    (z (normalize-vector (m* B x-vector)))
+                    (u (normalize-vector (m* B y-vector)))
+                    (theta1 (- pi/2 (phase (+ (sqrt (first lambdas)) (* #C(0 1) (sqrt (second lambdas)))))))
+                    (theta2 (- pi/2 (phase (+ (sqrt (first deltas)) (* #C(0 1) (sqrt (second deltas)))))))
+                    (U1 (make-row-major-matrix 2 2 (list (magicl:ref c 0 0) (magicl:ref d 0 0)
+                                                         (magicl:ref c 1 0) (magicl:ref d 1 0))))
+                    (U2 (make-row-major-matrix 2 2 (list (magicl:ref z 0 0) (magicl:ref u 0 0)
+                                                         (magicl:ref z 1 0) (magicl:ref u 1 0))))
+                    (V1 (make-row-major-matrix 2 2 (mapcar #'conjugate
```
stylewarning: this mapcar is extremely @ecpeterson

src/compilers/state-prep.lisp
ecpeterson: Yeah: the two operators don't make sense in isolation. Elsewhere in the code, I've used returning `values` to mean "you might need the other return values, but perhaps not" and returning a `list` to mean "you definitely need all the values, so you'd better be ready to destructure them". It helps me typecheck receiving code that I write.

stylewarning: can you comment this in the code as a nice reminder?

is it too much to `defstruct foo (:type list)` ?

ecpeterson: I pushed a version that uses `adt:defdata`.

PR #449, opened 2019-10-10
feature: addresser subclassing

ecpeterson:
Here's an attempt at a refactoring of the addresser that allows for easier specialization of its behavior, as well as a starting point on a fidelity-based such subclass.

It's a work in progress—but so is the PR it's being merged into.

Some ugly regions that could stand improvement:

* The way the addresser gets invoked is not very flexible: there are a few distinct function calls. It would be better to get CLOS involved in this, so that `compiler-hook` doesn't need to be manually modified if we want to switch addressing methods.
* `chip-specification` sometimes carries extra data used by a particular addressing method to make more intelligent decisions, like the `time-bound` key inside of various `hardware-object`'s `misc-data` slots. It would probably be better to make this part of (the subclasses of) `addresser-state`, since which addresser cares about what is really specific to the style of addressing being used. (But it's also conceivable that the compressor will want to see that same information and optimize along the same cost function...?) (I don't think the `cost-bound` field gets populated at all (or correctly?) for the fidelity addresser!)
* I did a bunch of monkey-patching around Corwin's contributions to the addresser (viz., the various bits of SWAP-searching), which involved passing different "versions" of the parent cost-function around with different argument specializations and so on. It's gross, and it sort-of breaks the abstraction of `addresser-state`.


tests/compiler-hook-tests.lisp
```diff
@@ -477,7 +477,7 @@ MEASURE 1
 (deftest test-clever-CCNOT-depth-reduction ()
   "Test that the ':GREEDY-QUBIT swap selection strategy brings CZ depth down to optimal for CCNOT."
   (let ((p (quil::compiler-hook (quil::parse-quil "
-PRAGMA INITIAL_REWIRING \"GREEDY\"
+PRAGMA INITIAL_REWIRING \"PARTIAL\"
```
appleby: Does it fail now with GREEDY?

src/compilation-methods.lisp
```diff
@@ -248,8 +248,9 @@ Returns a value list: (processed-program, of type parsed-program
 
          (touch-unpreserved-block (blk registrant)
            ;; actually process this block
-           (multiple-value-bind (initial-l2p chip-schedule final-l2p)
-               (do-greedy-temporal-addressing
+           (multiple-value-bind (chip-schedule initial-l2p final-l2p)
+               ;; TODO: here, and below, we need to find a way to not hard-code this method call
+               (do-greedy-fidelity-addressing
```
appleby: `(funcall *addressing-function* ...)`  ?

🏃 💨

tests/initial-rewiring-tests.lisp
```diff
@@ -117,6 +117,7 @@ HALT
     ;; It's questionable whether asserting an error is the right thing to do here, but it at least
     ;; documents the current behavior. And if we ever improve NAIVE rewiring so that this doesn't
     ;; error, this test will fail and remind us to update it to NOT-SIGNALS.
+    #+ignore
```
appleby: No longer failing, or just sufficiently annoying to disable? If the former, maybe flip it to `not-signals`.

src/addresser/embed-swap.lisp
```diff
@@ -0,0 +1,210 @@
...
+;;;;
+;;;; Author: Eric Peterson, Erik Davis
+
+(in-package #:cl-quil)
+
+;;; This file contains routines related to the selection and insertion of a
+;;; SWAP into a schedule, usually as part of migrating an active addresser
+;;; instance from one state to another.
+
+(defparameter *addresser-swap-search-type* :greedy-qubit
```
notmgsk: > quote

notmgsk: Or don't. Who cares

src/addresser/embed-swap.lisp
```diff
@@ -0,0 +1,210 @@
...
+the qubit-qubit distance array QQ-DISTANCES."
+  (loop :for i :across (rewiring-l2p rewiring)
+        :for j :across (rewiring-l2p target-rewiring)
+        :when i
+          :sum (aref qq-distances i j)))
+
+(defun cost-lowering-candidates (rewiring cost-function rewirings-tried chip-spec depth)
+  "Given a rewiring and a cost function, returns a list of swap links for which
+the cost of the rewiring is reduced."
+  (let ((best-cost-so-far nil)
```
notmgsk: explicit `nil` vs implicit

src/addresser/fidelity-addresser.lisp
```diff
@@ -0,0 +1,169 @@
...
+    ;; calculate log-infidelity coming from INSTR, using recombination:
+    ;;   + calculate the cut point
+    ;;   + calculate the infidelity of instructions since the cut point
+    ;;   + use only the additional infidelity penalty
+    (when (and instr (typep instr 'gate-application))
+      (let* ((chip-spec (addresser-state-chip-specification state))
+             (instruction-expansion
+               (let ((*compress-carefully* nil))
+                 (expand-to-native-instructions (list instr) chip-spec))))
+        (setf instr-cost (- (log (calculate-instructions-fidelity instruction-expansion chip-spec))))
```
notmgsk: `decf`

src/addresser/fidelity-addresser.lisp
```diff
@@ -0,0 +1,169 @@
...
+    (when gate-weights
+      (let ((qq-distances (addresser-state-qq-distances state)) ; populated with log-infidelities
+            (rewiring (addresser-state-working-l2p state))
+            (assigned-qubits nil)
+            (gate-count 0))
+        (dohash ((gate tier-index) gate-weights)
+          (when (and (<= tier-index 3)
+                     (typep gate 'application))
+            (let* ((logical-qubits (mapcar #'qubit-index (application-arguments gate)))
+                   (physical-qubits (mapcar (a:curry #'apply-rewiring-l2p rewiring) logical-qubits))
```
notmgsk: nice

src/addresser/fidelity-addresser.lisp
```diff
@@ -0,0 +1,169 @@
...
+                   (any-assigned? (some #'identity physical-qubits))
+                   (all-assigned? (every #'identity physical-qubits)))
```
notmgsk: imagine the efficiency improvements if we did `(and any-assigned? (every #'identity physical-qubits))` save TENS of cycles

src/addresser/fidelity-addresser.lisp
```diff
@@ -0,0 +1,169 @@
...
+                                 (rewired-instrs
+                                  (loop :for instr :in instrs
+                                        :for rewired-instr := (copy-instance instr)
+                                        :do (setf (application-arguments rewired-instr)
+                                                  (mapcar (constantly (qubit (first physical-qubits)))
+                                                          (application-arguments rewired-instr)))
+                                        :collect rewired-instr))
+                                 (fidelity (calculate-instructions-fidelity rewired-instrs chip-spec)))
+                     (incf gate-count)
+                     (incf gate-weights-cost (- (log fidelity))))))
```
notmgsk: `decf`

src/addresser/fidelity-addresser.lisp
```diff
@@ -0,0 +1,169 @@
...
+                     (let ((qq-distance (aref qq-distances p0 p1)))
+                       ;; we're using 2^(-depth) * (1 + 2^(1-dist)) so that distant
+                       ;; qubits exert weaker forces than nearby ones, encouraging
+                       ;; us to execute more quickly accomplishable gates sooner.
+                       ;; it's totally possible that dist alone is a good cost fn
+                       ;; on its own, and we should experiment with this.
+                       (assert (not (= qq-distance most-positive-fixnum)) ()
+                               "Multiqubit instruction requested between ~
+                                   disconnected components of the QPU graph: ~
+                                   ~A ."
```
notmgsk: trailing space

notmgsk: could also use `~/instruction-fmt/` or w/e it is

src/addresser/fidelity-addresser.lisp
```diff
@@ -0,0 +1,169 @@
...
+(defmethod assign-weights-to-gates ((state fidelity-addresser-state))
+  (multiple-value-bind (max-value value-hash)
+      (lscheduler-walk-graph (addresser-state-logical-schedule state)
+                             :base-value 0
+                             :bump-value (lambda (instr value)
+                                           (cond
+                                             ((typep instr 'gate-application)
+                                              (case (length (application-arguments instr))
+                                                (1
+                                                 (+ 1/10 value))
```
notmgsk: Why `1/10`?

src/addresser/logical-schedule.lisp
```diff
@@ -660,7 +609,8 @@ mapping instructions to their tags. "
             :do (setf fidelity
                       (fidelity-combinator (fidelity-bumper instr 1d0)
                                            (* fidelity (gethash instr value-hash)))) 
-            :finally (return (exp (- (sqrt (- (log fidelity))))))))))
+            :finally (return (values (exp (- (sqrt (- (log fidelity)))))
+                                     value-hash))))))
```
notmgsk: Could you document the return values in the docstr

src/addresser/temporal-addresser.lisp
```diff
@@ -523,550 +302,118 @@ cost-function associated to the current lschedule."
...
+;;; method overrides that augment the common addresser to deal with 1Q queues ;;;
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+
+;; flush the 1Q lines in the relevant events
+(defmethod dequeue-classical-instruction :before
+    ((state temporal-addresser-state) instr &optional dry-run-escape)
+  ;; handle our 1Q queues
+  (when dry-run-escape  ; every classical instruction can be handled
+      (funcall dry-run-escape))
+    (with-slots (lschedule chip-spec chip-sched working-l2p) state
```
notmgsk: weird indent

src/addresser/fidelity-addresser.lisp
ecpeterson: `decf`ing by a negative number makes my stomach turn

notmgsk: it would be by a positive number

notmgsk: Oh I see what you mean. Still, `(decf place amount)` looks better than `(decf place (- amount))` IMO. Agree to disagree

ecpeterson: `fidelity` is less than 1, so `(log fidelity)` is less than `0`

notmgsk: Hm. I guess I *can* see it from your side. `incf` makes it clear that we really do mean to increase the value.

src/addresser/fidelity-addresser.lisp
ecpeterson: i'll see if you remember to re-make this comment when we merge the actual fidelity addresser PR into master

src/addresser/logical-schedule.lisp
ecpeterson: nope: this is a dangling change that i'm going to revert instead

src/addresser/fidelity-addresser.lisp
ecpeterson: No good reason :( It'd be nice to deduce this value from some statistics in the chip-schedule—but there's no obvious way to do so, since the lscheduler is populated by logically addressed instructions.

ecpeterson: I'll make it a named magic constant instead of an unnamed magic constant.

tests/initial-rewiring-tests.lisp
ecpeterson: It's no longer failing, but I'm not sure if that's supposed to be the case or not. For the moment, I'm leaving in `#+ignore` as an eyesore, so that when the parent PR of this one finally settles out, we remember to really take a hard look at it.

src/compilation-methods.lisp
ecpeterson: Yeah, something like this... but I think I'd prefer to dispatch on `state` again? I'm not sure.

tests/compiler-hook-tests.lisp
ecpeterson: Yeah, puzzlingly. Again, this is a deliberate eyesore so that we circle back to it when doing the real clean-up.

src/addresser/embed-swap.lisp
```diff
@@ -0,0 +1,210 @@
...
+(defparameter *addresser-swap-search-type* :greedy-qubit
+  "The type of swap search the addresser should use.
+
+GREEDY-PATH: Assign links values based on whether they are on the shortest path
+between two qubits that need to be adjacent
+A*: Use A* search algorithm using the cost function as a heuristic
+GREEDY-QUBIT: Greedily choose the best link to swap according to the cost function.")
+
+(defparameter *addresser-move-to-rewiring-swap-search-type* :a*
+  "The type of swap search the addresser should use when doing move-to-rewiring. ")
```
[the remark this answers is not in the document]
ecpeterson: RIP Corwin

PR #453, opened 2019-10-16
Fix docstring typos for UPDATE-PRINT-PARSED-PROGRAM-GOLDEN-FILES

appleby:
And submit to the no-newlines-in-docstrings convention 😞.

Also, capitalize "pairs" in section heading of the golden files comment in utilities.lisp.

### DIFF
--- a/tests/utilities.lisp
+++ b/tests/utilities.lisp
@@ -193,7 +193,7 @@ If SUFFIX-P is non-nil, suffix the returned string with DELIMITER."
 ;;; other means. Just keep in mind the above comments about trailing newlines.
 ;;;
 ;;;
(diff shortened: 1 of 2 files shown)


tests/utilities.lisp
```diff
@@ -193,7 +193,7 @@ If SUFFIX-P is non-nil, suffix the returned string with DELIMITER."
 ;;; other means. Just keep in mind the above comments about trailing newlines.
 ;;;
 ;;;
-;;; ADDING NEW INPUT/OUTPUT pairs
+;;; ADDING NEW INPUT/OUTPUT PAIRS
```
kilimanjaro: THAT WAS A BIT odd

PR #456, opened 2019-10-21
Quilt extension (redux)

kilimanjaro:
This extends `cl-quil` with support for `quilt` instructions. The earlier PR was getting a bit unweildy. The most substantial change present here is that I have moved the Quilt code to a separate package, `cl-quil/quilt` (with tests in `cl-quil/quilt-tests`). This is entirely optional. The mechanism by which this is provided is twofold:

- Generic methods in `cl-quil` have been exposed, and `cl-quil/quilt` defines additional methods on these.
- The parser in `cl-quil` now has `*lexer-extensions*` and `*parser-extensions*` lists, which provide hooks for language extensions. 

I have also added a number of tests.

Here are the previously mentioned features/changes of `cl-quil/quilt`.

- AST objects for new quilt instructions and primitives (e.g. frames, waveforms)
- Extend `parser.lisp` to support quilt (this includes a few minor quality of life improvements as well)
- Refactor the high level behavior of `parse-quil`. Here we process includes for a list of "raw" quil instructions, then parse to an "unresolved program", and then resolve objects. 
- When converting raw quil to an unresolved program, we check and handle ambiguous definitions.
- Object resolution (previously application resolution) now includes things like frames and waveforms, which must have their definitions resolved.
- `expand-calibrations` is a new pass which uses much of the circuit expansion machinery (which has been moved to `expansion.lisp`). The behavior here is to produce simple quilt with all gate applications, measurements, etc resolved.
- The `type-check` pass has been extended to verify memory references in `CAPTURE` and `RAW-CAPTURE`
- An additional pass, `fill-delays`, can turn a quilt program with "empty" time intervals and return a program with an entire block of time accounted for (via additional `DELAY` instructions). Some of the machinery in here is written to also be useful for the pulse tracing qvm.

Anticipated remaining work:
-  the primary language feature that is left unaddressed here is how to support affine transformations of waveforms (or the equivalent, so that one can `CAPTURE` on a "projected" iq value). This will be a separate PR.
- quilc app may need some additional work on error reporting (e.g. a nice message if unexpected quilt instructions show up)


src/parser.lisp
```diff
@@ -110,7 +118,8 @@
    (return (tok ':CONTROLLED)))
   ((eager #.(string #\OCR_FORK))
    (return (tok ':FORKED)))
-  ("INCLUDE|DEFCIRCUIT|DEFGATE|MEASURE|LABEL|WAIT|NOP|HALT|RESET|JUMP\\-WHEN|JUMP\\-UNLESS|JUMP|PRAGMA|NOT|AND|IOR|MOVE|EXCHANGE|SHARING|DECLARE|OFFSET|XOR|NEG|LOAD|STORE|CONVERT|ADD|SUB|MUL|DIV|EQ|GT|GE|LT|LE|CONTROLLED|DAGGER|FORKED|AS|MATRIX|PERMUTATION"
+  ;; both Quil and Quilt keywords are handled here
+  ("INCLUDE|DEFCIRCUIT|DEFGATE|MEASURE|LABEL|WAIT|NOP|HALT|RESET|JUMP\\-WHEN|JUMP\\-UNLESS|JUMP|PRAGMA|NOT|AND|IOR|MOVE|EXCHANGE|SHARING|DECLARE|OFFSET|XOR|NEG|LOAD|STORE|CONVERT|ADD|SUB|MUL|DIV|EQ|GT|GE|LT|LE|CONTROLLED|DAGGER|FORKED|AS|MATRIX|PERMUTATION|PULSE|CAPTURE|RAW\\-CAPTURE|DELAY|FENCE|NONBLOCKING|DEFWAVEFORM|DEFCAL|DEFFRAME|SET\\-FREQUENCY|SET\\-PHASE|SHIFT\\-PHASE|SET\\-SCALE|SWAP\\-PHASE"
```
appleby: any way to shovel this into a `*lexer-extensions*` analog of `*parser-extensions*` ?

src/parser.lisp
```diff
@@ -961,19 +906,23 @@ result of BODY, and the (possibly null) list of remaining lines.
              (parse-arithmetic-tokens entry :eval t)))
     (mapcar #'simplify entries)))
 
-(defun parse-gate-entries (tok-lines)
+;;; This is used for both Quil and Quilt
+(defun parse-gate-or-waveform-entries (tok-lines &key require-indent)
```
appleby: would it make sense for this to have a generic name like `parse-indented-entries` or something and use `*parse-context*` to disambiguate in error messages?

src/analysis/resolve-objects.lisp
```diff
@@ -0,0 +1,98 @@
...
+;;; inside of a circuit or calibration.
+(defvar *in-definition-body* nil)
+
+(defgeneric resolve-instruction (instr parsed-program)
+  (:method ((instr unresolved-application) parsed-program)
+    (macrolet
+        ((assert-and-print-instruction (test-form &optional places datum &rest arguments)
+           `(assert ,test-form
+                    ,places
+                    (format nil "Error in resolving ~/quil:instruction-fmt/: ~a."
```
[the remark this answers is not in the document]
kilimanjaro: Seems reasonable

src/cl-quil.lisp
```diff
@@ -1,42 +1,147 @@
...
+;;; definitions (where the basic check is whether the region names are equal),
+;;; to a bit more involved for Quilt calibrations (where applicability is
+;;; generally determined by a pattern matching system, and so we have to
+;;; consider the arity and values of parameters and arguments).
+
+(defgeneric definition-signature (instr)
+  (:documentation "Computes a signature for a Quil definition such that if two definitions are equivalent for the purposes of name resolution, then their signatures are EQUALP.")
+  (:method ((instr gate-definition))
+    (list 'gate-or-circuit-definition
+          (intern (gate-definition-name instr))
```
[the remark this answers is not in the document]
kilimanjaro: Probably the latter is preferable if one anticipates the possibility of cross-extension signature collisions.

src/parser.lisp
kilimanjaro: Strongly agree with this

src/parser.lisp
kilimanjaro: Good q. This is a trickier business. With the parser, the action which `parse-program-lines` takes is based on a simple dispatch on token type, and so it's no big deal to add extra options here. For the lexer, the way that `alexa:define-string-lexer` works is that the the action taken by the resulting lexer is based on the longest substring match, with ties resolved by selecting the first of the options (as written in the the lexer definition). It's a little less obvious as to how to compose these in general, so I avoided it at first.

However, _assuming_ that the new syntax is just in the form of new keywords (this is the case for Quilt, and isn't too big of a compromise in general), I think a reasonable option is to hook in to the part of the lexer responsible for handling identifiers. So I'm updating this part of `line-lexer`
```
  ("{{IDENT}}"
   (return
     (cond
       ((string= "pi" $@) (tok :COMPLEX (constant pi)))
       ((string= "i" $@)  (tok :COMPLEX (constant #C(0.0d0 1.0d0))))
       (t                 (tok :NAME $@)))))
``` 
to check `*lexer-extensions*` before defaulting to `(tok :NAME $@)`.

appleby: Do you get enough context to resolve ambiguities in that case? Like if some one did you dirty and named a label `@DEFWAVEFORM` or something

kilimanjaro: Well, even before it was ok for a keyword to be a substring of a longer token. E.g. the lexer allowed for things like `@DEFGATE` or `FOODEFGATEBAR`, with the first being a valid label name and the second a valid name.

kilimanjaro: The general method for resolving ambiguities is just that it finds the longest match, and of these the first is selected. So `DEFGATE-DEFGATE` (or `DEFWAVEFORM-DEFWAVEFORM`) tokenizes to a name. This does feel a bit gross!

appleby: I'm probably just misunderstanding. What I meant was, if you're hooking into the `{{IDENT}}` rule, will the `*lexer-extensions*` (however that is handled) have enough context to know that `@DELAY` isn't a quilt keyword, and fallback to lexing it as an IDENT? (since the label rule is specified above as `\\@{{IDENT}}`).

Slightly less far-fetched example with param instead of label:

```
DEFCIRCUIT DELAYQ0(%DELAY):
	DELAY 0 %DELAY
```

appleby: Yep, I was confused. I see now that `{{IDENT}}` is a regex. I misunderstood the alexa syntax and thought this rule was defining how to parse the `{{IDENT}}` non-terminal, but I see now that is defined above as `:ident  some-regex-string`, and that this is just a separate parse rule.  The cases for `pi` and `i` should have tipped me off.

src/quilt/analysis/expand-calibrations.lisp
```diff
@@ -0,0 +1,350 @@
...
+  (let ((remake nil))
+    (flet ((instantiate-parameter (assoc)
+             (destructuring-bind (name . value) assoc
+               (let ((instantiated-value (quil::ensure-instantiated value param-value)))
+                 (cond ((eq value instantiated-value)
+                        assoc)
+                       (t
+                        (setf remake t)
+                        (cons name instantiated-value)))))))
+      (let ((updated-alist (mapcar #'instantiate-parameter
```
appleby: `cl-quil::flag-on-update` ?

src/quilt/analysis/fill-delays.lisp
```diff
@@ -0,0 +1,223 @@
...
+                  :when (plusp lag)
+                    :do (incf (local-time clocks f) lag)
+                    :and :collect (make-instance 'delay-on-frames
+                                                 :frames (list f)
+                                                 :duration (constant lag))))))))
+
+(defun fill-delays (parsed-program &key (omit-fences t) (synchronize-at-end t))
+  "Introduce any implicit DELAY instructions in the Quilt program PARSED-PROGRAM.
+
+If OMIT-FENCE is T, then FENCE instructions will be removed from the resulting program.
```
appleby:
```suggestion
If OMIT-FENCES is T, then FENCE instructions will be removed from the resulting program.
```

src/quilt/analysis/resolve-objects.lisp
```diff
@@ -0,0 +1,141 @@
...
+;;;; Author: Erik Davis
+
+(in-package #:cl-quil/quilt)
+
+(defun validate-waveform-parameters (waveform-ref expected-parameters)
+  "Determines whether the waveform reference WAVEFORM-REF has parameter names conforming to the list of EXPECTED-PARAMETERS."
+  (let ((actual (mapcar (a:compose #'param-name #'car)
+                        (waveform-ref-parameter-alist waveform-ref))))
+    (a:when-let ((missing (set-difference expected-parameters actual :test #'string=)))
+      (quil-parse-error "Expected parameter ~A in waveform ~A."
```
appleby:
```suggestion
      (quil-parse-error "Expected parameters ~{~A~^, ~} in waveform ~A."
```

src/quilt/analysis/resolve-objects.lisp
```diff
@@ -0,0 +1,141 @@
...
+                        (first missing)
```
appleby:
```suggestion
                        missing
```
contingent on above edit

src/quilt/analysis/type-safety.lisp
```diff
@@ -0,0 +1,56 @@
...
+                      (raw-capture-frame instr))))
+    (quil::enforce-mref-bounds mref mdesc)
+    (adt:match quil-type (memory-descriptor-type mdesc)
+      (quil-real
+       (a:if-let ((samples (raw-capture-num-real-samples instr)))
+         (if (> samples (quil::mref-available-length mref mdesc))
+             (quil-type-error "RAW-CAPTURE instruction target ~/quil:instruction-fmt/ must be a REAL ~
+                                 vector of length no less than ~A."
+                              mref
+                              samples))
```
appleby: round `samples` up to next largest integer here?

src/quilt/cl-quilt.lisp
```diff
@@ -0,0 +1,117 @@
...
+;;;; Author: Erik Davis
+
+(in-package #:cl-quil/quilt)
+
+(defmethod definition-signature ((instr waveform-definition))
+  (cons 'waveform-definition
+        (intern (waveform-definition-name instr) :keyword)))
+
+;;; When computing signatures for calibrations, we must account for the fact
+;;; that parrameter and argument names don't matter (e.g. `DEFCAL MEASURE q` is
```
appleby:
```suggestion
;;; that parameter and argument names don't matter (e.g. `DEFCAL MEASURE q` is
```

src/quilt/cl-quilt.lisp
```diff
@@ -0,0 +1,117 @@
...
+(defmethod definition-signature ((instr frame-definition))
+  (list 'frame-definition
+        (frame-hash (frame-definition-frame instr))))
+
+(defun raw-quilt-to-unresolved-program (code)
+  "This constructs a PARSED-PROGRAM object from the given list of Quilt instructions CODE, without any resolution of applications.
+
+This also signals ambiguous definitions, which may be handled as needed."
+  ;; Note: the processing below preserves the order of definitions.
+  (let ((gate-defs nil)
```
appleby:
```suggestion
  (let ((gate-defs '())
```

nitpick

src/parser.lisp
appleby: ^ In hindsight, not sure what I was thinking here since I was obviously confusing lexing vs parsing and thought I was looking at yacc grammar or something, whoops.

src/ast.lisp
```diff
@@ -1406,6 +1465,20 @@ For example,
             (application-arguments instr)))
 
   ;; The following are not actually instructions, but who cares.
+
+  (:method ((defn memory-descriptor) (stream stream))
+    (format stream "DECLARE ~a ~a"
```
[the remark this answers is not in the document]
appleby: oof owie ouch, my precious `~A`s

src/analysis/resolve-objects.lisp
```diff
@@ -0,0 +1,98 @@
...
+           instr)))))
+
+  (:method ((instr t) parsed-program)
+    (declare (ignore parsed-program))
+    instr))
+
+(defgeneric resolve-objects (unresolved-program)
+  (:documentation "Perform all object resolution within UNRESOLVED-PROGRAM, returning a PARSED-PROGRAM.")
+  (:method ((unresolved-program parsed-program))
+    ;; For straight quil, we need to resolve UNRESOLVED-APPLICATIONS. For Quilt,
```
stylewarning: capital Quil

src/parser.lisp
```diff
@@ -4,6 +4,31 @@
 
 (in-package #:cl-quil)
 
+;;; There are three main steps on the journey from Quil string to Quil program,
```
stylewarning: ty for docs

src/quilt/analysis/expand-calibrations.lisp
```diff
@@ -0,0 +1,350 @@
+;;;; expand-calibrations.lisp
+;;;;
+;;;; Author: Erik Davis
+
+(in-package #:cl-quil/quilt)
```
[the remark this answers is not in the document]
stylewarning: this is preferred, the dot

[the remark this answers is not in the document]
stylewarning: (but keep `/` for the system name)

src/quilt/analysis/resolve-objects.lisp
```diff
@@ -0,0 +1,141 @@
...
+      (resolve-frame (swap-phase-left-frame instr) defns)
+      (resolve-frame (swap-phase-right-frame instr) defns)))
+  instr)
+
+(defmethod resolve-instruction ((instr delay-on-qubits) parsed-program)
+  (declare (ignore parsed-program))
+  instr)
+
+(defmethod resolve-instruction ((instr delay-on-frames) parsed-program)
+  (unless quil::*in-definition-body*
```
stylewarning: I think we should file an issue that we write documentation on how to add a language extension. If I didn't know the code, I would never know that I need to do this binding. We don't need it for this PR, but could you add it and assign yourself?

src/quilt/analysis/type-safety.lisp
```diff
@@ -0,0 +1,56 @@
...
+        (* 2                            ; real, imag
+           (constant-value (raw-capture-duration instr))
+           (constant-value (frame-definition-sample-rate frame-defn)))
+        nil)))
+
+;; CAPTURE must target a REAL[2]
+(defmethod type-check-instr ((instr capture) memory-regions)
+  (let* ((mref (capture-memory-ref instr))
+         (mdesc (quil::find-descriptor-for-mref mref memory-regions)))
+    (quil::enforce-mref-bounds mref mdesc)
```
stylewarning: We should make a `quil.internal` that imports symbols from Quil and exports them publicly

src/quilt/analysis/expand-calibrations.lisp
kilimanjaro: There was this fussiness of the value instantiated living in a cons cell, but overall I suppose it's preferred to adopt a uniform mechanism like `flag-on-update` (since getting this wrong here would be a pernicious bug...).

src/quilt/analysis/type-safety.lisp
kilimanjaro: Nice catch. I'm going to compute the ceiling in `raw-capture-num-real-samples`. 

FWIW, what to do when there is a fractional number of samples (i.e. when pulse duration is not a multiple of the underlying frame frequency) is a bit ambiguous at the moment, and for all practical purposes isn't addressed until downstream in `pidgin` (where depending on options things get padded or raise an exception). Leaving this as a later issue for me to resolve (since I need to adopt a compatible strategy here).

src/quilt/analysis/type-safety.lisp
kilimanjaro: I created an issue for this.

tests/quilt/good-test-files/good-overloading-of-parameter-names.quil
```diff
@@ -0,0 +1,4 @@
+DEFCIRCUIT DELAYQ0(%DELAY):
+    DELAY 0 %DELAY
+
+DELAYQ0(1.0)
```
appleby: 😊 this is right up there with the time my "zonk" testcase made into SBCL

tests/quilt/good-test-files/good-quilt-defframe.quil
```diff
@@ -0,0 +1,11 @@
+DEFFRAME 0 1 "foo"
+
+DEFFRAME 0 1 "bar":
+    SAMPLE-RATE: 1.0
+
+DEFFRAME 0 1 "baz":
+    INITIAL-FREQUENCY: 1e8
+
+DEFFRAME 0 "frob":
```
appleby: I love a good `frob` as much as the next person, but [according to the spec](http://www.lispworks.com/documentation/HyperSpec/Body/01_daf.htm), this should clearly be `quux`.

tests/quilt/good-test-files/good-quilt-frame-mutations.quil
```diff
@@ -0,0 +1,9 @@
+DEFFRAME 0 "xy"
+DEFFRAME 0 1 "cz"
+
+SET-PHASE 0 "xy" pi/2
```
appleby: get crazy. throw a `cos` in there

src/ast.lisp
```diff
@@ -1503,47 +1610,24 @@ Examples:
              (terpri stream)))
       (map nil #'print-one-line seq))))
 
+(defgeneric print-parsed-program-generic (parsed-program stream)
```
ecpeterson: What's happening here? Why did this become generic?

src/parser.lisp
ecpeterson: I'd even appreciate something in the formal `docs` directory.

src/quilt/parser.lisp
```diff
@@ -0,0 +1,618 @@
...
+    ((:SAMPLE-RATE)
+     (parse-sample-rate value-toks))
+    ((:INITIAL-FREQUENCY)
+     (let ((freq (quil::parse-parameter-or-expression value-toks)))
+       (unless (and (constantp freq)
+                    (plusp (constant-value freq)))
+         (quil-parse-error "Expected INITIAL-FREQUENCY to be a positive real, but got ~/quil:instruction-fmt/."
+                           freq))
+       freq))
+    (otherwise
```
ecpeterson: In the interest of forward compatibility, I'd much rather this warn than error.

src/quilt/parser.lisp
```diff
@@ -0,0 +1,618 @@
...
+  "Parse a CAPTURE instruction from TOK-LINES. Returns the instruction object, and the remaining lines."
+  (quil::match-line ((op :CAPTURE) &rest rest-toks) tok-lines
+    (multiple-value-bind (frame rest-toks)
+        (parse-frame rest-toks)
+      (unless (= 1 (length (frame-qubits frame)))
+        (quil-parse-error "CAPTURE instruction is only applicable to single-qubit frames."))
+      (when (endp rest-toks)
+        (quil-parse-error "CAPTURE instruction is missing waveform reference."))
+      (multiple-value-bind (waveform-ref rest-toks)
+          (parse-waveform-ref rest-toks)
```
ecpeterson: This should parse an affine family of kernels, not just a single kernel reference.

src/ast.lisp
kilimanjaro: This will be covered in https://github.com/rigetti/quilc/issues/463 but the short of it is that Quilt has its own subclass of `parsed-program`, with some additional slots (e.g. for calibrations), and printing is generic so that we can handle both Quil and Quilt programs, along with any other possible language extensions.

src/analysis/expansion.lisp
```diff
@@ -0,0 +1,352 @@
...
+        (funcall transform obj)
+        obj)))
+
+;;; To REMAKE or not to REMAKE
+;;; that is the question....
+;;;
+;;; Basically, in circuit or calibration bodies some instructions might depend
+;;; on circuit parameters or arguments. For those, instantiation necessarily
+;;; must make a new instruction (since this may happen more than once). However,
+;;; in the absence of this we can avoid allocation and just return the original.
```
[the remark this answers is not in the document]
kilimanjaro: Doubtful, but the precedent for this sort of penny-pinching had already been set, and at the other extreme is unrestrained deepcopying across passes, which is also not very desirable.

src/quilt/parser.lisp
kilimanjaro: I mentioned this above in the PR overview (under "Anticipated remaining work"), but is a TODO item which I prefer to delay until after this PR is merged. The decision to omit from the present PR is due to the already large amount of changes + the amount of effort people have put into reviewing (which seems to grow quadratically w/ code size).

src/analysis/expansion.lisp
```diff
@@ -0,0 +1,352 @@
...
+(defmacro flag-on-update (flag op)
+  "Given a unary function OP, lift to a function which applies OP and sets FLAG to T if the result differs from the input."
+  (a:with-gensyms (x new-x)
+    `(lambda (,x)
+       (let ((,new-x (funcall ,op ,x)))
+         (unless (eq ,x ,new-x)
+           (setf ,flag t))
+         ,new-x))))
+
+(defgeneric instantiate-instruction (instr param-value arg-value)
```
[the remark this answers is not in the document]
kilimanjaro: I do not know

src/ast.lisp
```diff
@@ -10,6 +10,10 @@
   "A qubit address."
   (index nil :type unsigned-byte))
 
+(defun qubit= (x y)
+  "Do the qubits X and Y have equal indices?"
+  (= (qubit-index x) (qubit-index y)))
```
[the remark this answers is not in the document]
kilimanjaro: Good idea.

src/quilt/parser.lisp
kilimanjaro: Seems reasonable.

src/analysis/type-safety.lisp
```diff
@@ -30,10 +30,13 @@
       (memory-name (setf (memory-name-descriptor mref) desc)))
     desc))
 
+(defun mref-available-length (mref memory-descriptor)
```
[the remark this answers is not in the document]
kilimanjaro: Mostly this dumb little computation is used for a few checks both here and in the Quilt `type-safety.lisp`, so I'd prefer to keep it around. But I agree that the name could use some work, so I changed it to `(defun memory-segment-length (memory-descriptor &key offset) ...)`. I also added a docstring as well as appropriate `check-type`s.

src/quilt/waveform.lisp
```diff
@@ -0,0 +1,136 @@
...
+    `(progn
+       (defclass ,class-name (standard-waveform)
+         ,slot-specs
+         ,@(when documentation
+             `((:documentation ,documentation))))
+
+       (setf (gethash ,quilt-name *quilt-to-waveform-class*) ',class-name)
+       (setf (gethash ',class-name *waveform-class-parameter-alists*) ',quilt-param-alist)
+
+       nil)))
```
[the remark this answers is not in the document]
kilimanjaro: Agreed, I changed to `prog1` for this

tests/printer-test-files/gold-regex/plundered-regex-gold.quil
```diff
@@ -22,7 +22,6 @@ DEFGATE H1:
     0\.707\d+, 0\.707\d+
     0\.707\d+, -0\.707\d+
 
-
```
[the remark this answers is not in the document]
appleby: ^ btw, this was supposed to be "joke" referencing [this comment](https://github.com/rigetti/quilc/blob/28748216a3b4d0b33ad5ad7b2410baf16e9d736a/tests/printer-tests.lisp#L65) (in case that wasn't obvious). please don't feel the need to scrutinize each whitespace change! marking resolved.

src/quilt/cl-quilt.lisp
appleby: not sure if this fell through the cracks or what. feel free to mark resolved again if intentionally skipped

kilimanjaro: The strange thing is that I recall making this change at some point. It's starting to be to late to just shrug this off as spooky 👻! Regardless, I've updated it now.

src/quilt/parser.lisp
```diff
@@ -0,0 +1,618 @@
...
+                        rate))
+    (constant rate)))
+
+(defun parse-frame-definition-property-value (property-name value-toks)
+  "Parse and validate a value associated with PROPERTY-NAME. Consumes VALUE-TOKS, and returns the parsed value."
+  (case property-name
+    ((:SAMPLE-RATE)
+     (parse-sample-rate value-toks))
+    ((:INITIAL-FREQUENCY)
+     (let ((freq (quil::parse-parameter-or-expression value-toks)))
```
appleby: Is the intention here to allow arithmetic expressions as long as they eval to a constant? If so, the length check on line 358 below prevents something like `INITIAL-FREQUENCY: 1e6/2`

The [spec](https://github.com/rigetti/quil/blob/master/rfcs/analog/spec_changes.md) seems to indicate that any `Expression` is valid, but e.g. `parse-sample-rate` on line 318 explicitly disallows compound expressions...

```
DefFrame :: DEFFRAME Frame (: FrameSpec+ )?
FrameSpec :: Indent FrameAttr : Expression
FrameAttr :: SAMPLE-RATE | INITIAL-FREQUENCY
```

I'm not sure what my point is, just raising this issue for your consideration

kilimanjaro:
> 42/52 files viewed.
> 
> saved by the tiny test files!
> 
> should they have a `.quilt` file extension? asking the important questions.

For now, I am going to leave them as `.quil`. This is a legitimate question though, and I am curious for thoughts on the matter. Given that we are treating Quilt as something of a "language extension", another option is that there could be a pragma line indicating the use of this extension.

src/ast.lisp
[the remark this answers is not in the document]
kilimanjaro: Back by popular demand!

src/parser.lisp
kilimanjaro: I think that is a good suggestion.

src/quilt/cl-quilt.lisp
appleby: día de muertos isn't until tomorrow here, so as far as I'm concerned spooky references are still fair game

tests/printer-test-files/gold-regex/plundered-regex-gold.quil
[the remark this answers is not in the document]
kilimanjaro: Ha! I didn't get the joke, so I was waiting for an opportune moment to make an excuse for not scrutinizing.

src/quilt/parser.lisp
```diff
@@ -0,0 +1,618 @@
...
+
+;;; Frame Definitions
+
+(defun parse-frame-definition (tok-lines)
+  "Parse a frame definition from TOK-LINES."
+  (when (null tok-lines)
+    (quil-parse-error "EOF reached when frame definition was expected."))
+  (destructuring-bind (defframe-tok . frame-toks) (first tok-lines)
+    (unless (eql ':DEFFRAME (quil::token-type defframe-tok))
+      (quil-parse-error "DEFFRAME exoected. Got ~A."
```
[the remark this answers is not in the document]
kilimanjaro: you are a careful reaoer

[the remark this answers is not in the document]
appleby: and yet I average 4 typos / slack message

src/quilt/parser.lisp
```diff
@@ -0,0 +1,618 @@
...
+    (constant rate)))
+
+(defun parse-frame-definition-property-value (property-name value-toks)
+  "Parse and validate a value associated with PROPERTY-NAME. Consumes VALUE-TOKS, and returns the parsed value."
+  (case property-name
+    ((:SAMPLE-RATE)
+     (parse-sample-rate value-toks))
+    ((:INITIAL-FREQUENCY)
+     (let ((freq (quil::parse-parameter-or-expression value-toks)))
+       (unless (and (constantp freq)
```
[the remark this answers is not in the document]
kilimanjaro: !

src/quilt/parser.lisp
```diff
@@ -0,0 +1,618 @@
...
+                    (plusp (constant-value freq)))
+         (warn "Expected INITIAL-FREQUENCY to be a positive real, but got ~/quil:instruction-fmt/."
```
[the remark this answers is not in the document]
kilimanjaro: Agreed. I'm breaking it out into two checks. The frequency definitely must be real, otherwise an error is signal. The requirement that it is positive is a different story: it might meaningfully be negative, so this is a corner case that I need to think about wrt to Pidgin. For now I am going to `warn` about that, and am leaving myself a TODO item to get to the bottom of this.

src/quilt/cl-quilt.lisp
```diff
@@ -0,0 +1,117 @@
...
+
+This also signals ambiguous definitions, which may be handled as needed."
+  ;; Note: the processing below preserves the order of definitions.
+  (let ((gate-defs '())
+        (circ-defs '())
+        (memory-defs '())
+        (wf-defs '())
+        (calib-defs '())
+        (frame-defs '())
+        (exec-code '())
```
jmbr: Suggestion: create an empty instance of `parsed-quilt-program` and a generic function to update it whose methods dispatch on the type of `instr` to avoid/encapsulate the `typecase` in lines 70-77 below. Then return the (now) well-formed instance replacing lines 79-87.

src/quilt/cl-quilt.lisp
```diff
@@ -0,0 +1,117 @@
...
+
+(defun read-quilt-file (filespec)
+  "Read the Quilt file designated by FILESPEC, and parse it as if by PARSE-QUIL."
+  (check-type filespec (or string pathname))
+  (assert (and (uiop:file-pathname-p filespec)
+               (uiop:file-exists-p filespec))
+          (filespec)
+          "FILESPEC must be a path to a file that exists")
+  (parse-quilt (a:read-file-into-string filespec)
+               :originating-file (first (directory filespec))))
```
jmbr: There is a race condition here. The file `filespec` may exist at the time the `assert` form is evaluated and be gone (deleted, for example) by the time `parse-quilt` is run. If you discard the `assert` and instead wrap `parse-quilt` in a `handler-case` dealing with the file not existing condition (`sb-ext:file-does-not-exist`), etc., then  your code will not be vulnerable to the race condition.

src/quilt/package.lisp
```diff
@@ -0,0 +1,130 @@
...
+   #:frame-name                         ; READER
+   #:frame-qubits                       ; READER
+   #:frame=                             ; FUNCTION
+   #:frame-hash                         ; FUNCTION
+   #:frame-intersects-p                 ; FUNCTION
+   #:frame-on-p                         ; FUNCTION
+
+   #:waveform-ref                       ; STRUCTURE
+   #:waveform-ref-name                  ; READER
+   #:waveform-ref-args                  ; READER
```
jmbr: This is not declared. Did you mean `waveform-ref-parameter-alist`?

src/quilt/package.lisp
```diff
@@ -0,0 +1,130 @@
...
+   #:capture-frame                      ; READER
+   #:capture-waveform                   ; READER
+   #:capture-memory-ref                 ; READER
+
+   #:raw-capture                        ; CLASS
+   #:raw-capture-frame                  ; READER
+   #:raw-capture-duration               ; READER
+   #:raw-capture-memory-ref             ; READER
+
+   #:nonblocking-p                      ; READER
```
jmbr: This is declared in as an accessor, not a reader.

src/quilt/package.lisp
```diff
@@ -0,0 +1,130 @@
...
+
+   #:delay                              ; CLASS
+   #:delay-duration                     ; READER
```
jmbr: This is declared in as an accessor, not a reader.

src/quilt/package.lisp
```diff
@@ -0,0 +1,130 @@
...
+   #:delay-on-frames                    ; CLASS
+   #:delay-frames                       ; READER
+   #:delay-on-qubits                    ; READER
+   #:delay-qubits                       ; READER
```
jmbr: These are declared as accessors, not readers.

src/quilt/package.lisp
```diff
@@ -0,0 +1,130 @@
...
+
+   #:fence                              ; CLASS
+   #:fence-qubits                       ; READER
```
jmbr: This is declared in as an accessor, not a reader.

src/quilt/package.lisp
```diff
@@ -0,0 +1,130 @@
...
+
+   #:simple-frame-mutation              ; ABSTRACT CLASS
+   #:set-frequency                      ; CLASS
+   #:set-phase                          ; CLASS
+   #:shift-phase                        ; CLASS
+   #:set-scale                          ; CLASS
+   #:frame-mutation-target-frame        ; READER
+   #:frame-mutation-value               ; READER
```
jmbr: Accessors.

src/quilt/package.lisp
```diff
@@ -0,0 +1,130 @@
...
+
+   #:swap-phase                         ; CLASS
+   #:swap-phase-left-frame              ; READER
+   #:swap-phase-right-frame             ; READER
```
jmbr: Accessors.

jmbr:
When running the `cl-quil` test suite using `(asdf:test-system :cl-quil/quilt)`, I get a number of errors such as:
```
The value
  #S(PARAM :NAME "theta")
is not of type
  CL-QUIL:CONSTANT
   [Condition of type TYPE-ERROR]
```
All of the errors are of the form "The value [...] is not of type [...]" where the types are either `cl-quil:constant`, `cl-quil:qubit`, `cl-quil:memory-ref`, or `float`.

Charging ahead with the `continue` restart on all of those errors, I end up with the following output:
```
CL-QUIL.QUILT-TESTS (Suite)
  TEST-PARSING-GOOD-QUILT-TEST-FILES
    Testing good file good-circuit-expansion-formal-params
......................................[FAIL]
  TEST-PARSING-BAD-QUILT-TEST-FILES
    Testing bad file bad-quilt-default-waveform-missing-parameter
    Testing bad file bad-quilt-defframe
    Testing bad file bad-quilt-missing-frame-definition
    Testing bad file bad-quilt-waveform-indentation
    Testing bad file bad-unresolved-waveform
    Testing bad file delay-missing-frame-definition
.......................................[ OK ]
  TEST-PARSER-FAIL-WITHOUT-EXTENSIONS.....................................[ OK ]
  TEST-QUILT-DEFWAVEFORM-SAMPLE-RATE......................................[ OK ]
  TEST-PRINT-QUILT-OBJECTS................................................[ OK ]
  TEST-PARSE-AND-PRINT-QUILT-INSTRUCTIONS.................................[ OK ]
  TEST-PARSE-AND-PRINT-QUILT-DEFINITIONS..................................[FAIL]
  TEST-DEFINITION-SIGNATURE...............................................[FAIL]
  TEST-GATE-CALIBRATION-MATCHING..........................................[ OK ]
  TEST-MODIFIED-CALIBRATION-MATCHING......................................[ OK ]
  TEST-RECURSIVE-CALIBRATION..............................................[ OK ]
  TEST-INFINITELY-RECURSIVE-CALIBRATIONS..................................[ OK ]
  TEST-MEASUREMENT-CALIBRATION-MATCHING...................................[FAIL]
  TEST-STRICT-CALIBRATION-EXPANSION.......................................[FAIL]
  TEST-CASE-SENSITIVE-CALIBRATION-EXPANSION...............................[FAIL]
  TEST-QUILT-CIRCUIT-EXPANSION............................................[FAIL]
  TEST-QUILT-NAME-RESOLUTION..............................................[ OK ]
  TEST-QUILT-DURATION.....................................................[FAIL]
  TEST-CAPTURE-TYPE-SAFETY................................................[FAIL]
  TEST-RAW-CAPTURE-TYPE-SAFETY............................................[ OK ]
  TEST-FRAME-SET-LOGIC....................................................[ OK ]
  TEST-QUILT-INSTRUCTION-DURATION.........................................[FAIL]
  TEST-FILL-DELAYS-SWAP-IMPLICIT-SYNCHRONIZATION..........................[FAIL]
  TEST-FILL-DELAYS-FENCE-EXPANSION........................................[FAIL]

T
```

src/quilt/package.lisp
stylewarning: I'm OK if the intended export is as a READER, even if care wasn't taken (or convenience was desired) in not defining a separate writer

src/quilt/parser.lisp
```diff
@@ -0,0 +1,618 @@
...
+                               "NONBLOCKING"
+                               "DEFWAVEFORM"
+                               "DEFCAL"
+                               "DEFFRAME"
+                               "SET-FREQUENCY"
+                               "SET-PHASE"
+                               "SHIFT-PHASE"
+                               "SET-SCALE"
+                               "SWAP-PHASE")))
+    (if (member str quilt-token-strings :test #'string=)
```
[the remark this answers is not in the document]
stylewarning: pollute KW => mem leak

src/quilt/package.lisp
kilimanjaro: Nice catch

src/quilt/parser.lisp
kilimanjaro: This is an oversight on my part. I think the rejection of compound expressions (e.g. by `parse-sample-rate`) reflects the best of intentions, and I would like it to stay for now. I will also for the time being impose this constraint on `INITIAL-FREQUENCY`. I do not want to encourage arithmetic expressions here, since the actual precision is relevant, and frankly these are just going to be magic numbers supplied by a database somewhere.

With respect to the spec, I will update it to match this (barring any complaints from the committee).

jmbr:
> > When running the Quilt test suite using `(asdf:test-system :cl-quil/quilt)`, I get a number of errors such as:
> 
> Nevermind. This went away after erasing the contents of `$HOME/.cache/common-lisp` and recompiling.

Turns out I can reliably reproduce it using:
```
(declaim (optimize safety))
(asdf:test-system :cl-quil/quilt :force t)
```
This seems to be SBCL being prickly about types with a compilation policy with high `safety` values.

src/quilt/parser.lisp
```diff
@@ -0,0 +1,618 @@
...
+      (pop (first lines))
+      (return-from parse-frame-definition-body
+        (values argument-plist lines)))
+    (when (< 3 (length line))
+      (quil-parse-error "DEFFRAME body lines should be of the form <property>: <value>."))
+    (destructuring-bind (property-tok col-tok &rest value-toks) line
+      (unless (eql ':NAME (quil::token-type property-tok))
+        (quil::disappointing-token-error property-tok "an identifier"))
+      (unless (eql ':COLON (quil::token-type col-tok))
+        (quil::disappointing-token-error col-tok "a colon"))
```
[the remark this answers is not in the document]
kilimanjaro: It's a bit annoying here. I wrote `match-line` to ease the common case of parsing an instruction: one consumes a single line, returning some AST object and the remaining lines. It's less friendly here, since it introduces an implicit `values` form is at odds with the tail call.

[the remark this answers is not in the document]
jmbr: I see.

src/quilt/parser.lisp
```diff
@@ -0,0 +1,618 @@
...
+  "Parse a <param>:<value> pair, consuming TOKS."
+  (let ((name (first toks))
+        (colon (second toks))
+        (value-expr (rest (rest toks))))
+    (unless (and name
+                 colon
+                 value-expr
+                 (eql ':NAME (quil::token-type name))
+                 (eql ':COLON (quil::token-type colon)))
+      (quil-parse-error "Waveform parameters must be provided by name. Exected <name>: <value>."))
```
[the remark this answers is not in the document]
kilimanjaro: Good call

kilimanjaro:
> > > When running the Quilt test suite using `(asdf:test-system :cl-quil/quilt)`, I get a number of errors such as:
> > 
> > 
> > Nevermind. This went away after erasing the contents of `$HOME/.cache/common-lisp` and recompiling.
> 
> Turns out I can reliably reproduce it using:
> 
> ```
> (declaim (optimize safety))
> (asdf:test-system :cl-quil/quilt :force t)
> ```
> 
> This seems to be SBCL being prickly about types with a compilation policy with high `safety` values.

This is a good observation. I've updated the declared slot types to fix this. The biggest culprit were the `standard-waveform` classes defined in `waveform.lisp`. Here the declared `:type` was actually a reference to the RPCQ spec. I've changed these to be `:rpcq-type` now (the actual RPCQ type specifiers are not currently used, but are informative and may be used in the future).

src/quilt/analysis/fill-delays.lisp
```diff
@@ -0,0 +1,223 @@
...
+     (delay-on-qubits
+      (loop :for defn :in (parsed-program-frame-definitions parsed-program)
+            :for frame := (frame-definition-frame defn)
+            :when (frame-on-p frame (delay-qubits instr))
+              :collect frame))
+     (fence
+      (loop :for defn :in (parsed-program-frame-definitions parsed-program)
+            :for frame := (frame-definition-frame defn)
+            :when (frame-intersects-p frame (fence-qubits instr))
+              :collect frame)))))
```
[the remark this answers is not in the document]
kilimanjaro: It's not clear that much is gained or lost. Mostly I preferred the `typecase` in this instance because the resulting code is a bit more compact.

cl-quil.asd
```diff
@@ -194,3 +195,44 @@
...
+
+(asdf:defsystem #:cl-quil/quilt
+  :description "Quil language extensions for pulse level quantum control."
+  :license "Apache License 2.0 (See LICENSE.txt)"
+  :maintainer "Rigetti Computing"
+  :author "Rigetti Computing"
+  :depends-on (#:cl-quil)
+  :in-order-to ((asdf:test-op (asdf:test-op #:cl-quil/quilt-tests)))
+  :around-compile (lambda (compile)
+                    (let (#+sbcl(sb-ext:*derive-function-types* t))
```
notmgsk: missing space in `#+sbcl(sb-ext` 😷

tests/quilt/analysis-tests.lisp
```diff
@@ -0,0 +1,203 @@
...
+    SWAP-PHASE q \"xy\" q \"ff\"
+    PULSE q \"xy\" flat(iq: 1.0, duration: %theta)
+    CAPTURE q \"xy\" flat(iq: 1.0, duration: %theta) iq[0]
+    RAW-CAPTURE q \"xy\" %theta iq[0]
+    DELAY q %theta
+    DELAY q \"xy\" %theta
+    FENCE q
+
+FOO(1.0) 0"
+                         :transforms '(quil::expand-circuits)))
```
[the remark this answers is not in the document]
notmgsk: @appleby that's a funny typo IMO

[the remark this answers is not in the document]
appleby: `without-output-to-quiln't`

src/quilt/parser.lisp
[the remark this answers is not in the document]
kilimanjaro: 11/1/19 never forget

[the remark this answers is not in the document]
appleby: 🇺🇸

PR #460, opened 2019-10-25
Fix tests

notmgsk:


notmgsk:
Heads up: this slows things down drastically, and that should be fixed before considering merging into *master*.

PR #466, opened 2019-11-04
Use "quilc" rather than argv[0] for syslog name

notmgsk:
As is done in QVM. Could alternatively sanitise the name, but that's more work for little gain.

### DIFF
--- a/app/src/entry-point.lisp
+++ b/app/src/entry-point.lisp
@@ -305,17 +305,6 @@
       (format *error-output* "~&! ! ! Error: ~A~%" c)
       (uiop:quit 1))))
 
-(defun clamp-syslog-name (name)
-  "Produces a name like \"<A>...<B>\" where <A> is a 15-char chunk
-from the beginning of the name, and <B> is a 30-char chunk from the
-end of the name."
-  (let ((len (length name)))
-    (if (<= len 48)
-        name
-        (format nil "~A...~A"
-                (subseq name 0 15)
-                (subseq name (- (length name) 30))))))
-
 (defun %entry-point (argv)
   (let ((*program-name* (pop argv)))
     (command-line-arguments:handle-command-line
@@ -410,7 +399,7 @@ end of the name."
 
   (setf *logger*
         (make-instance 'cl-syslog:rfc5424-logger
-                       :app-name (clamp-syslog-name *program-name*)
+                       :app-name "quilc"
                        :facility ':local0
                        :maximum-priority *log-level*
                        :log-writer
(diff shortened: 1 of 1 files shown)


ecpeterson:
Whatever you think is best, it all sounds fine by me. (Re: internal discussion, I don't understand how this could be considered a compromise in security. Does it... have the potential to reveal read-protected directory structure?? (Not typically, since syslog messages are only available to administrators.))

appleby:
> Whatever you think is best, it all sounds fine by me. (Re: internal discussion, I don't understand how this could be considered a compromise in security. Does it... have the potential to reveal read-protected directory structure?? (Not typically, since syslog messages are only available to administrators.))

I'm not sure exactly what they had in mind either. Information leakage is certainly one possibility. Normally only admins have access to syslog, but if an attacker were to gain access, they might learn something about the system you didn't want them to. There are probably more interesting things in the logs than quilc's `argv[0]`, but who knows? Maybe you invoked quilc as `./super-secret-gubment-project/quilc` and now they realize you are a juicy target. Or maybe they found a bug in syslog that they can exploit by tricking someone into running quilc from a directory they control, etc, etc.

Often the idea is just that every user controlled/controllable input is a potential attack surface, and therefore anything that doesn't need to be user input shouldn't be. It's about switching from a mindset of "I don't see how this could be exploited", to a mindset of "I'm not close to certain this *couldn't* be exploited, so mitigate the risk if possible." I never fully internalized this philosophy myself, but I can see where the security folks are coming from. To be fair, the best exploits are almost by definition something you would not have guessed. Try as you might, you will probably not foresee the exact Lemony-Snicket's-series-of-unfortunate-events that some buffer-overflow-breathing ebpf-bytecode-injecting cold-blooded psychopath will manage to pull off to exploit that poor innocent `argv[0]` that was minding it's own business. So from that perspective, a default policy of paranoia around security issues makes sense.

jmbr:
> > Whatever you think is best, it all sounds fine by me. (Re: internal discussion, I don't understand how this could be considered a compromise in security. Does it... have the potential to reveal read-protected directory structure?? (Not typically, since syslog messages are only available to administrators.))
> 
> I'm not sure exactly what they had in mind either. Information leakage is certainly one possibility. Normally only admins have access to syslog, but if an attacker were to gain access, they might learn something about the system you didn't want them to. There are probably more interesting things in the logs than quilc's `argv[0]`, but who knows? Maybe you invoked quilc as `./super-secret-gubment-project/quilc` and now they realize you are a juicy target. Or maybe they found a bug in syslog that they can exploit by tricking someone into running quilc from a directory they control, etc, etc.
> 

This is the most likely scenario, in my opinion: an attacker launches `quilc` from `/this/extremely/long/directory/that/triggers/a/buffer/overflow/in/whatever/software/is/used/to/monitor/logs/...` where the path contains a suitable shellcode [1]. If the monitoring software used to manage the logs is vulnerable to a buffer overflow and the path name overruns, say, the stack, then this could give the attacker control of the machine were the monitoring code is executed (usually some computer deep in the network operations center).

[1] https://nets.ec/Alphanumeric_shellcode

jmbr:
Some examples of things gone awry when letting users control `argv[0]`: https://cve.mitre.org/cgi-bin/cvekey.cgi?keyword=%22argv%5B0%5D%22

PR #469, opened 2019-11-06
XY

ecpeterson:
Adds "XY" as a synonym for "PISWAP", as well as support for a native gateset built out of "XY/2".

### DIFF
--- a/src/compilers/translators.lisp
+++ b/src/compilers/translators.lisp
@@ -155,6 +155,27 @@
   (inst "RZ"   '(#.(/ pi -4)) q1)
   (inst "CNOT" ()             q0 q1))
 
+(define-compiler XY-to-XY/2 ((XY-gate ("XY" (theta) q1 q0)))
+  (inst "RZ" `(,(/ pi 4))  q0)
+  (inst "RZ" `(,(/ pi -4)) q1)
+  (inst "XY" `(,pi/2)      q1 q0)
+  (inst "RZ" `(,(param-*  0.5d0 theta)) q0)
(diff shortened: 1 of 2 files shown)


src/compilers/translators.lisp
```diff
@@ -155,6 +155,27 @@
   (inst "RZ"   '(#.(/ pi -4)) q1)
   (inst "CNOT" ()             q0 q1))
 
+(define-compiler XY-to-XY/2 ((XY-gate ("XY" (theta) q1 q0)))
+  (inst "RZ" `(,(/ pi 4))  q0)
```
deanna-abrams: Hmm, my signs are different. I get 
```
  (inst "RZ" `(,(/ pi -4))  q0)
  (inst "RZ" `(,(/ pi 4)) q1)
  (inst "XY" `(,pi/2)      q1 q0)
  (inst "RZ" `(,(param-*  -0.5d0 theta)) q0)
  (inst "RZ" `(,(param-* 0.5d0 theta)) q1)
  (inst "RZ" `(,pi/2)      q0)
  (inst "RZ" `(,-pi/2)     q1)
  (inst "XY" `(,pi/2)      q1 q0)
  (inst "RZ" `(,(/ pi -4)) q0)
  (inst "RZ" `(,(/ pi  4)) q1))
```

stylewarning: @ecpeterson

ecpeterson: I saw, and I haven't checked yet, but I'm not too concerned: you can imagine taking this sequence and applying complex conjugation to get a bunch of different signs, or flipping which of the to XY/2s is playing the role of -XY/2.

PR #473, opened 2019-11-08
exponentiation of Paulis

stylewarning:
Adds a routine not unlike `exponential_map` in pyQuil. Depends on #472.

Not sure if I got the signs right and will require some careful math review.

### DIFF
--- a/src/clifford/pauli.lisp
+++ b/src/clifford/pauli.lisp
@@ -33,7 +33,12 @@
                                 :initial-element 0))
 
   (defun phase-factor (p)
+    "Return the phase factor as an integer. This integer can be interpreted as a complex number via the function INTERPRET-PAULI-PHASE-FACTOR."
     (aref (pauli-components p) 0))
+
+  (defun interpret-pauli-phase-factor (n)
+    "Interpret the phase of a Pauli term (as given by PHASE-FACTOR) as a complex number."
+    (expt #C(0 1) n))
 
   (defun base4-list (p)
     (coerce (subseq (pauli-components p) 1) 'list))
@@ -194,8 +199,7 @@ phase factor)."
       (prog1 (%make-pauli :components c)
         (loop :for i :from 1
               :while s
-              :do (setf (aref c i) (sym-to-base4 (pop s)))
-              ))))
+              :do (setf (aref c i) (sym-to-base4 (pop s)))))))
 
   (defun pauli-from-string (s)
     "Make Pauli operator from a string, e.g., \"iXYZ\", \"-iZ\", \"Z\"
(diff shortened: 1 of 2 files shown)


src/clifford/pauli.lisp
```diff
@@ -33,8 +33,13 @@
                                 :initial-element 0))
 
   (defun phase-factor (p)
+    "Return the phase factor as an integer. This integer can be interpreted as a complex number via the function INTERPRET-PAULI-PHASE-FACTOR."
     (aref (pauli-components p) 0))
 
+  (defun interpret-pauli-phase-factor (n)
+    "Interpret the phase of a Pauli term (as given by PHASE-FACTOR) as a complex number."
+    (expt #C(1 0) n))
```
jmbr: `#C(1 0)`?

stylewarning: no way, i got that wrong!

stylewarning:
I'm really not sure about the case of an identity operator. I copied what we do in pyQuil. I'm also not sure I'm handling identity factors correctly within the more general case.

kilimanjaro:
This is a bit cryptic to me. Regardless, I would also like to have a matching set of unit tests.

Considering the power series of exp(X), for a matrix P satisfying P^2 = I one has exp(-itP) = cos(t)I - i sin(t)P. What's not clear to me is how this translates to a product of H, RX, CNOT gates.

stylewarning:
requesting math review from @eric @jmbr @kilimanjaro

src/clifford/pauli.lisp
```diff
@@ -323,3 +327,68 @@ hash-function."
...
+;;;     structure" by Seeley, Richard, and
+;;;     Love. (https://arxiv.org/pdf/1208.5986.pdf)
+(defun exp-pauli (p x)
+  "Compute exp(-ixp) as a list of Quil instructions."
+  (let ((phase (interpret-pauli-phase-factor (phase-factor p))))
+    (assert (realp phase) (p) "The given Pauli term has a complex phase of ~A." phase)
+    (setf phase (coerce (realpart phase) 'double-float))
+    (quil:with-inst ()
+      (cond
+        ((pauli-identity-p p)
```
ecpeterson: I know that this is what pyQuil does, but everyone agrees that it should do something different. What "different" means is somewhat split: some people want to keep this instruction sequence but find it uncomfortable that it always touches qubit `0`, which can have uncomfortable implications for QPU execution; others (incl. me) think that this sequence should just be completely elided.

To repeat the argument for elision: this sequence doesn't do anything material to the state. At most it records the value of `phase` for programmer reference / inspection—but the programmer also has access to the Pauli `p`, so idk what's stopping them from just inspecting that.

At any rate, I would prefer that CL-QUIL does not inherit this controversy.

src/clifford/pauli.lisp
```diff
@@ -323,3 +327,68 @@ hash-function."
...
+             (Y (quil:inst "RY" (list (quil::param-* (* 2.0d0 phase) x)) qubit))
+             (Z (quil:inst "RZ" (list (quil::param-* (* 2.0d0 phase) x)) qubit)))))
+
+        ;; The general case.
+        (t
+         (let* ((n (num-qubits p)))
+           (loop :with last-qubit := nil
+                 :for qubit :below n
+                 :for pauli-component := (base4-to-sym (aref (pauli-components p) (1+ qubit)))
+                 ;; This COLLECT builds the CNOT ladder needed to do the
```
ecpeterson: You might consider making this particular routine hot-swappable in some way. The role of these CNOTs is to "collect all the toggles" onto the qubit where RZ being applied, but (1) there are log-depth strategies for accomplishing the same task, and (2) this particular CNOT ladder isn't a semantic / intentional description of what's happening here, which makes it difficult for some future compiler to take this circuit and rewrite it as it sees fit. [For instance, you could emit a FORKED representation instead of a CNOT ladder, then let the FORKED-expander follow its heart.] [To be clear: (1) and (2) roughly respectively correspond to an ability to change the CNOT-expansion strategy here or an ability to make a decision about it later.]

PR #380, opened 2019-08-08
Add PROG-INITIAL-REWIRING-HEURISTIC

appleby:
Per ecpeterson's comments in #361, update `COMPILER-HOOK` to call a new function, `PROG-INITIAL-REWIRING-HEURISTIC`, to try and determine whether a `:NAIVE` initial rewiring strategy is likely to succeed, and otherwise fallback to `*INITIAL-REWIRING-DEFAULT-TYPE*`, which now defaults to `:PARTIAL`, rather than `:NAIVE`.

This commit also adds a bunch of Aspen-*.qpu files exported from `pyquil.api._devices.get_lattice`. The idea is that these QPU files can be used in testing as representative of real Aspen chipspecs.

Additionally, add a (very basic) chip-reader test and the stub of a regression test for #378.

Fixes #361

~~Note that the failing test in `QUILC-TESTS` is due to #375; therefore, this PR should not be merged until after #375.~~
Rebased onto the changes in #375, so tests are happy now.

PR #474, opened 2019-11-11
Include RESET-QUBIT and MEASURE in PROG-INITIAL-REWIRING-HEURISTIC

appleby:
Previously, `PROG-INITIAL-REWIRING-HEURISTIC` only considered `GATE-APPLICATION`s when attempting to guess the initial rewiring. This resulted in a bad initial rewiring guess of `:NAIVE` when all the program's `GATE-APPLICATION`s (possibly 0) were `%NAIVELY-APPLICABLE-P`, but the program included a `MEASURE` or `RESET-QUBIT` instruction that was not `%NAIVELY-APPLICABLE-P`.

See also:

https://github.com/rigetti/quilc/pull/380
https://github.com/rigetti/quilc/issues/361

### DIFF
--- a/src/addresser/initial-rewiring.lisp
+++ b/src/addresser/initial-rewiring.lisp
@@ -163,17 +163,13 @@ is found, then return NIL."
       (t
        (return-from prog-rewiring-pragma nil)))))
 
-(defun %naively-applicable-p (gate-app chip-spec
-                              &aux (qubit-indices (application-qubit-indices gate-app)))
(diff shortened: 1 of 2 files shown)


appleby:
This change was motivated by @karalekas reporting on slack that the program

```
RESET
DECLARE ro BIT[1]
MEASURE 0 ro[0]
```

fails to compile, producing the error: `User program incompatible with chip: naive rewiring crosses chip component boundaries.`

PR #480, opened 2019-11-16
Fix parametric RZ agglutination

### LINKED ISSUE #468: Arithmetic simplification is not performed for non-constant parameter values
More precisely, it is sometimes not recognized that a phase is identically zero.
For example `theta[0] + -1*1.0*theta[0]`.
This code
```python
from pyquil.gates import *
from pyquil import Program, get_qc

def gate_cancel(gate):
    c = get_qc('1q-qvm')
    p = Program()
    theta = p.declare('theta', memory_type='REAL')
    p += gate(theta, 0)
    p += gate(-theta, 0)
    print(c.compile(p, to_native_gates=True, optimize=True).program)

gate_cancel(RZ)
gate_cancel(RY)
```
outputs
```
DECLARE theta REAL[1]
RZ(theta[0] + -1*1.0*theta[0]) 0
HALT

DECLARE theta REAL[1]
RX(pi/2) 0
RZ(theta[0]) 0
RZ(-1*1.0*theta[0]) 0
RX(-pi/2) 0
HALT
```

notmgsk:
Closes #468.

Previously, an instruction sequence involving named parameters that might reasonably be reduced (e.g. `RZ(theta[0] - theta[0])`) would be missed by the compiler.  @karalekas introduced the `simplify-arithmetic` routine (#211) that is capable of reducing that type of parameter, but it was not installed as part of the compilation pipeline.

This PR uses some of the functionality introduced in #211 (namely `canonicalize-expression`) to simplify arithmetic expressions that involve named (delayed) parameters.

Old:
```
QUIL> (print-parsed-program
       (compiler-hook (parse-quil "DECLARE theta REAL[1]; RZ(theta) 0; RZ(-theta) 0;")
                      (build-nq-fully-connected-chip 2)))
DECLARE theta REAL

RZ((theta[0]+-(theta[0]))) 0            # Entering rewiring: #(0 1)
HALT                                    # Exiting rewiring: #(0 1)
```

New:
```
QUIL> (print-parsed-program
       (compiler-hook (parse-quil "DECLARE theta REAL[1]; RZ(theta) 0; RZ(-theta) 0;")
                      (build-nq-fully-connected-chip 2)))
DECLARE theta REAL

HALT                                    # Entering/exiting rewiring: (#(0 1) . #(0 1))
```

Follow-on work: fix the peephole rewriter missing optimizations in the following program
```
QUIL> (print-parsed-program
       (compiler-hook (parse-quil "DECLARE theta REAL[1]; RX(theta) 0; RX(-theta) 0;")
                      (build-nq-fully-connected-chip 2)))
DECLARE theta REAL

RZ(pi/2) 0                              # Entering rewiring: #(0 1)
RX(pi/2) 0
RZ(theta[0]) 0
RZ(-(theta[0])) 0
RX(-pi/2) 0
RZ(-pi/2) 0
HALT                                    # Exiting rewiring: #(0 1)
```

### DIFF
--- a/cl-quil-tests.asd
+++ b/cl-quil-tests.asd
@@ -7,6 +7,7 @@
   :author "Robert Smith <robert@rigetti.com>"
   :license "Apache License 2.0 (See LICENSE.txt)"
   :depends-on (#:cl-quil
+               #:quilc
                (:version #:qvm "1.9.0")
                #:magicl
                #:alexandria
--- a/cl-quil-tests.asd
+++ b/cl-quil-tests.asd
@@ -7,7 +7,6 @@
   :author "Robert Smith <robert@rigetti.com>"
   :license "Apache License 2.0 (See LICENSE.txt)"
   :depends-on (#:cl-quil
-               #:quilc
                (:version #:qvm "1.9.0")
                #:magicl
                #:alexandria
--- a/cl-quil-tests.asd
+++ b/cl-quil-tests.asd
@@ -40,6 +40,7 @@
                (:file "compilation-tests")
                (:file "clifford-tests")
                (:file "translator-tests")
+               (:file "rewrite-tests")
                (:file "state-prep-tests")
                (:file "compiler-hook-tests")
                (:file "benchmarking-procedures-tests")
--- a/src/build-gate.lisp
+++ b/src/build-gate.lisp
@@ -104,7 +104,7 @@ EXAMPLE: The Quil line \"CPHASE(pi) 2 3\" corresponds to the S-expression (build
                 :when (find (second head) zipped-tail :key #'second :test #'equal)
                   :do (error "Duplicate definition of lambda-param: ~A" (second head)))
           (if (or delayedp-1 delayedp-2)
-              (canonicalize-expression
+              (simplify-arithmetic
                (make-delayed-expression (mapcar #'first params-zipped)
                                         (mapcar #'second params-zipped)
                                         `(,op ,expression-1 ,expression-2)))
--- a/src/build-gate.lisp
+++ b/src/build-gate.lisp
@@ -104,9 +104,10 @@ EXAMPLE: The Quil line \"CPHASE(pi) 2 3\" corresponds to the S-expression (build
                 :when (find (second head) zipped-tail :key #'second :test #'equal)
                   :do (error "Duplicate definition of lambda-param: ~A" (second head)))
           (if (or delayedp-1 delayedp-2)
-              (make-delayed-expression (mapcar #'first params-zipped)
-                                       (mapcar #'second params-zipped)
-                                       `(,op ,expression-1 ,expression-2))
+              (canonicalize-expression
+               (make-delayed-expression (mapcar #'first params-zipped)
+                                        (mapcar #'second params-zipped)
+                                        `(,op ,expression-1 ,expression-2)))
               (funcall op expression-1 expression-2)))))))
 
 (defun param-+ (arg1 arg2)
--- a/src/cl-quil.lisp
+++ b/src/cl-quil.lisp
@@ -6,7 +6,7 @@
 (in-package #:cl-quil)
 
 (defvar *standard-post-process-transforms*
-  '(expand-circuits type-check)
+  '(expand-circuits type-check simplify-arithmetic)
   "The standard transforms that are applied by PARSE-QUIL.")
 
 (define-condition ambiguous-definition-condition ()
--- a/src/compressor/rewriting-rules.lisp
+++ b/src/compressor/rewriting-rules.lisp
@@ -46,11 +46,11 @@
 
 (defun full-rotation-p (param &key (full 2pi))
   "Return T if PARAM is an integer multiple of FULL. PARAM may be a DOUBLE-FLOAT, or a CONSTANT."
-  (or (and (typep param 'double-float)
-           (double= (/ param full) (round (/ param full))))
-      (and (typep param 'constant)
-           (let ((val (constant-value param)))
-             (double= (/ val full) (round (/ val full)))))))
+  (check-type param (or constant double-float))
+  (let ((val (or (and (is-constant param)
+                      (constant-value param))
+                 param)))
+    (double= (/ val full) (round (/ val full)))))
 
 (define-compiler eliminate-full-RX-rotations
     ((x ("RX" (theta) _)
--- a/src/compressor/rewriting-rules.lisp
+++ b/src/compressor/rewriting-rules.lisp
@@ -182,7 +182,8 @@
 
 (define-compiler eliminate-half-PISWAP
     ((x ("PISWAP" (theta) p q)
-        :where (full-rotation-p theta :full 2pi)))
+        :where (and (full-rotation-p theta :full 2pi)
+                    (not (full-rotation-p theta :full 4pi)))))
   (inst "RZ" '(#.pi) p)
   (inst "RZ" '(#.pi) q))
 
--- a/src/compressor/rewriting-rules.lisp
+++ b/src/compressor/rewriting-rules.lisp
@@ -44,12 +44,13 @@
 
 ;;; rewriting rules specialized to qubit types
 
-(defun full-rotation-p (param)
+(defun full-rotation-p (param &key (full 2pi))
+  "Return T if PARAM is an integer multiple of FULL. PARAM may be a DOUBLE-FLOAT, or a CONSTANT."
   (or (and (typep param 'double-float)
-           (double= (/ param 2pi) (round (/ param 2pi))))
+           (double= (/ param full) (round (/ param full))))
       (and (typep param 'constant)
            (let ((val (constant-value param)))
-             (double= (/ val 2pi) (round (/ val 2 pi)))))))
+             (double= (/ val full) (round (/ val full)))))))
 
 (define-compiler eliminate-full-RX-rotations
     ((x ("RX" (theta) _)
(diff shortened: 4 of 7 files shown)


src/compressor/rewriting-rules.lisp
```diff
@@ -44,10 +44,16 @@
 
 ;;; rewriting rules specialized to qubit types
 
+(defun full-rotation-p (param)
+  (or (and (typep param 'double-float)
+           (double= (/ param 2pi) (round (/ param 2pi))))
+      (and (typep param 'constant)
+           (let ((val (constant-value param)))
+             (double= (/ val 2pi) (round (/ val 2 pi)))))))
```
appleby:
```suggestion
             (double= (/ val 2pi) (round (/ val 2pi)))))))
```

?

tests/rewrite-tests.lisp
```diff
@@ -0,0 +1,27 @@
+(in-package #:cl-quil-tests)
+
+(defun %filter-halt (pp)
+  (remove-if (lambda (instr) (typep instr 'halt))
+             (parsed-program-executable-code pp)))
+
+(deftest test-rz-full-rotation-elimination ()
+  ;; Test the compiler directly.
+  (is (not (quil::eliminate-full-rz-rotations (quil::build-gate "RZ" (list 0d0) 0))))
+  (is (not (quil::eliminate-full-rz-rotations (quil::build-gate "RZ" (list (quil::constant 0d0)) 0))))
```
appleby: `not` -> `null` (or maybe `endp` idk)

tests/rewrite-tests.lisp
```diff
@@ -0,0 +1,27 @@
+(in-package #:cl-quil-tests)
+
+(defun %filter-halt (pp)
+  (remove-if (lambda (instr) (typep instr 'halt))
```
appleby:
```suggestion
  (remove-if #'quil::haltp
```

tests/rewrite-tests.lisp
```diff
@@ -0,0 +1,27 @@
...
+        (pps (list (quil::parse-quil "DECLARE theta REAL[1]; RZ(0.0) 0;")
+                   (quil::parse-quil "DECLARE theta REAL[1]; RZ(theta) 0; RZ(-theta) 0;"))))
+    (dolist (pp pps)
+      (is (zerop (length (%filter-halt (quil::compiler-hook pp chip))))))))
+
+(deftest test-rz-agglutination ()
+  (let ((chip (quil::build-nq-fully-connected-chip 2))
+        (pps (list
+              (parse-quil "RZ(1) 0; RZ(-1) 0;")
+              (parse-quil "DECLARE theta REAL[1]; RZ(theta) 0; RZ(-theta) 0"))))
```
appleby: identical to test on line 15 above?

src/build-gate.lisp
```diff
@@ -104,9 +104,10 @@ EXAMPLE: The Quil line \"CPHASE(pi) 2 3\" corresponds to the S-expression (build
                 :when (find (second head) zipped-tail :key #'second :test #'equal)
                   :do (error "Duplicate definition of lambda-param: ~A" (second head)))
           (if (or delayedp-1 delayedp-2)
-              (make-delayed-expression (mapcar #'first params-zipped)
-                                       (mapcar #'second params-zipped)
-                                       `(,op ,expression-1 ,expression-2))
+              (canonicalize-expression
```
ecpeterson: What happens when `canonicalize-expression` doesn't know how to act, like `RZ(a*b)` for memory references `a`, `b`? I think in the transform we installed a signal handler to smooth this over.

src/compressor/rewriting-rules.lisp
```diff
@@ -44,10 +44,16 @@
...
+
 (define-compiler eliminate-full-RX-rotations
     ((x ("RX" (theta) _)
-        :where (and (typep theta 'double-float)
-                    (double= (/ theta 2pi) (round (/ theta 2pi))))))
+        :where (full-rotation-p theta)))
```
ecpeterson: :) This is a good synonym.

FYI, PISWAP is only locally equivalent to the identity at 2pi and _actually_ the identity at 4pi. It might be nice to have a name that makes it clearer which for gates this predicate actually does what it sounds like.

tests/rewrite-tests.lisp
```diff
@@ -0,0 +1,27 @@
+(in-package #:cl-quil-tests)
+
+(defun %filter-halt (pp)
```
ecpeterson: I don't know if it's preferable to `%filter-halt` or to `:protoquil t`.

notmgsk: I guess that has been a regression in the behaviour of `compiler-hook` :(

```
QUIL> (print-parsed-program 
       (compiler-hook 
        (parse-quil "Z 0") 
        (build-nq-linear-chip 2)
        :protoquil t))
RZ(pi) 0                                # Entering rewiring: #(0 1)
HALT                                    # Exiting rewiring: #(0 1)
NIL
```

ecpeterson: Oh—maybe I'm just wrong. We also separately strip `HALT` in the quilc app, don't we?

notmgsk: Yes, `process-program :protoquil t` does that job.  But I forget whether it does some other stuff that I don't want.  Can't hurt to check.

src/compressor/rewriting-rules.lisp
notmgsk: I wonder if it is simpler to match the above compiler on `("RX" (0d0) _)` and then rely on `normalize-(rx|rz)-rotations` to produce a `theta` of `0d0` where appropriate.

notmgsk: Although that doesn't help when the input is a `constant`.

src/build-gate.lisp
notmgsk: Trying it out, it doesn't seem like an issue:
```
QUIL> (compiler-hook (parse-quil "DECLARE theta REAL[1]; DECLARE gamma REAL[1]; RZ(theta * gamma) 0;")
                     (build-nq-linear-chip 2))
#<PARSED-PROGRAM {1005319943}>
0
1/100
QUIL> (print-parsed-program *)
DECLARE theta REAL
DECLARE gamma REAL

RZ((theta[0]*gamma[0])) 0               # Entering rewiring: #(0 1)
HALT                                    # Exiting rewiring: #(0 1)
NIL
```

If that doesn't allay your concerns I can look into it more closely. LMK.

src/compressor/rewriting-rules.lisp
notmgsk: @ecpeterson I added a `:full` kwarg to `full-rotation-p` so you can specify what full should mean.  Good enough?

src/build-gate.lisp
notmgsk: Tried it out and it doesn't seem to be a problem:
```
QUIL> (compiler-hook (parse-quil "DECLARE theta REAL[1]; DECLARE gamma REAL[1]; RZ(theta * gamma) 0;")
                     (build-nq-linear-chip 2))
#<PARSED-PROGRAM {1005319943}>
0
1/100
QUIL> (print-parsed-program *)
DECLARE theta REAL
DECLARE gamma REAL

RZ((theta[0]*gamma[0])) 0               # Entering rewiring: #(0 1)
HALT                                    # Exiting rewiring: #(0 1)
NIL
```

If that doesn't allay your concerns, I can look into it more closely. LMK.

ecpeterson: That isn't a sufficient test case, since `param-whatever` never gets called. Try e.g. `RZ(theta*gamma) 0; RZ(gamma) 0`.

src/compressor/rewriting-rules.lisp
ecpeterson: Sounds good to me.

src/build-gate.lisp
notmgsk: Oh duh.  I extended `simplify-arithmetic` to take `constant`s and `delayed-expression`s and do the appropriate thing, and now use that instead of `canonicalize-expression`.
```
QUIL> (print-parsed-program
       (compiler-hook (parse-quil "DECLARE theta REAL[1]; DECLARE gamma REAL[1]; RZ(theta*gamma) 0; RZ(gamma) 0")
                      (build-nq-linear-chip 2)))
DECLARE theta REAL
DECLARE gamma REAL

RZ(((theta[0]*gamma[0])+gamma[0])) 0    # Entering rewiring: #(0 1)
HALT                                    # Exiting rewiring: #(0 1)
```

src/compressor/rewriting-rules.lisp
```diff
@@ -175,12 +181,14 @@
       (inst "PISWAP" (list reduced-theta) p q))))
 
 (define-compiler eliminate-half-PISWAP
-    ((x ("PISWAP" (#.2pi) p q)))
+    ((x ("PISWAP" (theta) p q)
+        :where (full-rotation-p theta :full 2pi)))
```
ecpeterson: This is dangerous / wrong: you only want to apply to x = 0 mod 2pi but x ≠ 0 mod 4pi.

cl-quil-tests.asd
```diff
@@ -7,6 +7,7 @@
   :author "Robert Smith <robert@rigetti.com>"
   :license "Apache License 2.0 (See LICENSE.txt)"
   :depends-on (#:cl-quil
+               #:quilc
```
appleby: > Introduces a quilc dependency but whomstsoever cares?

I don't care *that* much, but seems a tad overkill just to be able to drop the 2-line `%filter-halt`

notmgsk: 😡

notmgsk:
OK. I introduced *another* change here that should be vetted. If you directly fed in `RZ(a[0] - a[0]) 0` the compiler would not reduce this since it is never passes through an invocation of `simplify-arithmetic`. To that end, I have rewritten `simplify-arithmetic` to take in more object types (and simultaneously simplify its implementation), and added this transform to the list of standard transforms applied by `parse-quil`. That means the above `RZ(a - a) 0` is simplified at parse time, while something like `RZ(a) 0; RZ(-a) 0;` is simplified at compile time.

src/cl-quil.lisp
```diff
@@ -6,7 +6,7 @@
 (in-package #:cl-quil)
 
 (defvar *standard-post-process-transforms*
-  '(expand-circuits type-check)
+  '(expand-circuits type-check simplify-arithmetic)
```
appleby: pleasantly surprised the golden file tests didn't barf on this change

PR #483, opened 2019-11-16
Improve compiler patience

ecpeterson:
This PR changes the way compilation subroutines are preference-sorted. Previously, they were sorted first into two buckets, those that target the native gate set vs those that target an intermediate gate set (but which might appear on the road to the native gate set), then they were sorted within those buckets by the worst-case cost of their output, then those lists were concatenated with those compilers immediately targeting the native gate set getting preference.

This PR does away with the buckets: if a compiler might appear on the road to the native gateset and its worst case is to give good output, it shouldn't matter whether there are more compilers still to apply before reaching the native gate set.

This PR simultaneously resolves an awkward-but-not-lethal behavior: on a chip with XY (or PISWAP) and CZ both available, quilc would compile SWAP to a triple of CZs, because (1) SWAP has a short decomposition into CZ + ISWAP, (2) "ISWAP" isn't literally part of the native gate set, and instead has to pass through a conversion to XY(pi), & (3) there is also a decomposition of SWAP into 3*CZ, which ultimately got preference because CZ is in the target gateset. Eliminating the buckets makes this false preference go away.

One last word about the buckets: the compiler sorting mechanism was written by degrees. My _guess_ is that the buckets came before the sorting, and I didn't bother to see what would happen by just using the sorting and discarding the buckets. However, it _could_ also be that the buckets were introduced to dodge some infinite loop, which may or may not be lurking around a corner here. I've thought about this some, and I feel fairly confident that under a mild assumption*, this is a safe change. At any rate, the tests pass. :P

* - Precise assumption: for a gate set with n gates and a family of associated fidelities (f_j), the map (m1, ..., mn) -> prod_j f_j^(mj) is injective. This assumption is mild because it's almost always true: you have to try hard to make such discrete things collide inside of the real numbers (or even floating point numbers).


notmgsk:
https://www.youtube.com/watch?v=273eSvOwpKk

Issue #292, opened 2019-06-17
Error: The alien function "zuncsd_" is undefined


Roger-luo:
I'm trying to build quilc and run the benchmark, however, after I build the compiler and run with

```
./quilc --benchmark
```

I got

```
! ! ! Error: The alien function "zuncsd_" is undefined.
Benchmarking CL-QUIL-BENCHMARKING::BENCH-0020Q-0000654-JOHANNES-*DEPTH-VS-SWAPS-ASSN*-0020Q-SKEW-RECT...%
```

I'm not very familiar with lisp, anyone could help me fix it? I'm currently building v1.8.2 with `sbcl` 1.5.3 on MacOS.

notmgsk:
@Roger-luo what is the output of the command `./quilc --check-libraries`?

Roger-luo:
```

Roger-luo:
Does this mean it is not correctly linked to lapack? but I didn't see any instruction for linking against lapack... Is there any manual option for this?

notmgsk:
It looks like it is finding an unsupported version of LAPACK. Do you know what version of LAPACK you have? Mine, for example, `3.8.0`.

Roger-luo:
I'm not sure which version it is, but I think it should link to the mkl, which is 2018.3.039 on my laptop. I also got 3.8.0 by openblas in conda, but I guess that's not the lapack it is trying to link to, how can I change the lapack vendor/path for quilc?

stylewarning:
Some versions of LAPACK don't have ZUNCSD unfortunately. This includes MKL and Accelerate I believe. Could you try using a different version of LAPACK?

Roger-luo:
sure, is there a compile option for specific the lapack path then?

stylewarning:
Could you try changing your `LD_LIBRARY_PATH` to the path with the library of interest? A possible problem is that we statically determine the path to these libraries at build time. Are you building the binaries from scratch?

rongpenl:
Hi, any update on this one?

jmbr:
Thank you for reaching out, @rongpenl. Our linear algebra package, magicl was tested a few months ago on several platforms. The CSD decomposition routines work with OpenBLAS on Linux and with the Accelerate framework in macOS. Could you give us some details about your system settings (platform, operating system, available BLAS, versions)?

jmbr:
Closed by #485.

PR #485, opened 2019-11-19
Use alternative CS decomposition when it is missing from LAPACK

jmbr:


notmgsk:
` ;     Symbol "HERMITIAN-EIG" not found in the MAGICL package.`

jmbr:
> ` ; Symbol "HERMITIAN-EIG" not found in the MAGICL package.`

That's only available in MAGICL's master or, preferably, in https://github.com/rigetti/magicl/pull/65

notmgsk:
> > ` ; Symbol "HERMITIAN-EIG" not found in the MAGICL package.`
> 
> That's only available in MAGICL's master or, preferably, in [rigetti/magicl#65](https://github.com/rigetti/magicl/pull/65)

Ah. Then maybe we ought to make a MAGICL release so that I will be picked up here.

jmbr:
> Ah. Then maybe we ought to make a MAGICL release so that I will be picked up here.

Yeah, that's reasonable.

appleby:
Would it make sense for the `csd-compat` implementation to live in the magicl package, and have magicl fallback to it if zuncsd isn't available? Or maybe just provide it separately?  I only skimmed the diff, but IIUC it looks like csd.lisp only depends on `m*` from `cl-quil`, which might also be reasonably moved into `magicl`.

jmbr:
The reason this code presently lives in `cl-quil` is because it implements a specific case of the CSD (currently the one used in Quantum Shannon compilation). If it ever becomes a general-purpose CSD implementation, then I agree we might consider moving it to magicl.

jmbr:
Passes the test suite. Closes #292.

appleby:
Should we bump the dependency in `cl-quil.asd` to magicl 0.6.5?

jmbr:
That's right.

src/csd.lisp
```diff
@@ -0,0 +1,155 @@
...
+                         (m* (magicl:conjugate-transpose vh) (m*-diag-general sigma vh)))
+                        (:hermitian-unitary
+                         (m* u (m*-diag-general sigma (magicl:conjugate-transpose u)))))))
+      (values unitary hermitian))))
+
+(defun csd-2x1 (a3 a4)
+  "Compute the 2x1 Cosine-Sine decomposition of a unitary matrix A partitioned into blocks A1 A2 as shown below:
+
+       ⎡ A3 ⎤        ⎡ A3 ⎤   ⎡ U1     ⎤ ⎡ -S ⎤
+If A = ⎢    ⎥, then  ⎢    ⎥ = ⎢        ⎥ ⎢    ⎥ V2ᴴ,
```
notmgsk: 👍

appleby:
> Passes the test suite. Closes #292.

By "passes the test suite" do you mean you installed a version of LAPACK without ZUNCSD and ran the test suite, or just that the new csd tests you added + rest of the test suite passes when run against an LAPACK that includes ZUNCSD.

jmbr:
> > Passes the test suite. Closes #292.
> 
> By "passes the test suite" do you mean you installed a version of LAPACK without ZUNCSD and ran the test suite, or just that the new csd tests you added + rest of the test suite passes when run against an LAPACK that includes ZUNCSD.

The former: I used Apple's Accelerate framework (which does not include ZUNCSD) to run the entire test suite (I also double-checked that the new routines were picked up by `trace`ing them).

src/csd.lisp
```diff
@@ -0,0 +1,155 @@
...
+;;; Compute the 2x2 Cosine-Sine decomposition of an unitary matrix using a
+;;; modification of the algorithm in: E. S. Gawlik, Y. Nakatsukasa, and
+;;; B. D. Sutton, “A Backward Stable Algorithm for Computing the CS
+;;; Decomposition via the Polar Decomposition,” SIAM J. Matrix Anal. Appl.,
+;;; vol. 39, no. 3, pp. 1448–1469, Jan. 2018. DOI:10.1137/18M1182747
+
+(defun m*-diag-general (diag matrix)
+  "Returns a newly allocated matrix resulting from the product of DIAG (a diagonal real matrix) with MATRIX (a complex matrix)."
+  (declare (type magicl:matrix diag matrix)
+           (values magicl:matrix))
```
kilimanjaro: Is this portable? If not, should we wrap it in `#+sbcl` or equivalent?

src/csd.lisp
```diff
@@ -0,0 +1,155 @@
...
+            (values u1 u2 c s v2h)))))))
+
+(defun csd (matrix p q)
+  "Compute the 2x2 Cosine-Sine decomposition of MATRIX (assumed to be unitary and 2n×2n) partitioned into n×n blocks A1 A2 A3 A4 as shown below:
+
+       ⎡ A1  A3 ⎤        ⎡ A1  A3 ⎤   ⎡ U1     ⎤ ⎡ C  -S ⎤ ⎡ V1     ⎤H
+If A = ⎢        ⎥, then  ⎢        ⎥ = ⎢        ⎥ ⎢       ⎥ ⎢        ⎥,
+       ⎣ A2  A4 ⎦        ⎣ A2  A4 ⎦   ⎣     U2 ⎦ ⎣ S   C ⎦ ⎣     V2 ⎦
+
+where U1, U2, V1, and V2 are unitary and C^2 + S^2 = I. When the partition is p = q = 1, we have
```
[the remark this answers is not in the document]
kilimanjaro: (Also applies to some of the above docstrings.)

tests/csd-tests.lisp
```diff
@@ -0,0 +1,94 @@
...
+        (quil::csd a n n)
+      (let ((u1h (magicl:conjugate-transpose u1))
+            (u2h (magicl:conjugate-transpose u2))
+            (v1 (magicl:conjugate-transpose v1h))
+            (v2 (magicl:conjugate-transpose v2h)))
+        (is (m= id (quil::m* u1h u1)))
+        (is (m= id (quil::m* u2h u2)))
+        (is (m= id (quil::m* v1h v1)))
+        (is (m= id (quil::m* v2h v2)))
+        (multiple-value-bind (c s)
```
kilimanjaro: Could use something like `(magicl:diag (mapcar #'cos theta))` to construct these.

src/csd.lisp
```diff
@@ -0,0 +1,155 @@
...
+      (values unitary hermitian))))
+
+(defun csd-2x1 (a3 a4)
+  "Compute the 2x1 Cosine-Sine decomposition of a unitary matrix A partitioned into blocks A1 A2 as shown below:
+
+       ⎡ A3 ⎤        ⎡ A3 ⎤   ⎡ U1     ⎤ ⎡ -S ⎤
+If A = ⎢    ⎥, then  ⎢    ⎥ = ⎢        ⎥ ⎢    ⎥ V2ᴴ,
+       ⎣ A4 ⎦        ⎣ A4 ⎦   ⎣     U2 ⎦ ⎣  C ⎦
+
+where U1, U2, and V2 are unitary and C^2 + S^2 = I. When the partition is p = q = 1, we have
```
kilimanjaro: Should also note: here p,q are note even arguments.

src/csd.lisp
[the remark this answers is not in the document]
jmbr: The values of p and q determine the partitioning of the submatrices A1, A2, A3, A4. It follows the same convention as `lapack-csd` in Magicl.

src/csd.lisp
jmbr: The values of p and q here determine the sizes of A3 and A4.

The significance of the lowercase letters is to contrast complex numbers with complex (sub)matrices.

src/csd.lisp
jmbr: It compiles without errors in ECL, Clozure Common Lisp, and CLISP.

tests/csd-tests.lisp
jmbr: Good point!

src/csd.lisp
kilimanjaro: Good enough for me!

PR #489, opened 2019-11-20
Error if preserved block touches illegal qubits

### LINKED ISSUE #448: Improve the default chip-building functionality
There are two ways to construct a chip specification. The general case is through the machinery of `chip-reader.lisp`, which takes an ISA description in JSON format. There are also some helpful routines in `chip-specification.lisp`, which are used to construct some of the basic/default chip architectures (e.g. everything involved in `build-nq-linear-chip`). 

These latter routines are alluring but do not work robustly, For example, if one wanted a linear chip supporting both `CZ` and `SWAP` operations natively, adding the line
```
(install-link-onto-chip chip-spec i (1+ i) :architecture ':SWAP)
```
in the appropriate place in `build-nq-linear-chip` would _not_ work (one will see a later failure in `find-shortest-compiler-path`). 

Here I would request some general improvements in this area, allowing for easy procedural construction of chips without recourse to the JSON middleman.

### LINKED ISSUE #488: Referencing illegal qubits in preserved block should always signal an error.
@jlapeyre recently noticed that you can get the compiler to spin its wheels by passing in a program that uses `PRAGMA PRESERVE_BLOCK` and within that block touches some illegal (dead) qubit.  It seems reasonable to me that, since one of the intentions of a preserved block is to prevent rewiring logical qubits, if any illegal qubit is referenced an error should be raised (because it will never be rewired to a legal qubit).

notmgsk:
Closes #488.


src/chip/chip-reader.lisp
```diff
@@ -410,19 +410,45 @@
...
+  (:documentation "Error representing the use of an illegal qubit in a preserved block.")
+  (:report (lambda (c s)
+             (format s "Within a preserved block, instruction '~/quil:instruction-fmt/' uses the illegal qubits ~{~A~^, ~}."
+                     (illegal-qubits-used-error-instruction c)
+                     (illegal-qubits-used-error-illegal-qubits c)))))
+
+(defun check-instructions-skip-dead-qubits (instrs dead-qubits &key (condition-class 'illegal-qubits-used-error))
+  (map nil
+       (lambda (instr)
+         (when (typep instr 'application)
```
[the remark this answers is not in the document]
notmgsk: Hm, yes, I think so.  Good spot.

tests/compiler-hook-tests.lisp
```diff
@@ -509,3 +509,15 @@ CCNOT 0 1 2")
       (declare (ignore order index))
       (is (= (quil::hardware-object-native-instruction-p obj (quil::build-gate "CZ" () 1 2))
              (quil::chip-schedule-resource-carving-point sched (quil::make-qubit-resource 1 2)))))))
+
+(deftest test-using-illegal-qubits-signals-error ()
+  (let* ((chip (quil::build-nq-fully-connected-chip 4))
+         (chip-hash (quil::chip-specification-to-qpu-hash-table chip)))
+    (setf (gethash "dead" (gethash "2" (gethash "1Q" (gethash "isa" chiphash)))) "true")
```
appleby:
```suggestion
    (setf (gethash "dead" (gethash "2" (gethash "1Q" (gethash "isa" chip-hash)))) "true")
```

notmgsk: oof

src/analysis/compress-qubits.lisp
```diff
@@ -7,8 +7,8 @@
 ;;; This file contains an analysis to compute the qubits used by a
 ;;; program.
 
-(defgeneric %qubits-used (isn)
```
stylewarning: I don't like this change only because `%qubits-used` was intended to have the type `instruction -> [int]`, and now you've made it (more or less) `Any -> [int]`. It's not wrong, but doesn't adhere to what I consider to be "type driven" design.

src/chip/chip-reader.lisp
```diff
@@ -410,19 +410,45 @@
...
+
+(define-condition illegal-qubits-used-in-preserved-block-error (illegal-qubits-used-error)
+  ()
+  (:documentation "Error representing the use of an illegal qubit in a preserved block.")
+  (:report (lambda (c s)
+             (format s "Within a preserved block, instruction '~/quil:instruction-fmt/' uses the illegal qubits ~{~A~^, ~}."
+                     (illegal-qubits-used-error-instruction c)
+                     (illegal-qubits-used-error-illegal-qubits c)))))
+
+(defun check-instructions-skip-dead-qubits (instrs dead-qubits &key (condition-class 'illegal-qubits-used-error))
```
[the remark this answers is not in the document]
stylewarning: Useful to note, but not applicable here. The existence of links vs the liveness of qubits are sufficiently different

src/analysis/compress-qubits.lisp
stylewarning: OTOH you could argue you're now engaging in "protocol-driven design"

PR #490, opened 2019-11-21
Don't call APPLICATION-ARGUMENTS on potential non-application instr

appleby:
INSTR might be a MEASUREMENT or RESET-QUBIT instruction, in which case attempting to call APPLICATION-ARGUMENTS will result in an error like

    There is no applicable method for the generic function
      #<STANDARD-GENERIC-FUNCTION CL-QUIL:APPLICATION-ARGUMENTS (1)>
    when called with arguments
      (#<MEASURE 0 ro[0]>).

when attempting to compile a program with a MEASURE or RESET-QUBIT instruction and NAIVE rewiring.


appleby:
just waiting for pyquil tests to finish.  minimal reproduction:

```
$ echo MEASURE 0 | ./quilc
<134>1 2019-11-21T20:20:05Z BK-LOSX-24JLVCJ.local quilc 29643 LOG0001 - This is the latest version of the SDK.
! ! ! Error: There is no applicable method for the generic function
               #<STANDARD-GENERIC-FUNCTION CL-QUIL:APPLICATION-ARGUMENTS (1)>
             when called with arguments
               (#<MEASURE 0>).
See also:
  The ANSI Standard, Section 7.6.6
```

notmgsk:
oopsie poopsie

PR #492, opened 2019-11-22
Fix type checking for classical-equality instructions

notmgsk:
The type requirements in [the spec](https://github.com/rigetti/quil/blob/master/rfcs/typed-memory.md#classical-instructions) were not re**spec**ted.

### DIFF
--- a/src/analysis/type-safety.lisp
+++ b/src/analysis/type-safety.lisp
@@ -47,12 +47,13 @@
                      (memory-descriptor-length memory-descriptor))))
 
 (defun constant-or-mref-typep (obj quil-type memory-regions)
+  "Test whether OBJ (a CONSTANT or a MEMORY-REF) has a value that has type QUIL-TYPE (or one of those types if QUIL-TYPE is a list)."
   (etypecase obj
     (memory-ref
      (let ((memory-region (find-descriptor-for-mref obj memory-regions)))
-       (equalp quil-type (memory-descriptor-type memory-region))))
+       (member (memory-descriptor-type memory-region) (a:ensure-list quil-type))))
     (constant
-     (equalp quil-type (constant-value-type obj)))))
+     (member (constant-value-type obj) (a:ensure-list quil-type)))))
 
 (defun check-mref (obj)
   (unless (typep obj 'memory-ref)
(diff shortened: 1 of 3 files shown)


src/analysis/type-safety.lisp
```diff
@@ -47,12 +47,13 @@
...
 (defun constant-or-mref-typep (obj quil-type memory-regions)
+  "Test whether OBJ (a CONSTANT or a MEMORY-REF) has a value that has type QUIL-TYPE (or one of those types if QUIL-TYPE is a list)."
   (etypecase obj
     (memory-ref
      (let ((memory-region (find-descriptor-for-mref obj memory-regions)))
-       (equalp quil-type (memory-descriptor-type memory-region))))
+       (member (memory-descriptor-type memory-region) (a:ensure-list quil-type))))
     (constant
-     (equalp quil-type (constant-value-type obj)))))
+     (member (constant-value-type obj) (a:ensure-list quil-type)))))
```
appleby: why was the test here previously `equalp`? Does the `member` test now also need to be `equalp`?

tests/typed-memory-tests.lisp
```diff
@@ -92,3 +92,21 @@ STORE w y w
                               quil:classical-exclusive-or-octet/octet
                               quil:classical-load-real/real*/integer
                               quil:classical-store-bit*/integer/bit)))))
+
+(deftest test-classical-equality-immediate-types ()
+  (signals quil-type-error (parse-quil "DECLARE r BIT; DECLARE a BIT; EQ r a 1.0"))
+  (signals quil-type-error (parse-quil "DECLARE r BIT; DECLARE a OCTET; EQ r a 1.0"))
+  (signals quil-type-error (parse-quil "DECLARE r BIT; DECLARE a INTEGER; EQ r a 1.0"))
+  (signals quil-type-error (parse-quil "DECLARE r BIT; DECLARE a REAL; EQ r a 1"))
```
appleby: maybe throw in some test cases where right operand is an mref of the wrong type?

src/analysis/type-safety.lisp
notmgsk: I don't know, but no it doesn't need to be there: `constant-value-type` has type `quil-type` and those are comparable with `eq`.

notmgsk: `quil-type`s are something like singleton objects defined through `cl-algebraic-data-type`.

appleby: think of all the nanoseconds you could save with `:test #'eq` then

tests/typed-memory-tests.lisp
notmgsk: errrrrrrrgh I assumed `eq` was the default.

PR #495, opened 2019-12-04
Parse OpenQASM

notmgsk:
This (late) PR implements all functionality described in the [OpenQASM spec](https://github.com/Qiskit/openqasm/blob/master/spec-human/qasm2.pdf).

The following is a somewhat complex example, demonstrating a [ripple-carry adder OpenQASM program](https://github.com/Qiskit/openqasm/blob/master/examples/generic/adder.qasm):

```
(qvm-app::print-classical-memory (qvm:run-program 10 (parse-qasm-string "
// quantum ripple-carry adder from Cuccaro et al, quant-ph/0410184
OPENQASM 2.0;
include \"qelib1.inc\";
gate majority a,b,c 
{ 
  cx c,b; 
  cx c,a; 
  ccx a,b,c; 
}
gate unmaj a,b,c 
{ 
  ccx a,b,c; 
  cx c,a; 
  cx a,b; 
}
qreg cin[1];
qreg a[4];
qreg b[4];
qreg cout[1];
creg ans[5];
// set input states
x a[0]; // a = 0001
x b;    // b = 1111
// add a to b, storing result in b
majority cin[0],b[0],a[0];
majority a[0],b[1],a[1];
majority a[1],b[2],a[2];
majority a[2],b[3],a[3];
cx a[3],cout[0];
unmaj a[2],b[3],a[3];
unmaj a[1],b[2],a[2];
unmaj a[0],b[1],a[1];
unmaj cin[0],b[0],a[0];
measure b[0] -> ans[0];
measure b[1] -> ans[1];
measure b[2] -> ans[2];
measure b[3] -> ans[3];
measure cout[0] -> ans[4];")))
Classical memory (low -> high indexes):
    ans: 0 0 0 0 1
```

There are some things to address:
- [ ] There should be more error checking, e.g. making sure the sizes of registers are appropriate for the gate application.
- [x] In `parse-application` I explicitly define all of the gates defined in OpenQASM's `qelib1.inc`. This could be simplified (and instead just define the fundamental gates `U` and `CX`), but I'm not sure (because I haven't tried) whether that will Just Work.
- [ ] Automatically invoke the parser from `parse-quil` if the first (non-comment) line is `OPENQASM 2.0`.

I think the test suite is generally in a good state.  It covers all positive behaviours listed in the spec but it doesn't test for misuse.


src/qasm.lisp
```diff
@@ -0,0 +1,956 @@
...
+        (*line-number* 1)
+        (*line-start-position* 0))
+    (loop :for tok := (funcall f)
+          :until (null tok)
+          :collect tok)))
+
+(defun tokenize (string)
+  (let* (;; Forcefully swap semi-colons for newlines, to ease parsing
+         ;; down-the-line.
+         (string (substitute #\Newline #\; string))
```
stylewarning: These three string bindings ought to be factored out into a function, which you can write more efficiently avoiding SUBSTITUTE and REGEX-REPLACE-ALL. Something like:
```
(with-output-to-string (s)
  (loop :for ch :across string
        :do (case ch
              ((#\{ #\}) (format s "~%~C" ch))
              ((#\;) (terpri s)))))
```

src/qasm.lisp
```diff
@@ -0,0 +1,956 @@
...
+"))
+         (string (ppcre:regex-replace-all #\} string "
+}
+"))
+         (lines (nsplit ':NEWLINE (tokenize-line 'line-lexer string))))
+    lines))
+
+;; parse
+
+(define-condition q2q-parse-error (alexandria:simple-parse-error)
```
stylewarning: `qasm-parse-error` maybe?

src/qasm.lisp
```diff
@@ -0,0 +1,956 @@
...
+       (parse-application tok-lines))
+
+      ((:IF)
+       (parse-if tok-lines))
+
+      (otherwise
+       (q2q-parse-error "Got an unexpected token of type ~S ~
+                         when trying to parse a program." tok-type)))))
+
+(defmacro destructuring-token-bind (token-idents token-line &body body)
```
stylewarning: re-use?

src/qasm.lisp
```diff
@@ -0,0 +1,956 @@
...
+       ;; Some ugliness follows to remove ignored elements (those
+       ;; whose token name is the underscore).
+       (destructuring-bind ,(mapcar #'first (remove-if (lambda (tok) (eql '_ (car tok))) ts))
+           (loop :for (tok-name tok-type) :in ',ts
+                 :for tok :in ,token-line
+                 :unless (eql tok-name '_)
+                   :collect tok)
+         ,@body))))
+
+;; u1(alpha) qubit;
```
stylewarning: stray?

src/qasm.lisp
```diff
@@ -0,0 +1,956 @@
...
+        ;; While a qreg doesn't directly map to a quil instruction, we
+        ;; do need to keep track of named qubit registers so that we
+        ;; can properly translate instructions that reference qubit
+        ;; registers.
+        ;;
+        ;; e.g. the translation of "measure q[3] -> r[0];" might
+        ;; simply be "MEASURE 3 r[0]". But what about "measure q[3] ->
+        ;; r[0]; measure p[3] -> r[1];"? In OpenQASM the qubits "q[3]"
+        ;; and "p[3]" are not the same.
+        (setf (gethash name *qreg-names*) (list *qubit-count* length))
```
stylewarning: it would be good to check if the name exists, and if so, error, so we don't get weird errors clobbering this table

src/qasm.lisp
```diff
@@ -0,0 +1,956 @@
...
+        (incf *qubit-count* length)
+        (values nil
+                rest-toks)))))
+
+(defun parse-include (tok-lines)
```
stylewarning: it _might_ be worth trying to use Quil's include facility here.

src/qasm.lisp
```diff
@@ -0,0 +1,956 @@
...
+        (check-register-is-defined creg)
+        
+        (values (map-registers (lambda (src dest)
+                                 (make-instance 'quil:measure
+                                                :qubit src
+                                                :address dest))
+                               qreg creg)
+                (rest tok-lines))))))
+
+(defun quilify-qasm-comment (str)
```
stylewarning: Is this used? Delete?

src/qasm.lisp
```diff
@@ -0,0 +1,956 @@
...
+
+  (destructuring-bind (openqasm-toks . rest-toks)
+      tok-lines
+    (destructuring-token-bind ((_ :OPENQASM) (version :REAL))
+        openqasm-toks
+      (values (quil::make-pragma (list "OPENQASM"
+                                       (format nil "~A" (token-payload version))))
+              rest-toks))))
+
+(defun parse-cregister (tokens)
```
stylewarning: could this function benefit from the destructuring-token-bind macro?

src/qasm.lisp
```diff
@@ -0,0 +1,956 @@
...
+               (eql (token-type next) ':LEFT-SQUARE-BRACKET))
+          (destructuring-token-bind ((index :NNINTEGER) (_ :RIGHT-SQUARE-BRACKET))
+              (subseq tokens 2 4)
+            (values (creg (token-payload id)
+                          (token-payload index))
+                    (subseq tokens 4)))
+          (values (creg (token-payload id))
+                  (rest tokens))))))
+
+(defun parse-qregister (tokens)
```
stylewarning: destructuring-token-bind?

src/qasm.lisp
```diff
@@ -0,0 +1,956 @@
...
+(defun parse-params (tokens)
+  "Parse a list of qasm params (e.g. in the instruction  rx(0.5) q;). Returns a list of parameter values (type float), and a second value which is the remaining tokens (not including closing parenthesis)."
+  (q2q-check-unexpected-eof tokens "parameters")
+  (q2q-check-token-type (first tokens) :LEFT-PAREN)
+  
+  (let ((rp-pos (position ':RIGHT-PAREN tokens :key 'token-type)))
+    (unless rp-pos
+      (q2q-parse-error "Could not find token of type ':RIGHT-PAREN in parameter list ~A." tokens))
+    (let* ((subseq (subseq tokens 1 rp-pos))
+           ;; TODO Raise an error when parsing something like "(0, 0,)"
```
stylewarning: you could tell `split-sequence` to not remove empty subseqs, and error if there are any empty ones

src/qasm.lisp
```diff
@@ -0,0 +1,956 @@
...
+        ;; TODO Check rest-toks is nil, or refactor to not require
+        ;; that check.
+        (list name-tok params registers)))))
+
+(defun check-number-of-parameters (params number)
+  (unless (= number (length params))
+    (q2q-parse-error "Expected ~A parameters but found ~A."
+                     number (length params))))
+
+(defun build-u-gate (θ ϕ λ qubit)
```
stylewarning: gottem

src/qasm.lisp
```diff
@@ -0,0 +1,956 @@
...
+  (list
+   (quil::build-gate "RZ" `(,λ) qubit)
```
stylewarning: can you provide a reference somewhere to this gate?

src/qasm.lisp
```diff
@@ -0,0 +1,956 @@
...
+   (quil::build-gate "RY" `(,ϕ) qubit)
+   (quil::build-gate "RZ" `(,θ) qubit)))
+
+(defun map-registers (function register &rest more-registers)
+  "Apply FUNCTION to successive sets of registers.
+
+In OpenQASM a gate can be applied to an indexed register like so: rx(pi) q[0].  Alternatively, a gate can also be applied to an unindexed register: rx(pi) q.  In this case, the gate is applied to all the qubits in that register.  More complex cases occur where multiple registers are involved, and some subset of those are unindexed and so the gate must be expanded into multiple gate applications over the appropriate registers.
```
stylewarning: Wow... @_@

src/qasm.lisp
```diff
@@ -0,0 +1,956 @@
...
+  qreg p[3]; qreg q[2]; qreg r[3];
+
+  ccx p[0], q[1], r[0];
+  ccx p[1], q[1], r[1];
+  ccx p[2], q[1], r[2];
+
+It is an error to have unindexed registers of unequal size.
+
+Note: the above \"expansion\" is not performed when in a gate body."
+  (let ((registers (append (list register) more-registers)))
```
stylewarning: `let*`

src/qasm.lisp
```diff
@@ -0,0 +1,956 @@
...
+             (apply function (mapcar #'register-to-quil-object registers)))))))
+
+(defun parse-application (tok-lines)
+  (let ((application-toks (collect-single-application-from-tokens (first tok-lines)))
+        (rest-toks (rest tok-lines)))
+    (destructuring-bind (name (&rest params) (&rest registers))
+        application-toks
+      (declare (ignorable params registers))
+      (let ((name (token-payload name)))
+        (cond
```
stylewarning: this function is definitely asking for something to be refactored, but I think this is a good first stab. Not sure I'd hold up this PR to get this cleaned up, but it's definitely something `TODO`

src/qasm.lisp
```diff
@@ -0,0 +1,956 @@
...
+      (setf (gethash gate-name *gate-names*) t)
+      (values (quil::make-circuit-definition
+               gate-name
+               gate-params
+               (mapcar #'quil::formal gate-qargs)
+               (parse-gate-body (subseq tok-lines 2 close-pos))
+               :context nil)
+              (subseq tok-lines (1+ close-pos))))))
+
+(defun parse-opaque (tok-lines)
```
stylewarning: oddball construct. Would it be better to have something like
```
opaque foo ...
```
get converted into
```
PRAGMA QASM_OPAQUE
foo ...
```
? Or will this suffer from `foo` not being defined anywhere?

src/qasm.lisp
```diff
@@ -0,0 +1,956 @@
...
+                         (mapcar #'register-to-quil-object qregs)))
+                rest-lines)))))
+
+(defun parse-if (tok-lines)
+  (q2q-check-unexpected-eof tok-lines "if")
+
+  (destructuring-bind (if-toks . rest-lines)
+      tok-lines
+    (pop if-toks)
+    (q2q-check-token-type (pop if-toks) ':LEFT-PAREN)
```
stylewarning: sounds like we can maybe use token destructuring?

src/qasm.lisp
```diff
@@ -0,0 +1,956 @@
...
+              ;; little strange at first glance.  You might expect
+              ;; that you would compare bits of classical memory with
+              ;; something like `if(c[0] == 1) ...` but actually you
+              ;; compare the entire classical register against a
+              ;; number (bitwise) `if(c == 5) ...` tests
+              ;;
+              ;;     c[0] == 1 & c[1] == 0 & c[2] == 1 & c[3] == 0 
+              ;;
+              ;; etc.
+              (loop :for i :below (1+ (ceiling (log (token-payload val) 2)))
```
stylewarning: food for thought, we can use Quil's memory sharing to accomplish this, but it will be tricky to get it 100% right

```
DECLARE x_int INTEGER
DECLARE x BIT[5] SHARING x_int
```
now `x_int` can be used to understand the integer interpretation of `x`.

stylewarning: In real life, however, current control systems will probably do an `and` operation

src/qasm.lisp
```diff
@@ -0,0 +1,956 @@
...
+                    :for reg := (quil:copy-instance register)
+                    :do (setf (reg-index reg) i)
+                    :collect
+                    (make-instance 'quil::classical-equality-bit/bit/immediate
```
stylewarning: You might opt to simplify the control flow and reduce jumping, which is intrinsically expensive.

You want to jump only when all bits match. so you can AND all those together.

Example: For the reg `x[3]` and the number `#b110`, you'd want

```
DECLARE temp BIT
DECLARE result BIT
EQ temp x[0] 0
AND result result temp
EQ temp x[1] 1
AND result result temp
EQ temp x[2] 1
AND result result temp
JUMP-UNLESS @skip result
<body of if branch>
LABEL @skip
```

src/qasm.lisp
```diff
@@ -0,0 +1,956 @@
...
+(deftype token-type ()
+  '(or
+    qasm-keywords
+    (member :SEMI-COLON :LEFT-PAREN :RIGHT-PAREN :COMMA
+            :ARROW :LEFT-SQUARE-BRACKET :RIGHT-SQUARE-BRACKET :LEFT-CURLY-BRACKET
+            :RIGHT-CURLY-BRACKET :LEFT-ANGLE-BRACKET :RIGHT-ANGLE-BRACKET :PLUS
+            :MINUS :TIMES :DIVIDE :EXPT :EQUALSEQUALS :NNINTEGER :REAL :ID
+            :STRING :KEYWORD :OPENQASM :QREG :CREG :GATE :COMMENT :PI)))
+
+(defparameter *valid-functions*
```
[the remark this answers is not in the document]
notmgsk: These are actually not used in OpenQASM, so will delete.

src/qasm.lisp
```diff
@@ -0,0 +1,956 @@
...
+(defparameter *gate-params* nil
+  "When parsing the body of a gate declaration, this is a list of the parameter names.")
+
+(defparameter *creg-names* (make-hash-table :test #'equal)
+  "Maps a creg name to its size.")
+
+(defparameter *qreg-names* (make-hash-table :test #'equal)
+  "Maps a qreg name to the pair (offset . size). A qreg defined with `qreg q[size];` maps to Quil qubits (offset, offset + 1, offset + 2, ..., offset + size - 1). This complication maintains unique qubits in the Quil translation.")
+
+(defparameter *qubit-count* 0
```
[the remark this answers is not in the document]
notmgsk: If I do that, sbcl complains that `*qubit-count*` is not defined.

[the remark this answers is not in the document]
notmgsk:
```
CL-USER> (defvar *test*)
*TEST*
CL-USER> (setf (documentation *test* 'variable) "hi")
; Evaluation aborted on #<UNBOUND-VARIABLE *TEST* {1006F86D63}>.
```

src/qasm.lisp
```diff
@@ -0,0 +1,956 @@
...
+  (:method ((qreg qasm-qreg))
+    (with-slots (name index) qreg
+      (if *gate-applications-are-formal*
+          (quil:formal name)
+          (destructuring-bind (offset size)
+              (gethash name *qreg-names*)
+            (assert (< index size) ()
+                    "The index ~s is out-of-bounds for qreg ~s."
+                    index qreg)
+            (quil::qubit (+ offset index))))))
```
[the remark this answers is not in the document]
notmgsk: it is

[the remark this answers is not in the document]
notmgsk: it is

src/qasm.lisp
```diff
@@ -0,0 +1,956 @@
...
+   (return (tok :REAL (parse-float:parse-float $@))))
+  ((eager "{{INT}}")
+   (return (tok :NNINTEGER (parse-integer $@))))
+  ("{{IDENT}}"
+   (return (tok :ID $@)))
+  ("[^\\S\\n\\r]+"
+   nil))
+
+;;; SPLIT-SEQUENCE was too slow for 10k+ tokens.
+(defun nsplit (wedge list)
```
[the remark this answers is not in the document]
notmgsk: No. It is pretty specific to parsing Quil, AFAICT.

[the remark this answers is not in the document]
notmgsk: Actually, maybe.

src/qasm.lisp
```diff
@@ -0,0 +1,956 @@
...
+(defun qreg (name &optional index)
+  (make-instance 'qasm-qreg :name name :index index))
+
+(defclass qasm-creg (qasm-reg)
+  ())
+
+(defun creg (name &optional index)
+  (make-instance 'qasm-creg :name name :index index))
+
+(defun print-reg (reg &optional stream)
```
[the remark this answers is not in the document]
notmgsk: I opted to write a `reg-fmt` like `instruction-fmt`.

src/qasm.lisp
notmgsk: Gonna need some more exposition on this at some point :D

src/qasm.lisp
```diff
@@ -0,0 +1,956 @@
...
+  (let ((name (reg-name reg))
+        (index (reg-index reg)))
+    (if index
+        (format stream "~A[~A]" name index)
+        (format stream "~A" name))))
+
+(defgeneric register-to-quil-object (register)
```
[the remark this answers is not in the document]
notmgsk: I'm not sure I understand what you mean RE special variables -- isn't that what `*gate-applications-are-formal*` and `*qreg-names*` are?

src/qasm.lisp
```diff
@@ -0,0 +1,956 @@
...
+                     (setf pieces-last (cdr pieces-last))
+                     (setf start (cdr end))
+                     (rplacd last nil)
+                     (doit start start start))))
+                 ;; Keep on truckin'.
+                 (t
+                  (doit start end (cdr end))))))
+      (cdr (doit list list list)))))
+
+(defun tokenize-line (lexer string)
```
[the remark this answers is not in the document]
notmgsk: Not quite. These used to be the same, but the quilt parsing PR changed it to be more quil-specific (less general).

src/qasm.lisp
notmgsk: No. That macro doesn't allow for optional tokens. A cregister can look like `c` or `c[i]`, and I need to account for both of those possible branches.

src/qasm.lisp
notmgsk: No, for same reason as above.

src/qasm.lisp
notmgsk: The latter. I suppose I could just treat it as an empty DEFCIRCUIT, but I might also like to see in compiled output where my opaque (lol) gate would be applied (and an empty circuit application would just be stripped).

stylewarning:
@notmgsk Ping when re-review is desired.

src/qasm.lisp
```diff
@@ -0,0 +1,928 @@
...
+       (destructuring-bind ,(mapcar #'first (remove-if (lambda (tok) (eql '_ (car tok))) ts))
+           (loop :for (tok-name tok-type) :in ',ts
+                 :for tok :in ,token-line
+                 :unless (eql tok-name '_)
+                   :collect tok)
+         ,@body))))
+
+(defun parse-creg-definition (tok-lines)
+  (when (null tok-lines)
+    (qasm-parse-error "Unexpectedly reached end of program, expecting creg."))
```
appleby: `qasm-check-unexpected-eof`?

src/qasm.lisp
appleby: docstring for `*qreg-names*` says the hash value is a dotted pair, but this is a proper list. i liked a dotted pair for this but i support u either way

src/qasm.lisp
```diff
@@ -0,0 +1,928 @@
...
+(setf (documentation '*creg-names* 'variable)
+      "Maps a creg name to its size.")
+
+(defvar *qreg-names*)
+(setf (documentation '*qreg-names* 'variable)
+      "Maps a qreg name to the pair (offset . size). A qreg defined with `qreg q[size];` maps to Quil qubits (offset, offset + 1, offset + 2, ..., offset + size - 1). This complication maintains unique qubits in the Quil translation.")
+
+(defvar *qubit-count*)
+(setf (documentation '*qubit-count* 'variable)
+      "The number of qubits registered by qreg.")
```
appleby:
```suggestion
      "The total number of qubits registered by qregs in the program.")
```

src/qasm.lisp
```diff
@@ -0,0 +1,928 @@
...
+  ;; TODO Throw some lib-specific condition if destructuring-bind
+  ;; fails.
+  ;;
+  ;; TODO Allow for "fall-through" cases. i.e. if the first
+  ;; TOKEN-IDENTS doesn't match, try the next one, etc. I'm thinking
+  ;; this might be helpful when matching `measure a -> b` vs `measure
+  ;; a[0] -> b[0]`.
+  ;;
+  ;; TODO I wrote this while very tired. Untested.
+  (let ((ts (mapcar #'alexandria:ensure-list (remove '_ token-idents))))
```
appleby: feels like there should be a smattering of `once-only`s but I didn't check carefully

src/qasm.lisp
```diff
@@ -0,0 +1,928 @@
...
+         ,@body))))
+
+(defun parse-creg-definition (tok-lines)
+  (when (null tok-lines)
+    (qasm-parse-error "Unexpectedly reached end of program, expecting creg."))
+
+  (destructuring-bind (creg-toks . rest-toks)
+      tok-lines
+    (destructuring-token-bind ((_ :CREG) (name-tok :ID) (_ :LEFT-SQUARE-BRACKET)
+                               (length-tok :NNINTEGER) (_ :RIGHT-SQUARE-BRACKET))
```
appleby: If `destructuring-token-bind` allowed a dotted tail for `token-idents` then you wouldn't need to nest inside `destructuring-bind` but otoh it's probably a pain so I see where you're coming from. ok good talk.

src/qasm.lisp
```diff
@@ -0,0 +1,928 @@
...
+        include-toks
+      ;; TODO Some error checking.
+      (let ((file (uiop:read-file-string (token-payload path-tok))))
+        (values (parse-qasm-body file)
+                rest-toks)))))
+
+(defun parse-measure (tok-lines)
+  (qasm-check-unexpected-eof tok-lines "measure")
+
+  (let ((measure-toks (first tok-lines)))
```
appleby: `destructuring-token-bind`?

notmgsk: Wherever a qreg or creg is expected, they can appear in two possible formats: indexed (e.g. `q[0]`) or unindexed `q`).  `destructuring-token-bind` is too inflexible for this use-case.  It probably could be reworked, but I'm not sure it's worth the effort.

src/qasm.lisp
notmgsk: A future `TODO`.

src/qasm.lisp
```diff
@@ -0,0 +1,928 @@
...
+It is an error to have unindexed registers of unequal size.
+
+Note: the above \"expansion\" is not performed when in a gate body."
+  (let* ((registers (append (list register) more-registers))
+         (unindexed-registers (remove-if-not #'null registers :key #'reg-index)))
+    (cond ((and unindexed-registers
+                ;; This prevents us from expanding, e.g., `x q` in
+                ;; `gate a q { x q; };`. Perhaps this guard should be
+                ;; performed by the caller however, as it makes this
+                ;; function less general than its name would imply.
```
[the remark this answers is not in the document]
notmgsk: AFAIK, no.

src/qasm.lisp
notmgsk: Donezo.

src/qasm.lisp
```diff
@@ -0,0 +1,928 @@
...
+                                                          :operator (quil:named-operator name)
+                                                          :parameters params
+                                                          :arguments qubits))
+                         registers))
+              rest-toks)
+             (qasm-parse-error "Found unknown gate application '~A'."
+                              name))))))))
+
+(defun parse-gate-body (tok-lines)
+  (alexandria:flatten (mapcar #'parse-application (mapcar #'list tok-lines))))
```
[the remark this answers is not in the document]
notmgsk: @rigettizach made me feel bad about that, and now I think we should just `:import-from` 😬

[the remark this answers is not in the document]
appleby: I'm fine with whatever, but consistency is good, too.

src/qasm.lisp
```diff
@@ -0,0 +1,928 @@
...
+
+(defun collect-gate-decl (tok-line)
+  (let ((name (token-payload (second tok-line))))
+    (multiple-value-bind (params rest-toks)
+        (maybe-parse-params (subseq tok-line 2))
+      (dolist (p params) (push p *gate-params*))
+      (let ((qregs (mapcar #'reg-name (parse-qregisters rest-toks))))
+        (dolist (q qregs) (push q *gate-qregs*))
+        (list name params qregs)))))
```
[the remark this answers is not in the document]
notmgsk: Hm. I'm not saying I have adhered to this style entirely, but my sense of when to use `values` is when you have some data that is not necessarily the result of the computation performed by the function, or otherwise is not necessary to the utility of the function. In the `parse-*` functions, the tasty part of the function is really the thing that was parsed, and whatever remains can be ignored or not, depending on your interest. The `collect-*` functions return only useful info -- a gate declaration really is all of `name`, `params`, and `qregs`.

That said, using `values` everywhere gives a nice consistency.

src/qasm.lisp
notmgsk: Since the current impl works, I think this is worth a future `TODO`.

src/qasm.lisp
```diff
@@ -0,0 +1,928 @@
...
+              ;; little strange at first glance.  You might expect
+              ;; that you would compare bits of classical memory with
+              ;; something like `if(c[0] == 1) ...` but actually you
+              ;; compare the entire classical register against a
+              ;; number (bitwise) `if(c == 5) ...` tests
+              ;;
+              ;;     c[0] == 1 & c[1] == 0 & c[2] == 1 & c[3] == 0 
+              ;;
+              ;; etc.
+              (loop :for i :below (1+ (ceiling (log (token-payload val) 2)))
```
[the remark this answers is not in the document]
notmgsk: @appleby wow good spot!

notmgsk:
@appleby @stylewarning ping for re-review at you leisure

src/qasm.lisp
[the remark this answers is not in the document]
appleby: Interesting points, and I agree about `values` being useful for situations where values 1-n are supposed to be ignorable. I think this is one of those things that doesn't really matter and everyone has little preferences picked up via osmosis. To my mind: return a list when you actually want to do something with the returned list and or manipulate it as a list, or when you want to return an unknown or large number of values or something. Or when the returned value is inherently "listy" (can't argue with that). In this case, the only place this function is called, the list is immediately unpacked with `destructuring-bind`, so feels like an unnecessary `cons`. But now I'm assuming the compiler will return `(values x0 x1 .. xn)` in registers for small n, and this probably depends on your lisp impl and processor arch, etc. And I guess there is a real possibility that someone will accidentally ignore the additional values, so maybe I'm coming around to your way of seeing it.

In any case, I love these discussions because even though it's ultimately bikeshedding and doesn't matter, it got me to think about a previously unquestioned assumption/preference so cheers for that.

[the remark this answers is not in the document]
notmgsk: You honour me, good Sir.

src/qasm.lisp
appleby: marked resolved but maybe not resolved? (documented vs actual "type")

src/qasm.lisp
notmgsk: TODO

src/qasm.lisp
notmgsk: resolved resolved

src/qasm.lisp
```diff
@@ -0,0 +1,816 @@
...
+              ;; something like `if(c[0] == 1) ...` but actually you
+              ;; compare the entire classical register against a
+              ;; number (bitwise) `if(c == 5) ...` tests
+              ;;
+              ;;     c[0] == 1 & c[1] == 0 & c[2] == 1 & c[3] == 0
+              ;;
+              ;; etc.
+              ;;
+              ;; TODO Clean this up, possibly using Quil's SHARING.
+              (loop :for i :below (integer-length (token-payload val))
```
appleby: Does this do what you want when `val` is zero? For example, `(integer-length 0) => 0`, and this will not emit any comparison instrs, no?

notmgsk: Correct.  I assumed `(integer-length 0) => 1`.

src/qasm.lisp
```diff
@@ -0,0 +1,822 @@
...
+        (multiple-value-bind (gate-application rest-toks)
+            (parse-application (list rest-toks))
+          (assert (null rest-toks))
+          (let* ((cmp-name (string (gensym "CMP-")))
+                 (cmp-desc (quil::make-memory-descriptor
+                            :name cmp-name
+                            :type quil::quil-bit
+                            :length 1))
+                 (jmp-label (quil::label (string (gensym "JMP-")))))
+            (unless (<= (token-payload val)
```
appleby:
```suggestion
            (unless (<= (integer-length (token-payload val))
```

It's ya boi log2

notmgsk: Making quite the mess of this. I used `(expt 2 register-size)` instead, because part of me feels that multiplication is faster. But these are integers, so it probably doesn't matter.

appleby: works for me.  fwiw, it looks like `integer-length` compiles down a [bsr](https://c9x.me/x86/html/file_module_x86_id_20.html) instruction + a handful of other bit twiddling/bookkeeping instructions on sbcl / x86.  No actual log2s were injured in the making of this function.

but `(expt 2 n)` is also optimized to an `ash`, so no actual multiplication there either and probably still faster

src/cl-quil.lisp
```diff
@@ -133,18 +133,34 @@ This also signals ambiguous definitions, which may be handled as needed."
...
   "Parse and process the Quil string STRING, which originated from the file ORIGINATING-FILE. Transforms in TRANSFORMS are applied in-order to the processed Quil string.
 
 In the presence of multiple definitions with a common signature, a signal is raised, with the default handler specified by AMBIGUOUS-DEFINITION-HANDLER.
 "
-  (%parse-quil string
-               #'raw-quil-to-unresolved-program
-               :originating-file originating-file
-               :transforms transforms
-               :ambiguous-definition-handler ambiguous-definition-handler))
+  (if (%check-for-qasm-header string)
```
notmgsk: @stylewarning @appleby I'm not sure if this is the best place to do this.  Feels naughty to overload the meaning of `parse-quil`.

appleby: Not sure. Maybe the check should happen in `quilc`, which then calls `parse-quil` or `parse-qasm` as appropriate?

src/cl-quil.lisp
```diff
@@ -133,18 +133,42 @@ This also signals ambiguous definitions, which may be handled as needed."
         (dolist (xform transforms pp)
           (setf pp (transform xform pp)))))))
 
+;; A valid OpenQASM program requires the line "OPENQASM 2.0" possibly
+;; preceded by comments (//).
+(defun %check-for-qasm-header (string)
+  (labels ((line-begins-with (prefix line)
+             (let ((pos (search prefix (string-left-trim '(#\Space) line))))
```
rigettizach: Here's a slightly bummed version:

    (line-begins-with (prefix line)
      (when (<= (length prefix) (length line))
        (let* ((start2 (or (position #\Space line :test-not #'eql) 0))
               (end2 (+ start2 (length prefix))))
          (string= prefix line :start2 start2 :end2 end2))))

src/cl-quil.lisp
stylewarning: I think it should happen in CL-QUIL so QVM can work with it too without duplicating code everywhere.

src/qasm.lisp
```diff
@@ -0,0 +1,821 @@
...
+       (parse-include tok-lines))
+
+      ((:QREG)
+       (parse-qreg-definition tok-lines))
+
+      ((:CREG)
+       (parse-creg-definition tok-lines))
+
+      ((:BARRIER)
+       (values (quil::make-pragma
```
stylewarning: We could represent the pragma more nicely. Just `TODO` it.

src/qasm.lisp
```diff
@@ -0,0 +1,821 @@
...
+                (list "QASM_BARRIER")
+                (format nil "~{~/quil::instruction-fmt/~^, ~}"
+                        (mapcar (lambda (qreg)
```
stylewarning: i prefer to `let *gate-..*` outside of the MAPCAR

src/qasm.lisp
```diff
@@ -0,0 +1,821 @@
...
+
+      ((:OPAQUE)
+       (parse-opaque tok-lines))
+
+      ((:MEASURE)
+       (parse-measure tok-lines))
+
+      ((:RESET)
+       (values
+        (map-registers (lambda (qub) (make-instance 'quil::reset-qubit :target qub))
```
stylewarning: is this not exported? it should be

src/qasm.lisp
```diff
@@ -0,0 +1,821 @@
...
+        ;; and "p[3]" are not the same.
+        (when (gethash name *qreg-names*)
+          (qasm-parse-error "Attempting to redefine qreg ~A." name))
+        (setf (gethash name *qreg-names*) (list *qubit-count* length))
+        (incf *qubit-count* length)
+        (values nil
+                rest-toks)))))
+
+(defun parse-include (tok-lines)
+  ;; TODO Use cl-quil's include?
```
stylewarning: No, this should be resolved now because it's a security issue (allows potentially reading from the filesystem). Quil's INCLUDE has facilities to deal with safe places to read files from, and we need to remain in that Bubble of Safety.

src/qasm.lisp
```diff
@@ -0,0 +1,821 @@
...
+  (qasm-check-unexpected-eof tok-lines "include")
+
+  (destructuring-bind (include-toks . rest-toks)
+      tok-lines
+    (destructuring-token-bind ((_ :INCLUDE)
+                               (path-tok :STRING))
+        include-toks
+      ;; TODO Some error checking.
+      (let ((file (uiop:read-file-string (token-payload path-tok))))
```
stylewarning: `/!\   DEFCON 1   /!\`

src/qasm.lisp
```diff
@@ -0,0 +1,821 @@
...
+        (assert (null rest-toks))
+
+        (values (map-registers (lambda (src dest)
+                                 (make-instance 'quil:measure
+                                                :qubit src
+                                                :address dest))
+                               qreg creg)
+                (rest tok-lines))))))
+
+(defun parse-comment (tok-lines)
```
stylewarning: Is this needed? Isn't the lexer just nuking them out of orbit?

src/qasm.lisp
```diff
@@ -0,0 +1,821 @@
...
+
+  (destructuring-bind (openqasm-toks . rest-toks)
+      tok-lines
+    (destructuring-token-bind ((_ :OPENQASM) (version :REAL))
+        openqasm-toks
+      (values (quil::make-pragma (list "OPENQASM"
+                                       (format nil "~A" (token-payload version))))
+              rest-toks))))
+
+;; TODO parse-cregister and parse-qregister have a common theme. For
```
stylewarning: factor it out in this PR; it looks easy. You just need to pass `#'qreg` or `#'creg` to the factored function:

```
(defun parse-register (tokens reg-ctor)
  ...
  (funcall reg-ctor id idx)
  ...
)

(defun parse-cregister (tokens)
  (parse-register tokens #'creg))

(defun parse-qregister (tokens)
  (parse-register tokens #'qreg))
```

src/qasm.lisp
```diff
@@ -0,0 +1,821 @@
...
+                 (eql (token-type next-tok) ':COMMA)) :do
+                   (setf rest-toks (rest rest-toks))))
+
+(defun %stringify-token-payload (token)
+  (let ((payload (token-payload token)))
+    (cond ((and *gate-params*
+                (find payload *gate-params* :test #'equalp :key #'quil:param-name))
+           (format nil "(%~A)" payload))
+          ((eql (token-type token) ':LEFT-PAREN)
+           (format nil "("))
```
stylewarning: dont format nil

src/qasm.lisp
```diff
@@ -0,0 +1,821 @@
...
+          ((eql (token-type token) ':RIGHT-PAREN)
+           (format nil ")"))
```
stylewarning: don't format nil

src/qasm.lisp
```diff
@@ -0,0 +1,821 @@
...
+                    (find payload *gate-params* :test #'equalp))
+               (%formalize payload))
+              (t
+               (quil::param payload))))
+      ;; Otherwise we have a compound expression like lambda/2, in
+      ;; which case we need to recombine that into a single string,
+      ;; and then pass it through cl-quil's arithmetic parsing stuff.
+      ;; This is a bit hacky, but I don't want to duplicate the
+      ;; arithmetic stuff in cl-quil.
+      (let* ((str (format nil "~A"
```
stylewarning:
```
(with-output-to-string (s)
  (dolist (tok tokens)
    (write-string (%stringify... tok) s)))
```
is a little cleaner

src/qasm.lisp
```diff
@@ -0,0 +1,821 @@
...
+          (parse-qregisters rest-toks)
+        (declare (ignore rest-toks))
+        ;; TODO Check rest-toks is nil, or refactor to not require
+        ;; that check.
+        (list name-tok params registers)))))
+
+(defun check-number-of-parameters (params number)
+  (unless (= number (length params))
+    (qasm-parse-error "Expected ~A parameters but found ~A."
+                     number (length params))))
```
stylewarning: indent

src/cl-quil.lisp
```diff
@@ -133,18 +133,42 @@ This also signals ambiguous definitions, which may be handled as needed."
...
+               (and pos (= 0 pos))))
+           (commented-line-p (line) (line-begins-with "//" line))
+           (qasm-line-p (line) (line-begins-with "OPENQASM 2.0" line)))
+    (with-input-from-string (*standard-input* string)
+      (loop :for line := (read-line *standard-input* nil) :do
+        (cond ((qasm-line-p line)
+               ;; Certainly an OpenQASM program.
+               (return t))
+              ((or (string= "" line)
```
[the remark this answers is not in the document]
notmgsk: oops

src/cl-quil.lisp
```diff
@@ -133,18 +133,42 @@ This also signals ambiguous definitions, which may be handled as needed."
...
+               ;; cannot be legal OpenQASM.
+               (return nil)))))))
+
 (defun parse-quil (string &key originating-file
                             (transforms *standard-post-process-transforms*)
                             (ambiguous-definition-handler #'continue))
   "Parse and process the Quil string STRING, which originated from the file ORIGINATING-FILE. Transforms in TRANSFORMS are applied in-order to the processed Quil string.
 
 In the presence of multiple definitions with a common signature, a signal is raised, with the default handler specified by AMBIGUOUS-DEFINITION-HANDLER.
 "
```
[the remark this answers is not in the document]
notmgsk: @stylewarning rather than overload `parse-quil`, I'm using a `cl-quil:parse` function which does the dispatching. That feels cleaner to me.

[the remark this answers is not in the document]
notmgsk: See latest diffs.

src/qasm.lisp
notmgsk: Seem's like this hasn't been enforced in quilc. See https://github.com/rigetti/quilc/pull/500

notmgsk: This should be addressed now. I'll let you resolve the comment if you agree.

src/qasm.lisp
notmgsk: 😆

src/cl-quil.lisp
```diff
@@ -133,6 +133,41 @@ This also signals ambiguous definitions, which may be handled as needed."
...
+               ;; Neither OPENQASM 2.0, comment, or empty line, so
+               ;; cannot be legal OpenQASM.
+               (return nil)))))))
+
+(defun parse (string &key originating-file
+                       (transforms *standard-post-process-transforms*)
+                       (ambiguous-definition-handler #'continue))
+  "Parse the input STRING which can be either Quil or OpenQASM code."
+  (if (%check-for-qasm-header string)
+      (quil.qasm:parse-qasm string)
```
appleby: Not saying it has to happen for this PR, but would be 🆒 to unify the interfaces of `parse-quil` and `parse-qasm` w.r.t. keyword args, assuming concepts like "ambiguous definitions" still apply in qasm.

src/qasm.lisp
```diff
@@ -0,0 +1,807 @@
...
+(defmacro destructuring-token-bind (token-idents token-line &body body)
+  ;; TODO Throw some lib-specific condition if destructuring-bind
+  ;; fails.
+  ;;
+  ;; TODO Allow for "fall-through" cases. i.e. if the first
+  ;; TOKEN-IDENTS doesn't match, try the next one, etc. I'm thinking
+  ;; this might be helpful when matching `measure a -> b` vs `measure
+  ;; a[0] -> b[0]`.
+  ;;
+  ;; TODO I wrote this while very tired. Untested.
```
appleby: probably want to `once-only` `token-line` or else `(check-type token-line symbol)` or some such

src/cl-quil.lisp
notmgsk: TODO

src/qasm.lisp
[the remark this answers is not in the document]
notmgsk: TODO 2.0

src/qasm.lisp
notmgsk: TODO 2.0

src/qasm.lisp
```diff
@@ -0,0 +1,807 @@
...
+  (etypecase register
+    (qreg *qreg-names*)
+    (creg *creg-names*)))
+
+(defun find-register (register &key (error-if-undefined t))
+  "Find the associated register data (in the appropriate namespace) for the given REGISTER.  If ERROR-IF-UNDEFINED, signal error if REGISTER is not defined."
+  (let ((register-val (gethash (reg-name register)
+                               (register-namespace register))))
+    (cond
+      ((not (null register-val))
```
jmbr:
```suggestion
      (register-val
```

src/qasm.lisp
```diff
@@ -0,0 +1,807 @@
...
+       register-val)
+      ((and (null register-val)
+            error-if-undefined)
+       (qasm-parse-error "Undefined register ~A." (reg-name register)))
+      (t
+       nil))))
+
+(defun register-offset (register)
+  (check-type register qreg)
+  (first (find-register register :error-if-undefined t)))
```
jmbr: I'd say you don't have to add `:error-if-undefined` given that `t` it is the default.

src/qasm.lisp
```diff
@@ -0,0 +1,807 @@
...
+   (return (tok :REAL (parse-float:parse-float $@))))
+  ((eager "{{INT}}")
+   (return (tok :NNINTEGER (parse-integer $@))))
+  ("{{IDENT}}"
+   (return (tok :ID $@)))
+  ("[^\\S\\n\\r]+"
+   nil))
+
+;;; SPLIT-SEQUENCE was too slow for 10k+ tokens.
+(defun nsplit (wedge list)
```
jmbr: I wonder if `uiop:split-string` is also slow.

src/qasm.lisp
```diff
@@ -0,0 +1,807 @@
+ ;;;; qasm.lisp
+
+(in-package #:cl-quil.qasm)
+
+(defvar *line-start-position*)
+(defvar *line-number*)
+
+(defvar *gate-names*)
```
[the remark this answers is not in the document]
rigettizach: Ediware has a macro like that, but I don't think it's external. Here's an internal version from cl-interpol:

```
(defmacro defvar-unbound (variable-name documentation)
  "Like DEFVAR, but the variable will be unbound rather than getting
an initial value.  This is useful for variables which should have no
global value but might have a dynamically bound value."
  ;; stolen from comp.lang.lisp article <k7727i3s.fsf@comcast.net> by
  ;; "prunesquallor@comcast.net"
  `(eval-when (:load-toplevel :compile-toplevel :execute)
    (defvar ,variable-name)
    (setf (documentation ',variable-name 'variable)
            ,documentation)))```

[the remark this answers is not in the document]
Comment (jmbr) [reply src/qasm.lisp]: Great! I didn't know about it. That `variable-name` is asking to be wrapped in an `alexandria:once-only`, though :-)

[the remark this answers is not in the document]
Comment (appleby) [reply src/qasm.lisp]: `variable-name` kinda has to be symbol by design though.

There is also a slightly-different version of this macro in hunchentoot that makes the doc string optional and makes sure to return the symbol name like `defvar`

```lisp
(defmacro defvar-unbound (name &optional (doc-string ""))
    "Convenience macro to declare unbound special variables with a
documentation string."
    `(progn
      (defvar ,name)
      (setf (documentation ',name 'variable) ,doc-string)
      ',name))
```

src/qasm.lisp
```diff
@@ -0,0 +1,807 @@
...
+    (values
+     (quil::raw-quil-to-unresolved-program code)
+     *creg-names*
+     *qreg-names*
+     *qubit-count*)))
+
+(defun parse-qasm (string)
+  "Parse STRING into a PARSED-PROGRAM object, applying all transforms."
+  (let ((pp (quil::resolve-objects (parse-qasm-into-raw-program string))))
+    (setf pp (quil::transform 'quil::expand-circuits pp))
```
jmbr: Suggestion: fuse `setf` forms.

src/qasm.lisp
```diff
@@ -0,0 +1,807 @@
...
+  "Parse STRING into a raw, untransformed PARSED-PROGRAM object."
+
+  (let* ((*creg-names* (make-hash-table :test #'equalp))
+         (*qreg-names* (make-hash-table :test #'equalp))
+         (*qubit-count* 0)
+         (*gate-applications-are-formal* nil)
+         (*gate-params* nil)
+         (*gate-names* (make-hash-table :test #'equalp))
+         (code (parse-qasm-body string)))
+    (setf code (quil::process-includes code))
```
jmbr: Suggestion: compose `parse-qasm-body` with `quil::process-includes` in the `let*` form above and discard the `setf`.

src/qasm.lisp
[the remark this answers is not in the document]
kilimanjaro: I absorbed it into `split-lines` when adding support for semicolons as newlines. At the time it was only used for this purpose, so it seemed harmless.

src/qasm.lisp
```diff
@@ -0,0 +1,807 @@
...
+    ;; Return the parsed sequence of objects.
+    (values
+     (quil::raw-quil-to-unresolved-program code)
+     *creg-names*
+     *qreg-names*
+     *qubit-count*)))
+
+(defun parse-qasm (string)
```
kilimanjaro: Minor request: a keyword argument `transforms` which defaults a list of the three you currently use

src/qasm.lisp
stylewarning: im not a fan in general of this kind of change when it's not a true boolean

src/cl-quil.lisp
```diff
@@ -133,6 +133,41 @@ This also signals ambiguous definitions, which may be handled as needed."
         (dolist (xform transforms pp)
           (setf pp (transform xform pp)))))))
 
+;; A valid OpenQASM program requires the line "OPENQASM 2.0" possibly
+;; preceded by comments (//).
+(defun %check-for-qasm-header (string)
+  (labels ((line-begins-with (prefix line)
```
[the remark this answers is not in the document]
jmbr: I just realized `string-left-trim` seems to be what you'd need in conjunction with `uiop:string-prefix-p` to replace `line-begins-with` here. I'm not asking for any action to be taken, just sayin'.

PR #496, opened 2019-12-04
Write unit test for EXP-PAULI.

### LINKED ISSUE #481: Add tests for EXP-PAULI
`exp-pauli` isn't tested in any way and it should be.

(@stylewarning merged `exp-pauli` without writing some tests for it.)

jmbr:
Closes #481 .

### DIFF
--- a/cl-quil-tests.asd
+++ b/cl-quil-tests.asd
@@ -9,6 +9,7 @@
   :depends-on (#:cl-quil
                (:version #:qvm "1.9.0")
                #:magicl
+               #:magicl-transcendental
                #:alexandria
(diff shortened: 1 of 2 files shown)


notmgsk:
I think we have avoided using `magicl-transcendental` since it requires `expokit` and we currently do not ship this library with the SDK.  I wonder if @stylewarning and @ecpeterson have more context (other than it meaning one more dependency).

jmbr:
FWIW, the `magicl-tests` package does depend on `magicl-transcendental`. I guess the question in this case is whether the SDK users are expected to run the test suite or not.

PR #498, opened 2019-12-09
Feature: "DEFEXPI"

### LINKED ISSUE #422: Introduce support for DEFGATE by exponentiated Hamiltonian
Pursuant to https://github.com/rigetti/quil/issues/29 , we will want to take on the following tasks in CL-QUIL / quilc:

* [ ] Upgrade the CL-QUIL parser and syntax tree to consume DEFGATEs specified by a Hamiltonian.
* [ ] Appropriately extend the `gate-matrix` family of routines to produce dense matrices from such definitions, so that `quilc` continues to function as-is.

### LINKED ISSUE #423: Limited parametric compilation of exponentiated 2Q Hamiltonians
An internally-voiced user request is to be able to parametrically compile generic two-qubit interactions. In a fully general interpretation this is too arithmetically complex a task to do on-the-fly, but #422 would permit support for the special case of a two-qubit interaction specified by a time-independent Hamiltonian. One could conjugate the Hamiltonian into the diagonal subgroup, parametrically compile the resulting diagonal operator, and decompose the static conjugation operators. The performance characteristics of the result will be slightly worse (≤ 8 CZs rather than the ≤ 3 CZs in the nonparametric case), but it might be better than nothing.

ecpeterson:
[Back at it again.](https://www.youtube.com/watch?v=SunaZVRxHcs)

This PR adds support for:

* A new gate definition class, `DEFGATE ... AS PAULI-SUM`, which defines a (possibly parametric) gate as an exponentiated Pauli sum (with possibly parametrically-determined coefficients). For example, the following gives a gate definition for `CPHASE`:

```
DEFGATE CPHASE(%alpha) p q AS PAULI-SUM:
    ZZ(-%alpha/4) p q
    Z(%alpha/4) p
    Z(%alpha/4) q
```

* Parametric compilation routines for Pauli sum gate definitions which are linear in a single gate parameter.

Some possible follow-ons:

* [ ] Benchmark this against the Tweedledum diagonal compiler.
* [ ] Investigate the diagonal expansion strategy. Can it be made usefully topology-sensitive?
* [x] Write some tests.

Attendant PRs:

* https://github.com/rigetti/quil/pull/38
* https://github.com/rigetti/pyquil/pull/1125

Closes #422. Closes #423.


src/ast.lisp
```diff
@@ -1528,7 +1543,26 @@ For example,
...
+            (gate-definition-name gate)
+            (mapcar #'string (pauli-sum-gate-definition-parameters gate))
+            (mapcar #'formal-name (pauli-sum-gate-definition-arguments gate)))
+    (dolist (pauli-term (pauli-sum-gate-definition-terms gate))
+      (with-slots (pauli-word prefactor arguments) pauli-term
+        (format stream "    ~a(" pauli-word)
+        (typecase prefactor
+          (number
+           (format stream "~a" prefactor))
+          (cons
```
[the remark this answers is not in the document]
ecpeterson: `(or cons symbol)`

src/compilers/linear-paulis.lisp
```diff
@@ -0,0 +1,252 @@
...
+                         total))
+                      (otherwise
+                       (give-up-compilation))))
+                   (number
+                    0)
+                   (symbol
+                    1)
+                   (otherwise
+                    (give-up-compilation)))))
+        (dolist (term terms)
```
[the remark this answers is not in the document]
ecpeterson: Definitely this deserves a comment, ty.

src/matrix-operations.lisp
```diff
@@ -391,3 +391,34 @@ as needed so that they are the same size."
        (matrix-equality
         kroned-ref-mat
         (scale-out-matrix-phases kroned-mat kroned-ref-mat))))))
+
+(defun matrix-expt (m s &key hermitian?)
```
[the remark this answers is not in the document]
ecpeterson: what would u prefer

src/matrix-operations.lisp
```diff
@@ -391,3 +391,34 @@ as needed so that they are the same size."
...
+  "Computes EXP(M*S).  Only works for unitarily diagonalizable matrices M."
+  (multiple-value-bind (d u)
+      (if hermitian? (magicl:hermitian-eig m) (magicl:eig m))
+    (assert (matrix-equality m
```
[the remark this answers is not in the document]
ecpeterson: Good idea.

src/matrix-operations.lisp
```diff
@@ -391,3 +391,34 @@ as needed so that they are the same size."
...
+(defun print-polar-matrix (m &optional (stream *standard-output*))
+  (let ((*print-fractional-radians* nil)
+        (*print-polar-form* t)
+        (height (magicl::matrix-rows m))
+        (width (magicl::matrix-cols m)))
+    (format stream "~&")
+    (dotimes (i height)
+      (dotimes (j width)
+        (let* ((z (magicl:ref m i j))
+               (abs (if (double= 0d0 (abs z)) 0d0 (abs z)))
```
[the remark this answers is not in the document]
ecpeterson: This truncation is genuinely useful for pretty-printing.

src/compilers/linear-paulis.lisp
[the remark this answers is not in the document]
notmgsk: yw

src/matrix-operations.lisp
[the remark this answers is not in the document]
notmgsk: `is-hermitianp-p`

src/compilers/linear-paulis.lisp
```diff
@@ -0,0 +1,252 @@
...
+          (dolist (term nonlocal-terms)
+            (let ((Z-pos (loop :for letter :across (pauli-term-pauli-word term)
+                               :for arg :in (pauli-term-arguments term)
+                               :for j :from 0
+                               :when (and (eql #\Z letter)
+                                          (eql vote (position arg arguments :test #'equalp)))
+                                 :do (return j))))
+              (cond
+                (Z-pos
+                 (push (list term Z-pos) Zs))
```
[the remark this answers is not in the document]
ecpeterson: I don't think so. The gate itself carries a list of qubit formals `qn ... q0`, and then each Pauli summand carries a list of qubit formals `qjm ... qj0`, where each `qjk` is one of the `qi`s from the original list. `vote` remembers the position of which qubit formal from the parent list appears most often; `Z-pos` tracks whether there's a position in the term's list which matches the one selected by `vote`.

src/matrix-operations.lisp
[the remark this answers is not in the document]
appleby: `¿hermitian?`

src/compilers/linear-paulis.lisp
```diff
@@ -0,0 +1,252 @@
...
+                                         :parameters parameters
+                                         :arity arity
+                                         :dimension dimension
+                                         :name "Is"
+                                         :terms Is)))
+              (inst* I-gate
+                     (application-parameters instr)
+                     (application-arguments instr))))
+          ;; emit the Zs
+          (labels
```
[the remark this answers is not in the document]
ecpeterson: This could be broken apart. What it's doing is finding the qubit on which the most Z terms act, then systematically removing all such Z terms by (repeatedly) finding large collections of ZZ pairs, with one of the Zs being the original Z, and conjugating by CNOT to convert the original Z into an I. This could be done just once, rather than repeatedly, with control then remitted to the overall compilation apparatus.

I don't know if it's beneficial, but it is in the spirit of quilc.

notmgsk:
Um pequeno erro:
```
(quil:parse-quil "DEFGATE CPHASE(%alpha) p q AS PAULI-SUM:
    ZZ(-%alpha) p q
    Z(%alpha) p
    Z(%alpha) q

")
```

src/compilers/linear-paulis.lisp
```diff
@@ -0,0 +1,261 @@
...
+        (setf vote (array-argmax votes)
+              control-qubit (nth vote (application-arguments instr)))
+        ;; now we re-sort the nonlocal terms into two buckets: those with a Z in
+        ;; the voted-upon location, and those without
+        (let* ((Is nil) (Zs nil))
+          (dolist (term nonlocal-terms)
+            (let ((Z-pos (loop :for letter :across (pauli-term-pauli-word term)
+                               :for arg :in (pauli-term-arguments term)
+                               :for j :from 0
+                               :when (and (eql #\Z letter)
```
jmbr: Suggestion: replace `:when` + `:do (return j)` by `:until` + `:finally`.
```suggestion
                               :until (and (eql #\Z letter)
                                           (eql vote (position arg arguments :test #'equalp)))
                               :finally (return j))))
```

src/compilers/linear-paulis.lisp
```diff
@@ -0,0 +1,261 @@
...
+                                          (eql vote (position arg arguments :test #'equalp)))
+                                 :do (return j))))
+              (cond
```
jmbr: For this conditional statement an `if` form is more readable than a `cond`, IMHO.

src/compilers/linear-paulis.lisp
```diff
@@ -0,0 +1,261 @@
...
+        (dolist (term nonlocal-terms)
+          (loop :for letter :across (pauli-term-pauli-word term)
+                :for argument :in (pauli-term-arguments term)
+                :when (eql #\Z letter)
+                  :do (incf (aref votes (position argument arguments :test #'equalp)))))
+        (setf vote (array-argmax votes)
+              control-qubit (nth vote (application-arguments instr)))
+        ;; now we re-sort the nonlocal terms into two buckets: those with a Z in
+        ;; the voted-upon location, and those without
+        (let* ((Is nil) (Zs nil))
```
jmbr:
```suggestion
        (let ((Is nil)
              (Zs nil))
```
or even
```suggestion
        (let (Is Zs)
```
I prefer the former. I suggested the latter because I've seen you use `let` this way in lines 76-77.

src/compilers/linear-paulis.lisp
```diff
@@ -0,0 +1,261 @@
...
+                               :when (and (eql #\Z letter)
+                                          (eql vote (position arg arguments :test #'equalp)))
+                                 :do (return j))))
+              (cond
+                (Z-pos
+                 (push (list term Z-pos) Zs))
+                (t
+                 (push term Is)))))
+          ;; emit the Is
+          (unless (endp Is)
```
jmbr:
```suggestion
          (when Is
```

src/compilers/linear-paulis.lisp
```diff
@@ -0,0 +1,261 @@
...
+                                             :terms ZZ-terms
+                                             :name "ZZ-GATE"))
+                     (ZI-gate (make-instance 'exp-pauli-sum-gate
+                                             :arguments arguments
+                                             :parameters parameters
+                                             :arity arity
+                                             :dimension dimension
+                                             :terms (mapcar #'first ZIs)
+                                             :name "ZI-GATE")))
+                (unless (zerop (length ZZ-terms))
```
jmbr:
```suggestion
                (when ZZ-terms
```

src/compilers/linear-paulis.lisp
```diff
@@ -0,0 +1,261 @@
...
+                  (inst "CNOT" () control-qubit subvote-literal)
+                  (inst* ZZ-gate
+                         (application-parameters instr)
+                         (application-arguments instr))
+                  (inst "CNOT" () control-qubit subvote-literal))
+                (unless (zerop (length ZIs))
```
jmbr:
```suggestion
                (when ZIs
```

ecpeterson:
> The fact that compilation and, specifically, gate depth is non-deterministic surprises me.

This surprises me too! The only nondeterministic component I'm aware of is in the qubit routing code, where the heuristic function has some randomized components to help prevent deadlock. That could be the culprit—but if you were to run this example with a 4Q all-to-all chip and still saw the wiggle, then that would rule this out and I'd really be at a loss.

src/compilers/linear-paulis.lisp
```diff
@@ -0,0 +1,261 @@
...
+      (dolist (term terms)
+        (multiple-value-bind (Z-count Z-position)
+            (loop :for letter :across (pauli-term-pauli-word term)
+                  :for j :from 0
+                  :with pos := 0
+                  :with count := 0
+                  :when (eql #\Z letter)
+                    :do (setf count (1+ count)
+                              pos j)
+                  :finally (return (values count pos)))
```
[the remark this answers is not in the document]
ecpeterson: I don't see a potential for re-use, but I do see a potential to reduce clutter.

src/compilers/linear-paulis.lisp
ecpeterson: I'm in the habit of only using `if` for forms whose value I expect to use, but I see the appeal here.

src/compilers/linear-paulis.lisp
ecpeterson: This has different semantics: the original `loop` will return `nil` if no Z is ever encountered.

ecpeterson:
```
QUIL> (dotimes (i 10) (print (length (parsed-program-executable-code (quilc::process-program (parse-quil "
DEFGATE HAMILTONIAN(%theta) q0 q1 q2 q3 AS PAULI-SUM:
    ZZIZ(%theta) q0 q1 q2 q3

HAMILTONIAN(pi/4) 0 1 2 3") (build-nq-fully-connected-chip 4))))))

26 
26 
26 
31 
26 
31 
26 
26 
26 
31 
```
My _guess_ is that there's jitter in the numerical routines, and even perhaps that diagonalization isn't fully deterministic—but I do not know.

jmbr:
> This surprises me too! The only nondeterministic component I'm aware of is in the qubit routing code, where the heuristic function has some randomized components to help prevent deadlock. That could be the culprit—but if you were to run this example with a 4Q all-to-all chip and still saw the wiggle, then that would rule this out and I'd really be at a loss.

Thanks for offering your insight on this. I followed your suggestion and tried it with `build-nQ-fully-connected chip`. I still observe the same behavior:
```lisp
QUIL> (dotimes (i 10)
        (print (length (parsed-program-executable-code (quilc::process-program (parse-quil "
DEFGATE HAMILTONIAN(%theta) q0 q1 q2 q3 AS PAULI-SUM:
    ZZIZ(%theta) q0 q1 q2 q3

HAMILTONIAN(pi/4) 0 1 2 3") (quil::build-nq-fully-connected-chip 4))))))

26 
26 
26 
26 
31 
26 
31 
26 
26 
26 
nil
```

src/compilers/linear-paulis.lisp
jmbr: You're right!

src/compilers/linear-paulis.lisp
```diff
@@ -90,21 +94,19 @@
               control-qubit (nth vote (application-arguments instr)))
         ;; now we re-sort the nonlocal terms into two buckets: those with a Z in
         ;; the voted-upon location, and those without
-        (let* ((Is nil) (Zs nil))
+        (let* (Is Zs)
```
jmbr: `let*` -> `let` :-)

jmbr:
Something else that bothers me is that `exp-pauli` in `src/clifford/pauli.lisp` compiles to significantly shorter code (4 vs 10 CZ gates for the example above).

ecpeterson:
@jmbr: I'm not sure what's going on with these inflated CZ counts, but it's not in this area of the code:
```
QUIL> (labels ((do-diagonal-expansion (in &optional out)
                 (cond
                   ((endp in)
                    (reverse out))
                   ((and (typep (first in) 'gate-application)
                         (typep (gate-application-gate (first in)) 'exp-pauli-sum-gate))
                    (do-diagonal-expansion (append (parametric-diagonal-compiler (first in))
                                                   (rest in))
                      out))
                   (t
                    (do-diagonal-expansion (rest in) (list* (first in) out))))))
        (do-diagonal-expansion (coerce (parsed-program-executable-code (parse-quil "
DEFGATE HAMILTONIAN(%theta) q0 q1 q2 q3 AS PAULI-SUM:
    ZZIZ(%theta) q0 q1 q2 q3

HAMILTONIAN(pi/4) 0 1 2 3")) 'list)))
(#<CNOT 0 1> #<CNOT 1 3> #<RZ(pi/2) 3> #<CNOT 1 3> #<CNOT 0 1>)
```
I'll investigate tomorrow.

ecpeterson:
(This also seems to make the gate count jitter go away. 🤷‍♀)

jmbr:
> @jmbr: I'm not sure what's going on with these inflated CZ counts, but it's not in this area of the code:
> ~I'll investigate tomorrow.~ When I turn `*compiler-noise-stream*` on, I see that `recognize-UCR` gets called on the invocation of HAMILTONIAN, which results in a thrice FORKED RZ, which just compiles worse. After demoting `RECOGNIZE-UCR` further down the global compilers list, this problem goes away.

Thanks for the heads up. That's spooky indeed.

By the way, just the other day I saw a link to a directly relevant compiler debugging technique: http://blog.ezyang.com/2011/06/debugging-compilers-with-optimization-fuel/

ecpeterson:
This is a neat suggestion. We'd have to track down and eliminate the nondeterminism to make it work for us. That's not an unreasonable task, but one would have to think very carefully about the problem we solved in the addresser by injecting randomization.

stylewarning:
> When I turn `*compiler-noise-stream*` on, I see that `recognize-UCR` gets called on the invocation of HAMILTONIAN, which results in a thrice FORKED RZ, which just compiles worse. After demoting `RECOGNIZE-UCR` further down the global compilers list, this problem goes away.

Can we document this in the global list as an artifact for future visitors?

cl-quil.asd
```diff
@@ -80,8 +80,9 @@
                              (:file "state-prep")
                              (:file "translators")
                              (:file "modifiers")
+                             (:file "linear-paulis")
+                             ;; attic'd files / pedagogical purposes only
                              (:file "optimal-2q")
```
stylewarning: is this an attic'd file now too? Should it be `:static-file`d or are there tests we want to continue to run?

src/ast.lisp
```diff
@@ -386,6 +386,26 @@ If no exit rewiring is found, return NIL."
...
+          :documentation "List of PAULI-TERMs comprising the sum.")
+   (parameters :initarg :parameters
+               :reader exp-pauli-sum-gate-definition-parameters
+               :documentation "Ordered list of parameter names to be supplied to the definition, which can appear in arithmetical expressions weighting the definition's Pauli terms.")
+   (arguments :initarg :arguments
+              :reader exp-pauli-sum-gate-definition-arguments
+              :documentation "Ordered list of formal arguments appearing in the definition's Pauli terms."))
+  (:documentation "Represents a gate definition as the exponential of a weighted sum of Pauli matrices."))
+
+(defstruct (pauli-term)
```
stylewarning: please document this more thoroughly. Also I'm worried about duplication in the Clifford module.

src/build-gate.lisp
```diff
@@ -54,16 +54,32 @@ EXAMPLE: The Quil line \"CPHASE(pi) 2 3\" corresponds to the S-expression (build
 
 (define-global-counter **anonymous-gate-counter** get-anonymous-gate-counter)
 
-(defun anon-gate (operator matrix qubit &rest qubits)
+(defun anon-gate (operator-or-gate gate-or-parameters qubit &rest qubits)
```
stylewarning: @_@

src/build-gate.lisp
```diff
@@ -54,16 +54,32 @@ EXAMPLE: The Quil line \"CPHASE(pi) 2 3\" corresponds to the S-expression (build
...
   "Variant of BUILD-GATE for constructing anonymous gate applications."
```
stylewarning: can we document the basically two kinds of lambda lists here?

src/compilers/linear-paulis.lisp
```diff
@@ -0,0 +1,263 @@
...
+     (make-delayed-expression
+      (delayed-expression-params big-tree)
+      (delayed-expression-lambda-params big-tree)
+      (tree-substitute (delayed-expression-expression big-tree) substitution-table)))
+    ((assoc big-tree substitution-table)
+     (cdr (assoc big-tree substitution-table)))
+    (t
+     big-tree)))
+
+(defun array-argmax (arr)
```
stylewarning: technically `vector-argmax`

src/compilers/linear-paulis.lisp
```diff
@@ -0,0 +1,263 @@
...
+                    pos j)
+        :finally (return pos)))
+
+(defun term->count-and-last-Z-position (term)
+  "For TERM a PAULI-TERM, counts the number of Zs and returns the position of the last occuring Z."
+  (loop :for letter :across (pauli-term-pauli-word term)
+        :for j :from 0
+        :with pos := 0
+        :with count := 0
+        :when (eql #\Z letter)
```
stylewarning: is `letter` always a char? `char=`

stylewarning:
```
(values (count #\Z (pauli-term-pauli-word term))
        (position #\Z (pauli-term-pauli-word term) :from-end t))
```

if u don't mind a slight efficiency loss

src/compilers/linear-paulis.lisp
```diff
@@ -0,0 +1,263 @@
...
+                   :name (or new-name name)
+                   :terms (or new-terms terms))))
+
+(defun find-most-occurrant-Z (terms arguments &key except)
+  (let ((votes (make-array (length arguments) :initial-element 0)))
+    (dolist (term terms)
+      (loop :for letter :across (pauli-term-pauli-word term)
+            :for argument :in (pauli-term-arguments term)
+            :when (eql #\Z letter)
+              :do (incf (aref votes (position argument arguments :test #'equalp)))))
```
stylewarning: why not slap on a `:for i from 0`? The args appear to be going in order. Is there more than meets the eye here?

src/compilers/linear-paulis.lisp
```diff
@@ -0,0 +1,263 @@
...
+    (when except
+      (setf (aref votes except) 0))
+    (array-argmax votes)))
+
+(define-compiler parametric-diagonal-compiler
+    ((instr _
+            :where (and (typep (gate-application-gate instr) 'exp-pauli-sum-gate)
+                        (every (lambda (term) (every (lambda (letter) (or (eql letter #\Z) (eql letter #\I)))
```
stylewarning: please factor this out into its own function

src/gates.lisp
```diff
@@ -189,6 +190,28 @@
 (defmethod gate-matrix ((gate parameterized-gate) &rest parameters)
   (apply (parameterized-gate-matrix-function gate) parameters))
 
+(defclass exp-pauli-sum-gate (parameterized-gate)
```
stylewarning: This should implement `gate-matrix` so the QVM won't choke.

stylewarning: (QVM smoker's cough)

src/gates.lisp
```diff
@@ -189,6 +190,28 @@
...
+  ((parameters :initarg :parameters
+               :reader exp-pauli-sum-gate-parameters)
+   (arguments :initarg :arguments
+              :reader exp-pauli-sum-gate-arguments)
+   (terms :initarg :terms
```
stylewarning: maybe some description of the data type, or a pointer to where to find the description

src/matrix-operations.lisp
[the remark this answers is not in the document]
stylewarning: i like the occasional question mark

src/matrix-operations.lisp
```diff
@@ -391,3 +391,35 @@ as needed so that they are the same size."
        (matrix-equality
         kroned-ref-mat
         (scale-out-matrix-phases kroned-mat kroned-ref-mat))))))
+
+(defun matrix-expt (m s &key hermitian?)
+  "Computes EXP(M*S).  Only works for unitarily diagonalizable matrices M."
```
stylewarning: are we avoiding the bad boi that is magicl.transcendental

src/compilers/linear-paulis.lisp
ecpeterson: More than meets the eye: `(pauli-term-arguments term)` is a shuffled subset of `arguments`.

src/build-gate.lisp
ecpeterson: I agree, and I'd take suggestions as to how to name these two cases so that the functions can be split apart. The two use cases are:
```
(anon-gate "DUMMY-NAME" matrix qn ... q0)
(anon-gate parametric-gate-defn parameter-list qn ... q0)
```

src/gates.lisp
ecpeterson: I'm going to leave this up until we're thoroughly dis/convinced about the fresh `pauli-term` type.

src/gates.lisp
ecpeterson: It inherits `gate-matrix` from `parameterized-gate`.

src/ast.lisp
ecpeterson: I also don't like the duplication with the Clifford module, but there the phase factor is specialized to multiples of i and the encoding leans on that.

src/matrix-operations.lisp
ecpeterson: I suppose we are. We are calling out to MAGICL, which calls out to LAPACK, but for the diagonalization routines instead of the exponentiation routines.

Importing `MAGICL.TRANSCENDENTAL` requires the somewhat uncomfortable secondary import of expokit, which we've so far avoided actually using (and which this also wouldn't use).

src/compilers/linear-paulis.lisp
```diff
@@ -0,0 +1,263 @@
...
+          (unless (= 1 (crawl-parameter (pauli-term-prefactor term)))
+            (give-up-compilation))))
+      
+      ;; instantiate the Hamiltonian
+      (let ((H (magicl:make-zero-matrix dimension dimension)))
+        (dolist (term terms)
+          (setf H (m+ H (pauli-term->matrix term arguments (list 1d0) parameters))))
+        ;; orthogonally diagonalize it: H = O D O^T
+        (multiple-value-bind (diagonal O) (magicl:hermitian-eig H)
+          (let ((O (orthonormalize-matrix O)))
```
jmbr: The function `hermitian-eig` (LAPACK's `zheev` [1]) already returns the orthonormalized eigenvectors, so the call to `orthonormalize-matrix` is not necessary.

http://www.netlib.org/lapack/explore-html/df/d9a/group__complex16_h_eeigen_gaf23fb5b3ae38072ef4890ba43d5cfea2.html

src/compilers/linear-paulis.lisp
```diff
@@ -0,0 +1,263 @@
...
+  (with-slots (arguments parameters arity dimension name terms) gate
+    (make-instance 'exp-pauli-sum-gate
+                   :arguments arguments
+                   :parameters parameters
+                   :arity arity
+                   :dimension dimension
+                   :name (or new-name name)
+                   :terms (or new-terms terms))))
+
+(defun find-most-occurrant-Z (terms arguments &key except)
```
jmbr: `occurrant` -> `occurrent`? or `frequent` perhaps?

ecpeterson: I like `frequent`, and I appreciate the spelling correction!

src/build-gate.lisp
```diff
@@ -54,16 +54,32 @@ EXAMPLE: The Quil line \"CPHASE(pi) 2 3\" corresponds to the S-expression (build
...
   (push qubit qubits)
-  (let ((name (format nil "~A-~A" operator (get-anonymous-gate-counter))))
+  (let* ((name
+           (etypecase operator-or-gate
+             (string
+              (format nil "~A-~A" operator-or-gate (get-anonymous-gate-counter)))
+             (gate
+              (format nil "~A-~A" (gate-name operator-or-gate) (get-anonymous-gate-counter)))))
+         (gate
+           (cond
```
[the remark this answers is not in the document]
ecpeterson: I'd like to `etypecase`, but I'm `typep`ing two different arguments.

src/ast.lisp
stylewarning: can we at least write a note about the duplication here, why it's here, and that the intended purpose of this structure is _only_ for the construction of the gate?

src/build-gate.lisp
stylewarning: still asking for docs

src/gates.lisp
stylewarning: i can be convinced if it's just very very well documented (as stated earlier).

src/parser.lisp
```diff
@@ -870,21 +870,151 @@ If ENSURE-VALID is T, then a memory reference such as 'foo[0]' will result in an
...
+              :for arg :in (pauli-term-arguments term)
+              :for arg-position := (- arg-count 1 (position arg arguments :test #'equalp))
+              :for row-toggle := (ldb (byte 1 arg-position) col)
+              :do (ecase letter
+                    (#\X
+                     (setf row (dpb (- 1 row-toggle) (byte 1 arg-position) row))
+                     (setf entry (- entry)))
+                    (#\Y
+                     (setf row (dpb (- 1 row-toggle) (byte 1 arg-position) row))
+                     (setf entry (* entry (if (zerop row-toggle) (complex 0 1) (complex 0 -1)))))
```
```suggestion
                     (setf entry (* entry (if (zerop row-toggle) #C(0 1) #C(0 -1)))))
```

ecpeterson:
Added a test, which also fixed a bug introduced during PR review.

src/parser.lisp
```diff
@@ -870,21 +870,151 @@ If ENSURE-VALID is T, then a memory reference such as 'foo[0]' will result in an
...
+        (quil-parse-error "Pauli word ~a expected ~a qubit arguments, but got ~a: ~a"
+                          pauli-word (length pauli-word) (length qubit-list) qubit-list))
+      ;; the parameter body only refers to defined terms
+      ;; XXX
+      (make-pauli-term :pauli-word pauli-word
+                       :prefactor param
+                       :arguments qubit-list))))
+
+(defun pauli-term->matrix (term arguments parameters parameter-names)
+  (let* ((prefactor-fn
```
kilimanjaro: It's not clear why you compute `prefactor-fn` if all you do below is apply it to `parameters`.

jmbr:
> It's difficult to really verify the details on much of this (e.g. whether the compilation routines are correct, and so on), so one must rely on tests. In that regard, I think testing the output of `make-matrix-from-quil` before and after `compiler-hook` on exponentials of random Pauli terms is a nice idea, and gives some confidence. However, for legibility/hackability I would strongly prefer for a few additional tests which are more explicit and which exercise narrower code paths. In particular,
> 

I concur. The `exp-pauli` function (and its corresponding unit test) could provide a baseline for comparison.

stylewarning:
@ecpeterson please make sure any follow ons are filed as issues

PR #500, opened 2019-12-17
Use SAFELY-PARSE-QUIL on external input

notmgsk:
For the sake of security, `SAFELY-PARSE-QUIL` restricts the `INCLUDE` command to only accept paths relative to the value of `*safe-include-directory*`.

### DIFF
--- a/src/parser.lisp
+++ b/src/parser.lisp
@@ -1621,6 +1621,14 @@ When ALLOW-EXPRESSIONS is set, we allow for general arithmetic expressions in a 
 
 (defvar *safe-include-directory* nil)
 
+(define-condition resolve-filename-safely-condition (quil-parse-error)
+  ((filename :initarg :filename :reader resolve-safely-filename)
+   (message  :initarg :message  :reader resolve-safely-message))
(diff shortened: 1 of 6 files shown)


src/parser.lisp
```diff
@@ -1621,6 +1621,14 @@ When ALLOW-EXPRESSIONS is set, we allow for general arithmetic expressions in a
...
 
+(define-condition resolve-filename-safely-condition (quil-parse-error)
+  ((filename :initarg :filename :reader resolve-safely-filename)
+   (message  :initarg :message  :reader resolve-safely-message))
+  (:report (lambda (condition stream)
+             (format stream "Cannot safely resolve ~A: ~A."
+                     (resolve-safely-filename condition)
+                     (resolve-safely-message  condition)))))
+
 (defun resolve-safely (filename)
```
jmbr: In the documentation for quilc, I would **strongly** encourage users relying on this functionality to use chroot [1] in addition to it.

[1] http://man7.org/linux/man-pages/man1/chroot.1.html

PR #508, opened 2019-12-23
Remove stray (BREAK) when parsing U-gate application in qasm.lisp

appleby:
This caused the compiler to try to enter the debugger when parsing a direct `U`-gate application in program like so:

```
OPENQASM 2.0
qreg q[1]
U(0,0,0) q[0]
```

### DIFF
--- a/src/qasm.lisp
+++ b/src/qasm.lisp
@@ -616,8 +616,6 @@ Note: the above \"expansion\" is not performed when in a gate body."
 
           ((string= name "U")
            (check-number-of-parameters params 3)
-           (unless *gate-applications-are-formal*
-             (break))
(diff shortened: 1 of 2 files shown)


appleby:
Github isn't doing a great job of highlighting the diff in `qasm-tests.lisp`. I tried reordering things in the last commit, but it didn't help much. Basically, I added a testcase for a direct `U(0,0,0)` application, plus additional cases for the existing `u{1,2,3}` tests to verify they expand into something identical to the corresponding `U(theta, phi, lambda)`.

notmgsk:
Apologies 👍

appleby:
> Apologies 👍

It was a test of our code-review diligence. We failed u.

PR #510, opened 2019-12-26
Refactor %STRIP-HALTS-RESPECTING-REWIRINGS

appleby:
Take @ecpeterson's sage advice from #375 and refactor `%STRIP-HALTS-RESPECTING-REWIRINGS` to only deal with a single HALT instruction at the end of the program, rather than stripping HALTs anywhere in the program. For starters, as [ecp mentions in the comment thread](https://github.com/rigetti/quilc/pull/375#discussion_r311787805) of #375, this function is only called in protoquil mode, so there should only be one HALT at the end. Even if there were multiple HALTs, stripping away a mid-program HALT is not likely to leave you with an equivalent program, so it doesn't seem very useful to support that, even in theory.

As noted in the original thread, this refactoring actually makes the code slightly longer and doesn't really improve big-O complexity since we still need to copy the vector to strip off the final HALT. However it does simplify (subjective) the test case since it only needs to test trailing HALTs. The test code is now slightly longer, but exhaustively tests the "interesting" cases. The `TEST-COMPATIBLE` / `TEST-INCOMPATIBLE` tests are also more declarative and therefore (imo) easier to reason about what cases are actually being tested than the previous incarnation.

This refactoring is of questionable value other than as a christmas gift to myself because the old test made me want to throw up in my mouth a little.

### DIFF
--- a/app/src/entry-point.lisp
+++ b/app/src/entry-point.lisp
@@ -532,7 +532,7 @@ Returns a values tuple (PROCESSED-PROGRAM, STATISTICS), where PROCESSED-PROGRAM 
         (setf (gethash "final_rewiring" statistics)
               (quil::extract-final-exit-rewiring-vector processed-program))
         (setf (parsed-program-executable-code processed-program)
-              (%strip-halts-respecting-rewirings processed-program))
+              (strip-final-halt-respecting-rewirings processed-program))
 
         (let ((lschedule (quil::make-lscheduler)))
           (loop :for instr :across (parsed-program-executable-code processed-program)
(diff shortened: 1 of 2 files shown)


notmgsk:
`HALT` is weird.

I would've guessed that the reasonable place to see a `HALT` is mid-program (probably in some control-flow block), and not at the end of a program (since there a `HALT` is implied).

appleby:
Mid-program *is* the reasonable place to see a `HALT` if you're writing arbitrary Quil with control flow, but this function is only called when protoquil was requested, and stripping the final `HALT` is meant to satisfy the protoquil constraint. IIRC, the final `HALT` gets added in `RECONSTITUTE-BASIC-BLOCK` when reconstituting the program from the control-flow graph.  It's possible that could be changed to not append the `HALT` in protoquil mode, but I haven't looked closely to see what that would entail.

ecpeterson:
I’m opposed to stripping the final HALT in general. If you were to skip the program-final HALT in typical assembly programming, you’d get undefined behavior (as execution potentially proceeds into whatever garbage lives in memory not overwritten by program load).

stylewarning:
I agree that a final halt feels sanitary.

app/src/entry-point.lisp
```diff
@@ -597,37 +597,59 @@ Returns a values tuple (PROCESSED-PROGRAM, STATISTICS), where PROCESSED-PROGRAM
...
+  (let* ((instructions (parsed-program-executable-code processed-program))
+         (last-instruction (and (plusp (length instructions))
+                                (quil::nth-instr 0 processed-program :from-end t)))
+         (penultimate-instruction (and (< 1 (length instructions))
+                                       (quil::nth-instr 1 processed-program :from-end t)))
+         (must-transfer-comment-p (and (not (null penultimate-instruction))
+                                       (comment last-instruction))))
+
+    (unless (quil::haltp last-instruction)
+      (return-from strip-final-halt-respecting-rewirings instructions))
```
stylewarning: Should we exit early if PENULTIMATE-INSTRUCTION is NIL? Should we just return an empty program? We wouldn't need to deal with rewiring in that case.

app/src/entry-point.lisp
```diff
@@ -597,37 +597,59 @@ Returns a values tuple (PROCESSED-PROGRAM, STATISTICS), where PROCESSED-PROGRAM
...
+                   ;; PENULTIMATE-INSTRUCTION and LAST-INSTRUCTION have rewiring comments attached
+                   ;; which might be incompatible. We check to ensure that either one of the
+                   ;; rewirings is NULL, or else they are EQUALP and can safely be merged.
+                   (assert (or (or (null last-rewiring)
+                                   (null penultimate-rewiring))
+                               (equalp last-rewiring penultimate-rewiring))
+                           ()
+                           "Failed to strip final HALT. Instructions have incompatible ~A rewirings:~@
+                           LAST: ~A ~A~@
                            PREV: ~A ~A"
```
stylewarning: These are not good shrimps.

```
"Failed to ... rewirings:~%  ~
    LAST: ~A ~A~%  ~
    PREV: ~A ~A"
```

`~@` ensures a fresh line which isn't needed, and these gargantuan spaces will be included without the trailing `~`

app/src/entry-point.lisp
appleby: `MUST-TRANSFER-COMMENT-P` will be `NIL` if `PENULTIMATE-INSTRUCTION` is `NIL`, so we won't attempt to transfer the rewiring comment in that case, and will just skip down to the `subseq` at the end.

In some intermediate version, I wrote this as a big `cond` which did special-case a single `HALT` to return `#()`, but the `cond` version ended up repeating subclauses and being a little harder to follow (I thought) than this imperative version.

stylewarning: ok

app/src/entry-point.lisp
appleby: AFAIU `~@<newline>` is roughly equivalent to `~% ~<newline>`. It leaves the newline in place, but will ignore the following whitespace. `~:` on the other hand ignores the newline and leaves the whitespace in place.

stylewarning: whoa, i totally confused `~&` and `~@`. My bad!

PR #512, opened 2019-12-27
Rewire INSTR before EXPAND-TO-NATIVE-INSTRUCTIONS in COST-FUNCTION

### LINKED ISSUE #458: cnot-to-native-cnots does not guard against a path length between q1 and q0 being zero
And it may thus enter an infiniloop.

### LINKED ISSUE #511: COST-FUNCTION on non-native instruction can cause the compiler to loop
On a chip with live qubits 3 and 4, witness `CZ 0 1`.  This bug was introduced by the fidelity-addresser merge.

notmgsk:
`cost-function` expands its input instruction into a list of native instructions. If the instruction is not first rewired, then `expand-to-native-instructions` will get confused and infiniloop. For example, if a chip has live physical qubits 3 and 4, then the instruction `CZ 0 1` will be expanded into `CAN ... 0 1` will be expanded into `CZ 0 1` will be expanded into `CAN ... 0 1` ... etc.

My first thought was that `expand-to-native-instructions` should be aware of the current rewiring by providing it a `addresser-state` object containing the rewiring, or even just the rewiring itself. Until now, however, that hasn't been required and so maybe isn't the best solution.

Second (and current) solution, is to temporarily apply the current rewiring to the instruction in `cost-function` (being careful not to mutate the instruction beyond the invocation of `expand-to-...`).

Closes #511 and closes #458.

### DIFF
--- a/src/addresser/rewiring.lisp
+++ b/src/addresser/rewiring.lisp
@@ -309,3 +309,13 @@ BODY as an implicit PROGN."
               first-rewiring-string second-rewiring-string)
       (values (make-rewiring-from-string first-rewiring-string)
               (make-rewiring-from-string second-rewiring-string)))))
+
+(defun rewiring-available-p (rewiring lq)
+  "Test whether REWIRING contains a non-nil rewiring for logical qubit LQ."
+  (aref (rewiring-l2p rewiring) lq))
+
+(defun rewiring-available-for-instruction-p (rewiring instr)
+  "Test whether every logical qubit in INSTR has been rewired in REWIRING."
+  (every (a:curry #'rewiring-available-p rewiring)
+         (mapcar #'qubit-index
+                 (application-arguments instr))))
--- a/src/addresser/rewiring.lisp
+++ b/src/addresser/rewiring.lisp
@@ -310,12 +310,12 @@ BODY as an implicit PROGN."
       (values (make-rewiring-from-string first-rewiring-string)
               (make-rewiring-from-string second-rewiring-string)))))
 
-(defun rewiring-available-p (rewiring lq)
+(defun rewiring-assigned-for-qubit-p (rewiring lq)
   "Test whether REWIRING contains a non-nil rewiring for logical qubit LQ."
   (aref (rewiring-l2p rewiring) lq))
 
-(defun rewiring-available-for-instruction-p (rewiring instr)
+(defun rewiring-assigned-for-instruction-qubits-p (rewiring instr)
   "Test whether every logical qubit in INSTR has been rewired in REWIRING."
-  (every (a:curry #'rewiring-available-p rewiring)
-         (mapcar #'qubit-index
-                 (application-arguments instr))))
+  (every (a:compose #'qubit-index
+                    (a:curry #'rewiring-assigned-for-qubit-p rewiring))
+         (application-arguments instr)))
--- a/src/addresser/rewiring.lisp
+++ b/src/addresser/rewiring.lisp
@@ -316,6 +316,6 @@ BODY as an implicit PROGN."
 
 (defun rewiring-assigned-for-instruction-qubits-p (rewiring instr)
   "Test whether every logical qubit in INSTR has been rewired in REWIRING."
-  (every (a:compose #'qubit-index
-                    (a:curry #'rewiring-assigned-for-qubit-p rewiring))
+  (every (a:compose (a:curry #'rewiring-assigned-for-qubit-p rewiring)
+                    #'qubit-index)
          (application-arguments instr)))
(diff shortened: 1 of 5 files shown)


notmgsk:
Latest commit tries to apply a rewiring if one exists.

notmgsk:
I've tried to convince myself that I'm not breaking anything, and having learned a good deal about the addresser along the way: I'm not 100% but relatively confident.

tests/compilation-tests.lisp
```diff
@@ -167,3 +167,7 @@ CNOT 0 2"))
     ;; NOTE: Decomposing into fewer 2q gates is more of a regression
     ;; test on quality of compilation, and not on correctness.
     (is (>= 6 (length 2q-code)))))
+
+(deftest test-sohaib-fidelity-rewiring-regression ()
+  (is (compiler-hook (parse "CZ 0 1")
```
stylewarning: do you want `is` or `not-signals`?

src/addresser/rewiring.lisp
```diff
@@ -309,3 +309,13 @@ BODY as an implicit PROGN."
               first-rewiring-string second-rewiring-string)
       (values (make-rewiring-from-string first-rewiring-string)
               (make-rewiring-from-string second-rewiring-string)))))
+
+(defun rewiring-available-p (rewiring lq)
```
stylewarning: I prefer `rewiring-assigned-for-qubit-p`

src/addresser/rewiring.lisp
```diff
@@ -309,3 +309,13 @@ BODY as an implicit PROGN."
...
+  "Test whether REWIRING contains a non-nil rewiring for logical qubit LQ."
+  (aref (rewiring-l2p rewiring) lq))
+
+(defun rewiring-available-for-instruction-p (rewiring instr)
```
stylewarning: similarly, `rewiring-assigned-for-instruction-qubits-p`

src/addresser/rewiring.lisp
```diff
@@ -309,3 +309,13 @@ BODY as an implicit PROGN."
...
+  "Test whether every logical qubit in INSTR has been rewired in REWIRING."
+  (every (a:curry #'rewiring-available-p rewiring)
```
stylewarning: why not go whole hog and

```
(every (a:compose (a:curry ...) #'qubit-index)
       (application-arguments instr))
```

tests/compilation-tests.lisp
notmgsk: I'm not sure actually. The failure mode is either an infinite loop which blows the stack or memory exhaustion.

stylewarning: i would add a timeout with bordeaux-threads and a `not-signals` then

src/addresser/rewiring.lisp
notmgsk: now you're cooking with functions

src/addresser/rewiring.lisp
```diff
@@ -309,3 +309,13 @@ BODY as an implicit PROGN."
...
       (values (make-rewiring-from-string first-rewiring-string)
               (make-rewiring-from-string second-rewiring-string)))))
+
+(defun rewiring-assigned-for-qubit-p (rewiring lq)
+  "Test whether REWIRING contains a non-nil rewiring for logical qubit LQ."
+  (aref (rewiring-l2p rewiring) lq))
+
+(defun rewiring-assigned-for-instruction-qubits-p (rewiring instr)
+  "Test whether every logical qubit in INSTR has been rewired in REWIRING."
+  (every (a:compose #'qubit-index
```
stylewarning: is this what you want?

```
CL-USER> (funcall (alexandria:compose #'1+ #'sqrt) 3)
2.732051
CL-USER> (funcall (alexandria:compose #'sqrt #'1+) 3)
2.0
```

tests/compilation-tests.lisp
stylewarning: still this

notmgsk: Give me chance!

tests/compilation-tests.lisp
```diff
@@ -169,5 +169,7 @@ CNOT 0 2"))
     (is (>= 6 (length 2q-code)))))
 
 (deftest test-sohaib-fidelity-rewiring-regression ()
-  (is (compiler-hook (parse "CZ 0 1")
-                     (quil::build-chip-from-digraph '((2 3) (3 2))))))
+  (not-signals bt:timeout
+    (bt:with-timeout (1)
+      (is (compiler-hook (parse "CZ 0 1")
```
stylewarning: nix `is`

Issue #368, opened 2019-08-01
`test-translators` tests only half of the compilers


notmgsk:

ecpeterson:
My proposal here is to add a slot to the `compiler` class ( https://github.com/rigetti/quilc/blob/master/src/define-compiler.lisp#L367 ) which stores a generator for random instruction sequences to which the compiler applies. If this is empty, we can fall back on the logic already in `translator-tests.lisp`, but for cases where it's difficult to deduce what instructions will pass a guarded compiler, we can rely on the compiler author to manufacture them for us.

You could extend `define-compiler` to consume a global option which records this generator, but I think it'd look cleaner in the source to just `setf` the slot outside of the `define-compiler` form. A downside to this "cleanliness" is that re-defining the compiler during development runs the risk of destroying the contents of this slot. I don't know how I'd prefer to handle that.

PR #518, opened 2019-12-31
add manually-specified fuzzers

ecpeterson:
This PR adds support for manually specifying compiler fuzzers for use during `test-translators`. The fuzzers this includes brings the automatic testing from ~50% to ~70% of all compilers.

Two things left undone, which are each beyond the scope of this PR:

* I haven't written fuzzers for the state preparation routines. Some of these depend on knowing a wavefunction context, which isn't readily available; and some of these depend on the tester knowing to use a particular sort of matrix equality check, which it's currently ignorant of.
* I haven't written fuzzers for the guarded compilers whose only guard is "someone's qubits need to subsume someone else's qubits". This is a common enough pattern that we should augment the automated fuzzer to handle this case.

A separate concerning thing:

* I had to tweak some of the routines that process memory references / delayed expressions. I suspected that the `AS PAULI-SUM` PR had some AST abuses that escaped notice, and I think those are re-appearing now. It's probably better to fix them—in a separate PR—rather than work around them here.

Partially addresses #368.

### DIFF
--- a/src/build-gate.lisp
+++ b/src/build-gate.lisp
@@ -109,6 +109,12 @@ Comes in two flavors:
                arg-lambda-params
                arg-expression
                t))
+             ((memory-ref)
+              (values
+               nil
+               nil
+               arg
+               t))
              ((constant (value val))
               (values nil nil val nil))
              (_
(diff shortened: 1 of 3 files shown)


stylewarning:
file issues!

PR #525, opened 2020-01-02
Separately weight non/native gates

### LINKED ISSUE #373: Separately weigh non-native gates from infidelious native gates
`compute-applicable-compilers` et al. use a fidelity estimation heuristic to decide which collection of compilers is best for converting instructions into the target gateset. Currently, non-native gates are penalized by treating them as having extremely poor fidelity (e.g., 0.5), which causes the compiler crawler to avoid such paths when possible. This has the potential to cause conflict when there are actually "native gates" with miserable fidelity statistics.

I see three things to fix here:

1. Separately track the number of nonnative gates from native gates when producing the cost of a compilation chain. There's then no reason to fake any bad fidelities for non-native gates.
2. In the event that the compiler crawler can't find a path and throws `'no-compiler-path`, which is treated as a fatal error. (It also has the potential to infinite loop, or to expand so many search nodes that the process runs out of memory. These ought to be caught early and `'no-compiler-path` thrown instead.) Rather than crashing quilc, this condition could be caught, the hardware object marked as 'dead', and control passed back.
3. Figure out what to do with `+totally-awful-fidelity+`. The solutions to 1 and 2 might obviate us from defining such a constant; or it might make it safe to install a low value there.

ecpeterson:
quilc used to discourage itself from pursuing "missing" or "dead" gates by pretending they had sufficiently low fidelities and then letting the compiler crawler do its usual work. This occasionally misbehaved, in particular when Forest advertised as a gate with low fidelity as being actually present.

This PR updates the compiler crawler to separately track low-fidelity and absent gates. In a second commit, we additionally extend the compiler crawler with the ability to mark a hardware object as dead if it fails to find a path to the native gate set, rather than throwing an uncaught error.

Closes #373. I don't have an opinion any longer as to what to do about `+totally-awful-fidelity+`.


src/addresser/fidelity-addresser.lisp
```diff
@@ -15,6 +15,20 @@
 ;;;
 ;;; See DO-FIDELITY-ADDRESSING below for the main entry point.
 
+;;; lil utility
```
stylewarning: utilitito

src/addresser/fidelity-addresser.lisp
```diff
@@ -178,7 +192,9 @@
   (let ((distance-mapping (make-hash-table)))
     (loop :for object :across (chip-spec-links chip-spec)
           :do (setf (gethash object distance-mapping)
-                    (- (log (swap-fidelity chip-spec object)))))
+                    (if (gethash "dead" (hardware-object-misc-data object))
```
stylewarning: can we factor out this gethash with a function? `(hardware-object-dead-p obj)` or something that might be more aptly named?

src/chip/chip-specification.lisp
```diff
@@ -226,6 +226,10 @@ used to specify CHIP-SPEC."
         :unless (find qi dead-qubits)
           :collect qi))
 
+(defun chip-spec-link-dead? (chip-spec link-index)
```
stylewarning: this could be implemented in terms of the fabled `chip-spec-object-dead-p` function

src/define-compiler.lisp
```diff
@@ -518,16 +518,39 @@ Optionally constrains the output to include only those bindings of a particular
...
-          (when (binding-subsumes-p cost-key key)
-            (setf true-cost (gate-record-fidelity cost-val))))
-        (setf ret (* ret (expt true-cost val)))))))
+      (tagbody
+         (dohash ((cost-key cost-val) gateset)
+           (when (binding-subsumes-p cost-key key)
+             (setf (occurrence-table-cost-fidelity ret)
+                   (* (expt (gate-record-fidelity cost-val) val)
+                      (occurrence-table-cost-fidelity ret)))
+             (go :skip)))
```
stylewarning: capitalize tags plz

src/addresser/fidelity-addresser.lisp
ecpeterson: good advice

src/addresser/fidelity-addresser.lisp
```diff
@@ -15,6 +15,20 @@
 ;;;
 ;;; See DO-FIDELITY-ADDRESSING below for the main entry point.
 
+;;; utilitito
```
appleby: ❤️

src/chip/chip-specification.lisp
```diff
@@ -209,9 +209,11 @@ used to specify CHIP-SPEC."
     :for qubit-index :across (chip-spec-qubits-on-link chip-spec link-index)
     :append (remove link-index (coerce (chip-spec-links-on-qubit chip-spec qubit-index) 'list))))
 
+(defun hardware-object-dead-p (hw-obj)
```
appleby: I have nothing against `-p`, but maybe `hardware-object-dead?` to match naming convention of other predicates in this file (assuming there is one, I'm just going by what's visible in this diff)?

tests/qpu-test-files/Aspen-4-4Q-A-with-missing-gate.qpu
```diff
@@ -0,0 +1,74 @@
+{
+  "metadata": {
+    "name": "Aspen-4-4Q-A",
```
appleby:
```suggestion
    "name": "Aspen-4-4Q-A-with-missing-gate",
```

PR #529, opened 2020-01-02
Allow immediate integers for all STORE values

notmgsk:
The spec says that all STOREs should permit an immediate integer type
as the value operand. Prior to this, the following (legal) Quil would
fail:

    DECLARE ro REAL
    DECLARE index INTEGER
    STORE ro index 1

where instead one would have to write

    DECLARE ro REAL
    DECLARE index INTEGER
    DECLARE val REAL
    MOVE val 1
    STORE ro index val


ecpeterson:
@notmgsk Looks like you didn't catch all instances of `TYPECHECK-CONDITIONAL-INSTRUCTION`. Aside from that, this looks fine.

stylewarning:
great PR

PR #533, opened 2020-01-03
Ensure valid memory refs in classical instructions

notmgsk:
During parsing some memory references are not validated. I'm inclined to think that the function `parse-memory-or-formal-token` should have a default `t` for  the kwarg `ensure-valid`. Unless there was a reason it is `nil` by default?

### DIFF
--- a/src/parser.lisp
+++ b/src/parser.lisp
@@ -538,7 +538,7 @@ In accordance with the typical usage here, there are two values returned: the re
   (case (token-type qubit-tok)
     ((:NAME)
      (unless *formal-arguments-allowed*
-       (quil-parse-error "Unexpected formal argument ~A~@[ in ~A~]."
+       (quil-parse-error "Unexpected formal argument \"~A\"~@[ in ~A~]."
                          (token-payload qubit-tok)
                          *parse-context*))
      (formal (token-payload qubit-tok)))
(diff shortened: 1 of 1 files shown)


src/parser.lisp
```diff
@@ -549,7 +549,7 @@ In accordance with the typical usage here, there are two values returned: the re
                                     "a name or formal argument"
                                     "a name")))))
 
-(defun parse-memory-or-formal-token (tok &key ensure-valid)
+(defun parse-memory-or-formal-token (tok &key (ensure-valid t))
   "Parse the token TOK as a memory reference or formal argument.
 
 If ENSURE-VALID is T, then a memory reference such as 'foo[0]' will result in an error unless 'foo' has been DECLAREd."
```
```suggestion
If ENSURE-VALID is T (default), then a memory reference such as 'foo[0]' will result in an error unless 'foo' has been DECLAREd."
```

src/parser.lisp
```diff
@@ -549,7 +549,7 @@ In accordance with the typical usage here, there are two values returned: the re
                                     "a name or formal argument"
                                     "a name")))))
 
-(defun parse-memory-or-formal-token (tok &key ensure-valid)
+(defun parse-memory-or-formal-token (tok &key (ensure-valid t))
```
stylewarning: Github doesn't let me comment below, but can we give a better error message?

```
"Bad memory region name ~A~@[ in ~A~]"

==>

"Bad memory region name ~A~@[ in ~A~]. This is probably due to either:
    * a missing DECLARE for this memory,
    * a misspelling of the memory reference, or
    * a misspelling of the DECLAREd memory."

notmgsk: Might we also print out the known memory names?

stylewarning: Yes, you could, but I won't block this PR.

PR #547, opened 2020-01-14
More effciently calculate preceding schedule fidelity

notmgsk:


notmgsk:
I've compared the results of `chip-schedule-subschedule-after-resource` to those of `chip-schedule-from-carving-point` on a non-trivial program (compiled `sohaib.quil`) for different resources and have found them to agree. I need to write a more appropriate test case. Otherwise I feel good(ish) about the fix.

notmgsk:
One thing: not sure whether I should be looking at instruction end times rather than start times.

src/addresser/outgoing-schedule.lisp
```diff
@@ -119,6 +120,57 @@ If POINT is generated by CHIP-SCHEDULE-RESOURCE-CARVING-POINT, then the resultin
...
+                 (cond ((carving-point-p instr)
+                        (when (< latest-time (chip-schedule-start-time schedule instr))
+                          (setf latest-time (chip-schedule-start-time schedule instr)
+                                carving-point instr)))
+                       ((and (resources-intersect-p resource (instruction-resources instr))
+                             (<= latest-time (chip-schedule-start-time schedule instr)))
+                        (push instr sched)
+                        (dolist (instr (gethash instr earlier-instrs))
+                          (bottoms-up instr)))))))
+      (dolist (instr last-instrs)
```
stylewarning: `(map nil #'bottoms-up last-instrs)`

src/addresser/outgoing-schedule.lisp
```diff
@@ -119,6 +120,57 @@ If POINT is generated by CHIP-SCHEDULE-RESOURCE-CARVING-POINT, then the resultin
...
+               (unless (gethash instr seen)
+                 (setf (gethash instr seen) t)
+                 (cond ((carving-point-p instr)
+                        (when (< latest-time (chip-schedule-start-time schedule instr))
+                          (setf latest-time (chip-schedule-start-time schedule instr)
+                                carving-point instr)))
+                       ((and (resources-intersect-p resource (instruction-resources instr))
+                             (<= latest-time (chip-schedule-start-time schedule instr)))
+                        (push instr sched)
+                        (dolist (instr (gethash instr earlier-instrs))
```
stylewarning: `(map nil #'bottoms-up (gethash instr earlier-instrs))`

src/addresser/outgoing-schedule.lisp
```diff
@@ -119,6 +120,57 @@ If POINT is generated by CHIP-SCHEDULE-RESOURCE-CARVING-POINT, then the resultin
...
+(defun chip-schedule-subschedule-after-resource (schedule resource)
+  "Returns an unordered subsequence of instructions from SCHEDULE which intersect RESOURCE and which start on or after the latest time at which RESOURCE is used."
+  (let* ((lsched (chip-schedule-data schedule))
+         (last-instrs (lscheduler-last-instrs lsched))
+         (earlier-instrs (lscheduler-earlier-instrs lsched))
+         (latest-time 0)
+         (carving-point nil)
+         (seen (make-hash-table))
+         (sched '()))
+    (labels ((carving-point-p (instr)
```
stylewarning: lift this to a toplevel function

src/addresser/outgoing-schedule.lisp
```diff
@@ -119,6 +120,57 @@ If POINT is generated by CHIP-SCHEDULE-RESOURCE-CARVING-POINT, then the resultin
...
+                        (dolist (instr (gethash instr earlier-instrs))
+                          (bottoms-up instr)))))))
+      (dolist (instr last-instrs)
+        (bottoms-up instr))
+      ;; It may happen that before a better (later) carving point is
+      ;; found, instructions have already been collected which would
+      ;; otherwise be ignored (i.e. they would have a start time
+      ;; earlier than the better carving point). Thus it is necessary
+      ;; to do a final scan of the collected instructions and remove
+      ;; those that begin before the carving point. It's maybe
```
stylewarning: I don't _think_ it's possible because `latest-time` is continually being updated.

src/addresser/fidelity-addresser.lisp
```diff
@@ -63,16 +63,15 @@
...
-                                  (coerce (vnth 0 (hardware-object-cxns hardware-object)) 'list)))
-                 (carving-point (chip-schedule-resource-carving-point
-                                 (addresser-state-chip-schedule state)
-                                 resource))
-                 (subschedule (chip-schedule-from-carving-point
-                               (addresser-state-chip-schedule state) resource carving-point))
-                 (preceding-fidelity
-                   (calculate-instructions-log-fidelity subschedule
-                                                        (addresser-state-chip-specification state))))
+          (a:when-let*
```
stylewarning: can we lift out the non-`nil` bindings into a `let`?

src/addresser/outgoing-schedule.lisp
```diff
@@ -119,6 +120,57 @@ If POINT is generated by CHIP-SCHEDULE-RESOURCE-CARVING-POINT, then the resultin
...
+               ;; time is no earlier than the time of the latest known
+               ;; carving point. The carving point is updated as and
+               ;; when later ones are found.
+               (unless (gethash instr seen)
+                 (setf (gethash instr seen) t)
+                 (cond ((carving-point-p instr)
+                        (when (< latest-time (chip-schedule-start-time schedule instr))
+                          (setf latest-time (chip-schedule-start-time schedule instr)
+                                carving-point instr)))
+                       ((and (resources-intersect-p resource (instruction-resources instr))
```
jmbr: I'd suggest refactoring the test form here into a predicate like `carving-point-p` (e.g., `feasible-and-not-late-p` in this case).

src/addresser/fidelity-addresser.lisp
notmgsk: I guess, though the only non-nil thing here is `resource`. I think I will just not create a name for the resource since it is only used once, and collapse the two `when-let`s into one. WDYT?

notmgsk:
> Looks OK to me and I want those sweet sweet speedups!
> 
> Where exactly does the speedup come from? Mostly from removing the `sort`? Or also because we only examine earlier/last instructions instead of the full schedule (or is the union of those a superset of the full sched)? Or maybe because you're caching previously seen instrs? (D) All of the above.

Both `chip-schedule-resource-carving-point` and `chip-schedule-from-carving-point` where exploring the *entire* schedule (list of "approved" instructions) on every invocation, and over time that schedule grows quite large. It's only necessary to look so far back in time in the schedule.

src/addresser/outgoing-schedule.lisp
```diff
@@ -119,6 +120,57 @@ If POINT is generated by CHIP-SCHEDULE-RESOURCE-CARVING-POINT, then the resultin
...
+               ;; resources are a subset of RESOURCE *and* their start
+               ;; time is no earlier than the time of the latest known
+               ;; carving point. The carving point is updated as and
+               ;; when later ones are found.
+               (unless (gethash instr seen)
+                 (setf (gethash instr seen) t)
+                 (cond ((carving-point-p instr)
+                        (when (< latest-time (chip-schedule-start-time schedule instr))
+                          (setf latest-time (chip-schedule-start-time schedule instr)
+                                carving-point instr)))
```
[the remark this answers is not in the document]
notmgsk: Good catch. That was just for debugging.

src/addresser/outgoing-schedule.lisp
```diff
@@ -119,6 +120,57 @@ If POINT is generated by CHIP-SCHEDULE-RESOURCE-CARVING-POINT, then the resultin
...
+      (dolist (instr last-instrs)
+        (bottoms-up instr))
+      ;; It may happen that before a better (later) carving point is
+      ;; found, instructions have already been collected which would
+      ;; otherwise be ignored (i.e. they would have a start time
+      ;; earlier than the better carving point). Thus it is necessary
+      ;; to do a final scan of the collected instructions and remove
+      ;; those that begin before the carving point. It's maybe
+      ;; possible to do this during the meaty logic above.
+      (remove-if-not (lambda (instr) (< latest-time (chip-schedule-start-time schedule instr)))
```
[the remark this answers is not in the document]
notmgsk: Agreed.

src/addresser/outgoing-schedule.lisp
```diff
@@ -119,6 +120,57 @@ If POINT is generated by CHIP-SCHEDULE-RESOURCE-CARVING-POINT, then the resultin
...
+    (labels ((carving-point-p (instr)
+               ;; A "carving point" occurs when INSTRs resources (the
+               ;; qubits and memory used) overlap (though not exactly)
+               ;; with those of RESOURCE. This means that in the
+               ;; instruction sequence
+               ;;
+               ;;   H 0; CZ 0 1; H 0; CZ 0 1; CZ 0 2;
+               ;;
+               ;; the carving point of the resource of qubits 0 and 1
+               ;; would be the second CZ 0 1.
```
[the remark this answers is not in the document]
notmgsk: Both are carving points, but by the nature of the function (and as specified in the docstr) we only want the latest carving point in the schedule.

src/addresser/outgoing-schedule.lisp
notmgsk: True that

src/addresser/outgoing-schedule.lisp
```diff
@@ -119,6 +120,57 @@ If POINT is generated by CHIP-SCHEDULE-RESOURCE-CARVING-POINT, then the resultin
...
+               (and (resources-intersect-p (instruction-resources instr) resource)
+                    (not (resource-subsetp (instruction-resources instr) resource))))
+             (bottoms-up (instr)
+               ;; Recursively explorer earlier instructions in the
+               ;; schedule. Instructions are collected if their
```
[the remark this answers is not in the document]
ecpeterson: If you can talk me through what happens in this example here, I'm happy to approve / make concrete requests for change.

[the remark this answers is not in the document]
notmgsk: Both `H 1`s are part of the calculated subschedule in that case.

`CNOT 0 2` is found to be the latest carving point and so both `H 1`s are added to the subschedule.

src/addresser/outgoing-schedule.lisp
```diff
@@ -119,6 +120,57 @@ If POINT is generated by CHIP-SCHEDULE-RESOURCE-CARVING-POINT, then the resultin
    #'<
    :key (lambda (i) (gethash i (chip-schedule-times schedule)))))
 
+(defun chip-schedule-subschedule-after-resource (schedule resource)
```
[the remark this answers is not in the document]
notmgsk: go on

src/addresser/outgoing-schedule.lisp
```diff
@@ -97,12 +97,16 @@ BEFORE-INST to make use of RESOURCE."
     (chip-schedule-end-time schedule inst)
     0))
 
+(defun carving-point-p (resource instr)
+  "Returns T if the resources used by INSTR intersect with RESOURCE and are not a subset of RESOURCE, and NIL otherwise."
+  (and (resources-intersect-p (instruction-resources instr) resource)
+       (not (resource-subsetp (instruction-resources instr) resource))))
+
 (defun chip-schedule-resource-carving-point (schedule resource)
```
stylewarning: Is it OK that this function can return NIL? Should we return a number instead of NIL?

src/addresser/outgoing-schedule.lisp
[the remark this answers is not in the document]
appleby: Wouldn't the first `H 1`'s starting time be before the `CNOT 0 2` and therefore get excluded? (I might be misunderstanding what instr start time means in this context).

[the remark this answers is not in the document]
notmgsk: @appleby `H 1` and `CNOT 0 2` do not share resources, and so can be scheduled in parallel.

src/addresser/outgoing-schedule.lisp
notmgsk: I *think* it would be OK for `0` to replace `NIL`. For the moment I want to err on the side of not changing stuff that works.

stylewarning: @ecpeterson Thoughts?

appleby:
> One thing: not sure whether I should be looking at instruction end times rather than start times.

Making sure @ecpeterson sees this in case it got lost in the page scroll.

ecpeterson:
Mark and I talked in person, and I disconvinced him that we should be looking at times at all.

src/addresser/outgoing-schedule.lisp
ecpeterson: With these changes, is `chip-schedule-resource-carving-point` called anywhere?

src/addresser/outgoing-schedule.lisp
```diff
@@ -119,6 +123,26 @@ If POINT is generated by CHIP-SCHEDULE-RESOURCE-CARVING-POINT, then the resultin
...
+               ;; Recursively explorer earlier instructions in the
+               ;; schedule. Instructions are collected if their
+               ;; resources are a subset of RESOURCE.
+               (unless (gethash instr seen)
+                 (setf (gethash instr seen) t)
+                 (when (and (resources-intersect-p resource (instruction-resources instr))
+                            (not (carving-point-p resource instr)))
+                   (push instr sched)
+                   (map nil #'bottoms-up (gethash instr earlier-instrs))))))
+      (map nil #'bottoms-up last-instrs)
```
ecpeterson: The semantics of this looks good to me. One thing to be aware of: this (depth-first) crawl of the graph consumes stack space proportional to the circuit depth of the extracted subschedule. If someone submits a program with order 1000 instructions which are all on one or two qubits, this might blow the stack. Such circuits aren't completely unheard of: randomized benchmarking sequences can get pretty long (though I don't know if 'pretty long' is in the 100s or the 1000s).

You could deal with this by writing a non-recursive search with a manually-managed stack. You could also leave it alone until it actually breaks, perhaps with a note about what might eventually go wrong.

stylewarning:
> I prefer `nil` but to each their own and I just noticed that we use `'()` in a ton of places.

I prefer `'()` when it's a list you're g

src/addresser/outgoing-schedule.lisp
notmgsk: It is used in the temporal addresser. It's cousin `chip-schedule-from-carving-point` however is not used, but I was keeping it around for posterity's sake / to aid testing.

PR #548, opened 2020-01-14
KILL-THREAD-SLOWLY in RPCQ tests

appleby:
This is the corresponding change mirroring https://github.com/rigetti/rpcq/pull/113 to unblock EC2 gitlab runners for quilc.

### DIFF
--- a/app/tests/rpcq-tests.lisp
+++ b/app/tests/rpcq-tests.lisp
@@ -4,6 +4,21 @@
 
 (in-package #:quilc-tests)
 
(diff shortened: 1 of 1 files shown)


appleby:
According to @jmackeyrigetti the gitlab ci tags change is not need, so I've reverted it in the last commit. You can see the quilc tests had already passed on the EC2 runner here:

https://gitlab.com/rigetti/forest/quilc/-/jobs/403106728

appleby:
timeout while executing cl-quil-tests, but quilc-tests passed ok

PR #549, opened 2020-01-15
Be more tolerant of IMPOSSIBLY-FULL-RESOURCEs

### LINKED ISSUE #346: bench-0017q-0000061-*depth-vs-swaps-assn*-0020q-linear: Operation would result in a resource collection with an unbounded list of MEMORY-REGIONS
I’m seeing the following error when running one of the cl-quil benchmarks.

Repro steps:

```
(ql:quickload :cl-quil-benchmarking)
(cl-quil-benchmarking::bench-0017q-0000061-*depth-vs-swaps-assn*-0020q-skew-rect)
```

Backtrace:

```
Operation would result in a resource collection with an unbounded list of MEMORY-REGIONS.
   [Condition of type CL-QUIL.RESOURCE::RESOURCE-COLLECTION-UNBOUNDED-NAMES]

Restarts:
 0: [RETRY] Retry SLIME REPL evaluation request.
 1: [*ABORT] Return to SLIME's top level.
 2: [ABORT] abort thread (#<THREAD "repl-thread" RUNNING {1001A01BC3}>)

Backtrace:
  0: (CL-QUIL.RESOURCE:RESOURCE-DIFFERENCE #<CL-QUIL.RESOURCE::RESOURCE-COLLECTION IMPOSSIBLY FULL RESOURCE> #<CL-QUIL.RESOURCE::RESOURCE-COLLECTION :QUBITS {3, 9} ..)
  1: (CL-QUIL.RESOURCE:RESOURCE-SUBSETP #<CL-QUIL.RESOURCE::RESOURCE-COLLECTION IMPOSSIBLY FULL RESOURCE> #<CL-QUIL.RESOURCE::RESOURCE-COLLECTION :QUBITS {3, 9} ..)
  2: (CL-QUIL::CHIP-SCHEDULE-RESOURCE-CARVING-POINT #<CL-QUIL::CHIP-SCHEDULE {100A1556F3}> #<CL-QUIL.RESOURCE::RESOURCE-COLLECTION :QUBITS {3, 9} ..)
  3: (CL-QUIL::SELECT-COST-LOWERING-SWAP #<CL-QUIL::CHIP-SPECIFICATION of 20:23 objects> #<CL-QUIL::CHIP-SCHEDULE {100A1556F3}> T #<CLOSURE (FLET CL-QUIL::COST-FUNCTION :IN CL-QUIL::DO-GREEDY-TEMPORAL-ADDR..
  4: (CL-QUIL::DO-GREEDY-TEMPORAL-ADDRESSING (#<PRAGMA EXPECTED_REWIRING "#(5 1 2 3 4 0 ...)"> #<RZ(-pi/2) 5> #<RX(pi/2) 5> #<RZ(-pi/2) 10> #<RX(-pi/2) 10> #<CZ 5 10> ...) #<CL-QUIL::CHIP-SPECIFICATION of ..
  5: ((LABELS CL-QUIL::PROCESS-BLOCK :IN CL-QUIL:COMPILER-HOOK) #<CL-QUIL::BASIC-BLOCK ENTRY-BLK-811 len:61 in:0 out:term {100A144453}> NIL)
  6: (CL-QUIL:COMPILER-HOOK #<CL-QUIL:PARSED-PROGRAM {100A0C4D03}> #<CL-QUIL::CHIP-SPECIFICATION of 20:23 objects> :PROTOQUIL NIL :REWIRING-TYPE NIL)
  7: ((LAMBDA (#:G2 &REST #:G3) :IN "/Users/mappleby/src/repos/rigetti/quilc/benchmarking/rewiring-analysis.lisp") #<CLOSURE (LAMBDA NIL :IN CL-QUIL-BENCHMARKING::MEASURE-REWIRING-PERFORMANCE) {100A13E5CB}..
  8: (CL-QUIL-BENCHMARKING::MEASURE-REWIRING-PERFORMANCE (:PARTIAL #<FUNCTION #1=(LAMBDA (#:G2 &REST #:G3) :IN "/Users/mappleby/src/repos/rigetti/quilc/benchmarking/rewiring-analysis.lisp") {23A8B41B}> :A*..
  9: (CL-QUIL-BENCHMARKING::BENCH-0017Q-0000061-*DEPTH-VS-SWAPS-ASSN*-0020Q-SKEW-RECT)
 10: (SB-INT:SIMPLE-EVAL-IN-LEXENV (CL-QUIL-BENCHMARKING::BENCH-0017Q-0000061-*DEPTH-VS-SWAPS-ASSN*-0020Q-SKEW-RECT) #<NULL-LEXENV>)
 11: (EVAL (CL-QUIL-BENCHMARKING::BENCH-0017Q-0000061-*DEPTH-VS-SWAPS-ASSN*-0020Q-SKEW-RECT))
```

appleby:
Be more tolerant of IMPOSSIBLY-FULL-RESOURCEs in RESOURCE-DIFFERENCE and RESOURCE-SUBSETP. Previously, if RESOURCE-DIFFERENCE was given an IMPOSSIBLY-FULL-RESOURCE as it's first argument, it would signal RESOURCE-COLLECTION-UNBOUNDED-NAMES. Now it only does so if the second argument is not also "impossibly full".

RESOURCE-SUBSETP now explicitly checks for impossibly full resources, and only delegates to RESOURCE-DIFFERENCE in cases where it is guaranteed safe and returns NIL otherwise.

Closes #346

### DIFF
--- a/src/addresser/astar-rewiring-search.lisp
+++ b/src/addresser/astar-rewiring-search.lisp
@@ -249,7 +249,7 @@ should stop with the given rewiring and false otherwise."
 
           ;; did we run out of time?
     :when (>= iterations max-iterations)
-      :do (format *compiler-noise-stream* "SEARCH-REWIRING: Ran out of iterations")
+      :do (format *compiler-noise-stream* "SEARCH-REWIRING: Ran out of iterations~%")
           (return (values (active-state-swaps best-state) nil))
 
           ;; update that we've visited the state
--- a/src/resource.lisp
+++ b/src/resource.lisp
@@ -311,7 +311,9 @@ REGION-COMBINER to their commonly-named regions."
 
 (defun resource-subsetp (rc1 rc2)
   "Return T if all of the resources in RC1 are contained within RC2."
-  (resource-null-p (resource-difference rc1 rc2)))
+  (or (resource-all-p rc2)
+      (and (not (resource-all-p rc1))
+           (resource-null-p (resource-difference rc1 rc2)))))
 
 
 (defun map-sorted-memory-regions (fn regions1 regions2 &key default-value)


appleby:
These changes seem reasonable to me in isolation and resolve the error reported in #346, but maybe that error shouldn't be ignored and is an indication some higher-level invariant of the addresser was broken and the error really ought to be signaled.

benchmarking/suite.lisp
```diff
@@ -15,7 +15,7 @@
         ;; the the compilation.
         :for quil-file := (read-quil-file quil-file-path)
         :do
-           (handler-bind ((simple-error
+           (handler-bind ((error
```
appleby: Not sure if this change is desirable. This will cause any `error` to be silently swallowed, since `break-on-error` is `nil` when the benchmarks are generated. For example, the error in #346 would probably have gone unnoticed with this change. On the other hand, I'm not sure why this was previously restricted to `simple-error`.

src/resource.lisp
```diff
@@ -282,10 +282,10 @@ REGION-COMBINER to their commonly-named regions."
...
-  (cond ((resource-all-p rc1)
-         (error 'resource-collection-unbounded-names))
-        ((resource-all-p rc2)
+  (cond ((resource-all-p rc2)
          (make-null-resource))
+        ((resource-all-p rc1)
+         (error 'resource-collection-unbounded-names))
         ;; Short circuit memory region combination, if possible
         ((null (resource-collection-memory-regions rc1))
          (combine-collections-default-regions #'integer-bits-difference
```
appleby: This change is not actually needed to resolve #346, only the change to `resource-subsetp`, below.

`(RESOURCE-DIFFERENCE +IMPOSSIBLY-FULL-RESOURCE+ +IMPOSSIBLY-FULL-RESOURCE+) ==> NULL-RESOURCE` seems like on plausible outcome, but maybe this is meant to signal an error in the same way the `inf - inf == nan` in IEEE 754 arithmetic, or maybe `RC1` being "impossibly full" indicates an error elsewhere.

benchmarking/suite.lisp
notmgsk: 🤷‍♂

appleby:
Also note that the backtrace in #346 has `CHIP-SCHEDULE-RESOURCE-CARVING-POINT` as the caller of `RESOURCE-SUBSETP`, so this PR might interact with #547.

src/resource.lisp
ecpeterson: I'm divided about this change. I think there's room for robust semantics here where rc1 is allowed to be infinite and complementing rc2 in it still results in some finite resource, and the case where rc2 passes `resource-all-p` is definitely one such instance. On the other hand, the previous behavior where we just always threw an error clearly says "we're too lazy to deal with this", which might be preferable to "we're too lazy to deal with this, except in one case".

If this were to return a value for these inputs, I agree that this is the right value to return.

benchmarking/suite.lisp
appleby: This did make me think "we should have some generic `QUILC-ERROR` type that is a base for all errors raised in quilc."  Maybe I will open an issue and label it `[good first issue]`.

Do you think this should remain `simple-error`?

src/resource.lisp
appleby: Fair enough. I will take your advice here and revert this change.

benchmarking/suite.lisp
ecpeterson: I don't have a stance about this right now. This can change or not change, and either way it should get mentioned in the `[good first issue]` as something to revisit.

appleby: Reverted for now. Will make sure to link this thread in the new issue.

appleby: https://github.com/rigetti/quilc/issues/550

src/addresser/astar-rewiring-search.lisp
```diff
@@ -249,7 +249,7 @@ should stop with the given rewiring and false otherwise."
 
           ;; did we run out of time?
     :when (>= iterations max-iterations)
-      :do (format *compiler-noise-stream* "SEARCH-REWIRING: Ran out of iterations")
+      :do (format *compiler-noise-stream* "SEARCH-REWIRING: Ran out of iterations~%")
```
notmgsk: MISSING PERIOD

appleby: 😞

PR #554, opened 2020-01-16
fix formal shadowing issue

stylewarning:
This fixes the issue described in https://github.com/rigetti/qvm/issues/236

### DIFF
--- a/src/parser.lisp
+++ b/src/parser.lisp
@@ -647,22 +647,45 @@ If ENSURE-VALID is T (default), then a memory reference such as 'foo[0]' will re
       (otherwise
        (disappointing-token-error tok "a parameter")))))
 
+(defun parse-formal (tok)
+  "Parse a token TOK into a formal argument."
+  (case (token-type tok)
+    ((:NAME)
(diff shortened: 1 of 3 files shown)


stylewarning:
another global special never hurts, right?

src/parser.lisp
```diff
@@ -647,22 +647,42 @@ If ENSURE-VALID is T (default), then a memory reference such as 'foo[0]' will re
...
                            *parse-context*))
-       (if names-memory-region-p
-           (mref (token-payload tok) 0)
-           (formal (token-payload tok)))))
+       (cond
+         ((and *formal-arguments-allowed*
+               shadowing-formal)
+          shadowing-formal)
+         (names-memory-region-p (mref (token-payload tok) 0))
+         (t (formal (token-payload tok))))))
```
appleby: In what situation do you wind up in the default clause here?

stylewarning: Hopefully never. Maybe the least I can do is warn.

appleby:
> illusive

illusive

PR #561, opened 2020-01-21
Basic XY-to-CZ compiler

notmgsk:
In leui of a more complex and generic compilation method, here we have a simple XY->CZ translator.

### DIFF
--- a/src/compilers/translators.lisp
+++ b/src/compilers/translators.lisp
@@ -343,3 +343,18 @@ Note that if (= START-NODE TARGET-NODE) then (list START-NODE) is returned."
         (t
          (finish-compiler
           (build-PISWAP-string computed-path)))))))
+
+;; h/t ecp
+(define-compiler XY-to-CZs
+    ((xy-gate ("XY" (theta) p q)))
+
+  (inst "RZ"   (list (/ pi 4))  p)
+  (inst "H"    nil              p)
+  (inst "RZ"   (list (/ -pi 4)) q)
+  (inst "CNOT" nil              p q)
+  (inst "RY"   (list (param-* theta 1/2)) p)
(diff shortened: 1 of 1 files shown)


ecpeterson:
Here's a less grody circuit:

```
DEFCIRCUIT XY(%alpha) p q:
    T            p
    H            p
    DAGGER T     q
    CNOT         p q
    RY(%alpha/2) p
    RY(%alpha/2) q
    CNOT         p q
    H            p
    DAGGER T     p
    T            q
```

Writing  `DAGGER`ed gates inside of `inst` is a little obnoxious, and so you might replace the `T`s with `RZ(pi/4)` because of that.

stylewarning:
Do we have a test to get XY->CZ nicer? This is a brute force approach. @notmgsk

notmgsk:
> Do we have a test to get XY->CZ nicer? This is a brute force approach. @notmgsk

I'm not sure what you mean

stylewarning:
I thought we had an issue where XY wasn't efficiently translating into CZ, and this fix just is a brute-force inclusion of the "right" translation. It Would Be Nice if we had a way to test the rest of the compiler framework without this test to see that the translation is happening well.

notmgsk:
> I thought we had an issue where XY wasn't efficiently translating into CZ, and this fix just is a brute-force inclusion of the "right" translation. It Would Be Nice if we had a way to test the rest of the compiler framework without this test to see that the translation is happening well.

Non-parametric (and I guess this includes "efficient") `XY` compilation was supported. Parametric `XY` would result in a `compiler-does-not-apply` error, which is fixed by this PR.

stylewarning:
I see. My misunderstanding. Thanks!

PR #569, opened 2020-01-30
Allow on-the-fly redefinitions of compilers

notmgsk:
This is what was originally intended, but wasn't actually mutating the compiler list.

### DIFF
--- a/src/define-compiler.lisp
+++ b/src/define-compiler.lisp
@@ -430,14 +430,14 @@ OPTIONS: plist of options governing applicability of the compiler binding."
 (defun record-compiler (name)
   "Record (possibly overwriting) the existence of the compiler named by NAME."
   (check-type name symbol)
-  (let* ((c (fdefinition name)))
+  (let ((c (fdefinition name)))
     (assert (typep c 'compiler) (name) "The name ~S doesn't name a known compiler." name)
-    (let ((prev (find name **compilers-available** :key #'compiler-name)))
+    (let ((prev (member name **compilers-available** :key #'compiler-name)))
       (cond
         ((null prev)
          (push c **compilers-available**))
         (t
-         (substitute c prev **compilers-available**)))))
+         (rplaca prev c)))))
   (values))
 
 ;;; in what follows, we're very interested in two kinds of maps keyed on bindings:


notmgsk:
> `nsubstitute`

I brought this up with @stylewarning but he was of the opinion that:
1. depending on mutation is not good; and
2. `nsubstitute` is *not* required by the standard to mutate.

src/define-compiler.lisp
```diff
@@ -437,7 +437,8 @@ OPTIONS: plist of options governing applicability of the compiler binding."
         ((null prev)
          (push c **compilers-available**))
         (t
-         (substitute c prev **compilers-available**)))))
+         (setf **compilers-available**
+               (substitute c prev **compilers-available**))))))
```
stylewarning: Here I would use nsubstitute so that we don't get fully quadratic compilation times.

notmgsk: I have been fully bamboozled.

stylewarning: I would do this:

```
(defun record-compiler (name)
  "Record (possibly overwriting) the existence of the compiler named by NAME."
  (check-type name symbol)
  (let* ((c (fdefinition name)))
    (assert (typep c 'compiler) (name) "The name ~S doesn't name a known compiler." name)
    (let ((prev (member name **compilers-available** :key #'compiler-name)))
      (cond
        ((null prev)
         (push c **compilers-available**))
        (t
         (rplaca prev c)))))
  (values))
```

appleby:
According to the spec

> nsubstitute and nsubstitute-if are required to setf any car (if sequence is a list) or aref (if sequence is a vector) of sequence that is required to be replaced with newitem. If sequence is a list, none of the cdrs of the top-level list can be modified.

and

> nsubstitute and nsubstitute-if can be used in for-effect-only positions in code.

which I took to mean you can count on the side-effecting, but I might be misunderstanding.

src/define-compiler.lisp
notmgsk: I can't allow that `let*`.

stylewarning: [We should double check that `**COMPILERS-AVAILABLE**` is always freshly consed.]

appleby:
Note, however, that the entry for `nsubst` does specifically say that

> For nsubst, nsubst-if, and nsubst-if-not the original tree is modified and returned as the function result, but the result may not be eq to tree.

src/define-compiler.lisp
appleby: `rplaca`!

stylewarning:
> According to the spec
> 
> > nsubstitute and nsubstitute-if are required to setf any car (if sequence is a list) or aref (if sequence is a vector) of sequence that is required to be replaced with newitem. If sequence is a list, none of the cdrs of the top-level list can be modified.
> 
> and
> 
> > nsubstitute and nsubstitute-if can be used in for-effect-only positions in code.
> 
> which I took to mean you can count on the side-effecting, but I might be misunderstanding.

I'm wrong, maybe remember nunion or something

notmgsk:
OK Take a look again please.

stylewarning:
do @appleby and @jmbr approve?

src/define-compiler.lisp
```diff
@@ -432,12 +432,12 @@ OPTIONS: plist of options governing applicability of the compiler binding."
...
   (let* ((c (fdefinition name)))
     (assert (typep c 'compiler) (name) "The name ~S doesn't name a known compiler." name)
-    (let ((prev (find name **compilers-available** :key #'compiler-name)))
+    (let ((prev (member name **compilers-available** :key #'compiler-name)))
       (cond
         ((null prev)
          (push c **compilers-available**))
         (t
-         (substitute c prev **compilers-available**)))))
+         (rplaca prev c)))))
```
jmbr: Suggestion 1: replace `let` and `cond` by `if-let` from alexandria.

Suggestion 2: my own preference is to use `(setf (first prev) c)` instead of `rplaca` although the latter packs more punch because it has an onomatopoieitic ring to it.

stylewarning: using `rplac[ad]` is a hill i will die on

stylewarning: also disagree with if-let since `if` is not for side-effects

appleby: @jmbr is "lawful good", prefers `(setf (first prev) c)`

@stylewarning is "choatic good", prefers `(rplaca prev c)`.

this is what makes lisp great, room for everyone

stylewarning: neutral good: `(setf (car prev) c)`

appleby: I instinctively typed it this way, then noticed just before clicking "comment" that it was actually `first` not `car`.

PR #570, opened 2020-01-31
Allow SDK version query to fail silently

notmgsk:


app/src/versions.lisp
```diff
@@ -62,6 +62,10 @@
     (sb-bsd-sockets:socket-error (condition)
       (cl-syslog:rfc-log (*logger* :warning "Encountered a socket error when fetching latest SDK version. (~A)" condition)
         (:msgid "LOG0001"))
+      nil)
+    (error (condition)
+      (cl-syslog:rfc-log (*logger* :warning "Encountered an error when fetching latest SDK version. (~A)" condition)
+        (:msgid "LOG0001"))
```
appleby: I'm guessing it doesn't matter since nobody has complained, but is it valid to have multiple messages logging with the "application startup" msgid, or should these all be `:msgid "LOG0000"` ? There is already a "launching quilc" message logged as `LOG0001` in `process-options`. 

According to the "Coordinate Log Messages" spreadsheet, `LOG0000` is for "A general piece of info. Also an example of a coordinated log message."  Whereas `LOG0001` is described as "An application has started. The message should be logged as early as possible."

notmgsk: You're welcome to clean these up in a separate PR :))))))

appleby:
```
  ||
  ||
  ||
  ||
  ||
  ||
  ||     
  ||     
 /||\
/||||\
======         __|__
||||||        / ~@~ \
||||||       |-------|
||||||       |_______|
```

stylewarning: 0001 is not correct and should be changed

app/src/versions.lisp
```diff
@@ -62,6 +62,10 @@
     (sb-bsd-sockets:socket-error (condition)
```
stylewarning: This is a subclass of `error` so you might just replace this with `error` and change msgid to `0000`

PR #571, opened 2020-01-31
Fix compiler crawler wildcards

### LINKED ISSUE #567: Providing a wildcard to the "arguments" key of a qubit in an ISA leads to the compiler crawler missing appropriate compilation routines.
PyQuil will construct the following ISA for a `1q-qvm` device:
```
{
    "isa": {
        "1Q": {
            "0": {
                "gates": [
                    {
                        "operator": "I",
                        "parameters": [],
                        "arguments": [
                            "_"
                        ],
                        "fidelity": null,
                        "duration": null
                    },
                    {
                        "operator": "RX",
                        "parameters": [
                            0.0
                        ],
                        "arguments": [
                            "_"
                        ],
                        "fidelity": null,
                        "duration": null
                    },
                    {
                        "operator": "RX",
                        "parameters": [
                            1.5707963267948966
                        ],
                        "arguments": [
                            "_"
                        ],
                        "fidelity": null,
                        "duration": null
                    },
                    {
                        "operator": "RX",
                        "parameters": [
                            -1.5707963267948966
                        ],
                        "arguments": [
                            "_"
                        ],
                        "fidelity": null,
                        "duration": null
                    },
                    {
                        "operator": "RX",
                        "parameters": [
                            3.141592653589793
                        ],
                        "arguments": [
                            "_"
                        ],
                        "fidelity": null,
                        "duration": null
                    },
                    {
                        "operator": "RX",
                        "parameters": [
                            -3.141592653589793
                        ],
                        "arguments": [
                            "_"
                        ],
                        "fidelity": null,
                        "duration": null
                    },
                    {
                        "operator": "RZ",
                        "parameters": [
                            "theta"
                        ],
                        "arguments": [
                            "_"
                        ],
                        "fidelity": null,
                        "duration": null
                    },
                    {
                        "operator": "MEASURE",
                        "qubit": "_",
                        "target": "_",
                        "duration": null,
                        "fidelity": null
                    },
                    {
                        "operator": "MEASURE",
                        "qubit": "_",
                        "target": null,
                        "duration": null,
                        "fidelity": null
                    }
                ]
            }
        },
        "2Q": {
            
        }
    },
    "specs": {
        "1Q": {
            "0": {
                "f1QRB": 0.909,
                "f1QRB_std_err": 0.01,
                "f1Q_simultaneous_RB": 0.909,
                "f1Q_simultaneous_RB_std_err": 0.02,
                "fRO": 0.9,
                "T1": 3e-05,
                "T2": 3e-05,
                "fActiveReset": 0.909
            },
            "1": {
                "f1QRB": 0.909,
                "f1QRB_std_err": 0.01,
                "f1Q_simultaneous_RB": 0.909,
                "f1Q_simultaneous_RB_std_err": 0.02,
                "fRO": 0.9,
                "T1": 3e-05,
                "T2": 3e-05,
                "fActiveReset": 0.909
            }
        },
        "2Q": null
    }
}

```
This ought to read as: on qubit `0`, the gates `I`, `RX(+/- n pi)`, and `RX(theta)` are native. The compiler will use this information to construct a list of compilation routines that will move us toward the above native gateset (see `compute-applicable-compilers`) or keep us within the gateset but produce more optimal output (i.e. shorter gate sequences; see `compute-applicable-reducers`).

Given the above ISA, however, the compiler decides that the routine `agglutinate-RXs` is not appropriate. After some exploration, this is due to the wildcard in all of the `"arguments"` fields of the `"RX"` gates in the ISA. Changing the wildcard to be specific (i.e. in this case changing it to `0`) fixes the issue. Indeed, `(build-nq-fully-connected-chip 1)` produces an ISA whose `"arguments"` are specifically the qubit under which they are keyed.

There are a few options here:
* a. pyQuil is in error, and should not be providing wildcards under the arguments field
* b. quilc ought to "correct" the wildcard to the appropriate qubit
* c. quilc ought to use the wildcard, with some (possibly missing) logic to relax its constraints when picking applicable reducers/compilers

I'm not sure why the `"arguments"` key needs to exist at all. My understanding is that all gates listed under a particular qubit in the ISA (`"0"` above) apply to that qubit, so repeating that qubit index under `"arguments"` seems redundant. It's even less clear to me what a wildcard here could mean.

notmgsk:
TODO
- [x] Test

Closes #567.

### DIFF
--- a/src/define-compiler.lisp
+++ b/src/define-compiler.lisp
@@ -300,7 +300,7 @@ OPTIONS: plist of options governing applicability of the compiler binding."
 (define-condition cannot-concretize-binding (serious-condition) ())
 (define-condition no-binding-match (serious-condition) ())
 
-(defun prefer-concrete-items (a b)
+(defun prefer-concrete-atom (a b)
   "Return one of A or B, choosing the one that is more \"concrete\".
 
 What \"concrete\" means depends on the types of A and B:
@@ -334,14 +334,14 @@ What \"concrete\" means depends on the types of A and B:
          (error 'cannot-concretize-binding)
          a))
     (t
-     (mapcar #'prefer-concrete-items a b))))
+     (mapcar #'prefer-concrete-atom a b))))
 
 (defgeneric binding-meet (a b)
   (:documentation "Constructs a maximal binding subsumed by both A and B.")
   (:method (a b)
     (error 'cannot-concretize-binding))
   (:method ((a gate-binding) (b gate-binding))
-    (make-gate-binding :operator (prefer-concrete-items (gate-binding-operator a) (gate-binding-operator b))
+    (make-gate-binding :operator (prefer-concrete-atom (gate-binding-operator a) (gate-binding-operator b))
                        :parameters (prefer-concrete-lists (gate-binding-parameters a)
                                                           (gate-binding-parameters b))
                        :arguments (prefer-concrete-lists (gate-binding-arguments a)
--- a/src/define-compiler.lisp
+++ b/src/define-compiler.lisp
@@ -301,7 +301,18 @@ OPTIONS: plist of options governing applicability of the compiler binding."
 (define-condition no-binding-match (serious-condition) ())
 
 (defun prefer-concrete-items (a b)
+  "Return one of A or B, choosing the one that is more \"concrete\".
+
+What \"concrete\" means depends on the types of A and B:
+
+  * If A (B) is a SYMBOL, and B (A) is not, return B (A). In this case we are choosing a value that should not be a wildcard-type value.
+
+  * If neither are symbols and they are equal to one another, return either. In this case both have a non-wildcard-type value.
+
+  * If both are symbols, then return either. In this case both are wildcard symbols and neither is more \"concrete\" than the other."
   (cond
+    ((and (symbolp a) (symbolp b))
+     b)
     ((and (symbolp a) (not (symbolp b)))
      b)
     ((and (not (symbolp a)) (symbolp b))
@@ -330,7 +341,6 @@ OPTIONS: plist of options governing applicability of the compiler binding."
   (:method (a b)
     (error 'cannot-concretize-binding))
   (:method ((a gate-binding) (b gate-binding))
-    
     (make-gate-binding :operator (prefer-concrete-items (gate-binding-operator a) (gate-binding-operator b))
                        :parameters (prefer-concrete-lists (gate-binding-parameters a)
                                                           (gate-binding-parameters b))
@@ -342,7 +352,7 @@ OPTIONS: plist of options governing applicability of the compiler binding."
     b))
 
 (defgeneric instantiate-binding (binding)
-  (:documentation "When possible, construct the unique instruction on which BINDING will match.")
+  (:documentation "When possible, construct the unique instruction on which BINDING will match. If any of the binding's arguments are unspecified (i.e. match against any qubit), those arguments will be filled-in with unique qubit indices.")
   (:method (binding)
     (error 'cannot-concretize-binding))
   (:method ((binding gate-binding))
@@ -352,7 +362,11 @@ OPTIONS: plist of options governing applicability of the compiler binding."
               (some #'symbolp (gate-binding-parameters binding)))
       (error 'cannot-concretize-binding))
     (when (some #'symbolp (gate-binding-arguments binding))
-      (error 'cannot-concretize-binding))
+      (let* ((usedq (gate-binding-arguments binding))
+             (availableq (set-difference (a:iota (length usedq)) usedq)))
+        (setf binding (copy-binding binding
+                                    :arguments (mapcar (lambda (q) (if (symbolp q) (pop availableq) q))
+                                                       usedq)))))
     (apply #'build-gate
            (gate-binding-operator binding)
            (gate-binding-parameters binding)
@@ -710,8 +724,16 @@ N.B.: The word \"shortest\" here is a bit fuzzy.  In practice it typically means
                   sorted-compilers)))
           compilers-with-features)))))
 
-(defun compute-applicable-reducers (gateset)
-  "Returns all those compilers (including those which match on sequences of instructions rather than single instructions) which improve the brevity of a gate sequence without exiting a particular GATESET."
+(defun gates-that-match-binding (binding gateset)
+  (loop :for gate-binding :being :the :hash-keys :of gateset
+        :for gate := (handler-case (instantiate-binding (binding-meet gate-binding binding))
+                       (cannot-concretize-binding () nil)
+                       (no-binding-match () nil))
+        :when gate
+          :collect gate))
+
+(defun compute-applicable-reducers (gateset &key (compilers **compilers-available**))
+  "Returns all those compilers (including those which match on sequences of instructions rather than single instructions) which improve the brevity of a gate sequence without exiting a particular GATESET. The optional COMPILERS specifies the list of compilers from which the applicable reducers should be selected. By default COMPILERS is the globally-defined compilers list."
   (labels
       ((calculate-fidelity-of-concrete-gates (gates)
          (loop :with fidelity := 1d0
@@ -723,12 +745,6 @@ N.B.: The word \"shortest\" here is a bit fuzzy.  In practice it typically means
                          ;; only executes if no binding subsumes the gate
                          :finally (error 'no-binding-match))
                :finally (return fidelity)))
-       (gates-that-match-binding (binding)
-         (loop :for gate-binding :being :the :hash-keys :of gateset
-               :for gate := (handler-case (instantiate-binding (binding-meet gate-binding binding))
-                              (cannot-concretize-binding () nil)
-                              (no-binding-match () nil))
-               :when gate :collect gate))
        (consider-input-test-case (compiler &rest gates)
          (handler-case
              (let* ((input-fidelity (calculate-fidelity-of-concrete-gates gates))
@@ -770,7 +786,7 @@ N.B.: The word \"shortest\" here is a bit fuzzy.  In practice it typically means
                   (>= out-fidelity in-fidelity)))
            ;; if that falls through, try the exhaustive match
            (let ((matches (loop :for binding :in (compiler-bindings compiler)
-                                :for match := (gates-that-match-binding binding)
+                                :for match := (gates-that-match-binding binding gateset)
                                 :when (null match)
                                   :do (return-from consider-compiler nil)
                                 :collect match)))
@@ -779,7 +795,7 @@ N.B.: The word \"shortest\" here is a bit fuzzy.  In practice it typically means
              (every #'identity (apply #'a:map-product (a:curry #'consider-input-test-case compiler) matches)))))))
     
     (let (applicable-compilers)
-      (dolist (compiler **compilers-available** applicable-compilers)
+      (dolist (compiler compilers applicable-compilers)
         (when (consider-compiler compiler)
           (push compiler applicable-compilers))))))
 
--- a/tests/compilation-tests.lisp
+++ b/tests/compilation-tests.lisp
@@ -213,7 +213,13 @@ CNOT 4 8
   (let* ((chip (quil::read-chip-spec-file
                 (merge-pathnames *qpu-test-file-directory*
                                  "1q-wildcard-arguments.qpu")))
-         (progm (parse "RX(pi/2) 0; RX(-pi/2) 0")))
-    (is (m= (magicl:make-identity-matrix 2)
-            (parsed-program-to-logical-matrix
-             (compiler-hook progm chip))))))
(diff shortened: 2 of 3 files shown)


src/chip/chip-reader.lisp
```diff
@@ -282,6 +282,7 @@
                   (dohash ((binding record) gate-info)
                     (when (and (gate-binding-p binding)
                                (equalp (named-operator "RX") (gate-binding-operator binding))
+                               (and (numberp (first (gate-binding-parameters binding))))
```
stylewarning: delete and; comment on what this case is guarding against

src/define-compiler.lisp
```diff
@@ -325,17 +325,25 @@ OPTIONS: plist of options governing applicability of the compiler binding."
     (t
      (mapcar #'prefer-concrete-items a b))))
 
+(defun concretize-gate-binding-arguments (gate)
+  "If GATE's arguments are all wildcards (i.e. they match any qubit) then return a new GATE-BINDING object whose arguments are unique qubit integers. Does not mutate GATE."
+  (let ((gate-arguments (gate-binding-arguments gate)))
+    (if (every #'wildcard-pattern-p gate-arguments)
```
stylewarning: is there any reason we can't concretize partially wild args?

src/define-compiler.lisp
```diff
@@ -710,7 +718,14 @@ N.B.: The word \"shortest\" here is a bit fuzzy.  In practice it typically means
                   sorted-compilers)))
           compilers-with-features)))))
 
-(defun compute-applicable-reducers (gateset)
+(defun gates-that-match-binding (binding gateset)
+  (loop :for gate-binding :being :the :hash-keys :of gateset
+        :for gate := (handler-case (instantiate-binding (binding-meet gate-binding binding))
+                       (cannot-concretize-binding () nil)
```
stylewarning: is it valid to give an empty lambda list?

src/define-compiler.lisp
```diff
@@ -710,7 +718,14 @@ N.B.: The word \"shortest\" here is a bit fuzzy.  In practice it typically means
...
+                       (no-binding-match () nil))
+        :when gate :collect gate))
```
stylewarning: prefer collect gate on newline

src/define-compiler.lisp
```diff
@@ -710,7 +718,14 @@ N.B.: The word \"shortest\" here is a bit fuzzy.  In practice it typically means
...
+
+(defun compute-applicable-reducers (gateset &key (compilers **compilers-available**))
```
stylewarning: maybe document COMPILERS and its default value being all known global compilers

src/define-compiler.lisp
```diff
@@ -325,17 +325,25 @@ OPTIONS: plist of options governing applicability of the compiler binding."
...
   (:method (a b)
     (error 'cannot-concretize-binding))
   (:method ((a gate-binding) (b gate-binding))
-    
-    (make-gate-binding :operator (prefer-concrete-items (gate-binding-operator a) (gate-binding-operator b))
-                       :parameters (prefer-concrete-lists (gate-binding-parameters a)
-                                                          (gate-binding-parameters b))
-                       :arguments (prefer-concrete-lists (gate-binding-arguments a)
-                                                         (gate-binding-arguments b))))
+    (let ((a (concretize-gate-binding-arguments a))
```
stylewarning: this change violates the docstring; we are no longer returning a "maximal binding". Why does relaxing this constraint and concretizing just the arguments make sense?

I think this is my biggest question about the PR.

notmgsk:
Test runners are failing, though not locally -- fun.

src/define-compiler.lisp
notmgsk: > If var is not needed, it can be omitted. That is, a clause such as:
>
>  (typespec (var) (declare (ignore var)) form)

src/define-compiler.lisp
```diff
@@ -352,7 +362,11 @@ OPTIONS: plist of options governing applicability of the compiler binding."
               (some #'symbolp (gate-binding-parameters binding)))
       (error 'cannot-concretize-binding))
     (when (some #'symbolp (gate-binding-arguments binding))
-      (error 'cannot-concretize-binding))
+      (let* ((usedq (gate-binding-arguments binding))
```
notmgsk: this can probably be cleaned up

src/define-compiler.lisp
notmgsk: I can't think of one. I allow for that in the latest commit.

src/define-compiler.lisp
notmgsk: True. Fixed.

src/define-compiler.lisp
```diff
@@ -301,7 +301,18 @@ OPTIONS: plist of options governing applicability of the compiler binding."
...
 (defun prefer-concrete-items (a b)
+  "Return one of A or B, choosing the one that is more \"concrete\".
+
+What \"concrete\" means depends on the types of A and B:
+
+  * If A (B) is a SYMBOL, and B (A) is not, return B (A). In this case we are choosing a value that should not be a wildcard-type value.
+
+  * If neither are symbols and they are equal to one another, return either. In this case both have a non-wildcard-type value.
+
+  * If both are symbols, then return either. In this case both are wildcard symbols and neither is more \"concrete\" than the other."
```
appleby: So this is the function now signals `cannot-concretize-binding` only when both `a` and `b` are already concrete (not symbols) but also non-`equalp`? And if it's not able to concretize (i.e. both are wildcards), then `cannot-concretize-binding` is not signaled?

src/define-compiler.lisp
```diff
@@ -352,7 +362,11 @@ OPTIONS: plist of options governing applicability of the compiler binding."
               (some #'symbolp (gate-binding-parameters binding)))
       (error 'cannot-concretize-binding))
     (when (some #'symbolp (gate-binding-arguments binding))
-      (error 'cannot-concretize-binding))
+      (let* ((usedq (gate-binding-arguments binding))
+             (availableq (set-difference (a:iota (length usedq)) usedq)))
```
appleby: Why fill in the wildcards with {1..n} - `usedq`? What if qubits 1...n aren't present in the ISA?

Also, the docstring says this function returns "*the* unique instruction on which BINDING will match", but by filling in wildcards with arbitrary qubits, aren't we now returning "*an* ~~unique~~ instruction on which BINDING will match"?

I'm not worried about the docstring being incorrect so much as I'm just nervous that this might be breaking an assumption that this function would previously signal `cannot-concretize-binding` if it couldn't find a single unique instantiation, whereas now we seem to just pick one for wildcards. Doesn't help that I'm clueless about how this is all supposed to work 😬

build-app.lisp
```diff
@@ -2,7 +2,7 @@
 ;;;;
 ;;;; This file is loaded by the Makefile to produce a quilc[.exe] binary.
 ;;;;
-
+1
```
stylewarning: oops

src/define-compiler.lisp
notmgsk: > Why fill in the wildcards with {1..n} - usedq? What if qubits 1...n aren't present in the ISA?

I don't believe it matters. The fidelity calculation used to select a compiler occurs in `calculate-fidelity-of-concrete-gates`. You will notice that it doesn't use fidelity of qubit or gate information of the targeted qubits (i.e. those in the `gates` produced by `gates-that-match-binding`) but instead uses the fidelity information in the gateset. We call a compiler, e.g. `normalize-rz` with a gate e.g. `(build-gate "RZ" (list (* pi pi)) 1000)` -- the compiler doesn't care about its input arguments and indeed has no knowledge of an ISA. The fidelity is calculated by: compile the above gate with the above compiler, for each gate in the result look up its fidelity in the ISA and multiply them all together.

We can ignore the particulars of the qubit indices because they were given as wildcards in both the compiler and the ISA.

Reading through the above I don't think I've explained it very well.

> Also, the docstring says this function returns "the unique instruction on which BINDING will match", but by filling in wildcards with arbitrary qubits, aren't we now returning "an unique instruction on which BINDING will match"?

That's true, but you can't have a unique instruction if the binding has wildcards.

appleby: I'm still confused, but should probably go try to read code before firing off more random questions (although I'll probably still be confused). If you're happy, I'm happy. I mistakenly assumed this was part of the process of trying to find an applicable compiler for a given instruction, and that the concretized binding would then be pattern matched against the instruction to see if it was applicable or something.

It seems odd that the "arguments" field exists at all if the actual values don't matter, but maybe it just doesn't matter here. Again, I should rtfs.

> That's true, but you can't have a unique instruction if the binding has wildcards.

True, but the docstring starts with "When possible", and previously raised `cannot-concretize-binding` if the binding couldn't be made fully concrete, but maybe that was just a bug... so here we are, fixing the bug. I think I'm catching on :)

appleby:
Fyi, I was just waiting to `[Approve]` because I saw there was an outstanding test-related TODO (which would be swell).  Also the stray `1` in `build-app.lisp` that @stylewarning pointed out.

notmgsk:
It was a good q actually. It made me think of possible issues that I hadn’t considered. Usually ecp would’ve been that guy, but you’re The Man Now, Dog.

appleby:
> It was a good q actually. It made me think of possible issues that I hadn’t considered. Usually ecp would’ve been that guy, but you’re The Man Now, Dog.

*au contraire*, you're the one fixing all these hairy compiler bugs recently 😉

src/define-compiler.lisp
notmgsk: Yes to the first part. To the second part, when both are wildcards then return either.

tests/compilation-tests.lisp
```diff
@@ -207,3 +207,13 @@ CNOT 4 8
...
+
+(deftest test-rx-agglutination-with-wildcard-chip ()
+  ;; See #567
+  (let* ((chip (quil::read-chip-spec-file
+                (merge-pathnames *qpu-test-file-directory*
+                                 "1q-wildcard-arguments.qpu")))
+         (progm (parse "RX(pi/2) 0; RX(-pi/2) 0")))
+    (is (m= (magicl:make-identity-matrix 2)
+            (parsed-program-to-logical-matrix
+             (compiler-hook progm chip))))))
```
appleby: why do you want `m=` here rather than `matrix-equality` or `matrix-equals-dwim`?

also, in what sense is this testing RX agglutination? even if the compiler emitted two separate `RX` instructions, wouldn't they still get combined to an identity by `p-p-t-l-m`?

stylewarning: More specifically, the IS check is more of a consistency check (is the compilation a correct one), not a behavioral check (did compilation operate in the expected way).

src/define-compiler.lisp
notmgsk: I don't think it's a terrible name since you cannot "concretize" unequal numbers.

notmgsk:
This seems to have made the `modifiers-galore.quil` test unstable. Running that test on master repeatedly shows no issue, but on this branch it generally fails within ~10 runs.

notmgsk:
> This seems to have made the `modifiers-galore.quil` test unstable. Running that test on master repeatedly shows no issue, but on this branch it generally fails within ~10 runs.

Actually, I take that back. Looks like it is also flacky on master.

src/define-compiler.lisp
notmgsk: For my sake and maybe for yours, here is what is happening:

1. We have defined a bunch of compiler routines, that act on a particular (though possibly "generic") form instruction (or more than one) (and these are called the compiler bindings). The compilers output a particular form of instruction.
2. We have an ISA which describes the native gates the compiled output can use.
3. The compilers in (1) might have outputs that are not in (2) but nevertheless they might provide an optimized output that can then be passed into other compilers whose output is in (2). So the idea here is to find a path of compilers in (1) that will, when given the most generic instruction (i.e. wildcard name, wildcard parameters, wildcard arguments), produce good output that is in (2).

notmgsk: I think the "fully concrete" interpretation of that function not necessarily appropriate.

appleby: this comment was helpful, ty.

src/define-compiler.lisp
```diff
@@ -301,7 +301,18 @@ OPTIONS: plist of options governing applicability of the compiler binding."
...
+  * If A (B) is a SYMBOL, and B (A) is not, return B (A). In this case we are choosing a value that should not be a wildcard-type value.
+
+  * If neither are symbols and they are equal to one another, return either. In this case both have a non-wildcard-type value.
+
+  * If both are symbols, then return either. In this case both are wildcard symbols and neither is more \"concrete\" than the other."
   (cond
+    ((and (symbolp a) (symbolp b))
+     b)
     ((and (symbolp a) (not (symbolp b)))
      b)
```
jmbr: Unless I'm missing something, I suggest:
```suggestion
    ((symbolp a)
     b)
```

PR #575, opened 2020-02-03
Only disallow SSL for SDK builds

notmgsk:
This prevents the annoying situation where you've loaded e.g. drakma in your REPL with SSL enabled, and after that when you try to compile quilc you get the error about being unable to find the CL+SSL package.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,7 +7,6 @@ QUICKLISP_HOME=$(HOME)/quicklisp
 QUICKLISP_SETUP=$(QUICKLISP_HOME)/setup.lisp
 QUICKLISP=$(SBCL) --load $(QUICKLISP_HOME)/setup.lisp \
 	--eval '(push (truename ".") asdf:*central-registry*)' \
-	--eval '(push :drakma-no-ssl *features*)' \
 	--eval "(push (truename \"$(RIGETTI_LISP_LIBRARY_HOME)\") ql:*local-project-directories*)"
 QUICKLISP_BOOTSTRAP_URL=https://beta.quicklisp.org/quicklisp.lisp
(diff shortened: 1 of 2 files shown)


build-app.lisp
```diff
@@ -6,7 +6,8 @@
 (unless *load-truename*
   (error "This file is meant to be loaded."))
 
-(pushnew :drakma-no-ssl *features*)
+(when (find :forest-sdk *features*)
+  (pushnew :drakma-no-ssl *features*))
```
appleby: `#+forest-sdk` instead of `(when (find ...))`

Makefile
```diff
@@ -73,13 +72,12 @@ endif
...
 	$(SBCL) $(FOREST_SDK_FEATURE) \
 	        --eval "(setf sb-ext:\*on-package-variance\* '(:warn (:swank :swank-backend :swank-repl) :error t))" \
-		--eval '(push :drakma-no-ssl *features*)' \
 		--load "build-app.lisp" \
 		$(FOREST_SDK_OPTION) \
 		$(QUILC_UNSAFE_OPTION)
 
 
-quilc-sdk-base: FOREST_SDK_FEATURE=--eval '(pushnew :forest-sdk *features*)'
+quilc-sdk-base: FOREST_SDK_FEATURE=--eval '(pushnew :forest-sdk *features*)' --eval ' (push :drakma-no-ssl *features*)'
```
appleby: Isn't this redundant with the `pushnew` in `build-app.lisp`?

notmgsk: Maybe. I was trying to cover all my bases. I'm scurred to remove it.

appleby: 👨‍👧‍👦 we're here for you

Issue #577, opened 2020-02-04
State prep compiler broken for small program


notmgsk:
The following reliably trips up the compiler, raising an error due to broke colinearity:
```
(progn 
  (setf quil::*compress-carefully* t
        quil::*enable-state-prep-compression* t)
  (quil::compiler-hook
   (quil::parse-quil "H 0
H 1
X 0
CPHASE(pi/2) 0 1")
   (quil::build-nq-linear-chip 2 :architecture :cphase)))
```

The issue was introduced around `585d3f52`.

notmgsk:
Closed by #578.

PR #578, opened 2020-02-05
Revert to old 2Q state prep compiler

kilimanjaro:
This partially addresses #577. The most recent 2Q state prep compiler, based on the technique of Perdomo, fails inexplicably in certain special circumstances. I have added a test case for which the new technique fails. The precise nature of this failure is not clear to me after a short investigation, but it seems like something that might take a significant effort to root out (e.g. on part with reimplementing the state prep compiler).

As a stopgap, I propose that we leave the newer state prep code marked `#+#:pedagogical-purposes-only` and revert to the previous method.


tests/compiler-hook-tests.lisp
```diff
@@ -118,8 +118,9 @@ JUMP @a")))
   "Test whether the compiler hook preserves semantic equivalence for
 some test programs."
   (finish-output)
-  (dolist (state-prep '(nil t))
-    (let ((quil::*enable-state-prep-compression* state-prep))
+  (dolist (state-prep '(t nil))
+    (let ((quil::*enable-state-prep-compression* state-prep)
+          (quil::*compress-carefully* t))
```
kilimanjaro: I realize this is set by the suite, but it's nice to have it set here as well if one is running this specific test.

src/compilers/state-prep.lisp
```diff
@@ -217,6 +181,7 @@ Returns a pair (LIST C0 C1) of 2x2 matrices with (C0 (x) C1) SOURCE-WF = TARGET-
 ;; this method is due to Oscar Perdomo, and this implementation blindly follows his notes.
 ;; TODO: this should be made architecture-sensitive, with separate templates
 ;;       for ISWAP-based chips
+#+#:pedagogical-purposes-only
```
appleby: should `state-prep-within-local-equivalence-class` and `tensor-factorization` also be `#+#:pedagogical-purposes-only` now?

src/compilers/state-prep.lisp
```diff
@@ -373,9 +338,7 @@ Returns a pair (LIST C0 C1) of 2x2 matrices with (C0 (x) C1) SOURCE-WF = TARGET-
 ;; formulas in Oscar's. it would be nice to write out exactly what this meet is.
 ;;
 ;; some of the code in this has been pulled down to state-prep-within-local-equivalence-class
-;; below. if anyone wants to turn this routine back on, it's worth offloading some
-;; of the code in here to that subroutine.
```
appleby: offloading no bueno?

src/compilers/state-prep.lisp
```diff
@@ -94,99 +94,63 @@
...
-                (not (double= 0d0 (imagpart (aref v 0)))))
-      (let* ((theta (- (atan (realpart (aref v 1))
-                             (realpart (aref v 0)))))
-             (m (magicl:make-complex-matrix 4 4 (list (cos theta) (sin theta) 0 0
-                                                      (- (sin theta)) (cos theta) 0 0
-                                                      0 0 1 0
-                                                      0 0 0 1))))
-        (setf matrix (m* m matrix))
-        (setf v (nondestructively-apply-matrix-to-vector matrix vector))))
-    ;; make sure |v0| > |v1|
```
appleby: I take it everything from here down was only needed for the Perdomo version? If so, would it make sense to add a note to the comment below reminding any future state prepper who wants to re-enable the Perdomo version that this code needs to be resurrected as well?

src/compilers/state-prep.lisp
```diff
@@ -94,99 +94,63 @@
...
-                 (or (<= centered-phase -pi/2)
-                     (double= centered-phase -pi/2)
-                     (< pi/2 centered-phase)))
-        (let* ((m (magicl:make-complex-matrix 4 4 (list -1 0  0 0
-                                                        0 1  0 0
-                                                        0 0 -1 0
-                                                        0 0  0 1))))
-          (setf matrix (m* m matrix))
-          (setf v (nondestructively-apply-matrix-to-vector matrix vector)))))
+    (symbol-macrolet ((v0 (aref v 0))
```
notmgsk: School me. Why use `symbol-macrolet` here?

jmbr: My understanding is that @kilimanjaro  is updating `v` throughout the calls to `apply-rotation` and thus he needs to get the latest values of its components every time (the alternative of using `let` wouldn't work because the entries would not be updated automatically). The `symbol-macrolet` here also helps with readability.

src/compilers/state-prep.lisp
kilimanjaro: Good idea

src/compilers/state-prep.lisp
kilimanjaro: I didn't want to do this work myself, so I have instead opted for preserving these comments in their original form, but with an extra note clarifying the context.

src/compilers/state-prep.lisp
kilimanjaro: Adding a comment and referring to the commit that introduced these, to make it easier to bring them back.

src/compilers/state-prep.lisp
kilimanjaro: Yes, exactly that.

src/compilers/state-prep.lisp
appleby: works for me

PR #581, opened 2020-02-06
Omit circuit and gate definitions from protoquil responses

kilimanjaro:
This strips non-protoquil data (gate and circuit definitions, specifically) from the protoquil responses in quilc.


kilimanjaro:
I decided to move this change slightly (so that it was not lumped in with the `HALT` handling code)

PR #584, opened 2020-02-11
Quilt Scheduler

kilimanjaro:
**Note**: _I am keeping this a draft as I do some hacking on the rudimentary backend. This PR specifically serves to catch some hardware-agnostic aspects of scheduling and code analysis which should exist independently of the low-level code generation._

This adds a simple greedy scheduler for Quilt programs. Here "scheduling" means building up a data structure which tracks for each hardware object an list of `(instruction . time)` pairs, consistent with the basic semantics of Quilt. "Hardware object" is also a term of art: these are simply names (e.g. `"q0_ff"`) specified in Quilt frame definitions.

The other change of note here is that I am removing the requirement that waveform definitions include a sample rate. This was always a hack, and there's no principled reason for it. For context, waveform definitions look something like this:
```
DEFWAVEFORM cool_gaussian:
    <list of complex numbers>
```
It's impossible to compute the duration of something like `PULSE 0 "rf" cool_gaussian` without knowing the sample rate of the DAC. I think this is fine: plenty of program analyses do not need this information (it's not needed for name resolution, calibration expansion, type checking, etc), and the ones that do can just signal an error if this information is not present (in the corresponding `DEFFRAME` body).


src/quilt/analysis/schedule-to-hardware.lisp
```diff
@@ -0,0 +1,221 @@
...
+;;; Greedy Scheduling
+;;;
+;;; A basic relationship is that of an instruction 'obstructing' a frame:
+;;; informally, if instructions A and B obstruct a common frame, then A and B
+;;; must be scheduled on non-overlapping time intervals.
+;;;
+;;; A naive approach to scheduling is to maintain a clock for each hardware
+;;; object, tracking the next available time for an operation. For each
+;;; instruction, the set of obstructed frames is computed. The instruction
+;;; begins at the next time which sychronizes these, and the associated clocks
```
jmbr:
```suggestion
;;; begins at the next time which synchronizes these, and the associated clocks
```

src/quilt/analysis/schedule-to-hardware.lisp
```diff
@@ -0,0 +1,221 @@
...
+;;; and so the naive start time for B is 1.0, due to synchronization with 0 1 "cz".
+;;;
+;;; The approach adopted here is to only synchronize across 'properly'
+;;; obstructed frames (defined below). More or less this implies that a program
+;;; without 2Q instructions will have all 1Q operations scheduled independently.
+
+(defun properly-obstructed-frames (instr program)
+  "Get all the frames in PROGRAM which are 'properly' obstructed by INSTR.
+
+Here a frame is 'properly' obstructed if it is obstructed by instr and the frame
```
jmbr:
```suggestion
Here a frame is 'properly' obstructed if it is obstructed by INSTR and the frame
```

tests/quilt/scheduling-tests.lisp
```diff
@@ -0,0 +1,217 @@
...
+FENCE
+DELAY 0 1.0
+"))))
+    (flet ((num-instrs (hw)
+             (length (gethash hw schedule))))
+      (assert (= 4 (hash-table-count schedule)))
+      (assert (= 1 (num-instrs "q0_ro_tx")))
+      (assert (= 3 (num-instrs "q0_rf")))
+      (assert (= 1 (num-instrs "q1_rf")))
+      (assert (= 1 (num-instrs "q0_ff"))))))
```
jmbr: I suggest using `is` instead of `assert`.

tests/quilt/scheduling-tests.lisp
```diff
@@ -0,0 +1,217 @@
...
+;;; TODO: test swap-phase error
+
+(defun instruction-start-time (instr schedule)
+  (let ((hw-name (quilt::resolve-hardware-object instr)))
+    (loop :for (time . op) :in (gethash hw-name schedule)
+          :when (eql op instr)
+            :do (return-from instruction-start-time time))))
+
+(deftest test-schedule-nonblocking-pulses ()
+  (let* ((pp (calibrate "
```
jmbr: Friends don't `let` friends use `let*` for just one binding.

tests/quilt/scheduling-tests.lisp
kilimanjaro: This is what happens when I spend 4 weeks straight with pytest 😵

kilimanjaro:
> Looking good so far! I'm not hitting "Approve" because it seems to be a draft according to Github. Let us know when that changes.

src/quilt/analysis/expand-template-waveforms.lisp
```diff
@@ -11,6 +11,9 @@
   ;; we make epsilon 10x larger than floating point epsilon.
   (ceiling (- x (* x 10 double-float-epsilon))))
 
+(cffi:defcfun "erf" :double
```
jmbr: 💖

PR #593, opened 2020-02-27
Improve collinearity check and allow for continuing from error raised by CHECK-CONTEXTUAL-COMPRESSION-WAS-WELL-BEHAVED.

notmgsk:


src/matrix-operations.lisp
```diff
@@ -282,29 +282,20 @@ as needed so that they are the same size."
...
-      ;; component-wise calculation (loop ... :always (double= 0d0 (expt ...)))
-      ;; which has the potential to terminate early on a negative result
-      (and scalar
-           (let ((norm (sqrt (dot-product vect1 vect2 (cis scalar)))))
-             (double= 1d0 norm))))))
-
+    ;; rather than computing the full norm, this could be replaced by the
+    ;; component-wise calculation (loop ... :always (double= 0d0 (expt ...)))
+    ;; which has the potential to terminate early on a negative result
+    (let ((norm^2 (abs (dot-product vect1 vect2))))
```
jmbr: The name of `norm^2` is misleading because it is (the absolute value of) an inner product, not a norm.

src/matrix-operations.lisp
```diff
@@ -282,29 +282,20 @@ as needed so that they are the same size."
...
+      (values (double= 1d0 norm^2)
+              (- 1d0 norm^2)))))
```
jmbr:
```suggestion
              (1- norm^2)))))
```

src/matrix-operations.lisp
```diff
@@ -282,29 +282,20 @@ as needed so that they are the same size."
...
-  "Tests whether two (complex double) vectors of unit norm are collinear."
-  (flet ((dot-product (v1 v2 &optional (scalar 1d0))
-           (loop :for r :across v1
-                 :for s :across v2
-                 :sum (* (conjugate r) s scalar))))
+  "Tests whether two (complex double) vectors of unit norm are collinear. Returns pair of values: first is whether the vectors were collinear, and the second is the amount by which the dot product of the vectors differs from one."
+  ;; TODO Do not use dot product. Instead do component-wise
+  ;; calculation of the scalar relation, and check that it is the same
+  ;; for all entries.
+  (let ((vect1 (coerce vect1 'list))
```
jmbr: We probably want the latest and greatest MAGICL routine here instead of `dot-product`.

src/matrix-operations.lisp
```diff
@@ -282,29 +282,20 @@ as needed so that they are the same size."
     ret))
 
 (defun collinearp (vect1 vect2)
-  "Tests whether two (complex double) vectors of unit norm are collinear."
-  (flet ((dot-product (v1 v2 &optional (scalar 1d0))
-           (loop :for r :across v1
-                 :for s :across v2
-                 :sum (* (conjugate r) s scalar))))
+  "Tests whether two (complex double) vectors of unit norm are collinear. Returns pair of values: first is whether the vectors were collinear, and the second is the amount by which the dot product of the vectors differs from one."
+  ;; TODO Do not use dot product. Instead do component-wise
```
jmbr: This could be done with the `.*` element-wise product in the new MAGICL and then calculating the standard deviation of the resulting vector (if it is small enough, then the resulting vector is deemed to be constant, and we infer that the original vectors were collinear).

jmbr: On a second thought, it's better to loop and check whether the quotient of the entries is approximately the same and end the loop early if it isn't.

src/matrix-operations.lisp
notmgsk: This was a test, and you passed it. Congrats @jmbr

jmbr: Hooray!!

tests/compiler-hook-tests.lisp
```diff
@@ -157,10 +170,11 @@ RX(-pi/2) 2
 RZ(0.9293531094939129) 2
 RX(pi) 2
 "))))
-    (CL-QUIL::COMPRESS-INSTRUCTIONS-IN-CONTEXT
-     (coerce instructions 'list)
-     (quil::build-nQ-linear-chip 4 :architecture ':cphase)
-     (quil::set-up-compilation-context :qubit-count 4 :simulate t))))
+    (%with-loose-state-prep-compression
+      (CL-QUIL::COMPRESS-INSTRUCTIONS-IN-CONTEXT
```
jmbr: BTW, what's with all the SCREAMING here?

tests/compiler-hook-tests.lisp
```diff
@@ -114,6 +114,18 @@ JUMP @a")))
...
 
+(defmacro %with-loose-state-prep-compression (&body body)
+  `(progn
+     (handler-bind ((quil::state-prep-compression-tolerance-error
+                      (lambda (c)
+                        (with-slots (compilation-tolerance compilation-precision) c
+                          (when (< compilation-tolerance
+                                   compilation-precision
+                                   quil::+double-comparison-threshold-loose+)
+                            (let ((r (find-restart 'continue c)))
```
jmbr: I know @stylewarning frowns upon `alexandria:when-let` but using it here might escape his radar.

src/compressor/compressor.lisp
```diff
@@ -499,22 +507,22 @@ other's."
...
+             (unless (or (null decompiled-instructions)
+                         (and (eql final-wf ':not-simulated)
+                              (eql final-wf-reduced-prep ':not-simulated)))
+               (multiple-value-bind (col precision)
+                   (collinearp final-wf final-wf-reduced-prep)
+                 (unless col
+                   (cerror "continue with possibly incorrect compilation"
+                           'state-prep-compression-tolerance-error
+                           :compilation-tolerance +double-comparison-threshold-strict+
+                           :compilation-precision precision))))))
```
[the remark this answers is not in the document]
notmgsk: IMO `precision = target - actual` and in this case `precision = 1 - abs(x.y)`

kilimanjaro:
> I feel like we're getting the runtime vs test time strictness requirement backwards here. If anything, shouldn't the tests be more strict about tolerance, rather than less? The point of a test is to tell you when you broke something, but if we silently relax the tolerance during testing, then the same program that passes our tests might result in error when a user tries to compile.
> 
> I'm not opposed to relaxing the tolerance in general, but it feels like this approach is bound to result in bug reports (if anyone ever turns on state prep reductions, that is)

I think there are basically two things here. 

First, we have implementations of a number of compilation routines. Some of these involve nontrivial numerical manipulations (e.g. matrix decompositions), and we should have tests for each of these which assert whether they are within some agreed error tolerance (e.g. perhaps for a state prep compiler we expect floating point imprecision of a certain order, and so we have a localized test which explicitly allows for this). I would advocate for adding such tests for the state prep compilers in this PR.

Second, we have the global compressor check. It is hard to know exactly what the right error tolerance should be for this -- it is based on combining a large number of numerical routines with varying degrees of accuracy -- but I think this tolerance is necessarily looser than that of the individual compilation routines. I agree that this should be stricter than the runtime requirements. FWIW I don't even know what the runtime requirements are. If running on our QPU, gate fidelity is the dominant factor in program performance. If running on the QVM, floating point error is much more relevant.

notmgsk:
> I would advocate for adding such tests for the state prep compilers in this PR.

Happy to in a follow-on PR.

tests/compiler-hook-tests.lisp
notmgsk: Historical artefact. It's like an ugly tattoo that you don't want to remove because by now it has some sentimental value.

notmgsk:
@stylewarning requesting your input

src/compressor/compressor.lisp
```diff
@@ -435,6 +435,14 @@ other's."
         ;; otherwise, we're obligated to do full unitary compilation.
         (decompile-instructions-into-full-unitary)))))
 
+(define-condition state-prep-compression-tolerance-error (error)
```
stylewarning: documentation

src/compressor/compressor.lisp
```diff
@@ -435,6 +435,14 @@ other's."
...
+  ((compilation-tolerance :type double-float :initarg :compilation-tolerance :reader state-prep-compression-tolerance-error-tolerance)
```
stylewarning: i would just make these `real` numbers

src/matrix-operations.lisp
```diff
@@ -262,29 +262,21 @@ as needed so that they are the same size."
...
-      (and scalar
-           (let ((norm (sqrt (dot-product vect1 vect2 (cis scalar)))))
-             (double= 1d0 norm))))))
-
+    ;; rather than computing the full norm, this could be replaced by the
+    ;; component-wise calculation (loop ... :always (double= 0d0 (expt ...)))
+    ;; which has the potential to terminate early on a negative result
+    (let ((abs-overlap (abs (dot-product vect1 vect2))))
+      (values (double= 1d0 abs-overlap)
+              (1- abs-overlap)))))
```
stylewarning: `1-` or `(- 1 x)` or `(abs (1- x))`

src/compressor/compressor.lisp
```diff
@@ -435,6 +435,14 @@ other's."
...
+   (compilation-precision :type doubel-float :initarg :compilation-precision :reader state-prep-compression-tolerance-error-precision))
```
notmgsk: @appleby did you really miss this typo

tests/compiler-hook-tests.lisp
```diff
@@ -114,6 +114,18 @@ JUMP @a")))
      (quil::calculate-instructions-2q-depth (coerce (quil::parsed-program-executable-code proc-prog)
                                                     'list)))))
 
+(defmacro %with-loose-state-prep-compression (&body body)
+  `(progn
+     (handler-bind ((quil::state-prep-compression-tolerance-error
+                      (lambda (c)
+                        (with-slots (compilation-tolerance compilation-precision) c
```
stylewarning: i would just call the slot readers directly. u only use them once

tests/compiler-hook-tests.lisp
```diff
@@ -114,6 +114,18 @@ JUMP @a")))
...
+                          (when (< compilation-tolerance
+                                   compilation-precision
```
stylewarning: I would do `(abs comp-prec)` here since it might be negative. `comp-tol` is an absolute value

src/compressor/compressor.lisp
```diff
@@ -435,6 +435,14 @@ other's."
         ;; otherwise, we're obligated to do full unitary compilation.
         (decompile-instructions-into-full-unitary)))))
 
+(define-condition state-prep-compression-tolerance-error (error)
+  ((compilation-tolerance :type real :initarg :compilation-tolerance :reader state-prep-compression-tolerance-error-tolerance)
+   (compilation-precision :type real :initarg :compilation-precision :reader state-prep-compression-tolerance-error-precision))
```
stylewarning: I would note that this can be a negative number.

src/compressor/compressor.lisp
appleby: 😞

Issue #566, opened 2020-01-25
Check that exported symbols have documentation


stylewarning:
(Same with QVM and MAGICL.)

stylewarning:
(I think we should make a new system for this.)

notmgsk:
A new system?

Something like a linter would be cool.

stylewarning:
See also: https://github.com/rigetti/quilc/issues/40

jmbr:
I'm still advocating for a Rigetti-specific version of the lisp-critic.

rigettizach:
I wrote a new program to dig up this kind of info. Here's a report it can produce, with additional remarks added manually in *italics*.

Many of these are easy fixes. Two exceptions stand out:

* adt:defdata generates new constants and structures, and needs an update to attach documentation to those
* the protocol-less `<class>-<slot>` accessors need documented protocols, with defgenerics

A final issue to consider: sometimes external symbols are just syntax, and do not have a natural place to hang documentation. Consider `it` in anaphoric libraries, or `otherwise` in CL. It would take some extra thought to exclude them from a report on problematic symbols.

The report:

# cl-quil

## Class

* [circuit-definition](https://github.com/rigetti/quilc/blob/master/src/ast.lisp#L456)
* [controlled-gate](https://github.com/rigetti/quilc/blob/master/src/gates.lisp#L221)
* [dagger-gate](https://github.com/rigetti/quilc/blob/master/src/gates.lisp#L270)
* [forked-gate](https://github.com/rigetti/quilc/blob/master/src/gates.lisp#L247)

## Compiler-Macro

* [inst](https://github.com/rigetti/quilc/blob/master/src/define-compiler.lisp#L897)
* [inst*](https://github.com/rigetti/quilc/blob/master/src/define-compiler.lisp#L908)

## Constant

*These are self-evident, to me, but a missing docstring is a missing
docstring.*

* [2pi](https://github.com/rigetti/quilc/blob/master/src/utilities.lisp#L230)
* [4pi](https://github.com/rigetti/quilc/blob/master/src/utilities.lisp#L231)
* [-pi](https://github.com/rigetti/quilc/blob/master/src/utilities.lisp#L227)
* [pi/2](https://github.com/rigetti/quilc/blob/master/src/utilities.lisp#L228)
* [-pi/2](https://github.com/rigetti/quilc/blob/master/src/utilities.lisp#L229)

*These arise from adt:defdata.*

[...]

PR #594, opened 2020-02-28
Add some more docstrings

rigettizach:
Cleanups based on the report in #566

### DIFF
--- a/src/ast.lisp
+++ b/src/ast.lisp
@@ -35,6 +35,7 @@
   (descriptor nil :type (or null memory-descriptor)))
 
 (defun memory-ref= (a b)
+  "Do the memory refs A and B represent the same memory ref?"
   (check-type a memory-ref)
   (check-type b memory-ref)
   (and (string= (memory-ref-name a) (memory-ref-name b))
@@ -191,6 +192,7 @@ Keys are tested with EQ."
 The keys are typically INSTRUCTION instances and associated values are STRINGs.")
 
 (defun comment (x)
+  "Return the comment associated with X."
   (values (gethash x **comments**)))
 
 (defun (setf comment) (comment-string x)
@@ -531,6 +533,7 @@ This replicates some of the behavior of CL-QUIL.CLIFFORD::PAULI, but it extends 
 (defmethod mnemonic ((inst pragma)) (values "PRAGMA" (class-name (class-of inst))))
 
 (defun make-pragma (words &optional freeform)
+  "Return a pragma, possibly specialized based on WORDS and FREEFORM."
   (assert (and (listp words)
                (not (endp words))))
   (assert (stringp (first words))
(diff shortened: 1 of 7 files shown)


src/utilities.lisp
```diff
@@ -223,12 +223,14 @@ contains the bits of INTEGER. See http://www.cliki.net/ROTATE-BYTE"
...
-(defconstant 2pi   (* 2 pi))
-(defconstant 4pi   (* 4 pi))
+(macrolet ((define-self-documenting-constant (symbol value)
+             `(defconstant ,symbol ,value ,(string-downcase symbol))))
+  (define-self-documenting-constant pi    (coerce cl:pi 'double-float))
+  (define-self-documenting-constant -pi   (- pi))
+  (define-self-documenting-constant pi/2  (/ pi 2))
+  (define-self-documenting-constant -pi/2 (/ pi -2))
+  (define-self-documenting-constant 2pi   (* 2 pi))
+  (define-self-documenting-constant 4pi   (* 4 pi)))
```
appleby: :-)

src/chip/chip-specification.lisp
```diff
@@ -512,6 +512,7 @@ Compilers are listed in descending precedence.")
 ;; 6 --5-- 5 --4-- 4
 ;; all qubits Xhalves-type, all links CZ-type
 (defun build-8Q-chip (&key (architecture ':cz))
+  "Return an example 8Q chip."
```
appleby:
```suggestion
  "Return an example 8Q square chip."
```

src/ast.lisp
```diff
@@ -531,6 +533,7 @@ This replicates some of the behavior of CL-QUIL.CLIFFORD::PAULI, but it extends
 (defmethod mnemonic ((inst pragma)) (values "PRAGMA" (class-name (class-of inst))))
 
 (defun make-pragma (words &optional freeform)
+  "Return a pragma, possibly specialized based on WORDS and FREEFORM."
```
jmbr:
```suggestion
  "Return a pragma based on WORDS, possibly specialized on FREEFORM."
```

src/utilities.lisp
```diff
@@ -223,12 +223,14 @@ contains the bits of INTEGER. See http://www.cliki.net/ROTATE-BYTE"
     #+allegro excl:*infinity-double*
     #-(or ccl ecl sbcl allegro) (error "double-float-positive-infinity not available."))
 
-(defconstant pi    (coerce cl:pi 'double-float))
-(defconstant -pi   (- pi))
-(defconstant pi/2  (/ pi 2))
-(defconstant -pi/2 (/ pi -2))
-(defconstant 2pi   (* 2 pi))
-(defconstant 4pi   (* 4 pi))
+(macrolet ((define-self-documenting-constant (symbol value)
```
jmbr: I love this.

src/ast.lisp
rigettizach: I don't think that's quite an accurate rephrasing, since `make-pragma` will happily return a pragma that isn't influenced by the arguments at all (except as far as they become properties of the returned object). For example, `(make-pragma (list "blorp"))`. If anything, it's only really possibly specialized on `(first words)`, but that seems like an implementation detail.

jmbr: Thanks for clarifying that.

src/utilities.lisp
stylewarning: classic @rigettizach

PR #591, opened 2020-02-25
Fix rz/rx string reduction

notmgsk:
Closes #592.

PR #574, opened 2020-02-03
Update Magicl bindings for Magicl NG

colescott:
This is to update all magicl bindings to work with rigetti/magicl#63.

Benchmarks need to be done prior to merging to make sure there is not a significant performance hit.

PR #598, opened 2020-03-02
Bump version to 1.17.0

notmgsk:
# Changelog

# Bugfixes

* Fix support for wildcard 1Q operators in ISA (@notmgsk, #571).
* Fix incorrect compiler selection, leading to long strings of RX and RZ gates not being collapsed (@notmgsk, #591).

## Behind-the-scenes

* Docstrings (@rigettizach, #594).
* Update magicl bindings and magicl version dependency (@colescott, #574).
* Improve collinearity comparisons, and allow tests to reduce the comparison threshold (@notmgsk, #593).


notmgsk:
> don't forget to bump the `quicklisp_version` in the `Dockerfile` here as well once the rigetti quicklisp repo is setup

I will update those, but the quicklisp dist won't be ready for a day or two (maybe longer) since I have to wait for IT to provision stuff.

appleby:
> > don't forget to bump the `quicklisp_version` in the `Dockerfile` here as well once the rigetti quicklisp repo is setup
> 
> I will update those, but the quicklisp dist won't be ready for a day or two (maybe longer) since I have to wait for IT to provision stuff.

oic. So this PR is just to bump the version.txt, but we won't actually do the release for a couple of days until the new quicklisp repo is setup?

notmgsk:
> > > don't forget to bump the `quicklisp_version` in the `Dockerfile` here as well once the rigetti quicklisp repo is setup
> > 
> > 
> > I will update those, but the quicklisp dist won't be ready for a day or two (maybe longer) since I have to wait for IT to provision stuff.
> 
> oic. So this PR is just to bump the version.txt, but we won't actually do the release for a couple of days until the new quicklisp repo is setup?

No, we will release. It's not a blocker, we can just trigger the deploys manually I think (so they're not dependent on the tests passing).

appleby:
> > > > don't forget to bump the `quicklisp_version` in the `Dockerfile` here as well once the rigetti quicklisp repo is setup
> > > 
> > > 
> > > I will update those, but the quicklisp dist won't be ready for a day or two (maybe longer) since I have to wait for IT to provision stuff.
> > 
> > 
> > oic. So this PR is just to bump the version.txt, but we won't actually do the release for a couple of days until the new quicklisp repo is setup?
> 
> No, we will release. It's not a blocker, we can just trigger the deploys manually I think (so they're not dependent on the tests passing).

I might be misunderstanding how the Docker build works, but if the docker image is based on a quicklisp version with the old magicl, won't the quilc it produces be broken (and therefore the QMIs that upgrade to that quilc / qvm docker image)?

notmgsk:
> > > > > don't forget to bump the `quicklisp_version` in the `Dockerfile` here as well once the rigetti quicklisp repo is setup
> > > > 
> > > > 
> > > > I will update those, but the quicklisp dist won't be ready for a day or two (maybe longer) since I have to wait for IT to provision stuff.
> > > 
> > > 
> > > oic. So this PR is just to bump the version.txt, but we won't actually do the release for a couple of days until the new quicklisp repo is setup?
> > 
> > 
> > No, we will release. It's not a blocker, we can just trigger the deploys manually I think (so they're not dependent on the tests passing).
> 
> I might be misunderstanding how the Docker build works, but if the docker image is based on a quicklisp version with the old magicl, won't the quilc it produces be broken (and therefore the QMIs that upgrade to that quilc / qvm docker image)?

Yeah, you're right (as usual). I'll try to get the quicklisp dist available `ASAP`.

appleby:
I guess the other option is that we could "vendor" magicl similar to what we already do for rpcq and qvm, rather than building off the version that is in quicklisp.

Maybe that would cause conflicts or something since both qvm and quilc depend on magicl, and the quilc Dockerfile depends on qvm's docker image while the qvm Dockerfile also depends on quilc's docker image... maybe it could be "imported" in just one of the Dockerfiles.

Unclear if this would be an improvement over setting up our own quicklisp repo :)

Dockerfile
```diff
@@ -1,7 +1,7 @@
 # specify the dependency versions (can be overriden with --build_arg)
 ARG rpcq_version=2.7.3
```
notmgsk: @appleby do we need to update rpcq here?

appleby: We could, but I don't think it's required.

PR #609, opened 2020-03-13
COMPILER-HOOK *does not* mutate and *does* compress qubits

### LINKED ISSUE #54: Should COMPILER-HOOK mutate its input?
The argument of `compiler-hook` mutates currently as a part of the compilation process. Should it?

### LINKED ISSUE #487: Compress logical qubits before compilation
Right now the single instruction program `H 9` will not compile to an 8Q lattice. This could be easily remedied by applying the `compress-qubits` transform right after parsing in `quilc`. The only caveat is that certain pragmas would preclude such compression, e.g. `PRAGMA PRESERVE_BLOCK`

notmgsk:
Closes #487, closes #54.

### DIFF
--- a/src/compilation-methods.lisp
+++ b/src/compilation-methods.lisp
@@ -149,14 +149,43 @@
 ;; TODO: deal with classical control and basic-blocks
 (defun compiler-hook (parsed-program
                       chip-specification
-                      &key (protoquil nil)
-                        (rewiring-type (prog-initial-rewiring-heuristic parsed-program chip-specification)))
+                      &key
+                        (protoquil nil)
+                        (rewiring-type (prog-initial-rewiring-heuristic parsed-program chip-specification))
+                        (destructive nil))
   "Runs a full compiler pass on a parsed-program object.
+
+Arguments:
+
+  - PARSED-PROGRAM: The parsed program to compile
+
+  - CHIP-SPECIFICATION: The chip specification describing the target chip
+
+Keyword arguments:
+
+  - PROTOQUIL: Whether the input and output programs should conform to protoquil restrictions
+
+  - REWIRING-TYPE: The scheme by which logical qubits are mapped (rewired) to physical qubits
+
+  - DESTRUCTIVE: Default NIL. If T, the input program is mutated and after compilation contains the compiled program. Otherwise, a copy is made of the input program and the compiled copy is returned. Note that for large programs you may prefer to mutate the original program rather than waste time copying it. For example, the server interface will mutate the program since it has no use for the unmolested original program after compilation.
 
 Returns a value list: (processed-program, of type parsed-program
                        topological-swaps, of type integer
                        unpreserved-block-duration, of type real)"
   (format-noise "COMPILER-HOOK: entrance.")
+
+  (when destructive
+    (setf parsed-program (copy-instance parsed-program)))
+
+  ;; Technically this could be fused with the above, but remains
+  ;; separate for clarity. We should not compress if the rewiring type
+  ;; is naive, or the program uses more qubits than are available on
+  ;; the chip (duh), or if the program has blocks of preserved qubits.
+  (when (and (not (eql ':naive rewiring-type))
+             (> (qubits-needed parsed-program)
+                (length (chip-spec-live-qubits chip-specification)))
+             (not (parsed-program-has-preserve-blocks-p parsed-program)))
+    (setf parsed-program (compress-qubits parsed-program)))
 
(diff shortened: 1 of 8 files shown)


src/pragmas.lisp
```diff
@@ -116,3 +116,12 @@ Expected syntax: PRAGMA INITIAL_REWIRING [NAIVE|PARTIAL|GREEDY|RANDOM]")
...
   (:display-string
    (prin1-to-string (symbol-name rewiring-type))))
+
+(defun parsed-program-has-pragma-p (parsed-program pragma-type)
+  "Return T if PARSED-PROGRAM's executable code contains a pragma of type PRAGMA-TYPE."
+  (some (a:rcurry #'typep pragma-type)
+        (parsed-program-executable-code parsed-program)))
+
+(defun parsed-program-has-preserve-blocks-p (parsed-program)
+  "Return T if PARSED-PROGRAM's executable code contains a program of type PRAGMA-PRESERVE-BLOCK."
```
appleby:
```suggestion
  "Return T if PARSED-PROGRAM's executable code contains a pragma of type PRAGMA-PRESERVE-BLOCK."
```

appleby:
I guess the fact that we no longer mutate is technically a breaking change. We should make a note of it in the release notes.

src/compilation-methods.lisp
```diff
@@ -158,6 +158,14 @@ Returns a value list: (processed-program, of type parsed-program
                        unpreserved-block-duration, of type real)"
   (format-noise "COMPILER-HOOK: entrance.")
 
+  (setf parsed-program (copy-instance parsed-program))
```
stylewarning: I propose adding an option to `compiler-hook`:

`&key destructive`

and

```
(unless destructive
  (setf parsed-program (copy-instance parsed-program)))
```

I will not be happy when I find that compiling a 10,000 gate program is spending time deep copying an AST and I have no way around it.

src/compilation-methods.lisp
```diff
@@ -158,6 +158,14 @@ Returns a value list: (processed-program, of type parsed-program
...
+
+  (when (and (not (eql ':naive rewiring-type))
```
stylewarning: Add a note that technically this step and the copying step could be fused, but in the interest of no premature optimization, it wasn't done.

src/package.lisp
```diff
@@ -466,6 +466,8 @@
    #:pragma-matrix-entries              ; READER
    #:pragma-qubit-arguments             ; READER
    #:pragma-operator-name               ; READER
+   #:parsed-program-has-pragma-p        ; FUNCTION
```
stylewarning: why are these exported?

src/pragmas.lisp
```diff
@@ -116,3 +116,12 @@ Expected syntax: PRAGMA INITIAL_REWIRING [NAIVE|PARTIAL|GREEDY|RANDOM]")
             (error "Invalid PRAGMA INITIAL_REWIRING: ~A" rewiring-type-string)))))
   (:display-string
    (prin1-to-string (symbol-name rewiring-type))))
+
+(defun parsed-program-has-pragma-p (parsed-program pragma-type)
```
stylewarning: I'd expect this function to be unary, searching for any kind of pragma. Maybe make the pragma-type optional and default it to the root pragma class?

src/compilation-methods.lisp
notmgsk: Maybe there should exist three functions: `compile`, `compile!`, and `compiler-hook`. `compiler-hook` looks mostly as it is now, and the others are non-mutating/mutating entry-points?

notmgsk: i.e.
```
(defun compile (program &rest rest) (apply #'compiler-hook (copy-instance program) rest))
```

stylewarning: That would also be fine to me. &rest is sad because we don't know the kw's anymore in slime :(

src/package.lisp
notmgsk: Why not? I think they're useful. Happy to `C-_` if you disagree, however.

src/compilation-methods.lisp
appleby: Are we ok with shadowing `cl:compile`?  I guess we only use it in a handful of places...

notmgsk: Eh, I went the kwarg route.

notmgsk:
@stylewarning please take another look.

src/package.lisp
stylewarning: "I think they're useful" is not good enough!!

Issue #606, opened 2020-03-05
Concatenate PARSED-PROGRAM objects


jmbr:
Implement a function to splice `parsed-program` objects together in a reasonable way.

PR #612, opened 2020-03-19
WIP: add COMBINE-PROGRAMS and QUIL-TYPE=

stylewarning:
WIP. Addresses #606.

### DIFF
--- a/src/ast.lisp
+++ b/src/ast.lisp
@@ -1678,6 +1678,75 @@ For example,
         (%print-indented-string (with-output-to-string (s)
                                   (print-parsed-program parsed-program s))
                                 stream)))))
+(defun combine-programs (a b &key (if-memory-definitions-conflict ':error))
+  "Given two PARSED-PROGRAMS A and B, combine them into a new program object.
+
+In the event there are conflicting memory definition names, there are the following available policies that can be supplied for IF-MEMORY-DEFINITIONS-CONFLICT:
+
+    - :ERROR: Simply signal an error.
+    - :PREFER-LARGER: Keep the larger of the two memories.
+    - :PREFER-A: Prefer the definitions for A. (Note that this may be unsafe!)
+    - :PREFER-B: Prefer the definitions for B. (Note that this may be unsafe!)
+"
+  ;; gate def names might overlap
+  ;;     - perhaps the names clobber
+  ;;     - otherwise just a union
+  ;;
+  ;; circ defs simlar
+  ;;
+  ;; memory defs
+  ;;     - a little more sensitive, esp. with sizes
+  ;;
+  ;;     - ERROR
+  ;;     - PREFER A
+  ;;     - PREFER B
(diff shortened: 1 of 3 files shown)


src/ast.lisp
```diff
@@ -1627,6 +1627,76 @@ For example,
...
+                      (unless (quil-type= (memory-descriptor-type memdef) (memory-descriptor-type found))
+                        (error "Trying to combine programs with memories called ~S which ~
+                                have different types."
+                               (memory-descriptor-name memdef)))
+                      ;;
+                      (ecase if-memory-definitions-conflict
+                        (:PREFER-LARGER
+                         (when (> (memory-descriptor-length memdef)
+                                  (memory-descriptor-length found))
+                           (rplaca found? memdef)))
```
rigettizach: I'm surprised to see `rplaca` in new code.

appleby: you'll have to pry `rplaca` from [his cold dead hands](https://github.com/rigetti/quilc/pull/569#discussion_r373244735) :-)

rigettizach: What an odd position to take in collaborative construction of new Common Lisp programs.

stylewarning: Just to be clear, this code was literally collaboratively constructed. It was constructed on Zoom with pair programming, and it served in part to teach various aspects of Common Lisp to the person on the line. Including old things.

RPLACA was used to show that a _function_ exists which mutates one of Lisp's fundamental data types: the cons. It was also an exercise in understanding the return value of MEMBER, and how one can use MEMBER to destructively substitute.

(SETF CAR) and (SETF FIRST) were both discussed and shown, but in order to hide the "magic" of a SETF expansion, which to the naked eye doesn't look like an ordinary Lisp function, RPLACA was used.

For whatever it's worth, I don't think (SETF CAR) is unilaterally better than RPLACA. RPLACA has a funny but relatively descriptive name: RePLACe cAr. It's a function. And it's standard. And it's old.

rigettizach: I don't think it's worth tripping up every reviewer that is already familiar with typical modern Common Lisp.

stylewarning: Does RPLACA actually trip up people? Maybe I am just a caveman. :(

rigettizach: If nothing else, it reduces the clarity of local intent. There's no distinction between a list/tree operation as there is with first/cdr. (This is not a call for implementing `rplafi` or `rplare` to restore clarity.)

But it also is contrary to a few decades of advice about preferring setf in this context. 

It has some didactic value ("Here's the low-level thing you could use instead of `(setf car/first)`") but the lesson isn't very deep, as there are no parallels for arrays or hash tables or other structured objects - setf is the one and only option in standard code.

I don't think anyone is confused by what it does, but it is confusing to see it used in 2020.

stylewarning:
Rename `combine-` to `conatenate-`

PR #613, opened 2020-03-19
Remove extra sqrt in fidelity calculation

notmgsk:


src/addresser/logical-schedule.lisp
```diff
@@ -596,8 +596,7 @@ mapping instructions to their tags. "
              (expt (log fidelity) 2)))))
     (let ((running-fidelity 0d0))
       (dolist (instr (lscheduler-first-instrs lschedule))
-        (unless (gethash instr (lscheduler-earlier-instrs lschedule))
```
notmgsk: I think this is redundant -- `lscheduler-first-instrs` should give the earliest instructions (i.e. `lscheduler-earlier-instrs` should always be `nil`).

jmbr:
@notmgsk It might be a good idea to document the fact that this function relies on `exp(-sqrt(log(x)^2 + log(y)^2))` being approximately equal to `min(x, y)` for `x` and `y` in the unit interval. That might save the next person some time figuring things out.

src/addresser/logical-schedule.lisp
```diff
@@ -607,7 +607,7 @@ mapping instructions to their tags. "
 (defun lscheduler-calculate-fidelity (lschedule chip-spec)
   (multiple-value-bind (max-value value-hash)
```
jmbr:
```suggestion
(defun lscheduler-calculate-fidelity (lschedule chip-spec)
  "Calculate fidelity as the minimum fidelity of the individual instructions.
  
  This relies on the fact that the function $\exp\{-\sqrt{\log(x)^2 + \log(y)^2}\}$ is approximately equal to $\min\{x, y\}$ for $x, y \in (0, 1]$."
  (multiple-value-bind (max-value value-hash)
```

notmgsk:
Thanks @jmbr

PR #616, opened 2020-03-23
Add PRINT-OBJECT for PARSED-PROGRAMs which prints the program

notmgsk:
Off by default.

How do you feel about this?

### DIFF
--- a/src/ast.lisp
+++ b/src/ast.lisp
@@ -1661,8 +1661,8 @@ For example,
                (parsed-program-executable-code parsed-program)))
     pp))
 
-(defvar *print-parsed-program-objects* nil
-  "When T, PARSED-PROGRAM objects are printed with PRINT-PARSED-PROGRAM. Otherwise the standard object printer is used.")
+(defvar *print-parsed-program-text* nil
+  "When T, PRINT-OBJECT on a PARSED-PROGRAM will include the program text. Otherwise, only the number of instructions is printed.")
 
 (defun %print-indented-string (string stream &key (indentation "    "))
   (loop :with last := (1- (length string))
@@ -1677,7 +1677,7 @@ For example,
 (defmethod print-object ((parsed-program parsed-program) stream)
   (print-unreadable-object (parsed-program stream :type t :identity t)
     (format stream "of ~D instructions" (length (parsed-program-executable-code parsed-program)))
-    (when *print-parsed-program-objects*
+    (when *print-parsed-program-text*
       (terpri stream)
       (%print-indented-string (with-output-to-string (s)
                                 (print-parsed-program parsed-program s))
--- a/src/ast.lisp
+++ b/src/ast.lisp
@@ -1664,17 +1664,24 @@ For example,
 (defvar *print-parsed-program-objects* nil
   "When T, PARSED-PROGRAM objects are printed with PRINT-PARSED-PROGRAM. Otherwise the standard object printer is used.")
 
+(defun %print-indented-string (string stream &key (indentation "    "))
+  (loop :with last := (1- (length string))
+        :for c :across string
+        :for i :upto last
+        :when (zerop i)
+          :do (write-string indentation stream)
+        :do (write-char c stream)
+        :when (and (eql c #\Newline) (/= i last))
+          :do (write-string indentation stream)))
+
 (defmethod print-object ((parsed-program parsed-program) stream)
   (print-unreadable-object (parsed-program stream :type t :identity t)
     (format stream "of ~D instructions" (length (parsed-program-executable-code parsed-program)))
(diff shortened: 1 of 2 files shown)


src/ast.lisp
```diff
@@ -1627,6 +1627,14 @@ For example,
    :executable-code #())
   (:documentation "A representation of a parsed Quil program, in which instructions have been duly sorted into their various categories (e.g. definitions vs executable code), and internal references have been resolved."))
 
+(defvar *print-parsed-program-objects* nil
+  "When T, PARSED-PROGRAM objects are printed with PRINT-PARSED-PROGRAM. Otherwise the standard object printer is used.")
+
+(defmethod print-object ((parsed-program parsed-program) stream)
+  (if *print-parsed-program-objects*
+      (print-parsed-program parsed-program stream)
```
appleby: We should probably wrap this in a `print-unreadable-object`.

stylewarning: i agree with this;

Ideally it would print-unreadable, with ~% and indented contents

src/ast.lisp
```diff
@@ -1627,6 +1627,14 @@ For example,
...
+      (call-next-method)))
```
stylewarning: I would say that we should have a print-object _regardless_ which prints the # of instructions. Why not?

notmgsk: Indeed, why not.

src/ast.lisp
```diff
@@ -1627,6 +1627,21 @@ For example,
...
+  (print-unreadable-object (parsed-program stream :type t :identity t)
+    (format stream "of ~D instructions" (length (parsed-program-executable-code parsed-program)))
+    (when *print-parsed-program-objects*
+      (terpri stream)
+      (dolist (line (split-sequence:split-sequence
+                     #\Newline
+                     (with-output-to-string (s)
+                       (print-parsed-program parsed-program s))))
+        (write-string "    " stream)
+        (write-line line stream)))))
```
appleby: The `split-sequence` here is unfortunate. I started down the path of adding `prefix` arg to all of the `print-*` functions, but winds it's way down to `print-instruction-generic` which has a *lot* of methods, and it just seemed like a lot of churn for this special case, so I backtracked to gratuitous consing here.

Another option might be to hook into the pretty printer and use `print-logical-block` + `pprint-indent` or some such.

appleby:
rebased and resolved conflicts

src/ast.lisp
stylewarning: Another lazy option: Just print chars one at a time. 😀

src/ast.lisp
```diff
@@ -1661,6 +1661,21 @@ For example,
                (parsed-program-executable-code parsed-program)))
     pp))
 
+(defvar *print-parsed-program-objects* nil
```
stylewarning: Export this symbol

src/ast.lisp
stylewarning: (don't let this block this PR tho)

appleby: No, I think it's worth it and probably not much more code than the current approach. I have another PR I want to wrap up first, but I will circle back to this, a.k.a `DO NOT MERGE`

appleby: Due to mounting internal pressure, I retract my `DO NOT MERGE` and replace it with `MERGE WHEN AND AS YOU PLEASE`

src/ast.lisp
appleby: Exported. I also renamed it to `*PRINT-PARSED-PROGRAM-TEXT*` since we now provide a custom printer either way, but only print the program text when this var is set.

src/ast.lisp
appleby: Latest version includes a `%PRINT-INDENTED-STRING` utilitito to avoid the need to `SPLIT-SEQUENCE`. It still has to allocate the whole string up front of course.

Is better than the `SPLIT-SEQUENCE` version? Debatable. Maybe it's time for the BIG GUNZ.

src/ast.lisp
```diff
@@ -1661,6 +1661,28 @@ For example,
...
 
+(defvar *print-parsed-program-text* nil
+  "When T, PRINT-OBJECT on a PARSED-PROGRAM will include the program text. Otherwise, only the number of instructions is printed.")
+
+(defun %print-indented-string (string stream &key (indentation "    "))
+  (loop :with last := (1- (length string))
+        :for c :across string
+        :for i :upto last
+        :when (zerop i)
+          :do (write-string indentation stream)
```
appleby: This could be simplified by requiring `string` to start with a newline, which would be easy to enforce below, but made this seem less useful as a general utility. Not that anyone else will ever want to call it or remember that it's here if they do.

Issue #531, opened 2020-01-03
A* swap search takes "forever" when compiling some benchmarking files


appleby:
Compiling the following program takes a racoon's age (maybe infiniloop):

```lisp
(let ((quil::*addresser-swap-search-type* :a*))
  (time (quil::compiler-hook (quil::read-quil-file #P"benchmarking/quil-rewiring/0005q-0000160.quil")
			     (quil::build-nQ-linear-chip 20))))
```

I noticed this while looking into another `CL-QUIL-BENCHMARKING` issue (#346). I'm not certain whether the search is infinite looping or just taking a really long time. Definitely doesn't finish within ~5 mins, whereas compiling with the default `:GREEDY-QUBIT` swap search strategy takes ~1.5 seconds.

```lisp
(time (quil::compiler-hook (quil::read-quil-file #P"benchmarking/quil-rewiring/0005q-0000160.quil")
		           (quil::build-nQ-linear-chip 20)))
Evaluation took:
  1.621 seconds of real time
...
```

The same issue occurs when compiling other high-gate-depth files under benchmarking/quil-rewiring, like `0008q-0000418-qaoa-8-5.quil` and `0020q-0000654-johannes.quil`.

Git bisect points to the swap fidelity commit (30882c05). Prior to 30882c05, compiling with `:A*` swap search takes about the same time as `:GREEDY-QUBIT` does after:

```lisp
;; with git HEAD pointing to 942b107d
(let ((quil::*addresser-swap-search-type* :a*))
  (time (quil::compiler-hook (quil::read-quil-file #P"benchmarking/quil-rewiring/0005q-0000160.quil")
                             (quil::build-nQ-linear-chip 20))))
Evaluation took:
  1.290 seconds of real time
....
```

appleby:
`*COMPILER-NOISE-STREAM*` output appears to be stuck in loop applying `EULER-YZY-COMPILER H 2 -> RY-TO-XZX RY(-3*pi/4) 2 -> RY-TO-XZX RY(-pi/4) 2 -> EULER-YZY-COMPILER H 2 -> ...`

```
DEQUEUE-LOGICAL-TO-PHYSICAL: entering dequeueing phase.
APPLY-TRANSLATION-COMPILERS: Applying #<COMPILER EULER-YZY-COMPILER {10057303FB}> to H 2.
    RY(-3*pi/4) 2
    RZ(-pi) 2
    RY(-pi/4) 2
[...]

PR #617, opened 2020-03-25
Addresser Refactor and Cleanup

kilimanjaro:
The following is intended to clean up and clarify the basic addresser API. It is not intended to change any behavior.

## Background and Motivation

The quilc addresser is responsible for translating source programs, which may involve arbitrary "logical qubits" interacting in arbitrary ways (e.g. 5q gates), to programs which conform to the topological constraints of the hardware (e.g. using only physical qubits, and with 2Q gates only on live hardware links). At the core of the addresser are two data structures:
-  a "logical to physical rewiring", which determines what physical qubit a logical qubit is presently assigned to
- a "logical schedule", which represents the source instructions as a partially ordered set, with resource conflicts determining whether one instruction may occur before another.
The addresser walks the logical scheduler in topological order, updating the logical to physical rewiring as need be, and emitting instructions (to a "chip schedule").

The addresser handles gates on > 2 qubits by first translating them with `APPLY-TRANSLATION-COMPILERS`. This is relatively straightforward. 

The main difficulty in addressing is managing the logical to physical rewiring. At any given moment, logical qubits might i) not have an assigment to a physical qubit (and hence an assignment must be made), or ii) be assigned to physical qubits which are not adjacent (in which case something must be done in order for a gate involving them to be executed). 

The main technique for dealing with the second problem is to introduce SWAP operations, which can be used to shuffle the quantum state around in a way that reflects an update to the logical to physical assignment. This is a challenging task, because it can incur a sizeable overhead in the 2Q cost of the compiled program. The quilc addresser makes a lot of effort to assign qubits and select swaps intelligently.

The approach taken by quilc is to apply a few heuristics, which can be tuned or adapted as we see fit. These are split along two axes. The first axis is how one measures the "cost" of an instruction or set of instructions. The second is how one finds "good" swap operations. 

*One of the main goals of this PR is to clarify the relationship between these heuristics, and to allow for the (relatively easy) addition of new ones.*

## Cost Heuristics

Right now there are two sorts of costs: a duration-based cost (`temporal-addresser.lisp`) and a fidelity-based cost (`fidelity-addreser.lisp`). New cost heuristics may be implemented by subclassing `ADDRESSER-STATE` (`addresser-state.lisp`), and then implementing methods on a few associated generics:
- `COST-FUNCTION`, which computes the cost of the current logical schedule.
- `WEIGHTED-FUTURE-GATES`, which constructs a mapping from gate in the current logical schedule to a value indicating "how far" into the future they will occur
- `BUILD-WORST-COST`, which is basically a stand-in for `most-positive-fixnum` if the heuristic uses a custom or compound cost value.

In practice, both the temporal and fidelity addresser go a bit further than this (e.g. with non-real cost values), but the above reflects the bare minimum needed to add a new cost heuristic.

As a note, the reason for `WEIGHTED-FUTURE-GATES` is to allow parts of the addresser to effectively do a look-ahead. For example, in swap selection, choosing swaps which look good "right now" might be a poor strategy if the next few instructions force additional swaps to be inserted. As it stands, both the temporal and fidelity cost functions incorporate this look-ahead information.

## Search Heuristics

When the addresser decides that swaps are needed, there are usually a large number of candidates. In order to select the best, it now relies on the following two generics
- `SELECT-SWAPS-FOR-GATES`, used by the addresser to select swaps so that it may assign logical gates to physical qubits, and
- `SELECT-SWAPS-FOR-REWIRING`, used by the addresser to select swaps in order to update the current rewiring to a desired rewiring.

These represent the two main pieces of functionality needed to add a new search heuristic to the addresser. Currently they dispatch on the search type (a keyword symbol) and take the full addresser state as an argument. 

We currently have three implementations (in `path-heuristic.lisp`, `qubit-heuristic.lisp`, and `astar-rewiring-search.lisp`). Defaults are specified by this snippet from `addresser-common.lisp`:
```
;;; Search routines are implemented in
;;;   - astar-rewiring-search.lisp (for A*)
;;;   - qubit-heuristic.lisp (for GREEDY-QUBIT)
;;;   - path-heuristic (for GREEDY-PATH)
(deftype addresser-search-type () '(member :a* :greedy-qubit :greedy-path))

(defvar *addresser-swap-search-type* ':greedy-qubit
  "The type of swap search the addresser should use.")

(defvar *addresser-move-to-rewiring-swap-search-type* ':a*
  "The type of swap search the addresser should use when doing move-to-rewiring.")
```

## Additional Customization

I added an additional flag `*ADDRESSER-USE-1Q-QUEUES*` which allows one (when true) to stash 1Q gates away so that they will not influence qubit allocation, swap selection, and so on. Previously this was hardcoded as default behavior for the temporal cost heuristic but not for the fidelity cost heuristic (this was because the fidelity heuristic can actually place 1Q gates in intelligent manner, on good qubits, whereas the duration heuristic cannot, and so for the duration heuristic all effort is made to not get in the way of 2Q addressing).


src/addresser/addresser-common.lisp
```diff
@@ -658,13 +697,8 @@ If DRY-RUN, this returns T as soon as it finds an instruction it can handle."
               (setf best-logical-qubit logical-qubit
                     best-physical-qubit physical-qubit
                     best-cost cost))))
-
-        (cond
-          (best-logical-qubit
-           (rewiring-assign working-l2p best-logical-qubit best-physical-qubit)
-           t)
-          (nil
-           nil))))))
```
appleby: what happened to this charming lil `cond`?

appleby: I guess the point is that is that some `logical-qubit` must have had a cost less than the worst possible cost, and therefore `best-logical-qubit` should never be `nil`?

src/addresser/astar-rewiring-search.lisp
```diff
@@ -272,3 +272,36 @@ should stop with the given rewiring and false otherwise."
...
+  (with-slots (chip-spec chip-sched) addresser-state
+    (flet ((cost-function (rewiring &key instr (gate-weights gates-in-waiting))
+             (declare (ignore instr))
+             (let ((modified-state (copy-instance addresser-state)))
+               ;; TODO
+               (setf (addresser-state-working-l2p addresser-state) rewiring)
+               (* *addresser-a*-swap-search-heuristic-scale*
+                  (cost-flatten (cost-function modified-state :gate-weights gate-weights)))))
+           (done-function (rewiring2)
+             (prog2
```
appleby: \*steve irwin voice\*

a rare `prog2` in the wild. Isn't she beauuuutiful?

src/addresser/addresser-common.lisp
kilimanjaro: Yes. Although I suppose if the worst possible cost is not actually the worst possible cost, then this would behave differently than before.

kilimanjaro: I might be overthinking it but I decided to just hit it with `EXTREMUM` instead.

kilimanjaro:
calling in the big guns @ecpeterson

src/compilation-methods.lisp
```diff
@@ -252,17 +252,18 @@ Returns a value list: (processed-program, of type parsed-program
...
-                                  :chip-spec chip-specification
-                                  :initial-l2p (if registrant
-                                                   (basic-block-in-rewiring blk)
-                                                   initial-rewiring))
-                 (coerce (basic-block-code blk) 'list)
-                 :initial-rewiring (if registrant
-                                       (basic-block-in-rewiring blk)
-                                       initial-rewiring)
-                 :use-free-swaps (null registrant))
+               (let ((*addresser-use-1q-queues* (eq *default-addresser-state-class* 'temporal-addresser-state)))
```
notmgsk, line 255: Leave a comment about why you're doing this

src/addresser/fidelity-addresser.lisp
```diff
@@ -172,58 +175,48 @@
...
                                :bump-value #'weight-bumper
-                               :test-values #'max)
+                               :combine-values #'max)
       (declare (ignore max-value))
       value-hash)))
 
-(defmethod select-and-embed-a-permutation ((state fidelity-addresser-state) rewiring-tried)
+;;; We also specialize this, to set up some state before this stage of swap
+;;; selection.
+(defmethod select-and-embed-a-permutation ((state temporal-addresser-state) rewirings-tried)
```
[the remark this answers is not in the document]
kilimanjaro: typo

src/addresser/temporal-addresser.lisp
```diff
@@ -176,299 +199,40 @@
...
-      ;; flush the 1Q gates down the line
-      (flush-1q-instructions-after-wiring state left-line)
-      (flush-1q-instructions-after-wiring state right-line))))
-
 (defvar *compute-tight-recombination-bound* nil
   "If T, use the compressor to try to precompute a tighter recombination bound.")
 
+(defmethod select-and-embed-a-permutation ((state temporal-addresser-state) rewirings-tried)
+  ;; randomize cost function weights
+  ;; not sure exactly why -- possibly to break symmetry when
```
ecpeterson, line 211: précisement

src/addresser/temporal-addresser.lisp
```diff
@@ -176,299 +199,40 @@
...
-                    (if (hardware-object-dead-p object)
-                        most-positive-fixnum
-                        (permutation-record-duration (vnth 0 (hardware-object-permutation-gates object))))))
-    (setf (addresser-state-qq-distances instance)
-          (precompute-qubit-qubit-distances chip-spec distance-mapping)))
+  (setf (addresser-state-qq-distances instance)
+        (compute-qubit-qubit-distances
+         chip-spec
+         (lambda (object)
+           ;; TODO: is there a good reason this uses floats over COST objects?
```
ecpeterson, line 228: good q

src/compilation-methods.lisp
ecpeterson: In fact, I'm puzzled as to why you're doing this. It seems like it nubs out much of the expected functionality of `*addresser-use-1q-queues*`.

src/addresser/addresser-common.lisp
```diff
@@ -110,281 +120,232 @@ GREEDY-QUBIT: Greedily choose the best link to swap according to the cost functi
...
-
           ;; is it maximally resourceful?
           ((global-instruction-p instr)
-           ;; dequeue the instruction and set the dirty flag
-           (chip-schedule-append chip-sched instr)
-           (lscheduler-dequeue-instruction lschedule instr)
-           (setf dirty-flag t))
+           ;; unload the 1Q queues, dequeue the instruction,
+           ;; and set the dirty flag
+           (dotimes (qubit (chip-spec-n-qubits chip-spec))
```
ecpeterson, line 236: Did you drop some commands here?

src/addresser/addresser-common.lisp
```diff
@@ -110,281 +120,232 @@ GREEDY-QUBIT: Greedily choose the best link to swap according to the cost functi
...
-           (chip-schedule-append chip-sched instr)
-           (lscheduler-dequeue-instruction lschedule instr)
-           (setf dirty-flag t))
+           ;; clear relevant 1Q queues, dequeue the instruction
+           ;; and set the dirty flag
+           (partially-flush-1Q-queues state (instruction-resources instr)))
 
           ;; is it a local mixed pure/classical instruction?
+          ;;
+          ;; TODO: this currently does not do the clever 'threading'
```
ecpeterson, line 247: This isn't your TODO, but a counterargument is that what's being "threaded" are SWAPs and MEASUREs / RESETs. Doing a MEASURE before a SWAP (1) suffers 3 fewer CZs + (2) has the potential (if, say, MEASURE happens on both participating qubits) to rewrite the post-instruction SWAP as some Xes or to elide it entirely.

src/addresser/addresser-common.lisp
```diff
@@ -110,281 +120,232 @@ GREEDY-QUBIT: Greedily choose the best link to swap according to the cost functi
...
-                                      initial-l2p
-                                      *addresser-use-free-swaps*)
-           (setf dirty-flag t))
-
-          ;; is it a small-Q gate?
-          ((<= (length (application-arguments instr))
+Two other values are returned: a list of fully rewired instructions for later scheduling, and a list of partially-rewired instructions for later scheduling."
+  (with-slots (lschedule working-l2p initial-l2p chip-spec chip-sched)
+      state
+    (let ((dirty-flag nil)
```
ecpeterson, line 315: 2 each their own

src/addresser/addresser-common.lisp
```diff
@@ -110,281 +120,232 @@ GREEDY-QUBIT: Greedily choose the best link to swap according to the cost functi
...
+          (ready-instrs nil)
+          (partial-instrs nil))
+      (cond
+        ((and *addresser-use-1q-queues*
```
ecpeterson, line 319: The organization of this clause makes me a little wary. There's another file out there that's supposed to handle 1Q queues. It would be nice if this clause could be attached as a `:before` for this method when the 1Q queues are relevant. That isn't presently possible with the organization around a global parameter—but that also felt awkward over in `compilation-methods.lisp` for different reasons..

src/compilation-methods.lisp
kilimanjaro: Oh, I agree this should go but the earliest motivation was just to make my life easier with respect to testing (since this preserves exactly the behavior of the earlier version).

kilimanjaro:
> On the whole, it seems fine. I left some comments as things to chew on. Here's one further comment that didn't belong at any particular point in the source:
> 
>     * Are keywords like `':a*` collected anywhere? As these definitions spread further out, it might be good to aggregate them into an enumerative type.

Thanks for checking this out. For the keywords, I have this little section in `addresser-common.lisp`

```

;;; Search routines are implemented in
;;;   - astar-rewiring-search.lisp (for A*)
;;;   - qubit-heuristic.lisp (for GREEDY-QUBIT)
;;;   - path-heuristic (for GREEDY-PATH)
(deftype addresser-search-type () '(member :a* :greedy-qubit :greedy-path))

(defvar *addresser-swap-search-type* ':greedy-qubit
  "The type of swap search the addresser should use.")

(defvar *addresser-move-to-rewiring-swap-search-type* ':a*
  "The type of swap search the addresser should use when doing move-to-rewiring.")
```

src/addresser/addresser-common.lisp
kilimanjaro: @stylewarning has made me prefer this through trauma-bonding PR reviews

appleby:
Nice little bonus: this PR appears to have resolved #531, or at least greatly reduced the runtimes.

Two of the three rewiring benchmarks that previously took "forever" now finish in < 5s. `0020q-0000654-johannes.quil` is still a slowpoke, but does eventually complete.

| benchmark | prev | this PR |
|------------|------|---------|
| `0005q-0000160.quil` | ?? | ~1.5s |
| `0008q-0000418-qaoa-8-5.quil` | ?? | ~5.0s |
| `0020q-0000654-johannes.quil` | ?? | ~18 min |

src/addresser/astar-rewiring-search.lisp
```diff
@@ -272,3 +272,36 @@ should stop with the given rewiring and false otherwise."
...
+                               #'done-moving
+                               :max-iterations *addresser-a*-swap-search-max-iterations*)
+              rewirings-tried))))
+
+(defmethod select-swaps-for-gates ((search-type (eql ':a*)) rewiring gates-in-waiting addresser-state rewirings-tried)
+  (with-slots (chip-spec chip-sched) addresser-state
+    (flet ((cost-function (rewiring &key instr (gate-weights gates-in-waiting))
+             (declare (ignore instr))
+             (let ((modified-state (copy-instance addresser-state)))
+               ;; TODO
```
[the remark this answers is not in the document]
kilimanjaro: Sorry, it was a TODO  for myself. The prompt was that `modified-state` is constructed, and then immediately after `(setf (addresser-state-working-l2p addresser-state) rewiring)`, although `addresser-state` is not otherwise used. Smelled funny.

src/addresser/astar-rewiring-search.lisp
kilimanjaro: Yes, good catch. I have fixed this, making sure it applies to the current addresser's working-l2p. Adding a tiny macro, `with-rotatef`, which I wouldn't be surprised if existed in some form elsewhere

```
(defmacro with-rotatef (places &body body)
  "Evaluate BODY under a rotation of PLACES, then restore these to their original values."
  `(prog2
       (rotatef ,@places)
       (progn ,@body)
     (rotatef ,@(reverse places))))
```
(possibly should `unwind-protect`, but existing uses haven't been doing this)
With this we can avoid some of the `copy-instance` calls.

src/addresser/addresser-common.lisp
kilimanjaro: That's a fair point. For now I've moved the actual queue manipulation to a separate function in `1q-queus.lisp`. As far as organization around a parameter vs CLOS's system for method combination, I think I almost always prefer the former. It's dumb and in some ways less elegant, but you never find yourself trying to piece together what a method call _actually_ turns into.

src/addresser/addresser-common.lisp
kilimanjaro: Thanks for the clarification. I'm going to leave the old TODO in there, but add your commentary as well.

src/addresser/addresser-common.lisp
kilimanjaro: I forgot to drop some comments.

kilimanjaro:
Aside from the usual tests passing, as a sanity check I ran the `ibm_qx_mapping` benchmarks (restricted to those qasm files < 10kB in size, due to time constraints). Relative to what is on master, the difference in gate depth and compilation time averaged out to < 2% (< 5% for compilation time) respectively. Considering that the the cost functions take random weights, I would imagine this meets roughly my goal of "not changing any behavior."

I was a bit surprised by @appleby 's remark about `0005q-0000160.quil` and company, since I don't know where the performance gain would come from. Considering the current master branch, on my system I see both `0005q-0000160.quil` and `0008q-0000418-qaoa-8-5.quil` finishing in seconds.

ecpeterson:
"Performance gain" may well mean only "avoids deadlock", and changes that have subtle-to-no effect on the rest of the code's performance may still dislodge whatever wrench for these particular examples. I'm not too shocked.

src/addresser/README.md
```diff
@@ -0,0 +1,101 @@
...
+
+## Cost Heuristics
+
+Right now there are two sorts of costs: a duration-based cost
+(cf. `temporal-addresser.lisp`) and a fidelity-based cost
+(cf. `fidelity-addreser.lisp`). New cost heuristics may be implemented by
+subclassing `ADDRESSER-STATE` (cf. `addresser-state.lisp`), and then implementing
+methods on a few associated generics:
+- `COST-FUNCTION`, which computes the cost of the current logical schedule.
+- `WEIGHTED-FUTURE-GATES`, which constructs a mapping from gate in the current
```
notmgsk, line 49: from /a/ gate

src/addresser/README.md
```diff
@@ -0,0 +1,101 @@
...
+  logical schedule to a value indicating "how far" into the future they will
+  occur
```
notmgsk, line 51: period

src/addresser/README.md
```diff
@@ -0,0 +1,101 @@
...
+- `BUILD-WORST-COST`, which is basically a stand-in for `most-positive-fixnum`
+  if the heuristic uses a custom or compound cost value.
+
+In practice, both the temporal and fidelity addresser go a bit further than this
+(e.g. with non-real cost values), but the above reflects the bare minimum needed
+to add a new cost heuristic.
+
+As a note, the reason for `WEIGHTED-FUTURE-GATES` is to allow parts of the
```
notmgsk, line 59: for the existence of

src/addresser/1q-queues.lisp
```diff
@@ -0,0 +1,157 @@
...
+;;; 1Q Queue Management
+;;;
+;;; The addresser optionally (cf. *ADDRESSER-USE-1Q-QUEUES*) prevents 1Q
+;;; operations from influencing anything (qubit allocation, SWAP placement, and
+;;; so on) since 1Q operations can be easily moved around and are often not very
+;;; costly.
+;;;
+;;; This is done via a system of "1Q Queues", which stash the 1Q operations
+;;; during addressing and then are flushed at appropriate times.
+
```
stylewarning, line 12: Could you add a little more documentation here (or point to req. documentation) regarding what "state" is being threaded thru these functions?

stylewarning: It would be nice to have more overview exposition as well.

stylewarning: (I do see the README now)

src/addresser/README.md
```diff
@@ -0,0 +1,101 @@
...
+following two generics
+- `SELECT-SWAPS-FOR-GATES`, used by the addresser to select swaps so that it may
+  assign logical gates to physical qubits, and
+- `SELECT-SWAPS-FOR-REWIRING`, used by the addresser to select swaps in order to
+  update the current rewiring to a desired rewiring.
+
+These represent the two main pieces of functionality needed to add a new search
+heuristic to the addresser.
+
+The search type is represented as a keyword symbol,
```
stylewarning, line 78: maybe just say
>a keyword of type ADDRESSER-SEARCH-TYPE in `addresser-common.lisp`

src/addresser/addresser-common.lisp
```diff
@@ -76,20 +68,38 @@ SWAPping qubits into place.")
 (defvar *addresser-start-with-partial-rewiring* t
   "When T, starts with a partial rewiring that is filled in gradually.")
 
-(defvar *addresser-use-2q-tiers* t
-  "When T, uses the 2-qubit tiers rather than the general instruction tiers.")
+;; TODO: consider making this part of the state
```
stylewarning, line 71: I assume these would be read-only parts of the state.

src/addresser/README.md
```diff
@@ -0,0 +1,101 @@
+## Overview and Motivation
```
stylewarning, line 1: Maybe add a "tutorial" or instructions for adding a new heuristic?

src/addresser/addresser-common.lisp
```diff
@@ -110,282 +120,225 @@ GREEDY-QUBIT: Greedily choose the best link to swap according to the cost functi
...
-           ;; dequeue the instruction and set the dirty flag
-           (chip-schedule-append chip-sched instr)
-           (lscheduler-dequeue-instruction lschedule instr)
-           (setf dirty-flag t))
+           ;; clear relevant 1Q queues
+           (partially-flush-1Q-queues state (instruction-resources instr)))
 
           ;; is it a local mixed pure/classical instruction?
+          ;;
+          ;; TODO: this currently does not do the clever 'threading' that
```
stylewarning, line 244: can we also reference this in a GH issue?

src/addresser/addresser-common.lisp
```diff
@@ -636,35 +582,37 @@ If DRY-RUN, this returns T as soon as it finds an instruction it can handle."
...
-(defun try-to-assign-qubits (state instrs-partially-assigned)
+(defun unassigned-qubits (instrs rewiring)
+  "Get all qubits referenced in INSTRS which are not assigned in REWIRING."
+  (let ((unassigned-qubits nil))
+    (dolist (instr instrs unassigned-qubits)
+      (dolist (qubit (cl-quil.resource::resource-qubits-list
+                      (instruction-resources instr)))
+        (unless (or (apply-rewiring-l2p rewiring qubit)
+                    (member qubit unassigned-qubits))
+          (push qubit unassigned-qubits))))))
```
stylewarning, line 593: cut out MEMBER and use PUSHNEW

src/addresser/addresser-state.lisp
```diff
@@ -0,0 +1,192 @@
+(in-package #:cl-quil)
+
+;;; The main entry point for addressing is DO-GREEDY-ADDRESSING . However, this
+;;; manages a fair amount of state as it navigates the addressing process. The
+;;; struct below bundles this together for the sake of convenience. At any given
+;;; point of the execution there is only one of these objects, with values being
+;;; mutated along the way.
+
+(defclass addresser-state ()
```
stylewarning, line 9: add class docs

src/addresser/addresser-state.lisp
```diff
@@ -0,0 +1,192 @@
...
+                :documentation "The initial logical-to-physical rewiring."
+                :type rewiring)
+   (working-l2p :accessor addresser-state-working-l2p
+                :initarg :working-l2p
+                :documentation "The working / current logical-to-physical rewiring. NOTE: This get mutated a _lot_."
+                :type rewiring)
+   (qq-distances :accessor addresser-state-qq-distances
+                 :initarg :qq-distances
+                 :documentation "Precomputed SWAP penalties between separated qubits."
+                 :type array)
```
stylewarning, line 21: what kind of array? `(array real (* *))`?

src/addresser/addresser-state.lisp
stylewarning, line 9: I would remove initargs if they'll be overwritten by the init-instance; people an always finely construct this obj if need be

src/addresser/addresser-state.lisp
```diff
@@ -0,0 +1,192 @@
...
+                                                          #'>
+                                                          :key #'length)
+          (addresser-state-chip-schedule instance) (make-chip-schedule chip-spec)
+          (addresser-state-chip-specification instance) chip-spec
+          (addresser-state-1q-queues instance) (make-array (chip-spec-n-qubits chip-spec) :initial-element (list)))))
+
+
+(defmacro with-rotatef (places &body body)
+  "Evaluate BODY under a rotation of PLACES, then restore these to their original values."
+  `(prog2
```
stylewarning, line 70: nit: what about UNWIND-PROTECT for early exit?

stylewarning: also i would just allow two places; too confusing for 100 places. then you can do (approx)

```
(flet ((flip () (rotatef ,p1 ,p2)))
  (declare (dynamic-extent #'rotate))
  (flip)
  (unwind-protect (progn ,@body) (flip)))
```

src/addresser/addresser-state.lisp
```diff
@@ -0,0 +1,192 @@
...
+;;; weights and costs.
+;;;
+;;; NOTE: Addresser implementations MUST also fill the qq-distances slot with
+;;; appropriate values. On the one hand, these are necessary for the basic
+;;; functioning of the addresser routines and swap selection heuristics. On the
+;;; other hand, we do not at present wish to suppose exactly what this looks
+;;; like.
+
+(defgeneric weighted-future-gates (state)
+  (:documentation "Collect gates from the logical schedule, along with weights which indicate how far into the future the gate occurs.
```
stylewarning, line 87: be more descriptive in this generic?

src/addresser/addresser-state.lisp
```diff
@@ -0,0 +1,192 @@
...
+
+(defgeneric cost-< (val1 val2)
+  (:documentation "Generic comparison function for heuristic values.")
+  (:method ((val1 number) (val2 number))
+    (< (- val1 +double-comparison-threshold-strict+) val2)))
+
+(defgeneric cost-= (val1 val2)
+  (:documentation "Generic equality function for heuristic values.")
+  (:method ((val1 number) (val2 number))
+    (double= val1 val2)))
```
stylewarning, line 120: don't these need to be coerced to doubles?

src/addresser/addresser-state.lisp
```diff
@@ -0,0 +1,192 @@
...
+
+(defgeneric build-worst-cost (state)
+  (:documentation "Builds a POSITIVE-INFINITY type value for the COST-FUNCTION associated to STATE."))
+
+;;; Since sometimes costs are compound objects which encode various values, they
+;;; must implement the following:
+
+(defgeneric cost-< (val1 val2)
+  (:documentation "Generic comparison function for heuristic values.")
+  (:method ((val1 number) (val2 number))
```
stylewarning, line 114: only works for `real`

src/addresser/addresser-state.lisp
```diff
@@ -0,0 +1,192 @@
...
+    (< (- val1 +double-comparison-threshold-strict+) val2)))
+
+(defgeneric cost-= (val1 val2)
+  (:documentation "Generic equality function for heuristic values.")
+  (:method ((val1 number) (val2 number))
```
stylewarning, line 119: only works for `real`

src/addresser/README.md
kilimanjaro: gratuitous

src/addresser/1q-queues.lisp
kilimanjaro: Added a bit more to the documentation at the top, as well as `(check-type state addresser-state)` at the top of the relevant defuns.

src/addresser/addresser-state.lisp
kilimanjaro: adding a bit more to the docstring

src/addresser/addresser-state.lisp
kilimanjaro: rip our beloved `prog2`

src/addresser/addresser-state.lisp
kilimanjaro: I'm just going to specialize these cost methods on `double-float`. IMO narrowing conversions for equality and inequality checks are a bit worrisome (not saying that they can't have a place, but that place shouldn't exist by accident).

kilimanjaro: ... nevermind. We pretty often do cost arithmetic with mixed `rational` and `double-float` types. For now I will preserve the current semantics, which does not involve casting to doubles -- `double=` assumes nothing of the sort -- but since we might actually one day want to upgrade `double=` to something which assumes that its arguments are doubles, I have remove that call from this site.

src/addresser/README.md
kilimanjaro: I think this is a good idea in general, and I would be happy to do so, but not for this PR (mostly due to time constraints on my end). As it stands we're at least halfway there: now that the temporal and fidelity heuristics are on equal footing, developers can imagine what the addition of a third cost heuristic might look like. Similarly, the existing search heuristics serve as examples for implementation of new search heuristics. (I recognize that this is still quite burdensome for an outsider to grok.)

src/addresser/addresser-common.lisp
kilimanjaro: Yes, that would be the plan.

src/addresser/addresser-common.lisp
kilimanjaro: Added #635

PR #618, opened 2020-03-25
Add MAGICL constructor shims for default matrix type

### LINKED ISSUE #597: Set default tensor type to double-float

colescott:
Resolves #597 

This commit adds shims for all the MAGICL constructors to set the
default element type to '(COMPLEX DOUBLE-FLOAT). This is being done
remove the need to specify the type on every call inside CL-QUIL.
NOTE: To get a DOUBLE-FLOAT matrix you will now need to specify the TYPE.

The shims are a way of avoiding setting MAGICL::+DEFAULT-TENSOR-TYPE+
for the whole package, as that would then have global side effects,
affecting packages loaded afterwards (and likely cause errors).
A modification to MAGICL to allow for a default for each package to be
set was also considered but would have made calls too verbose.


src/matrix-operations.lisp
```diff
@@ -317,11 +313,11 @@ as needed so that they are the same size."
     (when *compress-carefully*
       (assert (matrix-equality m
                                (magicl:@ u
-                                         (magicl:from-diag d :type '(complex double-float))
+                                         (from-diag d)
                                          (magicl:conjugate-transpose u)))
               ()
               "MATRIX-EXPT failed to diagonalize its input."))
-    (let ((dd (magicl:from-diag
```
notmgsk: Should this one be changed? It wasn't specifying a type before.

src/clifford/benchmarking-procedures.lisp
```diff
@@ -178,8 +178,8 @@
     "Return a list of the n qubit pauli basis matrices. Note that this is the basis for group action, consisting of strings with one X or Z, not the basis for the vector space of complex matrices."
     (or (gethash n memo-table)
         (setf (gethash n memo-table)
-              (let ((X   (magicl:from-list '(0 1 1 0) '(2 2) :type '(complex double-float)))
-                    (Z   (magicl:from-list '(1 0 0 -1) '(2 2) :type '(complex double-float))))
+              (let ((X (quil::from-list '(0 1 1 0) '(2 2)))
```
stylewarning: why do you need the package prefix?

appleby:
> It looks like there are still a handful of places in the quilec tests that still call `magicl:zeros` (maybe others). For example, in `encode-linear-algebra` in `cleve-gottesman-tests.lisp`.
> 
> The quilec test suite let-binds `magicl::*default-tensor-type*` to complex double float in `test/quilec/suite.lisp`, which is probably why grepping for `:type '(complex double-float)` didn't catch these.
> 
> We should probably remove that let binding and update the tests to call the quilc shims insteads.

Wrong again! My code review game is weak today. Turns out that I only fetched but didn't actually check out your branch.  Should probably still remove the let binding though.

src/clifford/benchmarking-procedures.lisp
colescott: We are inside the clifford package. See lines below

src/matrix-operations.lisp
colescott: This one is fine as is. It previously inspected the first element of the list and chose the correct type (`(complex double-float)`). With the `cl-quil` bindings it will by default use that type (with the option to specify a different type if needed). Plus it's _very slightly_ faster!

colescott:
> Wrong again! My code review game is weak today. Turns out that I only fetched but didn't actually check out your branch. Should probably still remove the let binding though.

Removed in de1f282

src/clifford/benchmarking-procedures.lisp
stylewarning: oops yeah

PR #611, opened 2020-03-18
Allow freely parameterized RX gates

notmgsk:
Closes #610. In an ISA, under the "gates" key one can specify a parameter value for which the gate is native. For example, on Aspen chips `RX` is native only for parameter values of `0`, `pi/2`, `pi`, `-pi/2`, and `-pi`. This restriction is occasionally baked-into the compiler, rather than just in the ISA. I remove that restriction, allowing freely parameterized RXs as the output of compilation:
```
;; produce a 1Q program on qubit 0, with 10 interleaved RX or RZ gates
(print-parsed-program (compiler-hook (random-1q-program 0 10) chip))

RX(0.8883713366102477) 0                # Entering rewiring: #(0)
RZ(-1.8883138383394045) 0
RX(2.8851948642306953) 0
```
where chip is given by this ISA:
```
{
    "isa": {
        "1Q": {
            "0": {
                "gates": [
                    {
                        "operator": "I",
                        "parameters": [],
                        "arguments": [
                            "_"
                        ],
                        "fidelity": null,
                        "duration": null
                    },
                    {
                        "operator": "RX",
                        "parameters": [
                           "theta"
                        ],
                        "arguments": [
                            "_"
                        ],
                        "fidelity": null,
                        "duration": null
                    },
                    {
                        "operator": "RZ",
                        "parameters": [
                            "theta"
                        ],
                        "arguments": [
                            "_"
                        ],
                        "fidelity": null,
                        "duration": null
                    },
                    {
                        "operator": "MEASURE",
                        "qubit": "_",
                        "target": "_",
                        "duration": null,
                        "fidelity": null
                    },
                    {
                        "operator": "MEASURE",
                        "qubit": "_",
                        "target": null,
                        "duration": null,
                        "fidelity": null
                    }
                ]
            }
        },
        "2Q": {
            
        }
    },
    "specs": {
        "1Q": {
            "0": {
                "f1QRB": 0.909,
                "f1QRB_std_err": 0.01,
                "f1Q_simultaneous_RB": 0.909,
                "f1Q_simultaneous_RB_std_err": 0.02,
                "fRO": 0.9,
                "T1": 3e-05,
                "T2": 3e-05,
                "fActiveReset": 0.909
            },
            "1": {
                "f1QRB": 0.909,
                "f1QRB_std_err": 0.01,
                "f1Q_simultaneous_RB": 0.909,
                "f1Q_simultaneous_RB_std_err": 0.02,
                "fRO": 0.9,
[...]

PR #607, opened 2020-03-06
Use hermitian-eig rather than eig

notmgsk:
Closes #599.

PR #619, opened 2020-03-26
Bump version to 1.18.0

appleby:
## Additions and Improvements

- Synthesis of encoding circuits for stabilizer codes (@jmbr, #600)

  Added a new package `CL-QUIL.QUILEC` with support for stabilizer
  codes. See the README for more info.

  https://github.com/rigetti/quilc/blob/master/src/quilec/README.org

- COMPILER-HOOK *does not* mutate and *does* compress qubits (@notmgsk, #609)

  COMPILER-HOOK now takes an optional `:DESTRUCTIVE` keyword arg. If
  `NIL` (default), the input `PARSED-PROGRAM` is not mutated.

  Also, logical qubits in the input Quil program will be compressed
  when possible, so that compiling a program like "H 9" no longer
  fails when compiling for an ISA that has qubits numbered 0-8, e.g.
  Qubit compression is disabled when the requested rewiring mode is
  `:NAIVE` or when the program contains `PRAGMA PRESERVE_BLOCK`
  pragmas.

- Add SWANK server (@notmgsk, #615)

  A SWANK server can be enabled for debugging purposes via the
  `--swank-port` command line flag. The SWANK server is disabled for
  release builds.

- Better errors and coercing for param values (@rigettizach, #608)
- Don't print circuit definitions in batch mode (@notmgsk, #601)

## Bugfixes

- Allow freely parameterized RX gates (@notmgsk, #611)
- Remove extra sqrt in fidelity calculation (@notmgsk, #613)
- Use MAGICL:HERMITIAN-EIG rather than MAGICL:EIG in PARAMETRIC-PAULI-COMPILER (@notmgsk, #607)

## Misc

- If quilc is latest version, don't print the SDK version check log message (@notmgsk, #602)
- Install quilc to system in Docker builds (@notmgsk, #614)


notmgsk:
Nice detail <3

notmgsk:
@appleby Do we need to bump the RPCQ version in `Dockerfile`? I think we may.

appleby:
> @appleby Do we need to bump the RPCQ version in `Dockerfile`? I think we may.

I think you're right. Good catch.

appleby:
I'm just going to wait until the rpcq pipeline finishes so that the new rpcq Docker image is available

appleby:
Bumped the qvm version while we're at it.

PR #625, opened 2020-03-30
allow INST to be used dynamically

stylewarning:
This commit doesn't change the how any existing functionality works,
but does permit new functionality.

Currently, INST & co may only be used lexically within a WITH-INST,
and by extension, within DEFINE-COMPILER. If INST is used outside of
either of those contexts, there will be a _compile-time_ error it can
be detected, but definitely a run-time error.

This commit allows INST to be used anywhere within the dynamic extent
of WITH-INST (and DEFINE-COMPILER). This means it's now possible to
write functions like:

    (defun inst-hadamard (N)
      (dotimes (q N)
        (inst "H" () q)))

There are two downsides:

- Compile-time errors cannot be detected since any INST call may
      possible be valid.

- There may be a small performance penalty using dynamic
      features. However, what we lost in making it dynamic might be
      gained back by much smaller code. (INST and INST* aren't
      duplicated in every definition.)

### DIFF
--- a/src/define-compiler.lisp
+++ b/src/define-compiler.lisp
@@ -650,7 +650,7 @@ N.B.: The word \"shortest\" here is a bit fuzzy.  In practice it typically means
 
 (defun compute-applicable-compilers (target-gateset qubit-count) ; h/t lisp
   "Starting from all available compilers, constructs a precedence-sorted list of those compilers which help to convert from arbitrary inputs to a particular target gateset."
-  (flet (;; Checks whether PATH doubles back through the occurrence table START.         
+  (flet (;; Checks whether PATH doubles back through the occurrence table START.
          (path-has-a-loop-p (path start)
            (and (< 1 (length path))
                 (loop :for (table compiler) :on path :by #'cddr
@@ -659,7 +659,7 @@ N.B.: The word \"shortest\" here is a bit fuzzy.  In practice it typically means
                                                   :thereis (binding-subsumes-p (first start) gate)))
                         :do (return-from path-has-a-loop-p t)
                       :finally (return nil)))))
-    
+
     (let* ((target-gateset (blank-out-qubits target-gateset))
            (compilers (get-compilers qubit-count))
            (unconditional-compilers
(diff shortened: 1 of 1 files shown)


stylewarning:
(May be of interest to @ecpeterson.)

src/define-compiler.lisp
```diff
@@ -903,85 +903,94 @@ N.B.: This routine is somewhat fragile, and highly creative compiler authors wil
...
+  (throw +inst-tag+ (if retval-p retval (%collected-insts))))
+
+(defun inst (&rest xs)
+  "Emit the instructions XS in the current compiler context.
+
+INST can only be called within the dynamic extent of a compiler body."
+  (declare (dynamic-extent xs))
+  (unless (boundp '*inst-collection*)
+    (error "INST called outside of a WITH-INST context."))
+  ;; INST is a no-op if something is provided.
```
ecpeterson: if nothing**

stylewarning:
For N = 1 samples, we are actually 10 seconds _faster_.

before:
```
Evaluation took:
  462.272 seconds of real time
  463.079237 seconds of total run time (460.666083 user, 2.413154 system)
  [ Run times consist of 3.633 seconds GC time, and 459.447 seconds non-GC time. ]
  100.17% CPU
  92,424 forms interpreted
  13,567 lambdas converted
  1,109,454,218,200 processor cycles
  323,737,789,280 bytes consed
```

after:
```
  451.827 seconds of real time
  452.877278 seconds of total run time (450.658180 user, 2.219098 system)
  [ Run times consist of 3.461 seconds GC time, and 449.417 seconds non-GC time. ]
  100.23% CPU
  92,424 forms interpreted
  4,796 lambdas converted
  1,084,334,105,504 processor cycles
  320,138,089,840 bytes consed
```

stylewarning:
I still request @appleby looks at it.

src/define-compiler.lisp
```diff
@@ -903,85 +903,94 @@ N.B.: This routine is somewhat fragile, and highly creative compiler authors wil
...
+This macro is used for DEFINE-COMPILER."
+  `(let ((*inst-collection* (%new-inst-collection)))
+     (catch ',+inst-tag+
+       ,@body
+       (%collected-insts))))            ; FINISH-COMPILER would also be OK.
 
-FINISH-COMPILER is a local macro usable within a compiler body."
-  (error "FINISH-COMPILER can only be used in the body of a compiler."))
+(defun finish-compiler (&optional (retval nil retval-p))
+  "Finish compiling, by default returning a list of all instructed emitted with INST.
```
appleby: instructed -> instructions

stylewarning: too lazy to fix

PR #630, opened 2020-04-04
initial proposal for quilc backends

stylewarning:
(This PR is intended to be a proposal/discussion rather than a "let's merge ASAP" sort of thing.)
(Nomenclature: I'm using the words "frontend" and "backend" for less typing and fewer -'s.)

Quilc is a compiler that only has a frontend. It produces Quil code that adheres to a particular chip specification. It currently has no means to generate code for a particular computer or control system.

The way it's done at Rigetti right now is that this "native Quil" is handed off to proprietary services internally which compile this code for Rigetti's variety of control systems. These services comprise a compiler "backend" and are currently opaque to users of quilc.

This PR aims to extend quilc to allow backends to be implemented in quilc itself. This PR does not implement any non-trivial backends. The reasons we ought to do this is:

 

-    It follows traditional compiler design. (It's very odd—though justified for private reasons—that a compiler is splayed across several services.)
-    Quilt moves the user just a few hairs away from the low-level code gen. The more adoption Quilt has, the more quilc ought to know about towards what it's compiling programs.
-   Having backend code generation allows additional optimizations to happen, which are easier to write and more performant in Lisp.
-   With a frontend and backend comes "middle-ends" [sic], where lower-level IRs can be introduced (like Quilt) which permit finer-grained, non-quantum optimizations.
-   Code generation can be costly and quilc is a good platform to optimize such.

The design I present is very simple and accomplishes the following goals:

-   It provides an ontology for backends. They are subclasses of BACKEND. I chose this over bare protocol functions since we can use the class structure to deduce available backends.
-   It provides a protocol for executables. At the end of the day, an executable is something that is writable to a binary stream, and cryptographically signable. There is no assumption as to whether an executable "is" a file or a bundle thereof. There is an assumption that an executable can be expressed as a file (e.g., a tarball).
-   It provides a UI for specifying the backend at the command line, as well as a file to write to.

The entry point to the backend (within cl-quil) is compile-to-backend. It's the line-and-sinker to our beloved compiler-hook. (Yes, I did contemplate putting compiler-line-and-sinker to be quirky in the codebase, and in this very rare occasion, I decided against it. I can be convinced otherwise by @appleby and @notmgsk only.)

Implemented in this PR is a "dummy" backend called quil, where an "executable" is a UTF8-encoded file of Quil code. This is an example of using the new user-facing features.
```
$ echo 'CONTROLLED CONTROLLED H 0 1 2' | ./quilc --backend 'quil' -o 'q.out'
<134>1 2020-04-04T21:24:42Z cobbler.local quilc 9596 - - Generating native Quil.
<134>1 2020-04-04T21:24:43Z cobbler.local quilc 9596 - - Generating backend code.
<134>1 2020-04-04T21:24:43Z cobbler.local quilc 9596 - - Wrote executable.

$ cat q.out 
RZ(0.9485576243433194) 4                # Entering rewiring: #(6 4 5 0 1 2 3 7)
RX(pi/2) 4
RZ(-3*pi/4) 5
RX(pi/2) 5
RZ(0.5299027897489266) 5
RX(-pi/2) 5
RZ(1.0149377841503018) 5
CZ 4 5
RX(pi/2) 4
RZ(0.8588857587292011) 4
RX(-pi/2) 4
CZ 4 5
RZ(2.9676117939887945) 5
RX(pi/2) 5
RZ(2.100629020021203) 5
RX(-pi/2) 5
RZ(1.5617345686152926) 5
RZ(pi/2) 6
RX(pi/2) 6
RZ(-2.383295449271497) 6
RX(-pi/2) 6
CZ 5 6
RZ(pi) 5
RX(pi/2) 5
RZ(pi) 6
RX(pi/2) 6
CZ 5 6
RX(pi/2) 4
RZ(-2.1987649493772086) 5
RX(pi/2) 5
RZ(2.078609539841221) 5
RX(-pi/2) 5
CZ 4 5
RZ(2.16095302738569) 5
RX(pi/2) 5
RZ(3*pi/4) 5
RX(-pi/2) 5
RX(pi/2) 6
CZ 5 6
RZ(-0.6455362053461816) 4
RX(-pi/2) 4
RZ(-2.208413464564872) 5
RX(pi/2) 5
RZ(1.0338528234701385) 5
RX(-pi/2) 5
RZ(0.9665135945991623) 5
CZ 5 4
RZ(pi) 4
RX(pi/2) 4
RZ(pi) 5
RX(pi/2) 5
CZ 5 4
RX(pi/2) 5
RZ(-2.332896987297744) 5
RX(pi/2) 5
RZ(-2.3407240608221773) 6
RX(pi/2) 6
CZ 5 6
RX(pi/2) 5
RZ(pi/8) 5
RX(-pi/2) 5
CZ 5 6
RX(pi/2) 4
RZ(0.808695666292048) 4
RX(-pi/2) 4
RZ(-pi/4) 4
RZ(pi) 5
RX(-pi/2) 5
RZ(7*pi/8) 6
RX(pi/2) 6
HALT                                    # Exiting rewiring: #(4 5 6 0 1 2 3 7)
```

Some big TODO items:

-   Parse out RSA keys for signage.
-   Understand a bit better how a backend might be specified. The quil backend isn't very sophisticated and I suspect more complicated backends cannot be identified with just a "string".
-   Explore what's missing in terms of protocol functions. I think there must be something missing WRT memory models, but maybe not.
-   Should any other standard conditions be specified?
-   Allow backends to be listed. The backend classes can be listed with cl-quil::list-available-backends, but there's no correspondence between "friendly names" and these classes. Right now it's hacked up in quilc::parse-backend.

I would appreciate comments, questions, and discussion!

### DIFF
--- a/app/src/entry-point.lisp
+++ b/app/src/entry-point.lisp
@@ -44,6 +44,21 @@
      :optional t
      :initial-value "8Q"
      :documentation "set ISA to one of \"8Q\", \"20Q\", \"16QMUX\", \"bristlecone\", \"ibmqx5\", or path to QPU description file")
+    
+    (("backend")
+     :type string
+     :optional t
+     :documentation "set the backend for which to generate code (default: none). Note that an output file (-o) must be specified!")
+    
+    (("output-file" #\o)
+     :type string
+     :optional t
+     :documentation "the filename to which to write an executable (default: none)")
+    
+    (("private-signing-key")
+     :type string
+     :optional t
+     :documentation "the filename of the RSA private key used to sign executables (default: none)")
 
     (("enable-state-prep-reductions")
      :type boolean
@@ -382,6 +397,9 @@
                           (print-logical-schedule nil)
                           (verbose nil)
                           (isa nil)
+                          (backend nil)
+                          (output-file nil)
+                          (private-signing-key nil)
                           (enable-state-prep-reductions nil)
                           (protoquil nil)
                           (print-statistics nil)
@@ -492,6 +510,10 @@
 
          (cl-syslog:rfc-log (*logger* :info "Launching quilc.")
            (:msgid "LOG0001"))
+         (when (or backend output-file)
+           (cl-syslog:rfc-log (*logger* :warning "--backend and --output-file/-o options ~
+                                                  don't make sense when using server mode. ~
+                                                  Ignoring them.")))
          ;; launch the polling loop
          (start-rpc-server :host host
                            :port port
@@ -501,36 +523,88 @@
         ;; server modes not requested, so continue parsing arguments
         (t
          (setf *human-readable-stream* *error-output*)
-         (setf *quil-stream* *standard-output*)
+         ;; If we didn't specify a backend, then we will use -o to
+         ;; write out the front-end compiled Quil code. If a backend
+         ;; WAS specified, we set this to stdout so that the optional
+         ;; statistics are printed there.
+         ;;
+         ;; We will never print binary output to the terminal.
+         (setf *quil-stream* (if (and (not backend)
+                                      output-file)
+                                 (open output-file :direction ':output
+                                                   :if-does-not-exist ':create
+                                                   :if-exists ':supersede)
+                                 *standard-output*))
+         
+         (when (and backend (not output-file))
+           (error "You specified a backend, so you must specify an ~
+                   output file with --output-file or -o."))
 
-         (let* ((program-text (slurp-lines))
-                (program (safely-parse-quil program-text))
-                (original-matrix (when (and protoquil compute-matrix-reps)
-                                   (parsed-program-to-logical-matrix program))))
-           (multiple-value-bind (processed-program statistics)
-               (process-program program (lookup-isa-descriptor-for-name isa)
-                                :protoquil protoquil
-                                :verbose verbose
-                                :gate-whitelist (and gate-whitelist
-                                                     (split-sequence:split-sequence
-                                                      #\,
-                                                      (remove #\Space gate-whitelist)
-                                                      :remove-empty-subseqs t))
-                                :gate-blacklist (and gate-blacklist
-                                                     (split-sequence:split-sequence
-                                                      #\,
-                                                      (remove #\Space gate-blacklist)
-                                                      :remove-empty-subseqs t)))
-             (unless print-circuit-definitions
-               (setf (parsed-program-circuit-definitions processed-program) nil))
-             (print-program processed-program *quil-stream*)
-             (when (and protoquil print-statistics)
-               (print-statistics statistics *quil-stream*))
-             (when (and protoquil compute-matrix-reps)
-               (let* ((processed-program-matrix (parsed-program-to-logical-matrix processed-program :compress-qubits t)))
-                 (print-matrix-comparision original-matrix
-                                           (quil::scale-out-matrix-phases processed-program-matrix
-                                                                          original-matrix)))))))))))
+         (unwind-protect
+              (let* ((chip-spec (lookup-isa-descriptor-for-name isa))
+                     (program-text (slurp-lines))
+                     (program (safely-parse-quil program-text))
+                     (original-matrix (when (and protoquil compute-matrix-reps)
+                                        (parsed-program-to-logical-matrix program))))
+                (cl-syslog:rfc-log (*logger* :info "Generating native Quil."))
+                (multiple-value-bind (processed-program statistics)
+                    (process-program program chip-spec
+                                     :protoquil protoquil
+                                     :verbose verbose
+                                     :gate-whitelist (and gate-whitelist
+                                                          (split-sequence:split-sequence
+                                                           #\,
+                                                           (remove #\Space gate-whitelist)
+                                                           :remove-empty-subseqs t))
+                                     :gate-blacklist (and gate-blacklist
+                                                          (split-sequence:split-sequence
+                                                           #\,
+                                                           (remove #\Space gate-blacklist)
+                                                           :remove-empty-subseqs t)))
+                  (unless print-circuit-definitions
+                    (setf (parsed-program-circuit-definitions processed-program) nil))
+                  ;; If a backend was selected, we do *not* want to
+                  ;; print the processed program, but we *do* want to
+                  ;; print statistics if asked for.
+                  (unless backend
+                    (print-program processed-program *quil-stream*))
+                  (when (and protoquil print-statistics)
+                    (print-statistics statistics *quil-stream*))
+                  (when (and protoquil compute-matrix-reps)
+                    (let* ((processed-program-matrix (parsed-program-to-logical-matrix processed-program :compress-qubits t)))
+                      (print-matrix-comparision original-matrix
+                                                (quil::scale-out-matrix-phases processed-program-matrix
+                                                                               original-matrix))))
+                  (when backend
+                    (cl-syslog:rfc-log (*logger* :info "Generating backend code."))
+                    (backend-compile processed-program
+                                     chip-spec
+                                     (parse-backend backend)
+                                     output-file
+                                     private-signing-key) ; TODO: parse key
+                    (cl-syslog:rfc-log (*logger* :info "Wrote executable.")))))
+           (when output-file
+             (close *quil-stream*))))))))
+
+(defun parse-backend (backend-str)
+  ;; TODO: make this more flexible for "real" backends.
+  (alexandria:eswitch (backend-str :test #'string-equal)
+    ("quil" (make-instance 'cl-quil:quil-backend))))
+
+(defun backend-compile (program chip-spec backend out-file &optional signing-key)
+  (when signing-key
+    ;; The caller should parse the signing key from the file.
+    (error "signing isn't supported yet"))
+  (unless (cl-quil:backend-supports-chip-p backend chip-spec)
+    (error "The backend provided doesn't support this ISA."))
+  (let ((exec (cl-quil:compile-to-backend program chip-spec backend)))
+    (when signing-key
+      (cl-quil:sign-executable exec signing-key))
+    (with-open-file (stream out-file :direction ':output
+                                     :element-type '(unsigned-byte 8)
+                                     :if-exists ':supersede
+                                     :if-does-not-exist ':create)
+      (cl-quil:write-executable exec stream))))
 
 (defun process-program (program chip-specification
                         &key
--- a/src/backends/common.lisp
+++ b/src/backends/common.lisp
@@ -78,5 +78,5 @@
 
 (defconstant +rsa-recommended-size-bits+ 3072)
 
-(defun generate-key-pair ()
+(defun generate-rsa-key-pair (&optional (num-bits +rsa-recommended-size-bits+))
   (ironclad:generate-key-pair ':rsa :num-bits +rsa-recommended-size-bits+))
--- a/src/backends/common.lisp
+++ b/src/backends/common.lisp
@@ -7,7 +7,7 @@
 ;;; The BACKEND protocol
 ;;;
 ;;; The BACKEND class is used primarily to determine available
-;;; baackends primarily by computing subclasses.
+;;; backends primarily by computing subclasses.
 
 (defclass backend ()
   ()
--- a/src/backends/common.lisp
+++ b/src/backends/common.lisp
@@ -62,7 +62,7 @@
 (defgeneric compile-to-backend (program chip-spec backend)
   (:documentation "Compile the PROGRAM which comports to the CHIP-SPEC to an executable for BACKEND."))
 
-;;;;;;;;;;;;;;; Backend Construction Library Functiona ;;;;;;;;;;;;;;;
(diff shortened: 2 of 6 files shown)


src/backends/common.lisp
```diff
@@ -0,0 +1,82 @@
...
+
+(defgeneric backend-supports-chip-p (backend chip)
+  (:documentation "Does BACKEND support the chip CHIP-SPECIFICATION?"))
+
+(define-condition incompatible-memory-model-error (error)
+  ((backend :initarg :backend
+            :reader incompatible-memory-model-error-backend)
+   (descriptors :initarg :descriptors
+                :reader incompatible-memory-model-error-descriptors))
+  (:documentation "A condition signaled if a set of descriptors is incompatible with a backend."))
```
colescott, line 25: Nitpick: Doc string should mention the incompatibility is with the memory model

src/backends/common.lisp
```diff
@@ -0,0 +1,82 @@
...
+;;; implementer with e.g. tarballs.
+;;;
+;;; Many of the functioms below deal with signing the executable,
+;;; which verifies the identity of the creator of the executable.
+;;;
+
+(defgeneric write-executable (executable stream)
+  (:documentation "Write EXECUTABLE to the binary output STREAM."))
+
+(defgeneric signature (executable)
```
colescott, line 48: Discussion: Is the action of signing our outputs intrinsic in the backend role of the compiler? While I agree that each backend needs a way of specifying where to put and what the signature looks like, I don't know if it should be done within this class or if there's a more appropriate place for it.

app/src/entry-point.lisp
```diff
@@ -491,6 +509,10 @@
 
          (cl-syslog:rfc-log (*logger* :info "Launching quilc.")
            (:msgid "LOG0001"))
+         (when (or backend output-file)
```
colescott, line 512: Would it make sense to allow the quilc server to compile the program and return the resulting (signed) binary?

src/backends/common.lisp
```diff
@@ -0,0 +1,82 @@
...
+  ;; This is simpleminded and assumes there isn't some crazy
+  ;; inheritance structure.
+  (mapcar #'class-name (c2mop:class-direct-subclasses (find-class 'backend))))
+
+(defconstant +binary-hash-size+ 64)
+
+(defun binary-hash (binary)
+  (ironclad:digest-sequence ':sha3 binary))
+
+(defconstant +rsa-recommended-size-bits+ 3072)
```
colescott, line 79: While I generally agree with larger key sizes, what is the reason for this over 2048 or 4096 especially wrt security and speed?

src/backends/common.lisp
```diff
@@ -0,0 +1,82 @@
...
+;;;;;;;;;;;;;;; Backend Construction Library Functiona ;;;;;;;;;;;;;;;
+
+;;; The things below aren't part of the protocol, but generally useful.
+
+(defun list-available-backends ()
+  ;; This is simpleminded and assumes there isn't some crazy
+  ;; inheritance structure.
+  (mapcar #'class-name (c2mop:class-direct-subclasses (find-class 'backend))))
+
+(defconstant +binary-hash-size+ 64)
```
colescott, line 74: Should this be allowed to be overridden by the end user?

app/src/entry-point.lisp
kilimanjaro: It would, but these two would be specified as part of the server request rather than here.

app/src/entry-point.lisp
```diff
@@ -500,36 +522,88 @@
...
+                    (cl-syslog:rfc-log (*logger* :info "Wrote executable.")))))
+           (when output-file
+             (close *quil-stream*))))))))
+
+(defun parse-backend (backend-str)
+  ;; TODO: make this more flexible for "real" backends.
+  (alexandria:eswitch (backend-str :test #'string-equal)
+    ("quil" (make-instance 'cl-quil:quil-backend))))
+
+(defun backend-compile (program chip-spec backend out-file &optional signing-key)
```
kilimanjaro, line 593: One thing worth thinking about is that backends will likely need more data than what we track in the current `chip-spec` (and more than what is currently specified in the ISA). Perhaps it merits having an additional argument here, which is required to be compatible with chip-spec but which has more specific information to be consumed by the backend (with e.g. hardware configuration, device operating points, additional linker settings, hardware/firmware versions, etc). The alternative, of having all of this data stashed into `chip-spec`, may I think substantially complicate cross-compilation, given the general state of classical control hardware (rapidly changing, heterogenous systems).

src/backends/common.lisp
```diff
@@ -0,0 +1,82 @@
+;;;; common.lisp
+;;;;
+;;;; Author: Robert Smith
+
+(in-package #:cl-quil)
+
+;;; The BACKEND protocol
+;;;
+;;; The BACKEND class is used primarily to determine available
+;;; baackends primarily by computing subclasses.
```
kilimanjaro, line 10:
```suggestion
;;; backends primarily by computing subclasses.
```

src/backends/instructions.md
```diff
@@ -0,0 +1,33 @@
...
+
+The `backends/` folder contains the implementations of the backends, as well as common code useful for constructing a backend. Every backend should have its own folder in this directory.
+
+## How to add a new backend
+
+The `quil` backend is a minimal, dummy backend that emits Quil, and can be used to model a new, more complicated backend. The basic steps are this:
+
+1. Make a new directory for your backend.
+
+1. Implement the `BACKEND`, `EXECUTABLE`, and `COMPILATION` protocols in `common.lisp`.
```
kilimanjaro, line 17:
```suggestion
2. Implement the `BACKEND`, `EXECUTABLE`, and `COMPILATION` protocols in `common.lisp`.
```

src/backends/common.lisp
```diff
@@ -0,0 +1,82 @@
...
+  (:documentation "Does BACKEND support the chip CHIP-SPECIFICATION?"))
+
+(define-condition incompatible-memory-model-error (error)
+  ((backend :initarg :backend
+            :reader incompatible-memory-model-error-backend)
+   (descriptors :initarg :descriptors
+                :reader incompatible-memory-model-error-descriptors))
+  (:documentation "A condition signaled if a set of descriptors is incompatible with a backend."))
+
+;;; TODO: Should we add anything to the BACKEND protocol to deal with
```
kilimanjaro, line 27: Did you have anything in mind?

src/backends/common.lisp
```diff
@@ -0,0 +1,82 @@
...
+;;;
+;;; The BACKEND class is used primarily to determine available
+;;; baackends primarily by computing subclasses.
+
+(defclass backend ()
+  ()
+  (:documentation "Every backend must be represented by a subclass of this abstract base class.")
+  (:metaclass abstract-class))
+
+(defgeneric backend-supports-chip-p (backend chip)
```
kilimanjaro, line 17: This might also take an extra argument consisting of backend-specific configuration data.

src/backends/common.lisp
```diff
@@ -0,0 +1,82 @@
...
+                :reader incompatible-memory-model-error-descriptors))
+  (:documentation "A condition signaled if a set of descriptors is incompatible with a backend."))
+
+;;; TODO: Should we add anything to the BACKEND protocol to deal with
+;;; memory allocation, the memory model (cf. classical-memory.lisp),
+;;; etc.?
+
+;;; The EXECUTABLE protocol
+;;;
+;;; An executable is an artifact which may be written to a binary
```
kilimanjaro, line 33: Just a ramble. Thinking about the present day implementation in PyQuil, in addition to executable code, certain metadata related to the classical/quantum interface is managed: specifically PyQuil has a "patch table" and some hardcoded notion that output bits are associated with with the 'ro' register. In principle I could imagine having either of those two sorts of things (or others) expected, in addition to signed code. 

Of course, this is getting platform specific. So your choice that an executable is "an artifact which may be written to a binary output stream" is hard to argue with, and the rest of the details can be punted.

src/backends/common.lisp
```diff
@@ -0,0 +1,82 @@
...
+
+(defgeneric sign-executable (executable private-key)
+  (:documentation "Sign the EXECUTABLE with PRIVATE-KEY. This should set the signature with SIGNATURE."))
+
+(defgeneric verify-signature (executable public-key)
+  (:documentation "Verify the EXECUTABLE was signed by the owner of PUBLIC-KEY."))
+
+;;; The COMPILATION protocol
+
+(defgeneric compile-to-backend (program chip-spec backend)
```
kilimanjaro, line 62: Might also take extra data.

src/backends/common.lisp
```diff
@@ -0,0 +1,82 @@
...
+  (mapcar #'class-name (c2mop:class-direct-subclasses (find-class 'backend))))
+
+(defconstant +binary-hash-size+ 64)
+
+(defun binary-hash (binary)
+  (ironclad:digest-sequence ':sha3 binary))
+
+(defconstant +rsa-recommended-size-bits+ 3072)
+
+(defun generate-key-pair ()
```
appleby, line 81:
```suggestion
(defun generate-rsa-key-pair (&optional (num-bits +rsa-recommended-size-bits+))
```

Maybe `WARN` if `(< num-bits +rsa-recommended-size-bits+)`. In general I'm in favor of picking sane defaults, but letting the caller change the key size seems reasonable.

jmbr:
> Some big TODO items:
> 
>     * Parse out RSA keys for signage.
> 

Why use RSA at all? Do we want the signatures to be cracked by a future quantum computer?

appleby:
> > Some big TODO items:
> > ```
> > * Parse out RSA keys for signage.
> > ```
> 
> Why use RSA at all? Do we want the signatures to be cracked by a future quantum computer?

My take on this was that user has an escape hatch in that they can just not provide a key and quilc will skip the signing and then they can use whatever external tool or algorithm they want.

I don't have any opinion on whether RSA is a good choice for the default, but I do like the idea of just picking some default (or maybe very small set of choices), and then allowing the user to handle signing themselves if they don't like our defaults.

jmbr:
> > Why use RSA at all? Do we want the signatures to be cracked by a future quantum computer?
> 
> My take on this was that user has an escape hatch in that they can just not provide a key and quilc will skip the signing and then they can use whatever external tool or algorithm they want.

I'm on board with that but not necessarily with choosing RSA in particular.

appleby:
> > > Why use RSA at all? Do we want the signatures to be cracked by a future quantum computer?
> > 
> > 
> > My take on this was that user has an escape hatch in that they can just not provide a key and quilc will skip the signing and then they can use whatever external tool or algorithm they want.
> 
> I'm on board with that but not necessarily with choosing RSA in particular.

fwiw, it looks like ironclad also supports `ed25519`, which according to my extensive 5-minute google search just now is considered a Cryptographic Best Practice ™️

src/backends/common.lisp
stylewarning: It only has to do with backends insofar as whatever executable format you choose must be able to support embedding the signature in some way. I could see some other structure for signing (e.g., subclassing a `signable-mixin` class or so).

src/backends/common.lisp
stylewarning: No reason, I just read some recommendation somewhere and used it. Happy to defer to someone more expertly, such as yourself.

src/backends/common.lisp
stylewarning: I provide just an example hash function that works out of the box. Anybody could use their own.

app/src/entry-point.lisp
stylewarning: I expected the instance of `backend` would contain this info. For example, suppose we had something like a `generic-awg-backend`, then at some point the configuration of AWGs would be described by `backend`.

`backend` should be seem more than just a name; it should contain the full configuration details being compiled to. As such, you could imagine providing this short of thing at the command line, like `--backend my-awg-configuration.sexp` or whatever.

src/backends/common.lisp
stylewarning: (see above or wherever)

app/src/entry-point.lisp
```diff
@@ -44,6 +44,21 @@
      :optional t
      :initial-value "8Q"
      :documentation "set ISA to one of \"8Q\", \"20Q\", \"16QMUX\", \"bristlecone\", \"ibmqx5\", or path to QPU description file")
+    
+    (("backend")
+     :type string
+     :optional t
+     :documentation "set the backend for which to generate code (default: none). Note that an output file (-o) must be specified!")
```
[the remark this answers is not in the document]
stylewarning: I am going to guess we should keep them separate, but maybe not. `hmm.jpg.gz`

stylewarning:
I wouldn't take the signature stuff too seriously yet. The backend stuff is more important.

kilimanjaro:
> `ed25519`

This Ed guy really figured out everything.

appleby:
> > `ed25519`
> 
> This Ed guy really figured out everything.

Yes, but he was standing on the shoulders of Ed 25518

stylewarning:
Superseded by #674

PR #642, opened 2020-05-20
Cache chip specs between server requests

notmgsk:
This PR adds a chip cache used by the RPC endpoints, so the benefit is mostly felt by pyquil users. Produces a speedup between ~3s and ~4s based on some simple tests. You'll notice below that even a very simple compilation still takes an amount of time that scales with the size of the chip -- this is because the addresser state object is rebuilt on every compilation. Caching this also would provide a very nice win. For a future PR.

For example:
```
from pyquil import get_qc, Program
from time import time
qc = get_qc("30q-qvm", as_qvm=True)
p = Program("""CNOT 0 1""")
for i in range(4):
    t = time()
    print(qc.compile(p).program)
    print(f"iteration {i} took \t {time() - t}s")

RZ(-pi/2) 1
RX(pi/2) 1
CZ 1 0
RZ(pi) 0
RX(pi/2) 0
RX(-pi/2) 0
RX(-pi/2) 1
RZ(pi/2) 1
HALT

iteration 0 took 	 3.0537497997283936s
RZ(-pi/2) 1
RX(pi/2) 1
CZ 1 0
RZ(pi) 0
RX(pi/2) 0
RX(-pi/2) 0
RX(-pi/2) 1
RZ(pi/2) 1
HALT

iteration 1 took 	 0.9757657051086426s
RZ(-pi/2) 1
RX(pi/2) 1
CZ 1 0
RZ(pi) 0
RX(pi/2) 0
RX(-pi/2) 0
RX(-pi/2) 1
RZ(pi/2) 1
HALT

iteration 2 took 	 1.0833849906921387s
RZ(-pi/2) 1
RX(pi/2) 1
CZ 1 0
RZ(pi) 0
RX(pi/2) 0
RX(-pi/2) 0
RX(-pi/2) 1
RZ(pi/2) 1
HALT

iteration 3 took 	 0.9600250720977783s
```
and
```
qc = get_qc("Aspen-8", as_qvm=True)
p = Program("""CNOT 0 1""")
for i in range(4):
    t = time()
    print(qc.compile(p).program)
    print(f"iteration {i} took \t {time() - t}s")

RZ(pi/2) 33
RX(-pi/2) 33
CZ 33 32
RZ(pi) 32
RX(pi/2) 33
RZ(-pi/2) 33
HALT

iteration 0 took 	 0.44094109535217285s
RZ(-pi/2) 33
RX(pi/2) 33
CZ 33 32
RZ(pi) 32
RX(-pi/2) 33
RZ(pi/2) 33
HALT

iteration 1 took 	 0.11467194557189941s
RZ(-pi/2) 33
RX(pi/2) 33
CZ 33 32
RZ(pi) 32
RX(-pi/2) 33
RZ(pi/2) 33
HALT

iteration 2 took 	 0.11051273345947266s
RZ(-pi/2) 33
RX(pi/2) 33
CZ 33 32
RZ(pi) 32
RX(-pi/2) 33
RZ(pi/2) 33
HALT

iteration 3 took 	 0.11118412017822266s
```

### DIFF
--- a/app/src/chip-cache.lisp
+++ b/app/src/chip-cache.lisp
@@ -1,0 +1,70 @@
+;;;; chip-cache.lisp
+;;;;
+;;;; Chip caching for faster compilation requests
+;;;;
+;;;; Author: Mark Skilbeck
+;;;;
+
+(in-package #:quilc)
+
+(defclass cached-chip ()
+  ((last-accessed
+    :initarg :last-accessed
+    :type real
+    :accessor cached-chip-last-accessed)
+   (chip
+    :initarg :chip
+    :type chip-specification
+    :accessor cached-chip-chip)
+   (addresser-state
+    :initarg :addresser-state
+    :type quil::addresser-state
+    :accessor cached-chip-addresser-state))
+  (:documentation "Represents chip and addresser state cached between server requests."))
+
+(defvar *chip-cache*
+  (cons (make-hash-table :test #'equalp)
+        (bt:make-lock "chip cache lock"))
+  "Cached chip specifications. Pair of values (hash-table . lock). Large chips have a significant construction overhead, and caching chips between requests reduces or eliminates that overhead. The hash value for a given key is a pair (last-access . chip-spec). Note: use the lock when accessing.")
+
+(defvar *chip-cache-max-size* 10
+  "The maximum number of entries in the cache.")
+
+(defun get-internal-real-time-seconds ()
+  (/ (get-internal-real-time)
+     internal-time-units-per-second))
+
+(defun chip-cache-purge ()
+  "Purge the least recently used entries in *CHIP-CACHE* according to *CHIP-CACHE-MAX-SIZE*.
+
+After calling this function, *CHIP-CACHE* has at most *CHIP-CACHE-MAX-SIZE* entries."
+  (let ((cache (car *chip-cache*)))
+    (when (>= (hash-table-count cache) *chip-cache-max-size*)
+      (let* ((n (- (hash-table-count cache) *chip-cache-max-size*))
+             (lru (subseq (sort (a:hash-table-alist cache)
+                                #'<
+                                :key (a:compose #'cached-chip-last-accessed #'cdr))
+                          0 n)))
+        (dolist (cached lru)
+          (remhash (car cached) cache))))))
+
+(defun chip-cache-or-create (qpu-hash)
+  "Look up the chip described by QPU-HASH in the chip spec cache if it exists, otherwise create and cache it.
+
+This function has the added side-effect that it will purge old chips according to *CHIP-CACHE-MAX-SIZE*"
+  (let* ((cache (car *chip-cache*))
+         (cached-chip (gethash qpu-hash cache)))
+    (cond
+      (cached-chip
+       (setf (cached-chip-last-accessed cached-chip)
+             (get-internal-real-time-seconds))
+       cached-chip)
+      (t
+       (let* ((chip (quil::qpu-hash-table-to-chip-specification qpu-hash))
+              (cached-chip (make-instance 'cached-chip
+                                          :last-accessed (get-internal-real-time-seconds)
+                                          :chip chip)))
+         (setf (gethash qpu-hash cache)
+               cached-chip)
+         (chip-cache-purge)
+         cached-chip)))))
--- a/app/src/entry-point.lisp
+++ b/app/src/entry-point.lisp
@@ -218,6 +218,7 @@
                           (server-mode-rpc nil)
                           (host nil)
                           (port nil)
+                          (cache-chips nil)
                           time-limit
                           (help nil)
                           (log-level nil)
@@ -326,7 +327,8 @@
        (*quil-stream* (make-broadcast-stream))
        (*protoquil* protoquil)
        (*state-aware* enable-state-prep-reductions)
-       (quil::*safe-include-directory* safe-include-directory))
+       (quil::*safe-include-directory* safe-include-directory)
+       (*chip-cache-max-size* cache-chips))
 
     (when check-sdk-version
       (asynchronously-indicate-update-availability +QUILC-VERSION+ :proxy proxy))
--- a/app/src/rpc-server.lisp
+++ b/app/src/rpc-server.lisp
@@ -55,7 +55,9 @@
          (qpu-hash (a:plist-hash-table (list "isa" (rpcq::|TargetDevice-isa| target-device)
                                              "specs" (rpcq::|TargetDevice-specs| target-device))
                                        :test #'equal))
-         (chip-specification (cl-quil::qpu-hash-table-to-chip-specification qpu-hash))
+         (cache (bt:with-lock-held ((cdr *chip-cache*))
+                  (chip-cache-or-create qpu-hash)))
+         (chip-specification (cached-chip-chip cache))
          ;; Allow endpoint to override server's -P
          (protoquil (ecase protoquil
                       ((nil) *protoquil*)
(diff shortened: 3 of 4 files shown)


app/src/rpc-server.lisp
```diff
@@ -45,6 +45,40 @@
                  :|topological_swaps| (gethash "topological_swaps" statistics)
                  :|qpu_runtime_estimation| (gethash "qpu_runtime_estimation" statistics)))
 
+(defvar *chip-cache* (make-hash-table :test #'equalp)
+  "Cached chip specifications. Large chips have a significant construction overhead, and caching chips between requests reduces or eliminates that overhead. The hash value for a given key is a pair (last-access . chip-spec).")
+
+(defvar *chip-cache-purge-age* (* 60 60)
```
kilimanjaro, line 51: I would have a preference for a clear upper bound on the size of `*chip-cache*`, and so would suggest that rather than controlling this with a timeout we have a `*chip-cache-max-size*`, and when this is reached do a LRU purge.

kilimanjaro: Not to suggest that this is really a big issue either way -- doubtful that we are going to have a lot of different chip objects floating around.

app/src/rpc-server.lisp
```diff
@@ -45,6 +45,44 @@
                  :|topological_swaps| (gethash "topological_swaps" statistics)
                  :|qpu_runtime_estimation| (gethash "qpu_runtime_estimation" statistics)))
 
+(defvar *chip-cache* (make-hash-table :test #'equalp)
```
stylewarning, line 48: I would add a lock around all interactions with this table.

app/src/entry-point.lisp
```diff
@@ -92,6 +92,11 @@
      :initial-value 0
      :documentation "time limit (in seconds) for server requests (0 => unlimited)")
 
+    (("cache-chips")
```
stylewarning, line 95: can we dedicate a separate file to all the chip cache logic `chip-cache.lisp`?

app/src/rpc-server.lisp
```diff
@@ -45,6 +45,44 @@
...
+
+
+(defun chip-cache-or-create (qpu-hash)
+  "Look up the chip described by QPU-HASH in the chip spec cache if it exists, otherwise create and cache it.
+
+This function has the added side-effect that it will purge old chips according to *CHIP-CACHE-MAX-SIZE*"
+  (let ((cache (gethash qpu-hash *chip-cache*)))
+    (cond
+      (cache
+       (setf (car cache) (get-internal-real-time-seconds))
```
stylewarning, line 76: RPLACA or bust

)nobody ever agrees with me :(

appleby: did somebody say... RPLACA??

appleby: doit

notmgsk: no

notmgsk: (with respect)

app/src/chip-cache.lisp
```diff
@@ -0,0 +1,63 @@
...
+		  (remhash (car cached) cache))))))
+
+(defun chip-cache-or-create (qpu-hash)
+  "Look up the chip described by QPU-HASH in the chip spec cache if it exists, otherwise create and cache it.
+
+This function has the added side-effect that it will purge old chips according to *CHIP-CACHE-MAX-SIZE*"
+  (let* ((cache (car *chip-cache*))
+		 (cached-chip (gethash qpu-hash cache)))
+    (cond
+	  (cached-chip
```
ecpeterson, line 51: No collision check?

app/src/chip-cache.lisp
```diff
@@ -0,0 +1,63 @@
...
+     internal-time-units-per-second))
+
+(defun chip-cache-purge ()
+  "Purge the least recently used entries in *CHIP-CACHE* according to *CHIP-CACHE-MAX-SIZE*.
+
+After calling this function, *CHIP-CACHE* has at most *CHIP-CACHE-MAX-SIZE* entries."
+  (let ((cache (car *chip-cache*)))
+	(when (>= (hash-table-count cache) *chip-cache-max-size*)
+	  (let* ((n (- (hash-table-count cache) *chip-cache-max-size*))
+			 (lru (subseq (sort (a:hash-table-alist cache)
```
ecpeterson, line 37: indentation

app/src/rpc-server.lisp
```diff
@@ -217,4 +224,5 @@
     (rpcq:start-server :dispatch-table dt
                        :listen-addresses (list (format nil "~A://~A~@[:~A~]" protocol host port))
                        :logger logger
-                       :timeout time-limit)))
+                       :timeout time-limit
+					   :debug t)))
```
ecpeterson, line 228: indentation

app/src/chip-cache.lisp
notmgsk: What should happen in the event of a collision? Just avoid the caching part?

app/src/rpc-server.lisp
notmgsk: That's what I get for starting-from-scratch my emacs config.

app/src/chip-cache.lisp
ecpeterson: I take it back: I had in my head that you were maintaining an LRU-sorted list, dropping the last element, and testing for equality using a more destructive hash. Actually, you're using an actual hash table, which will manage its own collisions, and the keys you're using perfectly determine the generated chip-spec object.

app/src/rpc-server.lisp
```diff
@@ -55,8 +55,10 @@
          (qpu-hash (a:plist-hash-table (list "isa" (rpcq::|TargetDevice-isa| target-device)
                                              "specs" (rpcq::|TargetDevice-specs| target-device))
                                        :test #'equal))
-         (chip-specification (cl-quil::qpu-hash-table-to-chip-specification qpu-hash))
-         ;; Allow endpoint to override server's -P
+         (cache (bt:with-lock-held ((cdr *chip-cache*))
+                  (chip-cache-or-create qpu-hash)))
+         (chip-specification (cached-chip-chip cache))
+         ;; Allow endpoint to override server's -Pnnn
```
colescott, line 61: Looks like a mistake while trying to navigate down (emacs is hard)

app/src/chip-cache.lisp
```diff
@@ -0,0 +1,63 @@
+(in-package #:quilc)
```
colescott, line 1: There should be a file header here to be consistent with the rest of quilc

colescott (dismissed):
Aside from the two minor things, looks good to me. Those are very impressive benchmarks!

stylewarning:
Rebase & fix conflicts, test one more time, and merge.

app/src/chip-cache.lisp
```diff
@@ -0,0 +1,70 @@
...
+    :type chip-specification
+    :accessor cached-chip-chip)
+   (addresser-state
+    :initarg :addresser-state
+    :type quil::addresser-state
+    :accessor cached-chip-addresser-state))
+  (:documentation "Represents chip and addresser state cached between server requests."))
+
+(defvar *chip-cache*
+  (cons (make-hash-table :test #'equalp)
```
stylewarning, line 26: `(make-hash-table :test ... :size *max-size* :growth-factor 0)` (double check and read carefully the data type implications of the growth factor argument!)

erik sez the arg is `:rehash-size`

stylewarning: (maybe make the size `(* *max-size* *slop-factor*)` where `(> *slop-factor* 1.0)`

app/src/chip-cache.lisp
```diff
@@ -0,0 +1,70 @@
+;;;; chip-cache.lisp
+;;;;
+;;;; Chip caching for faster compilation requests
+;;;;
+;;;; Author: Mark Skilbeck
+;;;;
+
+(in-package #:quilc)
+
+(defclass cached-chip ()
```
stylewarning, line 10: _consider_ an alternative data structure instead of hash tables

stylewarning (commented):
I didn't really review the hash table code

notmgsk:
The table is not intended to grow. It's size is fixed and configured at launch. A hash is maybe overkill, but it is a simple implementation and if it's all the same I'd like to leave it as-is.

notmgsk:
Merging.

PR #647, opened 2020-06-17
Add GENERATE-NATURAL-QAOA-PROGRAM to benchmarking package

notmgsk:
Moves some benchmarking-related stuff into the benchmarking package.


benchmarking/qaoa-tests/qaoa-bench.lisp
```diff
@@ -27,7 +27,7 @@
     (format t "~&~A: ~F (~F)~%" name mean sdev)))
 
 (load (asdf:system-relative-pathname
```
kilimanjaro, line 29: If `generate-program.lisp` is part of the `cl-quil-benchmarking` system definition, why are we also loading it here?

notmgsk: I could remove that, but didn't want to change stuff. Really I only want the generate-... function exposed.

PR #649, opened 2020-07-09
Fix compiler path calculation

### LINKED ISSUE #643: Compiler generically always uses CZ instead of a mix of XY and CZ
If I use a device that supports both CZ and XY gates, e.g. (currently) Aspen-4, and if I provide explicit commands to use either the CZ or XY gate, then the compiler knows to use those gates. For example, `CZ 0 1` compiles to `CZ 0 1` and `XY(0.3) 0 1` compiles to `XY(0.3) 0 1`.

For generic programs that may not be written in terms of the native gate set, the compiler doesn't seem to make use of the XY gate. For example, the following QAOA circuit borrowed from https://arxiv.org/abs/1912.04424:
```
PRAGMA INITIAL_REWIRING "NAIVE"
DECLARE beta REAL[1]
DECLARE gamma REAL[1]
DECLARE ro BIT[4]
H 1
H 2
H 3
H 0
CNOT 1 2
RZ(0.9554352891425398*gamma) 2
CNOT 1 2
CNOT 1 0
RZ(1.0202671293389616*gamma) 0
CNOT 1 0
CNOT 2 3
RZ(-0.7027488587297126*gamma) 3
CNOT 2 3
CNOT 3 0
RZ(1.2459303490186089*gamma) 0
CNOT 3 0
H 1
RZ(2.0*beta) 1
H 1
H 2
RZ(2.0*beta) 2
H 2
H 3
RZ(2.0*beta) 3
H 3
H 0
RZ(2.0*beta) 0
H 0
MEASURE 1 ro[0]
MEASURE 2 ro[1]
MEASURE 3 ro[2]
MEASURE 0 ro[3]
```
compiles to a circuit that uses only CZ and no XYs, even though the paper claims that there are more optimal choices of the compilation if one uses both those gates.

### LINKED ISSUE #644: Compiler doesn't recognize SWAP ~ CZ+iSWAP trick
If I run a SWAP instruction on a device that has both CZ and XY (and therefore, XY(pi) = iSWAP) gates, then I should be able to use the trick identified in https://arxiv.org/abs/quant-ph/0209035 to use one CZ and one iSWAP to compile a SWAP gate, instead of 3 CZs or 3 XYs. However, the compiler generically just compiles a program like the following using 3 CZs, even though it has access to XY gates as well.

```
from pyquil import get_qc, Program

print(get_qc('Aspen-4', as_qvm=True).compile(Program("SWAP 0 1")).program)
```

notmgsk:
The computed compiler paths were:

  (1) not returning the shortest path, though the function claimed to do
  so; and
  (2) not including XY compilers, even if the target gateset included an
  XY gate that was better than a CZ gate on the same hardware object

This introduces a separate issue: arbitrarily long sequences of rx and
rz gates are not collapsed into at most five gates, but now collapse
into at most 7 gates. This is due to the compiler thinking that the
EULER-YZY compiler is better than the EULER-ZXZ compiler. I can live
with this regression for now, since the two extra gates don't cost much
in terms of fidelity, when compared to the improvement of allowing for
high-fidelity XY gates.

An extra nicety: we can now support SWAPs that compile to just two 2Q
gates (one XY and one CZ, cf. three XY or three CZ).

Closes #643. Closes #644.

### DIFF
--- a/src/compilers/euler-compile.lisp
+++ b/src/compilers/euler-compile.lisp
@@ -39,42 +39,42 @@
                (inst ,inner-gate `(,(* ,inner-prefactor 2 (first ,angles)))
                      q)
                (inst ,outer-gate `(,(* ,outer-prefactor
-				       (- (phase (magicl:tref ,u1 0 0))
-					  (phase (magicl:tref ,u0 0 0)))))
-		     q))))))))
+                                       (- (phase (magicl:tref ,u1 0 0))
+                                          (phase (magicl:tref ,u0 0 0)))))
+                     q))))))))
 
 (define-euler-compiler euler-YXY-compiler
-    :outer-gate "RY"
-    :inner-gate "RX"
-    :prefix-quil (list (build-gate "RX" `(,-pi/2) 0)
-                       (build-gate "RY" `(,-pi/2) 0)))
+  :outer-gate "RY"
+  :inner-gate "RX"
+  :prefix-quil (list (build-gate "RX" `(,-pi/2) 0)
+                     (build-gate "RY" `(,-pi/2) 0)))
 
 (define-euler-compiler euler-ZXZ-compiler
-    :outer-gate "RZ"
-    :inner-gate "RX"
-    :prefix-quil (list (build-gate "RZ" `(,-pi/2) 0)))
+  :outer-gate "RZ"
+  :inner-gate "RX"
+  :prefix-quil (list (build-gate "RZ" `(,-pi/2) 0)))
 
 (define-euler-compiler euler-XYX-compiler
-    :outer-gate "RX"
-    :inner-gate "RY"
-    :outer-prefactor -1
-    :prefix-quil (list (build-gate "RY" `(,-pi/2) 0)))
+  :outer-gate "RX"
+  :inner-gate "RY"
+  :outer-prefactor -1
+  :prefix-quil (list (build-gate "RY" `(,-pi/2) 0)))
 
 (define-euler-compiler euler-XZX-compiler
-    :outer-gate "RX"
-    :inner-gate "RZ"
-    :outer-prefactor -1
-    :inner-prefactor -1
-    :prefix-quil (list (build-gate "RY" `(,-pi/2) 0)
-                       (build-gate "RX" `(,-pi/2) 0)))
+  :outer-gate "RX"
+  :inner-gate "RZ"
+  :outer-prefactor -1
+  :inner-prefactor -1
+  :prefix-quil (list (build-gate "RY" `(,-pi/2) 0)
+                     (build-gate "RX" `(,-pi/2) 0)))
 
 (define-euler-compiler euler-YZY-compiler
-    :outer-gate "RY"
-    :inner-gate "RZ"
-    :inner-prefactor -1
-    :prefix-quil (list (build-gate "RX" `(,-pi/2) 0)))
+  :outer-gate "RY"
+  :inner-gate "RZ"
+  :inner-prefactor -1
+  :prefix-quil (list (build-gate "RX" `(,-pi/2) 0)))
 
 (define-euler-compiler euler-ZYZ-compiler
-    :outer-gate "RZ"
-    :inner-gate "RY"
-    :prefix-quil ())
+  :outer-gate "RZ"
+  :inner-gate "RY"
+  :prefix-quil ())
--- a/src/compilers/translators.lisp
+++ b/src/compilers/translators.lisp
@@ -358,3 +358,12 @@ Note that if (= START-NODE TARGET-NODE) then (list START-NODE) is returned."
   (inst "H"    nil              p)
   (inst "RZ"   (list (/ -pi 4)) p)
   (inst "RZ"   (list (/ pi 4))  q))
+
+(define-compiler SWAP-to-iSWAP+CNOT
+    ((swap-gate ("SWAP" () q0 q1)))
+  (inst "CNOT"  () q0 q1)
+  (inst "H"     () q1)
+  (inst "RZ"    (list -pi/2) q0)
+  (inst "RZ"    (list -pi/2) q1)
+  (inst "XY"    (list pi) q0 q1)
+  (inst "H"     () q0))
--- a/src/compressor/compressor.lisp
+++ b/src/compressor/compressor.lisp
@@ -617,8 +617,8 @@ other's."
       (let ((result-instructions
               (cond
                 ((and decompiled-instructions
-                      (< (calculate-instructions-duration reduced-decompiled-instructions chip-specification)
-                         (calculate-instructions-duration reduced-instructions chip-specification)))
+                      (>= (calculate-instructions-fidelity reduced-decompiled-instructions chip-specification)
+                          (calculate-instructions-fidelity reduced-instructions chip-specification)))
                  reduced-decompiled-instructions)
                 (t
                  reduced-instructions))))
--- a/src/define-compiler.lisp
+++ b/src/define-compiler.lisp
@@ -108,7 +108,7 @@ OPTIONS: plist of options governing applicability of the compiler binding."
      (a:when-let ((arguments (gate-binding-arguments obj)))
        (format stream "~{ ~A~}" arguments)))))
 
-(defun get-binding-from-instr (instr)
+(defun binding-from-instr (instr)
   "Constructs a COMPILER-BINDING object from an INSTRUCTION object, in such a way that if some auxiliary COMPILER-BINDING subsumes the output of this routine, then it will match when applied to the original INSTRUCTION object."
   (typecase instr
     (measure-discard
@@ -266,6 +266,8 @@ OPTIONS: plist of options governing applicability of the compiler binding."
                 (gate-binding-parameters binding)
                 (application-parameters gate)))
      ;; each (binding argument is not a wildcard => binding arg = gate arg)
+     (= (length (gate-binding-arguments binding))
+        (length (application-arguments gate)))
      (every (lambda (b g) (or (symbolp b) (equalp b (qubit-index g))))
             (gate-binding-arguments binding)
             (application-arguments gate))))
@@ -422,14 +424,16 @@ What \"concrete\" means depends on the types of A and B:
 (defun compiler-gateset-reducer-p (compiler)
   (getf (compiler-options compiler) ':gateset-reducer t))
 
-(defun get-compilers (qubit-bound)
+(defun get-compilers (&optional qubit-bound)
   "Returns a list of the available compilers which match on no more than QUBIT-BOUND qubits and which function as gateset reducers."
   (remove-if-not
    (lambda (compiler)
      (and (compiler-gateset-reducer-p compiler)
           (= 1 (length (compiler-bindings compiler)))
           (typep (first (compiler-bindings compiler)) 'gate-binding)
-          (>= qubit-bound (length (gate-binding-arguments (first (compiler-bindings compiler)))))
+          (if qubit-bound
+              (>= qubit-bound (length (gate-binding-arguments (first (compiler-bindings compiler)))))
+              t)
           (endp (compiler-binding-options (first (compiler-bindings compiler))))))
    **compilers-available**))
 
@@ -615,10 +619,9 @@ Optionally constrains the output to include only those bindings of a particular 
 Optionally restricts to considering only those gates and compilers which involve QUBIT-COUNT many qubits.
 
 N.B.: The word \"shortest\" here is a bit fuzzy.  In practice it typically means \"of least length\", but in theory this invariant could be violated by pathological compiler output frequency + fidelity pairings that push the least-length path further down the priority queue."
-  (let ((queue (make-instance 'cl-heap:priority-queue :sort-fun #'occurrence-table-metric-worsep)))
+  (let ((queue (make-instance 'cl-heap:priority-queue :sort-fun (complement #'occurrence-table-metric-worsep))))
     (flet ((collect-bindings (occurrence-table)
-             (loop :for key :being :the :hash-keys :of (occurrence-table-map occurrence-table)
-                   :collect key)))
+             (a:hash-table-keys (occurrence-table-map occurrence-table))))
       ;; initial contents: arbitrary gate, no history
       (loop :with visited-nodes := ()
             :for (task . history) := (list occurrence-table)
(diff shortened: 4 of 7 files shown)


src/compilers/euler-compile.lisp
```diff
@@ -39,42 +39,42 @@
                (inst ,inner-gate `(,(* ,inner-prefactor 2 (first ,angles)))
                      q)
                (inst ,outer-gate `(,(* ,outer-prefactor
-				       (- (phase (magicl:tref ,u1 0 0))
-					  (phase (magicl:tref ,u0 0 0)))))
-		     q))))))))
+                                       (- (phase (magicl:tref ,u1 0 0))
```
notmgsk, line 42: Just some indentation changes in this file.

src/compressor/compressor.lisp
```diff
@@ -617,8 +617,8 @@ other's."
       (let ((result-instructions
               (cond
                 ((and decompiled-instructions
-                      (< (calculate-instructions-duration reduced-decompiled-instructions chip-specification)
-                         (calculate-instructions-duration reduced-instructions chip-specification)))
+                      (>= (calculate-instructions-fidelity reduced-decompiled-instructions chip-specification)
```
notmgsk, line 620: This maybe ought to depend on the addresser type, but since it lives in the compressor I wasn't sure. (i.e. if `*default-addresser-state-class*` is `fidelity` or w/e, then use fidelity here; otherwise use duration)

src/define-compiler.lisp
```diff
@@ -108,7 +108,7 @@ OPTIONS: plist of options governing applicability of the compiler binding."
      (a:when-let ((arguments (gate-binding-arguments obj)))
        (format stream "~{ ~A~}" arguments)))))
 
-(defun get-binding-from-instr (instr)
+(defun binding-from-instr (instr)
```
notmgsk, line 111: Gratuitous name change. Will revert if too controversial

src/define-compiler.lisp
```diff
@@ -266,6 +266,8 @@ OPTIONS: plist of options governing applicability of the compiler binding."
                 (gate-binding-parameters binding)
                 (application-parameters gate)))
      ;; each (binding argument is not a wildcard => binding arg = gate arg)
+     (= (length (gate-binding-arguments binding))
```
notmgsk, line 269: This is here because `(every #'equalp (list 1 2 3) (list 1 2))` does not raise an error (lists of differing lengths), and evaluates to true.

src/define-compiler.lisp
```diff
@@ -422,14 +424,16 @@ What \"concrete\" means depends on the types of A and B:
...
-(defun get-compilers (qubit-bound)
+(defun get-compilers (&optional qubit-bound)
   "Returns a list of the available compilers which match on no more than QUBIT-BOUND qubits and which function as gateset reducers."
   (remove-if-not
    (lambda (compiler)
      (and (compiler-gateset-reducer-p compiler)
           (= 1 (length (compiler-bindings compiler)))
           (typep (first (compiler-bindings compiler)) 'gate-binding)
-          (>= qubit-bound (length (gate-binding-arguments (first (compiler-bindings compiler)))))
+          (if qubit-bound
```
notmgsk, line 434: Optional qubit bound. Just for convenience.

src/define-compiler.lisp
```diff
@@ -615,10 +619,9 @@ Optionally constrains the output to include only those bindings of a particular
 Optionally restricts to considering only those gates and compilers which involve QUBIT-COUNT many qubits.
 
 N.B.: The word \"shortest\" here is a bit fuzzy.  In practice it typically means \"of least length\", but in theory this invariant could be violated by pathological compiler output frequency + fidelity pairings that push the least-length path further down the priority queue."
-  (let ((queue (make-instance 'cl-heap:priority-queue :sort-fun #'occurrence-table-metric-worsep)))
+  (let ((queue (make-instance 'cl-heap:priority-queue :sort-fun (complement #'occurrence-table-metric-worsep))))
```
notmgsk, line 622: The queue was dequeueing "worse" occurrence tables, rather than "better" occurrence tables. This reverses that, and fixes the main issue of compiler paths being longer than they need to be.

src/define-compiler.lisp
```diff
@@ -648,89 +651,90 @@ N.B.: The word \"shortest\" here is a bit fuzzy.  In practice it typically means
        (dohash ((key val) item)
          (format t "~/cl-quil::binding-fmt/ -> ~A~%" key val))))))
 
+(defun candidate-special-compilers (target-gateset qubit-count)
```
notmgsk, line 654: These are lifted to aide in debugging. I find them useful per se, but will stuff them back in to `compute-applicable-compilers` if controversial.

src/define-compiler.lisp
```diff
@@ -648,89 +651,90 @@ N.B.: The word \"shortest\" here is a bit fuzzy.  In practice it typically means
...
-                    unconditional-compilers target-gateset
-                    (filter-occurrence-table-by-qubit-count (compiler-output-gates compiler) qubit-count)
-                    qubit-count))
-                 (special-cost (occurrence-table-metric-fidelity
-                                (evaluate-occurrence-table (first special-path) target-gateset))))
-            ;; did we in fact beat out the generic machinery?
-            (when (and (not (path-has-a-loop-p special-path (compiler-bindings compiler)))
-                       ;; TODO hmm
-                       (>= special-cost generic-cost))
+  (let* ((target-gateset (blank-out-qubits target-gateset))
```
notmgsk, line 678: Nothing has changed here, other than lifting the above functions.

notmgsk, line 678: Nothing has changed here other than lifting the above functinos.

tests/compilation-tests.lisp
```diff
@@ -250,7 +250,7 @@ CNOT 4 8
                      (%read-test-chipspec "Aspen-6-2Q-A.qpu")
                      (%read-test-chipspec "Aspen-7-28Q-A.qpu"))))
     (dolist (chip chips)
-      (%test-reduction-with-chip 5 chip))))
+      (%test-reduction-with-chip 7 chip))))
```
notmgsk, line 253: This is a lil sad, but will be remedied in a future PR.

notmgsk:
Ping @ecpeterson

src/compressor/compressor.lisp
ecpeterson: This is a nice find; it indicates that "address state class" is a misnomer. It'd be better to preserve that class instance for use here and to deploy the relevant cost-function.

src/define-compiler.lisp
ecpeterson: lgtm

src/define-compiler.lisp
ecpeterson: I'd hope that the first guard + the Quil parser would capture this in all relevant cases, but I could be wrong, and there's no harm in being safe.

src/define-compiler.lisp
ecpeterson: I'm surprised / dismayed / skeptical that this error survived for a year. How is it possible that the compiler has been deliberately choosing worse compiler paths and no one noticed? I hope that there's a unit test introduced in this PR which guards against this kind of mistake in the future.

src/define-compiler.lisp
```diff
@@ -615,10 +619,9 @@ Optionally constrains the output to include only those bindings of a particular
 Optionally restricts to considering only those gates and compilers which involve QUBIT-COUNT many qubits.
 
 N.B.: The word \"shortest\" here is a bit fuzzy.  In practice it typically means \"of least length\", but in theory this invariant could be violated by pathological compiler output frequency + fidelity pairings that push the least-length path further down the priority queue."
-  (let ((queue (make-instance 'cl-heap:priority-queue :sort-fun #'occurrence-table-metric-worsep)))
+  (let ((queue (make-instance 'cl-heap:priority-queue :sort-fun (complement #'occurrence-table-metric-worsep))))
     (flet ((collect-bindings (occurrence-table)
-             (loop :for key :being :the :hash-keys :of (occurrence-table-map occurrence-table)
-                   :collect key)))
+             (a:hash-table-keys (occurrence-table-map occurrence-table))))
```
ecpeterson, line 624: :) 👍 I've since learned about some `alexandria` shorthands that I didn't know about in 2019.

src/define-compiler.lisp
ecpeterson: I think `c-a-c` (and some other functions in this file) are too large for their own good, and splitting them apart is fine.

tests/compilation-tests.lisp
ecpeterson: This makes me disbelieve in this PR. Is this related to the "too many 1Q gates" behavior you mentioned?

src/define-compiler.lisp
ecpeterson: It looks like #591 flipped some of the inequalities in the definition of `occurrence-table-metric-worsep`, but not all of them. Was that deliberate? Could that mean that neither `o-t-m-w` nor its complement are appropriate, bc the definition itself is curiously halfway right?

src/define-compiler.lisp
notmgsk: This is actually unrelated to the current PR. I was writing some stuff where I used this function directly, thus sidestepping the parser, etc. and so had to include the guard.

src/define-compiler.lisp
notmgsk: >I'm surprised / dismayed / skeptical that this error survived for a year. How is it possible that the compiler has been deliberately choosing worse compiler paths and no one noticed? I hope that there's a unit test introduced in this PR which guards against this kind of mistake in the future.

I think (to allay your skepticism) that it is true that longer paths have been an issue, as shown in [this gist](https://gist.github.com/notmgsk/1ed9cf7eaca68df8eff6546e22a6317c), but it didn't necessarily matter because in `compute-applicable-compilers` there is a second pass at finding appropriate compilers which would then include compilers not in the "shortest" compiler path. e.g. in the above gist, the shortest path (actually a rather long path) goes from a generic gate up to XY gates, without using the CZ-outputting compilers.

I will add UT, though in a follow-on PR.

> It looks like #591 flipped some of the inequalities in the definition of occurrence-table-metric-worsep, but not all of them. Was that deliberate? Could that mean that neither o-t-m-w nor its complement are appropriate, bc the definition itself is curiously halfway right?

Yeah. The original reads strange to me. It says [that an occurrence table "costs" more if it has better fidelity](https://github.com/rigetti/quilc/pull/591/files#diff-f9cd5b27a0930abbf5b0372a7fc40159L548-L549)? Anyway, even prior to that PR, `find-shortest-compiler-path` was not producing the shortest compiler path.

ecpeterson: That may be on me: some places the code uses fidelity and multiplication, some places it uses log-infidelity and addition, and the 'worse' inequality is opposite between the two. Perhaps I forgot which was in play here.

tests/compilation-tests.lisp
notmgsk: @ecpeterson Yes. Before #591, the shortest compiler path for a typical x-halves chip would be:
```
(#<OCCURRENCE-TABLE 
  RX (1.5707963267948966d0) _ -> 2
  RZ (_) _ -> 3
  RX (-1.5707963267948966d0) _ -> 2>
 #<COMPILER RY-TO-XZX {1003333A9B}>
 #<OCCURRENCE-TABLE 
  RY (_) _ -> 2
  RZ (_) _ -> 1>
 #<COMPILER EULER-YZY-COMPILER {10033336BB}>
 #<OCCURRENCE-TABLE 
  
_  _ -> 1>)
```
and the full compiler list would be:
```
(#<COMPILER STATE-PREP-1Q-COMPILER {10033326CB}>
 #<COMPILER ELIMINATE-IDENTITIES {1008184C7B}>
 #<COMPILER PHASE-TO-RZ {1003333ADB}> #<COMPILER RY-TO-XZX {1003333A9B}>
 #<COMPILER RX-TO-ZXZXZ {1003333ABB}>
 #<COMPILER EULER-YZY-COMPILER {10033336BB}>)
```
Which is missing the `euler-zxz` compiler. In total, you'd end up with long strings of single qubit gates.

That was "fixed" in #591 at the expense of breaking the "shortest" claim:
```
(#<OCCURRENCE-TABLE 
  RZ (_) _ -> 63
  RZ (1.5707963267948966d0) _ -> 21
  RX (1.5707963267948966d0) _ -> 21
  RX (-1.5707963267948966d0) _ -> 21
  RZ (-1.5707963267948966d0) _ -> 21>
 #<COMPILER RX-TO-ZXZXZ {100818BEDB}>
 #<OCCURRENCE-TABLE 
  RZ (_) _ -> 42
  RX (_) _ -> 21>
 #<COMPILER EULER-ZXZ-COMPILER {100818BA7B}>
 #<OCCURRENCE-TABLE 
  RZ (_) _ -> 14
  RY (_) _ -> 7>
 #<COMPILER EULER-ZYZ-COMPILER {100818BAFB}>
 #<OCCURRENCE-TABLE 
  RX (1.5707963267948966d0) _ -> 2
  RZ (_) _ -> 2
  RX (-1.5707963267948966d0) _ -> 2
  RX (_) _ -> 1>
 #<COMPILER RY-TO-XZX {100818BEBB}>
 #<OCCURRENCE-TABLE 
  RY (_) _ -> 2
  RX (_) _ -> 1>
 #<COMPILER EULER-YXY-COMPILER {100818BA5B}>
 #<OCCURRENCE-TABLE 
  
_  _ -> 1>)
```
Which includes the `euler-zxz` compiler, thus being able to better reduce long strings of single qubit gates. Tho it still wasn't perfect in retrospect. I just now see that in both cases if you fed in a string of alternating *native* RZ/RXs you would end up with a similarly long result; but #591 fixed the case where the RXs were *non-native*. Weird.

Anyway, this PR makes the upper-bound 7 (on both native and non-native RZ/RX strings). I solemnly promise to fix this in a follow-on PR.

src/compilers/translators.lisp
```diff
@@ -358,3 +358,12 @@ Note that if (= START-NODE TARGET-NODE) then (list START-NODE) is returned."
   (inst "H"    nil              p)
   (inst "RZ"   (list (/ -pi 4)) p)
   (inst "RZ"   (list (/ pi 4))  q))
+
+(define-compiler SWAP-to-iSWAP+CNOT
```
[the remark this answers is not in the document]
notmgsk: @ecpeterson yes, that is the case.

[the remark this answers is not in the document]
notmgsk: Again, happy to investigate the reason why but would rather it didn't gate approval of this PR.

notmgsk:
I got out-of-band approval of this from Eric. Have made a few follow-on issues to address his comments.

PR #655, opened 2020-07-21
Account for number of gates of unknown fidelity in EVAL-OCC-TABLE

colescott:
Previously, a gate with unknown fidelity and n occurrences was
recorded in the metric as having one unknown. This means an occurrence
table of `{RX: 1, RY: 50}` with `RX` known and `RY` unknown would have a
better metric than `{RX: 2, RY: 1}` since the `RY` in both cases is recorded
as 1 unknown. This causes issues in determining short compiler paths.

This change now records the former table with 50 unknowns
and the latter with 1 unknown. This change also adds additional
documentation to the `OCCURRENCE-TABLE` structure to make this clear.

`TEST-RX-RZ-STRINGS-REDUCE` has been returned to its initial expectation
of five gates.

### DIFF
--- a/tests/compilation-tests.lisp
+++ b/tests/compilation-tests.lisp
@@ -250,7 +250,7 @@ CNOT 4 8
                      (%read-test-chipspec "Aspen-6-2Q-A.qpu")
                      (%read-test-chipspec "Aspen-7-28Q-A.qpu"))))
     (dolist (chip chips)
-      (%test-reduction-with-chip 7 chip))))
+      (%test-reduction-with-chip 5 chip))))
 
 ;; (deftest test-free-rx-rz-strings-reduce ()
 ;;   (%test-reduction-with-chip 3 (%read-test-chipspec "1q-free-rx.qpu")))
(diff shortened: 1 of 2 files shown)


tests/compilation-tests.lisp
```diff
@@ -250,7 +250,7 @@ CNOT 4 8
                      (%read-test-chipspec "Aspen-6-2Q-A.qpu")
                      (%read-test-chipspec "Aspen-7-28Q-A.qpu"))))
     (dolist (chip chips)
-      (%test-reduction-with-chip 7 chip))))
+      (%test-reduction-with-chip 5 chip))))
 
 ;; (deftest test-free-rx-rz-strings-reduce ()
 ;;   (%test-reduction-with-chip 3 (%read-test-chipspec "1q-free-rx.qpu")))
```
notmgsk, line 256: You can re-enable this one too.

colescott: Re-enabled in 9dff029.

Issue #667, opened 2020-08-04
Parametric compilation broken


notmgsk:
For some qubits on Aspen-8 the following program raises a `compiler-does-not-apply` error.

```
import pyquil
qc = pyquil.get_qc('Aspen-8', as_qvm=True)
qc.compile(pyquil.Program("""DECLARE t REAL
RX(((1.0)*t[0])) 3"""))
```

```
RPCError: Unhandled error in host program:
Condition CL-QUIL::COMPILER-DOES-NOT-APPLY was signalled.
```

@colescott

stylewarning:
Was the "broken after" determined by bisection?

notmgsk:
Yes

notmgsk:
Possibly a red herring, however. I have a different program that causes the issue before #655.

stylewarning:
@notmgsk can you confirm your test case is fixed?

notmgsk:
Confirmado

PR #669, opened 2020-08-05
ensure applicable symbolic-accepting compilers are included

stylewarning:
When we changed the way occ-tbl metrics were calculated, it caused
quilc to find particularly good collections of compilers ("compiler
paths") for certain gate sets. These particularly good compiler paths
were better than before! But there was a hitch... the better compilers
were also less flexible in what they matched against, though that
wasn't apparent from the binding patterns themselves. In particular,
in some cases, the better compilers were ones that also did _not_
accept symbolic parameters (e.g., a memory reference). This means
these better compilers failed to take into account inputs like
`RX(theta) 0` even though the pattern `RX(_) _` was compilable.

The fix in this commit has two facets:

- First, we use heuristics to determine if a compiler can accept
  symbolic parameters. This heuristic can be improved, but works
  fine for one-parameter compilers, which are the real-world cases
  that failed.

- Second, even if we find a better compiler path when warming the
  chip spec, as long as a symbolic-accepting path was found, it's
  included in the list of compilers. This may still fail, if the
  entire path doesn't allow symbolic compilers. This was too
  complicated to check for this commit with too little gain.

This fixes both issues reported in issue #667.

### DIFF
--- a/src/define-compiler.lisp
+++ b/src/define-compiler.lisp
@@ -353,6 +353,24 @@ What \"concrete\" means depends on the types of A and B:
   (:method ((a wildcard-binding) (b gate-binding))
     b))
 
+(defvar *instantiate-binding-with-symbolic-parameters* nil
+  "When calling INSTANTIATE-BINDING and encountering a symbolic parameter, instead of erroring, supply a symbolic parameter.")
+
+(defun symbolic-param-generator ()
+  (let ((n -1))
+    (lambda ()
+      (make-delayed-expression nil nil `(* 1.0 ,(mref "G" (incf n)))))))
+
+(defun instantiate-parameters (params)
+  ;; Instantiate PARAMS with symbolic arguments. Used as a helper
+  ;; below.
+  (let ((gen (symbolic-param-generator)))
+    (mapcar (lambda (p)
+              (if (symbolp p)
+                  (funcall gen)
+                  p))
+            params)))
+
 (defgeneric instantiate-binding (binding)
   (:documentation "When possible, construct the unique instruction on which BINDING will match. If any of the binding's arguments are unspecified (i.e. match against any qubit), those arguments will be filled-in with unique qubit indices.")
   (:method (binding)
@@ -360,8 +378,10 @@ What \"concrete\" means depends on the types of A and B:
   (:method ((binding gate-binding))
     (when (symbolp (gate-binding-operator binding))
       (error 'cannot-concretize-binding))
-    (when (or (symbolp (gate-binding-parameters binding))
-              (some #'symbolp (gate-binding-parameters binding)))
+    (when (symbolp (gate-binding-parameters binding))
+      (error 'cannot-concretize-binding))
+    (when (and (some #'symbolp (gate-binding-parameters binding))
+               (not *instantiate-binding-with-symbolic-parameters*))
       (error 'cannot-concretize-binding))
     (when (some #'symbolp (gate-binding-arguments binding))
       (let* ((usedq (gate-binding-arguments binding))
@@ -371,7 +391,7 @@ What \"concrete\" means depends on the types of A and B:
                                                        usedq)))))
     (apply #'build-gate
            (gate-binding-operator binding)
-           (gate-binding-parameters binding)
+           (instantiate-parameters (gate-binding-parameters binding))
            (gate-binding-arguments binding))))
 
 
@@ -405,6 +425,15 @@ What \"concrete\" means depends on the types of A and B:
     :initarg :output-gates
     :reader compiler-output-gates
     :documentation "Information automatically extracted about the target gate set.")
+   (info-plist
+    :initform nil
+    :accessor compiler-info-plist
+    :documentation "A plist of inessential information. You will likely find cached data here. Data includes:
+
+    :SYMBOLIC <boolean>
+        Accepts symbolic parameters.
+
+N.B. This is *NOT* the same as COMPILER-OPTIONS.")
    (%function
     :initarg :function
     :reader compiler-%function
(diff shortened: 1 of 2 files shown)


src/define-compiler.lisp
```diff
@@ -353,15 +353,35 @@ What \"concrete\" means depends on the types of A and B:
   (:method ((a wildcard-binding) (b gate-binding))
     b))
 
+(defvar *instantiate-binding-with-symbolic-parameters* nil
+  "When calling INSTANTIATE-BINDING and encountering a symbolic parameter, instead of erroring, supply a symbolic parameter.")
+
+(defun symbolic-param-generator ()
+  (let ((n -1))
+    (lambda ()
+      (make-delayed-expression nil nil `(* 1.0 ,(mref "G" (incf n)))))))
```
ecpeterson, line 362: Gross expressions like this one make me feel the AST stuff around delayed expressions should be seriously rethought. There's nothing to do about it here, but someone (perhaps me) should open an issue about it.

src/define-compiler.lisp
```diff
@@ -405,6 +425,15 @@ What \"concrete\" means depends on the types of A and B:
     :initarg :output-gates
     :reader compiler-output-gates
     :documentation "Information automatically extracted about the target gate set.")
+   (info-plist
```
ecpeterson, line 428: Beware: `misc-data` was not a great idea in `chip-specification`, and the short-form docs for this sound OK (cached info! sure!) the details are sketchier (in what sense is `:SYMBOLIC` caching anything?). Are we sure this tag doesn't belong at top-level in this object?

src/define-compiler.lisp
```diff
@@ -668,6 +697,75 @@ N.B.: The word \"shortest\" here is a bit fuzzy.  In practice it typically means
                      (compiler-bindings x)))
              (get-compilers qubit-count)))
 
+(defun compiler-allows-symbolic-parameters-p (compiler)
+  "Does COMPILER accept instructions with symbolic parameters?
+
+Compilers matching gates which don't have parameters don't count as accepting parameters."
+  (let ((info (getf (compiler-info-plist compiler) ':SYMBOLIC '#1=#:unknown)))
```
ecpeterson, line 704: IMO `getf` should return a secondary value like `gethash` does. Where do I file a complaint?

src/define-compiler.lisp
```diff
@@ -668,6 +697,75 @@ N.B.: The word \"shortest\" here is a bit fuzzy.  In practice it typically means
...
+        (unless (some (lambda (p) (symbolp (first p))) parameter-lists)
+          (return nil))
+        ;; Ok, so, here's what we now know:
+        ;;
+        ;;     - All bindings take 1 parameter
+        ;;
+        ;;     - At least one binding takes an arbitrary parameter.
+        (let ((*instantiate-binding-with-symbolic-parameters* t))
+          (let ((prototype-instructions (map 'list #'instantiate-binding bindings)))
+            ;; We are ready to lock & load.
```
ecpeterson, line 751: lock & load the hounds

src/define-compiler.lisp
```diff
@@ -668,6 +697,75 @@ N.B.: The word \"shortest\" here is a bit fuzzy.  In practice it typically means
...
+      ;; compilation of symbolic 2Q operators with wildcard 1Q gates
+      ;; surrounding, but we don't have this kind of thing anywhere.
+      (unless (every (lambda (b) (typep (gate-binding-operator b) 'operator-description )) bindings)
+        (return nil))
+      ;; Now we descend to look at the bindings' parameter lists for
+      ;; heuristic conformance.
+      (let ((parameter-lists (map 'list #'gate-binding-parameters bindings)))
+        ;; HEURISTIC ALERT: All bindings have fewer than two
+        ;; parameters. This rule doesn't necessarily mean a compiler
+        ;; doesn't accept symbolic parameters, but it is unlikely in
```
ecpeterson, line 731: Is this also unlikely when using parametric Pauli sums?

src/define-compiler.lisp
```diff
@@ -708,10 +813,14 @@ N.B.: The word \"shortest\" here is a bit fuzzy.  In practice it typically means
...
-                   (>= special-cost generic-cost))
-              ;; then store it!
-              (setf (gethash compiler compiler-hash) special-cost))))
+          ;;
+          ;; or does this special compiler actually allow a broader
+          ;; set of inputs than the generic ones?
+          (when (or (compiler-allows-symbolic-parameters-p compiler)
+                    (and (not (path-has-a-loop-p special-path (compiler-bindings compiler)))
+                         (>= special-cost generic-cost)))
+            ;; then store it ya dingus!
```
ecpeterson, line 822: ❤️

tests/compilation-tests.lisp
```diff
@@ -277,3 +277,42 @@ CNOT 4 8
 
 (deftest test-free-rx-rz-strings-reduce ()
   (%test-reduction-with-chip 3 (%read-test-chipspec "1q-free-rx.qpu")))
+
+(deftest test-symbolic-parameter-compiles-in-examples ()
```
colescott, line 281: Is it possible to split this up into tests for the new symbolic parameter detection and then another for compiler hooks directly from #667?

ecpeterson:
I remembered this morning that the "symbolic" compilers for PAULI-SUM applications permit several parameters but only one memory reference. How close is that to being covered with what's already here?

src/define-compiler.lisp
notmgsk: `/dev/null`

src/define-compiler.lisp
stylewarning: I agree.

stylewarning:
I think it’s a TODO to file issues with ECP’s comments.

PR #672, opened 2020-08-17
Add POST_LOAD_ASDF_SYSTEMS option to build

colescott:
Add POST_LOAD_ASDF_SYSTEMS which adds systems to be built and loaded in
the executable file. This allows for external sysytems to extend quilc
and be built into the executable without being part of the quilc system.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,12 +9,6 @@ QUICKLISP=$(SBCL) --load $(QUICKLISP_HOME)/setup.lisp \
 	--eval '(push (truename ".") asdf:*central-registry*)' \
 	--eval "(push (truename \"$(RIGETTI_LISP_LIBRARY_HOME)\") ql:*local-project-directories*)"
 QUICKLISP_BOOTSTRAP_URL=https://beta.quicklisp.org/quicklisp.lisp
-
-
-# Extra ASDF systems to load after quilc as part of the executable.
-export ASDF_SYSTEMS_TO_LOAD=quilc $(POST_LOAD_ASDF_SYSTEMS)
-SYSTEM_LOAD_STRING=$(foreach system,$(ASDF_SYSTEMS_TO_LOAD),--eval '(ql:quickload "$(system)")')
-
 UNAME_S=$(shell uname -s)
 PREFIX ?= /usr/local
 
@@ -34,8 +28,8 @@ $(QUICKLISP_SETUP):
 system-index.txt: $(QUICKLISP_SETUP)
 	$(QUICKLISP) \
 		$(FOREST_SDK_FEATURE) \
-		$(SYSTEM_LOAD_STRING) \
-		--eval '(ql:write-asdf-manifest-file "system-index.txt")'
+		--eval "(ql:quickload '(quilc ${POST_LOAD_ASDF_SYSTEMS}))" \
+		--eval "(ql:write-asdf-manifest-file \"system-index.txt\")"
 
 ###############################################################################
 # DEPENDENCIES
@@ -76,7 +70,8 @@ endif
 
 .PHONY: quilc
 quilc: system-index.txt
-	$(SBCL) $(FOREST_SDK_FEATURE) \
+	ASDF_SYSTEMS_TO_LOAD="quilc ${POST_LOAD_ASDF_SYSTEMS}" \
+        $(SBCL) $(FOREST_SDK_FEATURE) \
 	        --eval "(setf sb-ext:\*on-package-variance\* '(:warn (:swank :swank-backend :swank-repl) :error t))" \
 		--load "build-app.lisp" \
 		$(FOREST_SDK_OPTION) \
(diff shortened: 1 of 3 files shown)


stylewarning:
This mostly helps in building the quilc executable. Obviously one can make new ASDF systems that _depend_ on CL-QUIL and/or QUILC.

build-app.lisp
```diff
@@ -40,7 +40,15 @@
            (subseq version 0 (position #\- version :test #'eql))))
     (load-systems-table)
     (push #'local-system-search asdf:*system-definition-search-functions*)
-    (asdf:load-system "quilc")
+
+    ;; Load systems defined by environment variable $ASDF_SYSTEMS_TO_LOAD,
+    ;; or if that does not exist just load quilc
+    (let ((systems (uiop:getenv "ASDF_SYSTEMS_TO_LOAD")))
+      (mapcar (lambda (sys)
```
stylewarning, line 47: MAP NIL

build-app.lisp
```diff
@@ -40,7 +40,15 @@
...
+                (unless (uiop:emptyp sys)
+                  (asdf:load-system sys)))
+              (uiop:split-string (or systems "quilc")
```
stylewarning, line 50: I would always load QUILC afterward; check that it's not already loaded tho.

stylewarning: (it doesn't hurt if it's already loaded, but double loading is a double no-bueno)

colescott: Well, all systems to load should be in order already. If someone passes a list without quilc in it I don't think there's a reasonable place to insert quilc. Maybe we should warn if quilc is not in the list and error if there are duplicates.

stylewarning: I don't get the complication. The point of this is to pre-load ASDF systems. Why not just unconditionally load quilc after all of that, so a pre-load of `""` is tolerated?

stylewarning: (s/pre/post/)

colescott: Line 46 uses the env var `ASDF_SYSTEMS_TO_LOAD` which is created in the Makefile. This is done so if for some reason in the future we want to provide systems to load _before_ quilc it will be supported.

stylewarning: great point, you're right, I think everything is OK now

colescott:
> can you document this new variable in the README?

Added a new section in the README with a table for build options

stylewarning (dismissed):
lgtm

Makefile
```diff
@@ -9,6 +9,12 @@ QUICKLISP=$(SBCL) --load $(QUICKLISP_HOME)/setup.lisp \
 	--eval '(push (truename ".") asdf:*central-registry*)' \
 	--eval "(push (truename \"$(RIGETTI_LISP_LIBRARY_HOME)\") ql:*local-project-directories*)"
 QUICKLISP_BOOTSTRAP_URL=https://beta.quicklisp.org/quicklisp.lisp
+
+
+# Extra ASDF systems to load after quilc as part of the executable.
+export ASDF_SYSTEMS_TO_LOAD=quilc $(POST_LOAD_ASDF_SYSTEMS)
```
notmgsk, line 15: Good for the soul if we are careful about strings here. What if `POST_LOAD_ASDF_SYSTEMS=rm -rf /`?

Makefile
```diff
@@ -9,6 +9,12 @@ QUICKLISP=$(SBCL) --load $(QUICKLISP_HOME)/setup.lisp \
...
+SYSTEM_LOAD_STRING=$(foreach system,$(ASDF_SYSTEMS_TO_LOAD),--eval '(ql:quickload "$(system)")')
```
notmgsk, line 16: likewise here

build-app.lisp
```diff
@@ -40,7 +40,15 @@
            (subseq version 0 (position #\- version :test #'eql))))
     (load-systems-table)
     (push #'local-system-search asdf:*system-definition-search-functions*)
-    (asdf:load-system "quilc")
+
+    ;; Load systems defined by environment variable $ASDF_SYSTEMS_TO_LOAD,
+    ;; or if that does not exist just load quilc
+    (let ((systems (uiop:getenv "ASDF_SYSTEMS_TO_LOAD")))
+      (map nil (lambda (sys)
```
notmgsk, line 47: `dolist`?

notmgsk (changes requested):
V cool addition

Makefile
notmgsk: Definitely here.

build-app.lisp
colescott: Yeah. looks much better

Makefile
colescott: Should now not try to eval a string (I forgot the `make` syntax :( )

stylewarning:
@notmgsk  ping

notmgsk (approved):

PR #674, opened 2020-08-27
Framework for implementing quilc backends

stylewarning:
Tell me, O compiler, of that ingenious backend who travelled far and wide after he had sacked the famous town of Aho-Ullman-Sethi-Lam. This PR introduces some protocols to formalize the notion of a "backend". QUILC so far has acted as a front-end for Quil. Included is a simple "no op" backend that just writes out Quil as UTF8 bytes.

We don't suggest that this will be the end-all be-all protocol; we have use-cases in mind and in development, and we expect that as actual backends get implemented, the shape of this code will change. As such, consider this a "germ" of sorts.

The PR includes:

- New command line options `--backend`, `--backend-options`, `--list-backends`, `--compile`, `--output`, as well as short versions.

- A new directory `backends` for implementations of backends, including the `quil-backend`, which is a re-imagining of current quilc functionality.

- A backend protocol, from `backends/common.lisp`.

The basic idea for implementing a backend:

1. Impelement a subclass of `backend`, which will be populated with requisite data needed to compile for that backend.

2. Implement the executable format.

3. Implement `backend-compile` to actually do the dirty work.

If you do this, quilc should "automatically know".

This supersedes PR #630.

### DIFF
--- a/app/src/entry-point.lisp
+++ b/app/src/entry-point.lisp
@@ -413,7 +413,7 @@
                (unless output
                  (error "Output must be provided when compilation is enabled. Specify an output file with -o or --output."))
 
-               (let ((backend-class (quil:parse-backend backend)))
+               (let ((backend-class (quil:find-backend backend)))
                  (unless backend-class
                    (error "The backend value '~a' does not name an available backend. For a list of available backends, run 'quilc --list-backends'." backend))
 
@@ -438,7 +438,7 @@
                                      :element-type '(unsigned-byte 8)
                                      :if-exists ':supersede
                                      :if-does-not-exist ':create))
-           (let ((executable (quil:compile-executable program chip-spec backend)))
+           (let ((executable (quil:backend-compile program chip-spec backend)))
              (quil:write-executable executable stream)))
       (when stream
         (close stream)))))
--- a/app/src/entry-point.lisp
+++ b/app/src/entry-point.lisp
@@ -99,7 +99,9 @@
           (mapcar 'sb-alien::shared-object-pathname sb-sys:*shared-objects*))
   (unless (magicl.foreign-libraries:foreign-symbol-available-p "zuncsd_"
                                                                'magicl.foreign-libraries:liblapack)
-    (format t "The loaded version of LAPACK is missing functionality. The compiler will still work with your current LAPACK but it is advisable to install a more complete version.~%")
+    (format t "The loaded version of LAPACK is missing ~
+    functionality. The compiler will still work with your current ~
+    LAPACK but it is advisable to install a more complete version.~%")
     (uiop:quit 1))
   (format t "Library check passed.~%")
   (uiop:quit 0))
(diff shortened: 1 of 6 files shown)


stylewarning:
CC @kilimanjaro

colescott:
Note re:#630: We have decided to drop the signing feature because it does not seem like something that belongs in the compiler. If the user wants to sign executables then an external tool such as gpg should be used. Dropping this responsibility onto an external tool provides better support for more signing options and reduces the technical overhead on quilc's side.

notmgsk:
Are we all for "backend"? What about "target"?

stylewarning:
> ~Are we all for "backend"? What about "target"?~ I thought I had a reason for disliking "backend" but it didn't stand up to the heat.

Oh yeah I forgot we had a discussion about that. I guess there isn't a single good answer. Maybe one might say that a compiler backend compiles code to a specific "target" or "architecture", but I'm not sure "backend" is all that problematic.

app/src/entry-point.lisp
```diff
@@ -358,16 +384,81 @@
...
-               (print-statistics statistics *quil-stream*))
-             (when (and protoquil compute-matrix-reps)
-               (let* ((processed-program-matrix (parsed-program-to-logical-matrix processed-program :compress-qubits t)))
-                 (print-matrix-comparision original-matrix
-                                           (quil::scale-out-matrix-phases processed-program-matrix
-                                                                          original-matrix)))))))))))
+
+             ;; NOTE: This flow is deprecated and will be merged with
+             ;; quil-backend in the future.
+             (unless compile
```
notmgsk, line 390: In the future you don't want to support `echo "H 0" | quilc` (instead requiring `echo "H 0" | quilc --compile ...`)?

app/src/entry-point.lisp
```diff
@@ -358,16 +384,81 @@
...
+                                (parse-backend-options backend-option))))
+                   (unless (quil:backend-supports-chip-p backend chip-spec)
+                     (error "The backend provided does not support this ISA."))
+
+                   (backend-compile-program processed-program chip-spec backend output)))))))))))
+
+(defun parse-backend-options (options)
+  "Parse backend options to keyword init-args for making backend."
+  (mapcan (lambda (opt)
+            (let ((ret (uiop:split-string opt :separator '(#\=))))
```
notmgsk, line 445: This syntax isn't described anywhere. Would you add it to the `--help` output?

app/src/options.lisp
```diff
@@ -43,6 +43,34 @@
...
+    (("compile" #\c)
+     :type boolean
+     :optional t
+     :initial-value nil
+     :documentation "compile the program to an executable using the specified backend")
+
+    (("backend" #\b)
+     :type string
+     :optional t
+     :documentation "backend used to compile the program. Only has effect when --compile or -c are provided.")
```
notmgsk, line 55: To end sentences with periods or not end sentences with periods, that is the question that `--help` ignores entirely.

app/src/entry-point.lisp
colescott: The idea was to automatically choose the quil backend and have it support the currently supported options either through the backend options or through the same cli interface. The main thing was to try to get away from writing quil-specific code in entry-point and push that responsibility onto backends.

app/src/entry-point.lisp
colescott: Added to backend-option help string

notmgsk:
Gitlab says no, but travis says yes. Going with travis on this one.

stylewarning:
What's the status of this? Is this ready to roll?

ecpeterson (commented):
seems like a fine direction to head in to me

notmgsk (approved):

PR #677, opened 2020-10-15
Include IRC #qlisp in README

### LINKED ISSUE #671: Slack status unclear
README seems to suggest that the Slack `#dev` channel is available to hang out in, but if you try to join it seems to only accept people with an `@rigetti.com` or `@qxbranch.com` email address.

notmgsk:
Closes #671

### DIFF
--- a/README.md
+++ b/README.md
@@ -259,14 +259,11 @@ make benchmark-qasm
 
 # Get involved!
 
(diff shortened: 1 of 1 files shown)


stylewarning:
Why was this closed?

PR #681, opened 2020-11-04
Do not use a default value in EYE

kilimanjaro:
`EYE` wraps `MAGICL:EYE`, which may be provided with both a type and a value for the diagonal. Currently, `MAGICL:EYE` will `COERCE` the value to the type, if need be. However, `(complex double-float)` cannot be coerced to `double-float`, even if the imaginary part is zero. Every so often quilc tries to construct a double-float identity matrix, e.g. here https://github.com/rigetti/quilc/blob/master/src/compilers/approx.lisp#L201, and this will cause an error.

### DIFF
--- a/src/magicl-constructors.lisp
+++ b/src/magicl-constructors.lisp
@@ -23,7 +23,7 @@
   "CL-QUIL version of MAGICL:RAND"
   (magicl:rand shape :type type :layout layout :distribution distribution))
 
-(defun eye (shape &key (value #C(1d0 0d0)) (type +default-magicl-type+) layout)
+(defun eye (shape &key value (type +default-magicl-type+) layout)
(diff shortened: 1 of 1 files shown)


colescott (dismissed):

notmgsk (approved):

kilimanjaro:
Original commit was the wrong thing, since `MAGICL:EYE` will prioritize the type of `:VALUE` over a user-specified `:TYPE`. It's better to just not use any default value here.

colescott (commented):
It might be good to remove the other defaults in this file that are just copies of the magicl defaults to avoid confusion like this in the future. (L34,40,46)

I think the original purpose for leaving those in is to allow the caller to see what the default will be, however this might not be worth the added weirdness.

stylewarning (approved):

PR #682, opened 2020-11-25
Allow --verbose for server mode

notmgsk:


stylewarning:
Print a warning that `--verbose` is not really appropriate for multithreaded server environments at the top of quilc startup output (in server mode only).

app/src/entry-point.lisp
```diff
@@ -373,7 +376,7 @@
            (multiple-value-bind (processed-program statistics)
                (process-program program chip-spec
                                 :protoquil protoquil
-                                :verbose verbose
+                                :verbose (when verbose *human-readable-stream*)
```
stylewarning, line 379: please use `if` with explicit else-branch

stylewarning (changes requested):
just this and looks GTM

app/src/entry-point.lisp
notmgsk: donezo

app/src/entry-point.lisp
```diff
@@ -352,6 +352,9 @@
            (cl-syslog:rfc-log (*logger* :warning "--backend and --output-file/-o options ~
                                                   don't make sense when using server mode. ~
                                                   Ignoring them.")))
+         (when verbose
+           (setf quil::*compiler-noise* *error-output*)
+           (warn "--verbose output is not appropriate for production multithreaded environments, and should b used when debugging and with care"))
```
stylewarning, line 357: I'm not going to hold this PR up but WARN takes a format string so you

```
"can break lines ~
 like this."
```

stylewarning (approved):

Issue #653, opened 2020-07-14
Compressor should use appropriate cost function when deciding on which compression strategy to follow


notmgsk:
See https://github.com/rigetti/quilc/pull/649/files#r452514653 where it is (now) hardcoded to use the program fidelity as the optimization strategy. Compare that with the addresser, which allows for either fidelity or duration to be used, depending on some configuration.

PR #684, opened 2020-12-03
Only use "reducers" that strictly improve fidelity

notmgsk:
In `compute-applicable-reducers` we hope to narrow the list of `compilers` down to those that "improve the brevity of a gate sequence" subject to the restriction "without exiting a particular `gateset`". (The word "brevity" here is maybe a hold-over from a time when program length (duration?) was the cost function, and now it means fidelity?) In reality, it would also include compilers that did *at least as good* as others. This seems to cause trouble for certain ISAs and compilers.

For example, if your ISA contains `CZ 0 1`, `CPHASE(θ) 0 1`, *and* each has the same fidelity `fid` then the compiler can get into a loop where a `CPHASE(θ) 0 1` is compiled with `half-cphase-to-cz` and then the resulting `CZ` are compiled with `cz-to-cphase`, etc. Both of these compilers are present in the list computed by `compute-applicable-reducers` where I think only one of them needs to be present. By making the `compute-applicable-reducers` comparison strict, the reducers then are only applied if they improve the fidelity, and the compiler can support ISAs whose gates may be equivalent (CZ = CPHASE(pi)) and of equal fidelity.

Am I missing something? :D

### DIFF
--- a/src/compressor/compressor.lisp
+++ b/src/compressor/compressor.lisp
@@ -240,10 +240,10 @@ other's."
                                           context)
   "Applies peephole rewriter rules from a CHIP-SPECIFICATION to a sequence of INSTRUCTIONS, using CONTEXT to activate context-sensitive rules."
   (labels
-      ( ;; utility for calculating how many instructions a rewriting rule requests
+      (;; utility for calculating how many instructions a rewriting rule requests
        (rewriting-rule-count (compiler)
          (length (compiler-bindings compiler)))
-       
+
        ;; let the context know that we've passed inspection of NODE, so that the
        ;; effect of that instruction is visible during inspection of the next node
        (update-context (node)
@@ -303,7 +303,7 @@ other's."
        ;; successfully, we rewind by the peephole window and try again. if it
        ;; fails, we fall through, step through to the next node, and try again.
        (outer-instruction-loop (node)
-         (do ( ;; for each instruction...
+         (do (;; for each instruction...
               (node #1=(peephole-rewriter-node-next node) #1#))
              ((null node))
            (update-context node)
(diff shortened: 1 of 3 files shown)


ecpeterson:
I don't think you're forgetting anything; I think this was an accidental consequence of both the `cost-function` changes and accomplishing the original by uncoordinatedly tweaking things 'til they worked.

src/compressor/compressor.lisp
```diff
@@ -240,10 +240,10 @@ other's."
                                           context)
   "Applies peephole rewriter rules from a CHIP-SPECIFICATION to a sequence of INSTRUCTIONS, using CONTEXT to activate context-sensitive rules."
   (labels
-      ( ;; utility for calculating how many instructions a rewriting rule requests
+      (;; utility for calculating how many instructions a rewriting rule requests
```
ecpeterson, line 243: If someone tells me how to disable this with M-q, I would do it in a heartbeat.

ecpeterson (approved):

notmgsk:
> I don't think you're forgetting anything; I think this was an accidental consequence of both the `cost-function` changes and accomplishing the original by uncoordinatedly tweaking things 'til they worked.

@ecpeterson Should the `compute-applicable-*` functions be generic? Right now they only consider fidelity, but we also support (elsewhere in the compiler) duration as a cost metric.

ecpeterson:
I think something like this is already an issue: #653 .

ecpeterson:
Yes, something like this should be done. Flexibility about cost metric is a problem throughout the compression, cf. your #653 . One hopes that this doesn't mean a further refactor of the addresser...

PR #687, opened 2020-12-15
Don't drop an empty RESET-BLOCK when processing a PRESERVED-BLOCK

notmgsk:
Take a look at the included test for a program where (currently) a `RESET` that precedes a `PRAGMA PRESERVE-BLOCK` is dropped.


ecpeterson (dismissed):
Looks fine for a surgical fix. This area of the code could overall use some careful thought and retooling; you might spend some time opening some issues for ways it strikes you as gross.

PR #690, opened 2021-01-23
Fix test-memoize-pauli-basis-matrices timing issues.

mhdavid-hrl:
Test was failing on a fast machine because the "slow" case took 1/2
msec or less on every iteration. Time-it macro would round the time
down to zero in an effort to return time in milliseconds, needlessly
throwing away the extra precision in internal time units.

There's no need for any particular time units (such as msec), since
the time is only used for comparison.

Furthermore, it should use get-internal-run-time, not -real-time,
i.e., should not be affected by other processes running.

Changed time-it to return time in terms of internal time units and to
be in terms of run time.  Also, added some documentation of what it
does.

Also, the test code only tested whether the slow case took longer at
all. Improved it to test that the slow test takes longer by some
minimum slow-to-fast time ratio (8x).

For comparison, ratios of 100x were typically achieved With SBCL,
standard optimization, on MacOS, on a Macbook Pro, 16-inch 2019, with
processor speed 2.6 GHz.

In the future, if machines get a lot faster, and/or a port is needed
to a fast Lisp with msec internal time units, basis-size could be
bumped higher, to slow the slow case.

### DIFF
--- a/tests/benchmarking-procedures-tests.lisp
+++ b/tests/benchmarking-procedures-tests.lisp
@@ -100,17 +100,18 @@
                                  :always (pauli-matrix-p (magicl:scale pauli (* 2 phase)))))))))
 
 (defmacro time-it (&body body)
+  ;; Time evaluation of BODY, and return two values: (1) the first
+  ;; value returned by body, and (2) elapsed internal runtime.
   (let ((start (gensym "START"))
         (end (gensym "END"))
         (body-value (gensym "BODY-VALUE")))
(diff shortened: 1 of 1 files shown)


notmgsk (approved):

stylewarning:
LGTM; I think only @notmgsk has the merge privileges

mhdavid-hrl:
Oh, did you not see? I think Mark took in the PR.

PR #696, opened 2021-01-26
Do not mutate input to INSTANTIATE-DEFINITION-BODY

notmgsk:
This caused a bug where invocations of a circuit would mutate the
circuit body for subsequent invocations if the original invocation had
modifiers. For example, the circuit

```
DEFCIRCUIT FOOBAR:
    X 0
```

would be mutated by the instruction `DAGGER FOOBAR` to be

```
DEFCIRCUIT FOOBAR:
    DAGGER X 0
```

and further use of the circuit would include the unwanted `DAGGER`.


src/analysis/expansion.lisp
```diff
@@ -76,13 +76,16 @@
...
+  "Returns a list of instructions which are the instructions in the BODY of DEFN instantiated with PARAMS and ARGS for DEFN-PARAMs and DEFN-ARGS respectively. Does not mutate its input."
   (assert (= (length params)
              (length defn-params)))
   (assert (= (length args)
              (length defn-args)))
 
-  (let ((*expansion-depth* (1+ *expansion-depth*)))
+  (let ((*expansion-depth* (1+ *expansion-depth*))
+        (body (mapcar #'copy-instance body))
+        (defn-params (mapcar #'copy-instance defn-params))
```
stylewarning, line 87: Do the params and args need to be copied? I'd have thought just the body.

stylewarning (approved):

src/analysis/expansion.lisp
notmgsk: You're right, I was being cautious :3

stylewarning:
LGTM

PR #700, opened 2021-01-29
Add test test-split-lines, testing parser function split-lines.

mhdavid-hrl:
While split-lines is heavily tested as part of testing the parser,
this adds tests focused specifically on split-lines. In addition, a
near-future proposed change to split-lines is planned, so this adds
testing needed for confidence in such a change.

add missing colon to make = a keyword within LOOP


mhdavid-hrl:
Changed commit log and this PR description just to removed stray line 'add missing colon to make = a keyword within LOOP'.

stylewarning:
This looks good to me @notmgsk. More tests are always better, much to the chagrin of the runners.

notmgsk (approved):

PR #701, opened 2021-02-01
Fix quilt test TEST-FILL-DELAYS-FENCE-EXPANSION

kilimanjaro:
It was observed that one of the `cl-quil/quilt` tests was failing. I'm not sure how this got through the original PR process (perhaps I goofed something in a rebase??), but here's an easy fix. It's worth noting two things:

- As far as I know, `cl-quil/quilt` is not currently used by anyone.
- The change here reflects the minor work needed to get the test to pass, and it is at least consistent with the stated intent of the analysis pass (previously it was inserting a spurious `DELAY`). But I'm not otherwise verifying the correctness of `cl-quil/quilt` with respect to the proposed Quilt standard or the Rigetti pidgin implementation.

There's another Quilt test, `test-definition-signature`, that fails under some circumstances due to how the hash method for frames is currently implemented. That may be addressed in a separate PR.


kilimanjaro:
On second glance, this is probably a consequence of the system-dependent behavior of the frame hash function. Closing for now.

PR #702, opened 2021-02-02
Load foreign libraries at run-time

notmgsk:
This allows the library-loading system to not depend solely on the
library name that was used at compile time. Example: If libffi.so.6
was used at compile time but is not present at run-time, then look for
alternatives rather than just failing.

Closes https://github.com/rigetti/pyquil/issues/1287

H/T https://gist.github.com/realark/1b6970a13615a6e389294ab9480d1e09

### DIFF
--- a/build-app.lisp
+++ b/build-app.lisp
@@ -10,34 +10,57 @@
 
 (require 'asdf)
 
+(defun unload-libraries ()
+  "Unloads foreign libraries using CFFI and returns a list of the closed libraries"
+  ;; A dumped lisp image will use the *exact* foreign library name
+  ;; that was provided at the time the image was dumped. If the
+  ;; library is moved/renamed, CFFI will not try to load an
+  ;; alternative, thus we unload all foreign libraries and tell CFFI
+  ;; to load them at run-time allowing it to pick up alternatives.
+  ;;
+  ;; For a concrete example: when we build the SDK packages for
+  ;; debian, the FFI library available is typically libffi.so.6. On
+  ;; newer versions of debian the library is named libffi.so.7, so
+  ;; when we try to load the lisp image on a newer debian version, it
+  ;; fails with an error about not finding libffi.so.6.
+  (let ((foreign-libraries
+          (mapcar (lambda (library) (funcall (find-symbol "FOREIGN-LIBRARY-NAME" :cffi) library))
+                  (funcall (find-symbol "LIST-FOREIGN-LIBRARIES" :cffi) :loaded-only t))))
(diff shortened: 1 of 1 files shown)


build-app.lisp
```diff
@@ -57,9 +80,10 @@
       (format t "~&Using unsafe entry point~%")
       (setf entry-point "quilc::%entry-point"))
     (force-output)
+    (unload-libraries)
     (sb-ext:save-lisp-and-die output-file
                               :compression #+sb-core-compression t
-                                           #-sb-core-compression nil
```
stylewarning, line 62: oof

stylewarning (changes requested):
My somewhat non-constructive request is to try to break this out into other toplevel functions. This is starting to resemble the entry point of the apps, which are behemoths.

stylewarning:
I forgot to say, I think this is a _great_ idea, even if it's ever so slightly error prone (because the library could theoretically mismatch).

build-app.lisp
```diff
@@ -10,34 +10,63 @@
...
+  ;; alternative, thus we unload all foreign libraries and tell CFFI
+  ;; to load them at run-time allowing it to pick up alternatives.
+  ;;
+  ;; For a concrete example: when we build the SDK packages for
+  ;; debian, the FFI library available is typically libffi.so.6. On
+  ;; newer versions of debian the library is named libffi.so.7, so
+  ;; when we try to load the lisp image on a newer debian version, it
+  ;; fails with an error about not finding libffi.so.6.
+  (let ((foreign-libraries
+          (mapcar (lambda (library) (funcall (read-from-string "cffi:foreign-library-name") library))
```
stylewarning, line 27: find-symbol

stylewarning (changes requested):
just this one thing then it LFTM

build-app.lisp
notmgsk: `(find-symbol "LIST-FOREIGN-LIBRARIES" :cffi)` 

```
The name :CFFI does not designate any package.
```

notmgsk: Using `(find-symbol "LIST-FOREIGN-LIBRARIES" :CFFI)` or `(find-symbol "CFFI:LIST-FOREGIN-LIBRARIES")` evaluates to `NIL`.

stylewarning:
```
* (ql:quickload :cffi)
To load "cffi":
  Load 1 ASDF system:
    cffi
; Loading "cffi"
.
(:CFFI)
* (find-symbol "LIST-FOREIGN-LIBRARIES" :cffi)
CFFI:LIST-FOREIGN-LIBRARIES
:EXTERNAL
```

stylewarning (approved):

PR #705, opened 2021-03-22
Fix for magicl update and other sundries

stylewarning:
This PR does two things:

1. ~Adds a forwards and backwards compatible fix for the upcoming changes to MAGICL. it'll search for the `LAPACK-CSD` function from multiple packages.~ Updates MAGICL to 0.9.0 and removes deprecated systems.
2. Fixes a few errors caught by SBCL at compile-time.

### DIFF
--- a/cl-quil-tests.asd
+++ b/cl-quil-tests.asd
@@ -8,8 +8,8 @@
   :license "Apache License 2.0 (See LICENSE.txt)"
   :depends-on (#:cl-quil
                (:version #:qvm "1.9.0")
-               #:magicl
-               #:magicl-transcendental
+               (:version #:magicl/core "0.9.0")
+               #:magicl/ext-expokit
                #:alexandria
                #:fiasco
                #:uiop
--- a/src/compressor/compressor.lisp
+++ b/src/compressor/compressor.lisp
@@ -348,15 +348,13 @@ other's."
 
     ;; strip out all the NOPs.
     (let* ((instructions (remove-if (lambda (x) (typep x 'no-operation)) instructions))
-           (head (instrs->peephole-rewriter-nodes instructions))
-           (node head))
+           (head (instrs->peephole-rewriter-nodes instructions)))
 
       ;; actually, we need to prepend a dummy head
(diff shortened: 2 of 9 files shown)


src/compressor/compressor.lisp
```diff
@@ -348,15 +348,13 @@ other's."
 
     ;; strip out all the NOPs.
     (let* ((instructions (remove-if (lambda (x) (typep x 'no-operation)) instructions))
-           (head (instrs->peephole-rewriter-nodes instructions))
-           (node head))
```
stylewarning, line 352: @ecpeterson Can you verify deleting `NODE` variable is OK? Looks OK, but maybe it's here as some sort of QUILC koan.

colescott (dismissed):
Looks good :+1:

src/compressor/compressor.lisp
notmgsk: look reasonable 2 me

stylewarning:
Tests pass locally. The CI system doesn't know about MAGICL 0.9.0 yet...

colescott:
> Tests pass locally. The CI system doesn't know about MAGICL 0.9.0 yet...

I think this is an issue I was having when working on magicl-ng. The issue was that the quilc tests only use the versions available in the current quicklisp dist but that has not been updated since the release of magicl 0.9.0

colescott (approved):

stylewarning:
@notmgsk ^

cl-quil-tests.asd
```diff
@@ -8,8 +8,8 @@
   :license "Apache License 2.0 (See LICENSE.txt)"
   :depends-on (#:cl-quil
                (:version #:qvm "1.9.0")
-               #:magicl
-               #:magicl-transcendental
+               (:version #:magicl/core "0.9.0")
```
colescott, line 11: If we just pull in `magicl` here shouldn't it pull in core and expokit?

stylewarning: yes but we don't want to depend on this functionality; see the magicl `R E A D M E`

PR #709, opened 2021-04-29
qasm: support #pragma directives

notmgsk:
Treats #prama <thing> as a Quil PRAGMA <thing>

### DIFF
--- a/src/qasm.lisp
+++ b/src/qasm.lisp
@@ -142,6 +142,8 @@
    (return (tok :ARROW)))
   ("OPENQASM|qreg|creg|barrier|measure|reset|opaque|include|if|gate"
    (return (tok (intern (string-upcase $@) :keyword))))
+  ("#pragma"
+   (return (tok :PRAGMA)))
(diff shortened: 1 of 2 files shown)


src/qasm.lisp
```diff
@@ -142,6 +142,8 @@
    (return (tok :ARROW)))
   ("OPENQASM|qreg|creg|barrier|measure|reset|opaque|include|if|gate"
    (return (tok (intern (string-upcase $@) :keyword))))
+  ("#pragma"
```
erichulburd, line 145: Worth filtering the supported pragma statements so that a pragma intended for another OpenQASM consumer doesn't result in an error in Quilc? Or does Quilc ignore unexpected pragma statements as is?

notmgsk:
@stylewarning lmkwyt

src/qasm.lisp
notmgsk: They're not *ignored* -- they are added to the AST -- but they have no effect.

PR #711, opened 2021-05-12
drop 2q coded statement

erichulburd:


notmgsk (dismissed):
oops

notmgsk (approved):
Thanks for catching this @erichulburd. I think this managed to slip through the cracks because the tests were broken in a different way that masked this issue.

PR #714, opened 2021-06-02
add quantikz circuit plotting

kilimanjaro:
This adds rudimentary plotting via LaTeX + Quantikz, usable a few ways. For example, in a Jupyter notebook

plot-circuit-example

or in a SLIME session
plot-circuit-slime

## Backends

In general,  we always convert the instructions of a parsed program to a `diagram`, which is basically just a set of lists corresponding to operations on qubit lines, and then emit this as latex source. 

What happens next depends on the choice of "backend", cf. methods on the `RENDER` generic. We currently have three
- `:latex` just passes the generated latex through
- `:jupyter-svg` uses `pdflatex` and `pdf2svg` to generate a SVG file, and then constructs a `jupyter:svg` object from this
- `:slime-png` uses `pdflatex` and `convert` to generate a PNG file, and then pushes this via SWANK to the slime repl, as described in https://righele.it/2012/02/22/display-images-in-slime-from-common-lisp/ (ht @colescott )

See `src/tools/README.md` for installation details.


kilimanjaro:
FWIW this implementation is heavily inspired by the pyquil version I wrote in 2019. Perhaps I should make a note about that somewhere...

colescott:
Looks great! 

Are there any plans to have multiple frontends for this (e.g. slime using something like [this](https://righele.it/2012/02/22/display-images-in-slime-from-common-lisp/))?
If not I am happy to draft up a quick PR for that.

kilimanjaro:
> Looks great!
> 
> Are there any plans to have multiple frontends for this (e.g. slime using something like [this](https://righele.it/2012/02/22/display-images-in-slime-from-common-lisp/))?
> If not I am happy to draft up a quick PR for that.

Good question -- I'm still slowly but surely iterating on the features for this, but will take at the possibility of multiple render targets.

stylewarning (approved):

stylewarning:
Awesome!

PR #717, opened 2021-06-07
Nonrecursive contiguous subschedule

erichulburd:
Rebases https://github.com/quil-lang/quilc/pull/693 which should fix tests


stylewarning:
Both @stylewarning (me!) and @kilimanjaro will take a look.

src/addresser/outgoing-schedule.lisp
```diff
@@ -108,54 +108,36 @@ BEFORE-INST to make use of RESOURCE."
...
-(defun chip-contiguous-subschedule-from-last-instructions (schedule resource) 
-  "Returns an unordered contiguous subsequence of instructions from SCHEDULE where each instruction is in the dependency graph of the last instructions in SCHEDULE and touches only resources in RESOURCE."
+(defun mark-predecessors-seen (instr earlier-instrs seen)
+  "Performs a depth first search on INSTR as the root node and marks all of its descendants as true in hash-table SEEN."
+  (let ((stack (list instr)))
+    (loop :while stack :do
+      (let ((predecessors (gethash (pop stack) earlier-instrs)))
+        (dolist (predecessor predecessors)
+          (unless (gethash predecessor seen)
+            (setf (gethash predecessor seen) t)
```
kilimanjaro, line 118: This is different from the earlier version -- previously, we had

```
(labels ((mark-predecessors-seen (instr)
               (cond ((resource= (instruction-resources instr)
                                 resource)
                      (setf (gethash instr seen) t))
                     ((resource-subsetp (instruction-resources instr)
                                        resource)
                      (map nil #'mark-predecessors-seen (gethash instr earlier-instrs)))))
...
```
which only marked those predecessors of `instr` which acted on a resource set equal to `resource`. Here you are marking everything -- is this the intent?

src/addresser/outgoing-schedule.lisp
```diff
@@ -108,54 +108,36 @@ BEFORE-INST to make use of RESOURCE."
...
-      sched)))
+    (loop :while leaves :do
+      (let ((instr (pop leaves)))
+        (unless (gethash instr seen)
+          (setf (gethash instr seen) t)
+          (cond
+            ((and (resources-intersect-p resource (instruction-resources instr))
+                  (resource-subsetp (instruction-resources instr) resource)) 
+             (push instr sched)
+             (dolist (earlier-instr (gethash instr earlier-instrs))
```
kilimanjaro, line 136: Does the ordering here matter?

It seems that previously there was an explicit `reverse`, but it wasn't entirely clear to me how the ordering of `earlier-instructions` was exploited (and e.g. I'm not sure whether the logical scheduler "officially" makes any promise about this ordering)? I realize that your code has gotten rid of the above `reverse`, but the use of a stack has the same effect (i.e. you push them one by one and then pop them in a reversed order), so if we are doing something which depends on the order I would ask that we leave a comment to that affect (an easy check for this is just to try the above iteration in reverse order).

kilimanjaro (changes requested):
It would help if we added another test with more than "H" gates (just using "H 0" and "H 1" does not allow us to tell whether the "H 0" gates get reordered accidentally), and perhaps some multi-qubit ops and classical instructions as well.

notmgsk:
> I'm a bit sketched out by the original code, and also by the changes here.

Unfriendly language IMO

kilimanjaro:
> > I'm a bit sketched out by the original code, and also by the changes here.
> 
> Unnecessarily unfriendly language IMO

I didn't mean it as a strong statement, but rather that it's hard for me to look at the code (e.g. even as originally written) and say if it is correct or not (or if it would break if the logical scheduler code changed, etc).

Issue #712, opened 2021-05-14
Print a logical-schedule as a Hasse diagram


stylewarning:
It is hard to debug these hash table monstrosities without some visualization. Maybe we can visualize them with a Hasse diagram.

PR #720, opened 2021-06-28
Add subsystem cl-quil/tools -- cl-quil language tools for developers

mhdavid-hrl:
This is a stub for now, but we'll add to it later.

In anticipation of adding some dev-tools type modules, this introduces
a new subsystem cl-quil/tools that new modules can hang off of. For
now this is a stub, but later we may add, e.g., hasse-stuff (a little
utility for doing Hasse diagrams for logical schedulers) and circ-viz
(some tools for circuit visualization), maybe with different names.

The two anticipated new modules, and more to come later probably, are
not being added to cl-quil (core) system because they are
nice-to-haves but non-core. In addition, these two in particular have
weird dependencies like GraphViz and LaTeX, so cordoning them off
keeps the core a bit cleaner.

Issue "Print a logical-schedule as a Hasse diagram #712" relates
loosely to this in that Hasse stuff is planned to be added to this
stub system, so it was a motivater for this general approach.

### DIFF
--- a/cl-quil.asd
+++ b/cl-quil.asd
@@ -289,3 +289,28 @@
                (:file "suite")
                (:file "stabilizer-group-tests")
                (:file "cleve-gottesman-tests")))
+
+(asdf:defsystem #:cl-quil/tools
+  :description "Tools for cl-quil developers."
+  :license "Apache License 2.0 (See LICENSE.txt)"
+  :depends-on (#:cl-quil)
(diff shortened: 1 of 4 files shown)


stylewarning:
Looks fine to me

notmgsk (approved):

mhdavid-hrl:
Hi @notmgsk can I ping you about merging this in? I have a follow-on commit with the Hasse stuff I'd like to get going on, so ideally I could do that off of master. Thanks

Issue #723, opened 2021-07-08
TEST-MEASURE-SEMANTICS test fail - unknown keyword :AMPLITUDES


mhdavid-hrl:
When you evaluate
```
(asdf:test-system :cl-quil)
```
or (if the system is loaded), simply
```
(cl-quil-tests::test-measure-semantics)
```
you get an error message like the following:
```
  Invalid initialization argument:
    :AMPLITUDES
  in call for class #<STANDARD-CLASS QVM:DENSITY-QVM>.
```
This happens with the most recent version qvm: v1.14.0

mhdavid-hrl:
In qvm repo, the commit de95ead "density-qvm: provide missing kwargs to DENSITY-QVM" changed qvm:make-density-qvm function to pass along any old keyword you give it as args to make-instance on QVM:DENSITY-QVM. That turned out to be a breaking change for the caller invoked by this test TEST-MEASURE-SEMANTICS in quil. 

Now, this test function has been passing in unknown keyword :AMPLITUDES since many moons ago. Here's the code: (from: quilc: tests/misc-tests.lisp)

```
(defun %make-density-qvm-initialized-in-basis (num-qubits basis-index)
  "Make a DENSITY-QVM that is initialized in the basis state described by BASIS-INDEX.

To put the density matrix into the basis state, e.g., |01><11|, we would choose BASIS-INDEX = 7. In general, the basis state |a><b| is prepared by choosing BASIS-INDEX = (2^N * a + b)."
  (let ((amps (make-array (expt 2 (* 2 num-qubits))
                          :element-type 'qvm:cflonum
                          :initial-element (qvm:cflonum 0))))
    (setf (aref amps basis-index) (qvm:cflonum 1))
    (qvm:make-density-qvm num-qubits :amplitudes amps)))
```

But it seems it was being ignored silently, whereas now it's getting passed in, and that results in the immediate error.
The forward-trace from the test function is as follows: (leaving out package prefixes)
```
test-measure-semantics 
  -> %test-measure-semantics 
  -> %make-density-qvm-initialized-in-basis 
  -> make-density-qvm
```
[...]

PR #725, opened 2021-07-13
[#723] fix error in test-measure-semantics due to change in qvm

mhdavid-hrl:
This fixes issue 723 in quil (not qvm). A subfunction of test function
test-measure-semantics in quil (%make-density-qvm-initialized-in-basis
in tests/misc-tests.lisp) had been calling qvm system and qvm package
function make-density-qvm with :amplitudes keyword. It turns out that
this argument was being ignored. Its being ignored was apparently not
a problem for this test.  However, in a more recent version of QVM,
the keyword arg began being passed along to make-instance on
DENSITY-QVM, and this is not a supported init arg, so the test
produced an error message like the following:

  Invalid initialization argument:
    :AMPLITUDES
  in call for class #<STANDARD-CLASS QVM:DENSITY-QVM>.

The new code performs the desired functionality without the keyword
arg, and is actually simpler.

Resolves: #723

Note: this bug caused quilc CI failures due to the CI using qvm
version 1.17.2.  The CI was subsequently patched to use the older
version of qvm. This commit will work with both the old version
(v1.17.1) and the new (v1.17.2).

Note: the change in qvm that broke the quilc test was not caught by
the CI until now because quilc's CI used an older version of qvm
(qvm-v1.17.1) until just recently.  The version chosen by the CI is
determined by Quicklisp, which is loosely tied to version of QVM, but
my have a lag of weeks or longer.

### DIFF
--- a/tests/misc-tests.lisp
+++ b/tests/misc-tests.lisp
@@ -297,11 +297,10 @@
   "Make a DENSITY-QVM that is initialized in the basis state described by BASIS-INDEX.
 
 To put the density matrix into the basis state, e.g., |01><11|, we would choose BASIS-INDEX = 7. In general, the basis state |a><b| is prepared by choosing BASIS-INDEX = (2^N * a + b)."
-  (let ((amps (make-array (expt 2 (* 2 num-qubits))
-                          :element-type 'qvm:cflonum
-                          :initial-element (qvm:cflonum 0))))
-    (setf (aref amps basis-index) (qvm:cflonum 1))
-    (qvm:make-density-qvm num-qubits :amplitudes amps)))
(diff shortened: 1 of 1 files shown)


kilimanjaro (changes requested):
The indentation here seems different from elsewhere, so I'd ask that it be made consistent. But otherwise, this looks good to me.

kilimanjaro (dismissed):

mhdavid-hrl:
> The indentation here seems different from elsewhere, so I'd ask that it be made consistent. But otherwise, this looks good to me.

Ooh, good catch. Fixed.

stylewarning (approved):

PR #740, opened 2021-10-08
Replace Tweedledum classical (permutation) gate synthesis routines with native Lisp routines. 

karlosz:
This change reimplements the Tweedledum permutation gate synthesis algorithm using decomposition with a native Lisp one.
The specialized native Lisp routine offers the same 30x speedup in compilation time that Tweedledum gives over our generic routines on the Bernstein-Vazirani test program, incurring only a 5%-10% instruction increase over Tweedledum's synthesis code. This is due to lack of specialized synthesis routines for multiple control Toffoli gates. I believe that once specialized MCT synthesis algorithms are written (as Tweedledum has), we should be able to generate better code as well, since native quilc routines have access to more information, such as target chip architecture and more knowledge about the rest of the compilation pipeline.
A big plus is that we have one less external (foreign!) dependency, also notorious for constantly being rewritten.

### DIFF
--- a/src/compilers/permutation.lisp
+++ b/src/compilers/permutation.lisp
@@ -169,7 +169,7 @@
            ;; controlled Toffoli gate, so we didn't do anything and
            ;; should give up.
            (when (and code (null (rest code)))
-             (give-up-compilation :acts-trivially))
+             (give-up-compilation :because :acts-trivially))
            code))
         (t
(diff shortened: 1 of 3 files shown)


stylewarning:
You can clean up commit history if you want. I'll merge this afternoon.

PR #744, opened 2021-10-15
draft: Draft benchmark nq and rebase

ghost:
Same as PR https://github.com/quil-lang/quilc/pull/733, but has been rebased with master, particularly in order to retest CI to see if we're getting the same error during testing and, if so, to see the effect of commit https://github.com/quil-lang/quilc/commit/d1768822a6d36076d791715e09276c992169d0e1

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -147,7 +147,7 @@ test-ccl:
 ###############################################################################
 # BENCHMARKS
 ###############################################################################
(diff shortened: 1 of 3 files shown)


mhdavid-hrl:
This was used as an experiment and is no longer useful, so I'm closing it now.

PR #747, opened 2021-10-19
Teach the "old gate definition model" about SWAP.

### LINKED ISSUE #745: Allow & test SWAP as a native gate

karlosz:
Fidelity picked completely arbitrarily. Fixes #745


karlosz:
Will leave this on hold until the higher order objects pull request is through, since that has revealed some points about SWAPs and permutation records that I might want to come back to.

Issue #74, opened 2019-02-08
Cannot compile "parallel" programs on disconnected components


ecpeterson:
@ntezak  was trying to run a simple parallel 1Q program

```
DECLARE ro BIT[3]

RX(pi/2) 1
RX(pi/2) 2
RX(pi/2) 3
MEASURE 1 ro[0]
MEASURE 2 ro[1]
MEASURE 3 ro[2]
```

on the disconnected ISA

```
{'1Q': {'0': {'dead': True}, '1': {}, '2': {}, '3': {}}, '2Q': {}}
```

which resulted in a QUILC error:

```
QuilcError(result['status']) compiler_server.quilc.QuilcError: User program used too many qubits: 3 used and 1 available in the largest connected component
```

This is because the compiler will blanket-reject programs that it can't fit onto a single connected component of the device. This is unnecessary: it should only reject programs if there's no assignment to connected components that permits the requested two-qubit operations. This change would require being a little more delicate in the guts of the addresser loop, which is why we were using this simpler heuristic instead.

PR #759, opened 2021-10-29
Teach the compiler how to compile parallel programs.

karlosz:
Using disconnected components.

Previously, we only supported allocating qubits in the program onto
the single largest component, precluding parallel program compilation,
regardless of whether the QPU topology supports it. Fix this and use a
greedy allocation scheme allowing more classes of programs to get
compiled onto QPUs with multiple connected components.

Conceptually, this is as if we combined mutliple QPUs into a
"multi-core" system, and we are now scheduling quil programs to use
them intelligently.

Fixes issue #74.

### DIFF
--- a/src/addresser/addresser-common.lisp
+++ b/src/addresser/addresser-common.lisp
@@ -566,23 +566,22 @@ If DRY-RUN, this returns T as soon as it finds an instruction it can handle."
              (lscheduler-replace-instruction lschedule instr compiled-seq)))))
       t)))
 
-(defun best-qubit-position (state gates-in-waiting logical &key locations)
+(defun best-qubit-position (state gates-in-waiting logical)
   "Finds the best location for an unassigned logical under the given future schedule."
-  (with-slots (working-l2p chip-spec qubit-cc) state
-    (let ((locations (or locations qubit-cc)))
-      (assert (not (apply-rewiring-l2p working-l2p logical)) (logical)
-              "Qubit ~a already assigned" logical)
-      (let ((best-cost (build-worst-cost state))
-            (best-physical nil))
-        (dolist (physical locations)
-          (unless (or (apply-rewiring-p2l working-l2p physical)
-                      (chip-spec-qubit-dead? chip-spec physical))
-            (let ((cost (with-rewiring-assign working-l2p logical physical
-                          (cost-function state :gate-weights gates-in-waiting))))
-              (when (cost-< cost best-cost)
-                (setf best-cost cost
-                      best-physical physical)))))
-        (values best-physical best-cost)))))
+  (with-slots (working-l2p chip-spec) state
+    (assert (not (apply-rewiring-l2p working-l2p logical)) (logical)
+            "Qubit ~a already assigned" logical)
+    (let ((best-cost (build-worst-cost state))
+          (best-physical nil))
+      (dolist (physical (find-physical-component state logical))
+        (unless (or (apply-rewiring-p2l working-l2p physical)
+                    (chip-spec-qubit-dead? chip-spec physical))
+          (let ((cost (with-rewiring-assign working-l2p logical physical
+                        (cost-function state :gate-weights gates-in-waiting))))
+            (when (cost-< cost best-cost)
+              (setf best-cost cost
+                    best-physical physical)))))
+      (values best-physical best-cost))))
 
 (defun unassigned-qubits (instrs rewiring)
   "Get all qubits referenced in INSTRS which are not assigned in REWIRING."
@@ -640,7 +639,7 @@ Optional arguments:
   (:method (state instrs &key (initial-rewiring nil) (use-free-swaps nil))
     (format-noise "DO-GREEDY-ADDRESSING: entrance.")
     (let ((*addresser-use-free-swaps* (or use-free-swaps (not initial-rewiring))))
-      (with-slots (lschedule working-l2p chip-sched initial-l2p) state
+      (with-slots (chip-spec lschedule working-l2p chip-sched initial-l2p) state
         ;; This is governed by an FSM of the shape
         ;;
         ;; [ do-greedy-addressing ]
@@ -681,6 +680,9 @@ Optional arguments:
                          (assert (> *addresser-max-swap-sequence-length* (length rewirings-tried)) ()
                                  "Too many SWAP instructions selected in a row: ~a" (length rewirings-tried))
                          (setf rewirings-tried (select-and-embed-a-permutation state rewirings-tried)))))
+          (setf (addresser-state-l2p-components state)
+                (greedy-prog-ccs-to-chip-ccs (chip-spec-live-qubit-cc chip-spec)
+                                             (instr-used-qubits-ccs instrs)))
 
           ;; build the logically parallelized schedule
           (append-instructions-to-lschedule lschedule instrs)
@@ -697,7 +699,7 @@ Optional arguments:
           (format-noise "DO-GREEDY-ADDRESSING: departure.")
           ;; get rid of any floaters
           (when *addresser-use-1q-queues*
-            (dotimes (qubit (chip-spec-n-qubits (addresser-state-chip-specification state)))
+            (dotimes (qubit (chip-spec-n-qubits chip-spec))
               (unless (endp (aref (addresser-state-1q-queues state) qubit))
                 (flush-1q-instructions-after-wiring state qubit))))
           ;; finally, return what we've constructed
--- a/src/addresser/addresser-common.lisp
+++ b/src/addresser/addresser-common.lisp
@@ -573,7 +573,7 @@ If DRY-RUN, this returns T as soon as it finds an instruction it can handle."
             "Qubit ~a already assigned" logical)
     (let ((best-cost (build-worst-cost state))
           (best-physical nil))
-      (dolist (physical (find-physical-component state logical))
+      (dolist (physical (find-physical-component-in-state state logical))
         (unless (or (apply-rewiring-p2l working-l2p physical)
                     (chip-spec-qubit-dead? chip-spec physical))
           (let ((cost (with-rewiring-assign working-l2p logical physical
@@ -680,9 +680,6 @@ Optional arguments:
                          (assert (> *addresser-max-swap-sequence-length* (length rewirings-tried)) ()
                                  "Too many SWAP instructions selected in a row: ~a" (length rewirings-tried))
                          (setf rewirings-tried (select-and-embed-a-permutation state rewirings-tried)))))
-          (setf (addresser-state-l2p-components state)
-                (greedy-prog-ccs-to-chip-ccs (chip-spec-live-qubit-cc chip-spec)
-                                             (instr-used-qubits-ccs instrs)))
 
           ;; build the logically parallelized schedule
           (append-instructions-to-lschedule lschedule instrs)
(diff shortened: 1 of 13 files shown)


tests/addresser-tests.lisp
```diff
@@ -100,3 +100,17 @@
                    (quil::chip-contiguous-subschedule-from-last-instructions
```
stylewarning, line 100: Can you add a test using two qubit gates, and another test demoing failure to compile across a connected component? (Use `signals` for the latter.)

stylewarning (changes requested):
One bit of preliminary feedback.

tests/addresser-tests.lisp
karlosz: Will do. I think I'm just going to go ahead and use the condition system more aggressively in this part of the code.

stylewarning (dismissed):
It seems good to me. But I'd like @kilimanjaro to also give a passing review too.

src/addresser/addresser-common.lisp
```diff
@@ -681,6 +680,9 @@ Optional arguments:
                          (assert (> *addresser-max-swap-sequence-length* (length rewirings-tried)) ()
                                  "Too many SWAP instructions selected in a row: ~a" (length rewirings-tried))
                          (setf rewirings-tried (select-and-embed-a-permutation state rewirings-tried)))))
+          (setf (addresser-state-l2p-components state)
+                (greedy-prog-ccs-to-chip-ccs (chip-spec-live-qubit-cc chip-spec)
+                                             (instr-used-qubits-ccs instrs)))
```
kilimanjaro, line 685: I think the CC assignment has to take into account the `initial-rewiring`. If provided this contains a (partial) logical -> physical qubit assignment, and so implies a corresponding (partial) assignment of logical ccs to physical ccs. I think you should add a test which evaluates the behavior of `do-greedy-temporal-addressing` directly, e.g. providing such an initial rewiring. We'd want a nice error if the rewiring is impossible to satisfy (right now, we'd get an error, but probably during the middle of addressing), and if it is possible it should be allowed.

tests/addresser-tests.lisp
```diff
@@ -100,3 +100,58 @@
...
                     sched (quil::make-qubit-resource 0 2))
                    expected-subschedule))))))
+
+;;; Check that we can compile "parallel" programs onto disconnected
+;;; chip components.
+(deftest test-addresser-multiple-components ()
+  (let ((program "DECLARE ro BIT[3]
```
kilimanjaro, line 107: Right now, these examples only exercise the `PARTIAL` initial rewiring (no assignments exist). It's probably a good idea to cook up a test using `GREEDY` and `NAIVE` as well, verifying that they behave as expected with parallel programs. `PRAGMA INITIAL_REWIRING GREEDY` etc.

src/addresser/initial-rewiring.lisp
```diff
@@ -185,112 +173,93 @@ Otherwise, return *INITIAL-REWIRING-DEFAULT-TYPE*."
...
+       ;; in a connected component on the chip.
+       (unless (loop :for prog-connected-component in prog-connected-components
+                     :always (loop :for chip-connected-component in chip-connected-components
+                                     :thereis (subsetp prog-connected-component chip-connected-component)))
+         (error 'naive-rewiring-crosses-chip-boundaries))
+       (make-rewiring n-qubits))
+      (:partial
+       (make-partial-rewiring n-qubits))
+      (:random
+       (generate-random-rewiring n-qubits))
```
kilimanjaro, line 223: It might be reasonable to update this for disconnected chips. I've never used the RANDOM initial rewiring, but if we wanted it around I'd prefer that it can at least try to satisfy CC constraints (otherwise, a program using 4 qubits, running on a device with two disjoint 4-qubit components, with random initial rewiring would fail to compile something like 97% of the time just due to the initial assignment bridging the components).

kilimanjaro (changes requested):

tests/addresser-tests.lisp
karlosz: I think with GREEDY specifically, since it makes all the assignments, nothing bad happens since the addresser proper never needs to assign qubits, and hence it can't do anything incompatible with the connected component mapping.

src/addresser/addresser-common.lisp
karlosz: How would such a partial rewiring get provided? Do we even want to allow the user to supply a partial rewiring? It seems hard to force the logic in non-identity cases to check and allocate the components around arbitrary qubit correspondences. Greedy simply tries to assign everything from the start, using a greedy assignment from logical to physical ccs, so this machinery never fires in the addresser proper.

kilimanjaro: The rewiring is already provided here, in the form of the `initial-l2p` slot on the addresser state (and also the `initial-rewiring` keyword arg... not sure why we have both of these, but `compiler-hook` currently passes the same data through both of these). 

This is not necessarily user supplied; `do-greedy-addressing` gets called for blocks in the control flow graph, and so in principle the initial rewiring of one block is the final rewiring of its predecessor block. The machinery at this point should be agnostic to whatever 'rewiring strategy' is used for constructing the initial rewiring.

src/addresser/initial-rewiring.lisp
karlosz: Do you think anyone cares if we gut the RANDOM rewiring? It doesn't make any sense to me; we already have enough in-determinism in the compiler to deal with. Was this added for a purpose or just for fun?

kilimanjaro: Personally I'd be fine with gutting it. An adequate substitute (for a different PR) would be to add the option to explicitly set the initial rewiring; syntactically this might look like `PRAGMA INITIAL-REWIRING EXPLICIT "#(0 2 1 4 3)"`. This would be intended for power users, who'd be able to cook up random rewirings and attach them if so desired. Wonder what @stylewarning thinks.

ecpeterson: I don't feel strongly about whether it's kept, but it was originally introduced as a benchmarking convenience: if the addresser is constrained by some garbage assignment rather than beingn able to pre-assign qubits willy-nilly, how performant is its output? Starting with a random assignment tests that pretty aggressively.

karlosz: Great, that use case makes a lot more sense. It shouldn't be difficult to try and just make :random satisfy the right connected component conditions, and that machinery might be largely the same as the bookkeeping needed to implement Erik's explicit rewiring suggestion anyway (to ensure compatibility with chip ccs).

karlosz:
Updated PR to calculate the component mapping globally, making the component mapping determined together with the global initial rewiring.

stylewarning:
@karlosz is this ready for re-review?

karlosz:
Yes.

src/addresser/initial-rewiring.lisp
```diff
@@ -185,112 +169,107 @@ Otherwise, return *INITIAL-REWIRING-DEFAULT-TYPE*."
...
-              :for (other-location . distance) :in (chip-spec-live-qubit-bfs chip-spec location)
-              :do (push (cons qubit distance) (aref p2l-distances other-location)))))
-
-        rewiring))))
+  "Find an initial rewiring for a program, ensuring all used qubits in the program can fit on the connected components of the QPU compatibly, and return the connected component assignment."
+  (let ((n-qubits (chip-spec-n-qubits chip-spec))
+        (chip-connected-components (chip-spec-live-qubit-cc chip-spec))
+        (prog-connected-components (prog-used-qubits-ccs parsed-prog)))
+    (when (> (qubits-needed parsed-prog) n-qubits)
+      (error 'chip-insufficient-qubits
```
kilimanjaro, line 216: Although you're consistent here with the prior behavior, I'm not that I agree with it.

Considering
```
(prog-initial-rewiring (parse-quil "X 100") (build-nq-linear-chip 3))
```
we'd have an error, because `qubits-needed` returns 101 (1+ the max qubit index seen). I don't see how this is sensible behavior; in particular, the program at this point involves "logical qubits" (we haven't passed it through the addresser yet) so the numbers shouldn't mean anything at all (they should be considered more as variable names than hardware registers).

My feeling is that this should only be an error for `naive` rewirings. Note that it appears the issue has been skirted in` compiler-hook`, because it calls `compress-qubits`, which tries to relabel qubit indices starting from 0, and hence subsequent `max` calculations work just fine in most cases.

src/addresser/initial-rewiring.lisp
```diff
@@ -76,59 +75,44 @@ impossible using that rewiring.")
...
+                 (setf (aref connected-component-map index) index))))
+           (merge-qubit-components (qubit1 qubit2)
+             (let ((component1 (aref connected-component-map (qubit-index qubit1)))
+                   (component2 (aref connected-component-map (qubit-index qubit2))))
+               (unless (= component1 component2)
+                 (dotimes (index n-qubits)
+                   (when (= (aref connected-component-map index) component2)
+                     (setf (aref connected-component-map index) component1)))))))
+      (loop :for inst :across (parsed-program-executable-code parsed-prog)
+            :do (typecase inst
```
kilimanjaro, line 97: I'd like to handle the `RESET <q>` case here as well, even though it probably gets dealt with implicitly because anybody resetting a qubit is likely to use it later. My advice is to not do a typecase but instead rely on a suitable generic, e.g. `(remove-if-not #'qubit-p (arguments instr))`.

src/addresser/initial-rewiring.lisp
kilimanjaro: (Also, I think for this PR it's ok if we stick with the prior behavior, instead making an issue re: this point.)

karlosz: I agree this behavior is weird; I did a double take when I first tried using `qubits-needed` &co. We can change this behavior later, but that can be done independently of this change, as `qubits-needed` seems to imply relabeling per se.

stylewarning: QUBITS-NEEDED is defined by what a "quantum abstract machine" defines as needed, so the behavior is correct.

But I could be on board with renaming that to MAXIMUM-QUBIT-PRESENT and creating a function DISTINCT-QUBITS-USED, or something.

karlosz: In any case, we can rename the functions later in a separate change.

src/addresser/initial-rewiring.lisp
karlosz: Hm, I'm pretty sure the way I do it here matches the old behavior when groveling for live qubit dependencies in the program. Plus, I tried doing it the way you said and I think that actually breaks the measurement tests, since `arguments` doesn't actually correspond to what is needed. (PRAGMA doesn't even have a method defined for it.)

kilimanjaro: Well, regardless of whether it's done with an explicit typecase or a suitable generic, we should handle every instruction involving qubit arguments. What's missing here is a case for `reset-qubit`. I'd say that the previous implementation was just wrong, but we don't use the `RESET <q>` instruction very often, and even then we usually do something with the qubit afterwards, so it just slipped under our radar.

kilimanjaro: (And with that change, this PR looks good to me, assuming tests pass.)

PR #761, opened 2021-11-01
Use the qvm in the quilc compressor

kilimanjaro:
This factors out a reasonable subset of `cl-quil` into `cl-quil/frontend`. In particular, the frontend includes
- AST and parsing code
- clifford stuff
- analysis routines & transforms
- some oddball miscellaneous stuff (utilities, magicl constructors, etc) needed by the rest of the frontend

We also introduce the `cl-quil.frontend` package, exporting the symbols needed by `cl-quil` and `qvm`. 
This addresses #758 and (and along the way, #760).

Finally, we update some of the compressor routines to use the `unitary-qvm` for matrix computations.

This is a WIP in part because I am still fussing with the `cl-quil.frontend` package.

To get a sense of the the performance implications, I ran the benchmark code written by @mhdavid-hrl. Here's what I see on the current master (3ec4629) vs this branch (only looking at a few smaller chips)

`(on master)`
```
+----------------------------+----------+-------+----------+
|        NAME                | TIME (s) | SWAPS | 2Q DEPTH |
+----------------------------+----------+-------+----------+
| STATIC/FULLY-CONNECTED 10  |     2.20 |     0 |        2 |
| STATIC/FULLY-CONNECTED 15  |     3.89 |     0 |        2 |
| STATIC/FULLY-CONNECTED 20  |     6.59 |     0 |        2 |
+----------------------------+----------+-------+----------+
+----------------------------+----------+-------+----------+
|        NAME                | TIME (s) | SWAPS | 2Q DEPTH |
+----------------------------+----------+-------+----------+
| BELL/FULLY-CONNECTED 10    |     3.90 |     0 |        9 |
| BELL/FULLY-CONNECTED 15    |     7.27 |     0 |       14 |
| BELL/FULLY-CONNECTED 20    |    11.75 |     0 |       19 |
+----------------------------+----------+-------+----------+
+----------------------------+----------+-------+----------+
|        NAME                | TIME (s) | SWAPS | 2Q DEPTH |
+----------------------------+----------+-------+----------+
| QFT/FULLY-CONNECTED 10     |     8.64 |     0 |       37 |
| QFT/FULLY-CONNECTED 15     |    19.08 |     0 |       57 |
| QFT/FULLY-CONNECTED 20     |    35.08 |     0 |       77 |
+----------------------------+----------+-------+----------+
+----------------------------+----------+-------+----------+
|        NAME                | TIME (s) | SWAPS | 2Q DEPTH |
+----------------------------+----------+-------+----------+
[...]

PR #767, opened 2021-11-09
support AllegroCL 10.1 (Part I)

franzinc:
This changeset provides support for Allegro CL 10.1 (modern cased
`mlisp` not supported). The updates are mostly concerned with floats
printer compatibility.

All tests passed for both Allegro and SBCL in a 64-bit Linux
environment, except these 4 take siginificantly longer time to even finish in Allegro:

* `test-compiler-hook`
* `test-sohaib-gh-361-regression`
* `test-compiler-hook-random-4q`
* `test-chp-files`

Therefore, in this _Part I_ PR, we only present updates that are only
concerned with "correctness". Hopefully, in _Part II_, we will present
optmizations that solves the performance problem.

Change-Id: I25ff3ac8e9a696b8550bebb828b87e0858a7268c

### DIFF
--- a/src/analysis/type-safety.lisp
+++ b/src/analysis/type-safety.lisp
@@ -51,9 +51,11 @@
   (etypecase obj
     (memory-ref
      (let ((memory-region (find-descriptor-for-mref obj memory-regions)))
-       (member (memory-descriptor-type memory-region) (a:ensure-list quil-type))))
+       (apply 'member (memory-descriptor-type memory-region) (a:ensure-list quil-type)
+              #+allegro (list :test 'equalp))))
     (constant
-     (member (constant-value-type obj) (a:ensure-list quil-type)))))
+     (apply 'member (constant-value-type obj) (a:ensure-list quil-type)
+            #+allegro (list :test 'equalp)))))
 
 (defun check-mref (obj)
   (unless (typep obj 'memory-ref)
--- a/src/ast.lisp
+++ b/src/ast.lisp
@@ -1308,21 +1308,26 @@ N.B. This slot should not be accessed directly! Consider using GATE-APPLICATION-
   "Print the real number R nicely to the stream STREAM."
   (check-type r real)
   (check-type stream stream)
-  (cond
-    ((or (double~ (abs r) 0) (not *print-fractional-radians*))
-     (format stream "~F" r))
-    (t
-     (loop :with r-abs := (abs r)
-           :for rr :across **reasonable-rationals**
-           :for numer := (numerator rr)
-           :for denom := (denominator rr)
-           :when (double~ r-abs (/ (* pi numer) denom)) :do
-             ;; Pretty-print "reasonable" integer and fractional multiples of pi.
-             (format stream "~:[~;-~]~:[~D*~;~*~]~A~:[/~D~;~*~]"
-                     (minusp r) (= 1 numer) numer *pi-literal* (= 1 denom) denom)
-             (return-from format-real))
-     ;; If we cannot find a nice fraction of pi, just print the real number.
-     (format stream "~F" r))))
+  (flet ((%format (r)
+           (if (floatp r)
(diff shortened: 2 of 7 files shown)


macdavid313:
Hello @stylewarning , I will explain many details in the PR soon.

macdavid313:
The most notable change is wrapping `(format nil "~F" ...)` forms with a lexical binding of a special variable `*read-default-float-format*`.  In Allegro, printing floats follows the ANSI standard which is described [here](https://franz.com/support/documentation/10.1/ansicl/subsubsu/printin0.htm). Note that:

>  If the format of the number does not match that specified by *read-default-float-format*, then the exponent marker for that format and the digit 0 are also printed.

For example, as `*read-default-float-format*` evaluates to `'single-float` by default, this works as expected:

```lisp
CL-USER(1): (format t "~F" 1f0)
1.0 
NIL
```

However, if it's printing a double float, the [exponent marker](https://franz.com/support/documentation/10.1/ansicl/glossary/e.htm#exponentmarker) will be printed, because `*read-default-float-format*` is different from `'double-float` :

```lisp
CL-USER(2): (format t "~F" 1d0)
1.0d0
NIL
```

Binding `*read-default-float-format*` with `'double-float` will print the expected result:

```lisp
CL-USER(3): (let ((*read-default-float-format* 'double-float)) (format t "~F" 1d0))
1.0
NIL
```

This is why you will see a lot of re-binding of `*read-default-float-format*` in this PR. Hope it makes sense to you.

src/analysis/type-safety.lisp
```diff
@@ -51,9 +51,11 @@
   (etypecase obj
     (memory-ref
      (let ((memory-region (find-descriptor-for-mref obj memory-regions)))
-       (member (memory-descriptor-type memory-region) (a:ensure-list quil-type))))
+       (apply 'member (memory-descriptor-type memory-region) (a:ensure-list quil-type)
+              #+allegro (list :test 'equalp))))
```
macdavid313, line 55: We still don't know if this is a bug in Allegro or not. It's quite hard to reproduce the issue in a simple script. But here's a gist of what happened ...

The list `quil-type` is a list of constants, e.g. `(list quil-bit)`. Due to optimization of `list`, `(list quil-bit)` is actually using the compile-time (instead of load-time) value of `quil-bit`. And that leads to a problem.

After a GC, `quil-bit` may be moved, so the default `:test` of `member`, which is `'eql`, is not useful for comparing structs anymore. The standard only says a constant should be `immutable`, and for a non-immediate value like a struct, *immutable* is rather tricky to interpret.

Hope it makes sense to you.

macdavid313:
Hello @stylewarning, I have left a few more detailed explanations. Feel free to leave comments and suggest edits later. Thank you.

stylewarning (approved):

karlosz:
@franzinc I have recently discovered in #761 that quilc spends almost all of its time calling `COMPILE` several times. This has huge performance implications, since the speed of a quilc compilation is now tied to how fast the Lisp compiler does its job at runtime. I suspect this may be the root of some of your performance troubles. I suggest trying again once #761 is finished, since that will remove these runtime calls to `COMPILE`.

macdavid313:
@karlosz Hello Zhang, thanks for the advice! Is it possible to show me your experiment code? I'm currently profiling `(cl-quil-tests::test-compiler-hook)` but it doesn't show any clues related to `COMPILE`.

karlosz:
I sampled a couple seconds of the whole test suite and used an sbcl specific statistical profiler to produce the profile linked. Not sure but when you run the test suite I expect there to be something related to using the compiler at runtime.

mhdavid-hrl:
The bug seems to be caused by code changes explained by @macdavid313 in this PR files comment: https://github.com/quil-lang/quilc/pull/767/files#r745348410
The code was changed to use APPLY, but there's no list of args in the non-allegro case, which leads to the bug.

PR #774, opened 2021-11-19
Addresser cleanups

karlosz:


stylewarning (approved):
looks OK to me

src/utilities.lisp
```diff
@@ -264,3 +264,49 @@ contains the bits of INTEGER. See http://www.cliki.net/ROTATE-BYTE"
   (calculate-instructions-fidelity
    (coerce (parsed-program-executable-code program) 'list)
    chip))
+
+;;;; Queues, taken with permission from https://bitbucket.org/tarballs_are_good/lisp-random/raw/a595eb662c2dce87f362dd0dd2541a93efe9c902/stack-queue.lisp
```
kilimanjaro, line 268: Can we put this stuff in its own `queue.lisp` or `stack-queue.lisp`? While we can in principle put anything in `utilities.lisp`, I think it's nice for something like a fully-fledged ADT to not get stashed in here with the rest of the rabble.

kilimanjaro (changes requested):
This looks good to me, aside from one minor suggestion/request re: where we stash the queue implementation.

src/utilities.lisp
stylewarning: I'll invite this as a separate PR. I agree with it.

stylewarning:
(sorry, I merged this a bit too quickly, since Erik depends on these changes)

Issue #486, opened 2019-11-19
"Efficient template matching in quantum circuits" paper


stylewarning:
There's a wonderful [paper](https://arxiv.org/abs/1909.05270) by Raban Iten, David Sutter, Stefan Woerner on compressor-like functionality for quantum circuits. At minimum, we should review it. At maximum, we should implement it. What do you folks think?

ecpeterson:
I think that their method is superior whenever it overlaps ours (viz., peephole rewriting input circuits of a fixed shape mapping onto output circuits of a fixed shape). However, the compressor broadly does additional things that I think will take significant effort (or which I don't even understand how) to fit into their framework:

1. The compressor selects _unstructured_ subgraphs of instructions to apply its (instructions -> single matrix -> instructions) logic to. I feel like this is something their searching methods could be modified to handle, but I don't think it's spelled out and it's not clear to me how to intelligently do it in tandem with the more usual rewrites.
2. One of their selling points is that their method allows for template reuse: for example, if U1 ... Un = 1 is recorded as a template and Ui ... Uj is found as a subsequence, then it can be replaced by U(i-1)^-1 ... U1^-1 Un^-1 ... U(j+1)^-1 (& similar generalizations for subgraphs). It's less clear to me to what degree this really has legs when working with fancier "templates", like those with parametric gates, or those which accept a wide variety of gates (e.g., the tweedledum permutation compiler) and the shape of whose output is conditional on the precise input given (e.g., the tweedledum permutation compiler).
3. The compressor optionally tracks partial quantum state through the program. This gets somewhat fussy when de-linearizing the traversal order: given a subgraph of instructions, you have to make sure you've used the entire "lightcone" of instructions behind a given subgraph to generate the present state. Making sure that this is so, that this is stable under graph transformations, & that it can be partially memoized will require care.

It's possible to dodge all 3 of these by retaining the compressor logic and only swapping out the peephole optimizer for their method. However, I think that their method is robust enough and has enough overlap with the broader compressor that I'd want to put serious thought into how much of the compressor can be outright replaced. Of course, I would also want to have proposed partial solutions to each of the items above before proceeding with an implementation.

ecpeterson:
[...]

PR #783, opened 2021-12-09
Some additional support for higher order objects

karlosz:


tests/compilation-tests.lisp
```diff
@@ -195,6 +195,13 @@ CNOT 0 2"))
     ;; test on quality of compilation, and not on correctness.
     (is (>= 6 (length 2q-code)))))
 
+(deftest test-native-ccnot-compilation ()
```
ecpeterson, line 198: 🎉

src/addresser/addresser-state.lisp
```diff
@@ -177,20 +177,33 @@ INSTR is the \"active instruction\".
...
-                                                        instrs-compressed)
-                      (funcall hardware-op hw))))))
+                    (unless (gethash "no-compiler-path-found"
+                                     (hardware-object-misc-data hw))
+                      (handler-case
+                        ;; This code is optional; it's only to needed
+                        ;; to help compute a bound on the single
+                        ;; operation cost on this chip. This may be a
+                        ;; source of non-determinism.
+                        (let* ((instr (apply #'anon-gate "FLEX" (random-special-unitary (expt 2 qubits))
```
ecpeterson, line 187: I know you're just refactoring something gross I wrote, but I want to put on the record that this is a very poor kludge. I might open an issue.

src/compilation-methods.lisp
```diff
@@ -296,11 +296,14 @@ Returns a value list: (processed-program, of type parsed-program
...
                       (fully-native-quil (expand-to-native-instructions straight-line-quil chip-specification))
                       (processed-quil fully-native-quil))
-                 (dotimes (n *compressor-passes*)
-                   (format-noise "COMPILER-HOOK: Compressing, pass ~D/~D." (1+ n) *compressor-passes*)
-                   (setf processed-quil
-                         (compress-instructions processed-quil chip-specification
-                                                :protoquil (null registrant))))
+                 ;; order >1 hardware objects not yet supported by
+                 ;; compressor.
+                 (when (<= (length (chip-specification-objects chip-specification)) 2)
```
ecpeterson, line 301: Prefer `(>= 2 (length (chip-specification-objects chip-specification)))`.

src/compilation-methods.lisp
```diff
@@ -296,11 +296,14 @@ Returns a value list: (processed-program, of type parsed-program
...
+                   (dotimes (n *compressor-passes*)
+                     (format-noise "COMPILER-HOOK: Compressing, pass ~D/~D." (1+ n) *compressor-passes*)
+                     (setf processed-quil
+                           (compress-instructions processed-quil chip-specification
```
ecpeterson, line 305: Another note for another PR: it's probably the case both that the compressor as-is is going to be gnarly to modify to support higher-order hardware objects and that the variant suggested in #486 would be relatively straightforward. Consider including that ticket in this arc.

src/chip/chip-specification.lisp
```diff
@@ -314,6 +379,40 @@ used to specify CHIP-SPEC."
     ;; return the link
     obj))
 
+(defun build-hardware-object (qubits &key type gate-information)
```
ecpeterson, line 382: I see that `qubits` isn't bounded from above, but the contents of `type` make it sound specialized to order 2. Do you really want `qubits` to be unbounded? Or would you prefer something like `build-triplet` (or something similar)? Does it make sense to introduce a nontrivial `type` kwarg when that's the "legacy" data path?

ecpeterson (commented):
This looks like it takes the first few steps in a good direction. I left some comments, a couple of them trivial. I don't feel up to speed on quilc anymore, so don't take my 👍 as a clear sign that there's nothing left to think about, but I hope it's some comfort that there's nothing glaring to me.

I'm glad you picked up on how the permutation record stuff is just kinda dangling, and I hope you find a way to make good use of it!

src/addresser/addresser-state.lisp
ecpeterson: #786

Issue #762, opened 2021-11-02
Prototype `DEFGATE .. AS GATE-SEQUENCE` or similar


stylewarning:
There is currently no way to define a gate based off of a sequence of other gates. Let's prototype syntax and semantics for something like this:

```
DEFGATE HADAMARD q AS GATE-SEQUENCE:
    RZ(pi/2) q
    RX(pi/2) q
    RZ(pi/2) q
```

Parameters could also be specified. The `AS GATE-SEQUENCE` is just a suggestion. Perhaps `AS GATES` instead?

The point of this feature is to _define a gate_, and not to define a circuit. While the compiler might decide to expand the given gates, it doesn't need to, and can treat it completely opaquely.

There are a couple questions that arise in implementing such a feature:

1. What about mutually dependent definitions? (Non-self) recursive definitions?
2. Should gate sequences be a more first-class definable concept? At some point we will want user-installable gate sequence expansions in QUILC, and maybe there's overlap here.

**Note**: This would require a change to the Quil spec.

**Note**: For prototyping, the syntax is pretty close to `AS PAULI-SUM` in terms of the declaration, and so that code path may be able to be used.

PR #785, opened 2022-01-07
Defgate as sequence

Spin1Half:
Proposed implementation of defgate as sequence outlined in issue #762 

This adds syntax to Quil and rudimentary integration with the compilation pipeline.

### DIFF
--- a/cl-quil.asd
+++ b/cl-quil.asd
@@ -165,7 +165,7 @@
                              (:file "rewrite-arithmetic")
                              (:file "fusion")
                              (:file "simplify-arithmetic")
-                             (:file "validate-seqence-gate")))))
+                             (:file "validate-sequence-gate")))))
 
 ;;; Contribs
 
--- a/cl-quil.asd
+++ b/cl-quil.asd
@@ -164,7 +164,8 @@
                              (:file "expand-circuits")
                              (:file "rewrite-arithmetic")
                              (:file "fusion")
-                             (:file "simplify-arithmetic")))))
+                             (:file "simplify-arithmetic")
+                             (:file "validate-seq-gate")))))
 
 ;;; Contribs
 
--- a/cl-quil.asd
+++ b/cl-quil.asd
@@ -165,7 +165,7 @@
                              (:file "rewrite-arithmetic")
                              (:file "fusion")
                              (:file "simplify-arithmetic")
-                             (:file "validate-seq-gate")))))
+                             (:file "validate-seqence-gate")))))
 
 ;;; Contribs
 
--- a/src/analysis/resolve-objects.lisp
+++ b/src/analysis/resolve-objects.lisp
@@ -98,10 +98,10 @@
       ;; resolve defgate as sequence definitions
       (map nil (lambda (sd)
                  (when (typep sd 'sequence-gate-definition)
-                       (let ((*in-definition-body* t))
-                         (resolve-instruction-sequence
-                          (sequence-gate-definition-sequence sd)))
-                       (validate-resolved-seq-gate-definition sd)))
+                   (let ((*in-definition-body* t))
+                     (resolve-instruction-sequence
+                      (sequence-gate-definition-sequence sd)))
+                   (validate-resolved-seq-gate-definition sd)))
            (parsed-program-gate-definitions unresolved-program))
       unresolved-program)))
 
--- a/src/analysis/resolve-objects.lisp
+++ b/src/analysis/resolve-objects.lisp
@@ -97,11 +97,11 @@
            (parsed-program-circuit-definitions unresolved-program))
       ;; resolve defgate as sequence definitions
       (map nil (lambda (sd)
-                 (cond ((typep sd 'sequence-gate-definition)
-                        (let ((*in-definition-body* t))
-                          (resolve-instruction-sequence
-                           (sequence-gate-definition-sequence sd)))
-                        (validate-resolved-seq-gate-definition sd))))
+                 (when (typep sd 'sequence-gate-definition)
+                       (let ((*in-definition-body* t))
+                         (resolve-instruction-sequence
+                          (sequence-gate-definition-sequence sd)))
+                       (validate-resolved-seq-gate-definition sd)))
            (parsed-program-gate-definitions unresolved-program))
       unresolved-program)))
 
--- a/src/analysis/resolve-objects.lisp
+++ b/src/analysis/resolve-objects.lisp
@@ -95,4 +95,13 @@
                    (resolve-instruction-sequence
                     (circuit-definition-body cd))))
            (parsed-program-circuit-definitions unresolved-program))
+      ;; resolve defgate as sequence definitions
+      (map nil (lambda (sd)
+                 (cond ((typep sd 'sequence-gate-definition)
+                        (let ((*in-definition-body* t))
+                          (resolve-instruction-sequence
+                           (sequence-gate-definition-sequence sd)))
+                        (validate-resolved-seq-gate-definition sd))))
+           (parsed-program-gate-definitions unresolved-program))
       unresolved-program)))
+
--- a/src/analysis/validate-seq-gate.lisp
+++ b/src/analysis/validate-seq-gate.lisp
@@ -27,6 +27,13 @@
 (defun seq-contains-only-gate-definitions-p (sequence)
   (every #'gate-application-p sequence))
 
+(define-transform validate-defgate-loops (validate-defgate-loops)
+  "This transform traverses all defgate as sequence objects to verify there are no circular references.")
+
+(defun validate-defgate-loops (parsed-program)
+  (verify-no-loops (graph-of-seq-def parsed-program))
+  parsed-program)
+
 (defgeneric graph-of-seq-def (program)
   (:method ((program parsed-program))
     (graph-of-seq-def (parsed-program-gate-definitions program)))
@@ -50,16 +57,21 @@
     nil))
 
 (defun verify-no-loops (graph &optional (path NIL))
-  (if graph
-      (if path
-          (map nil (lambda (x) (if (member x path)
-                                   (error "graph contains loops")
-                                   (verify-no-loops graph (cons x path))))
-               (neighboors-of graph (car (last path))))
-          (map nil (lambda (x) (verify-no-loops graph (list (first x)))) graph)) NIL))
+  (if (and (not (null (cdr path)))
+           (null (first path)))
+      NIL
+      (if graph
+          (if path
+              (map nil (lambda (x) (if (member x path)
+                                       (error (format nil "Defgate sequence dependencies contain loops: ~a " path))
+                                       (verify-no-loops graph (cons x path))))
+                   (neighboors-of graph (first path)))
+              (map nil (lambda (x) (verify-no-loops graph (list (first x)))) graph))
+          NIL)))
 
 (defun neighboors-of (graph key)
   (if graph
-      (if (= (first (first graph)) key)
+      (if (string= (first (first graph)) key)
           (rest (first graph))
-          (neighboors-of (rest graph) key)) NIL))
+          (neighboors-of (rest graph) key))
+      NIL))
--- a/src/analysis/validate-seq-gate.lisp
+++ b/src/analysis/validate-seq-gate.lisp
@@ -12,14 +12,14 @@
              (validate-argument (arg)
                (cond
                  ((qubit-p arg)
-                  (error "argument cannot be qubit"))
+                  (quil-parse-error "argument cannot be qubit"))
                  ((not (is-formal arg))
-                  (error "argument is not formal"))
+                  (quil-parse-error "argument is not formal"))
                  ((not (find arg args :test #'formal=))
-                  (error "argument ~a is not in gatedef arguments" arg))
+                  (quil-parse-error "argument ~a is not in gatedef arguments" arg))
                  (T T))))
       (unless (seq-contains-only-gate-definitions-p seq)
-        (error (format nil "Sequence gate definition body (~a) contains non gate-application" (gate-definition-name gate-def) )))
+        (quil-parse-error (format nil "Sequence gate definition body (~a) contains non gate-application" (gate-definition-name gate-def) )))
       (loop :for isn :in seq
             :for isn-args := (application-arguments isn)
             :do (map nil #'validate-argument isn-args)))))
@@ -34,14 +34,13 @@
   (verify-no-loops-in-gate-defs (parsed-program-gate-definitions parsed-program))
   parsed-program)
 
-(defun verify-no-loops-in-gate-defs (gate-defs &optional (path NIL))
+(defun verify-no-loops-in-gate-defs (gate-defs &optional (ancestors NIL))
   "Takes list of gate defs, verifys there are no circular dependencies between sequence gate defs."
-  (mapcar (lambda (x) (when (typep x 'SEQUENCE-GATE-DEFINITION)
-                        (let ((gate-name (gate-definition-name x)))
-                          (if (member gate-name path)
-                              (error (format nil "Defgate sequence dependencies contains a loop: ~a " (reverse (cons gate-name path))))
-                              (verify-no-loops-in-gate-defs (neighbors-of-sequence-gate-def x) (cons gate-name path))))))
-          gate-defs))
+  (dolist (gate-def gate-defs) (when (typep gate-def 'SEQUENCE-GATE-DEFINITION)
+                                 (let ((gate-name (gate-definition-name gate-def)))
+                                   (if (member gate-name ancestors)
+                                       (quil-parse-error (format nil "Defgate sequence dependencies contains a loop: ~a " (reverse (cons gate-name ancestors))))
+                                       (verify-no-loops-in-gate-defs (neighbors-of-sequence-gate-def gate-def) (cons gate-name ancestors)))))))
 
 (defun neighbors-of-sequence-gate-def (gate-def)
   "Returns all sequence gate defs that another sequence gate def depends on."
--- a/src/analysis/validate-seq-gate.lisp
+++ b/src/analysis/validate-seq-gate.lisp
@@ -57,21 +57,21 @@
     nil))
 
 (defun verify-no-loops (graph &optional (path NIL))
-  (if (and (not (null (cdr path)))
-           (null (first path)))
-      NIL
-      (if graph
-          (if path
-              (map nil (lambda (x) (if (member x path)
+  "Accepts a graph of structure ((node1 neighboora neighboorb ..) (node2 ..)) and verifies no loops exist. If path is provided, it prunes the search space to verify no loops exist for all unique paths in the graph beginning with that path."
+  (when (and graph
+             (or (null (cdr path))
+                 (first path)))
+    (if path
+        (map nil (lambda (x) (if (member x path)
                                        (error (format nil "Defgate sequence dependencies contain loops: ~a " path))
                                        (verify-no-loops graph (cons x path))))
-                   (neighboors-of graph (first path)))
-              (map nil (lambda (x) (verify-no-loops graph (list (first x)))) graph))
-          NIL)))
+             (neighbors-of graph (first path)))
+        (map nil (lambda (x) (verify-no-loops graph (list (first x)))) graph))))
 
-(defun neighboors-of (graph key)
-  (if graph
+
+(defun neighbors-of (graph key)
+  "Returns all neighboors of a node (KEY) for a GRAPH of structure ((node1 neighboora neighboorb ..) (node2 ..))"
+  (when graph
       (if (string= (first (first graph)) key)
           (rest (first graph))
-          (neighboors-of (rest graph) key))
-      NIL))
+          (neighboors-of (rest graph) key))))
--- a/src/analysis/validate-seq-gate.lisp
+++ b/src/analysis/validate-seq-gate.lisp
@@ -1,0 +1,65 @@
+;;;; validate-seq-gate.lisp
+;;;;
+;;;; Author: Parker Williams
+
+(in-package #:cl-quil.frontend)
+
+(defun validate-resolved-seq-gate-definition (gate-def)
+  (declare (type sequence-gate-definition gate-def))
+  (let ((seq (sequence-gate-definition-sequence gate-def))
+        (args (sequence-gate-definition-arguments gate-def)))
+    (labels (
+             (validate-argument (arg)
+               (cond
+                 ((qubit-p arg)
+                  (error "argument cannot be qubit"))
+                 ((not (is-formal arg))
+                  (error "argument is not formal"))
+                 ((not (find arg args :test #'formal=))
+                  (error "argument ~a is not in gatedef arguments" arg))
+                 (T T))))
+      (unless (seq-contains-only-gate-definitions-p seq)
+        (error "sequence gate definition body contains non gate-application"))
+      (loop :for isn :in seq
+            :for isn-args := (application-arguments isn)
+            :do (map nil #'validate-argument isn-args)))))
+
+(defun seq-contains-only-gate-definitions-p (sequence)
+  (every #'gate-application-p sequence))
+
+(defgeneric graph-of-seq-def (program)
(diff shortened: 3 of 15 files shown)


ecpeterson:
Responding to the general RFC from the Quil stack meeting a couple weeks ago:

I very much like the spirit of this PR / extension to the Quil spec. I don't much care for DEFCIRCUIT, and I think this captures many of the parts of that which are in common use while avoiding some of the snares.

One main concern: what does the second instance of `U` mean in the following?

```
DEFGATE U q AS GATE-SEQUENCE:
    U q
```

Other `DEFGATE` instances don't suffer from this potential recursion, and I think it's legal in a DEFCIRCUIT (though the compiler's "expand" pass may complain). I believe there's something semantic to pin down here related to name resolution — which, AFAIK, is part of the compiler internals but whose behavior is not expressly prescribed at the level of the language spec.

Spin1Half:
@ecpeterson The plan right now is to consider any circularity in the definitions invalid (i.e. its fine for one defgate to depend on another, as long as there are no circular dependencies).  I have made a validation routine to identify any circular dependencies and produce an error in this case (not pushed yet, stay tuned).

ecpeterson:
Reasonable 👍, but also restrictive — my understanding from the meeting was that this was to be one way to describe “translators” to the compiler, and the quoted case is such a translator which is presently used.

Just something to think about. Should gates be permitted to have multiple compatible definitions — some potentially “self”-referencing another definition? Maybe! Sounds like a lot of work to make use of, but it could be worth consciously avoiding making the present guts deeply incompatible with such a future.

stylewarning:
I think it's a reasonable restriction if these must act as unambiguous gate *definitions*. I think the restriction could be lifted if we go in the direction of adding a keyword (eg DEFGATE blah AS ALTERNATIVE blah) or a pragma (I'm less enthusiastic) which won't install the gate as a definition, but rather as an identity to be dwelled upon. Something like ALTERNATIVE implies there was already an unambiguous definition somewhere.

Spin1Half:
I believe this PR to be *functionally* complete, so I will take it out of draft. Though I certainly expect there are spots that can be improved. Looking forward to any feedback!

src/analysis/resolve-objects.lisp
```diff
@@ -95,4 +95,13 @@
                    (resolve-instruction-sequence
                     (circuit-definition-body cd))))
            (parsed-program-circuit-definitions unresolved-program))
+      ;; resolve defgate as sequence definitions
```
karlosz, line 98: Two things: Could this map get folded into the map above?

And here, with `(cond ...)` and only one leg, the more idiomatic usage is `(when <cond> <body>)`.

src/analysis/validate-seq-gate.lisp
```diff
@@ -0,0 +1,77 @@
...
+(defgeneric sequence-gate-def-to-graph-node (gate-def)
+  (:method ((gate-def sequence-gate-definition))
+    (cons (gate-definition-name gate-def) (sequence-gate-def-to-graph-node (sequence-gate-definition-sequence gate-def))))
+  (:method ((gate-def list))
+    (cons (operator-description-name (application-operator (first gate-def)))
+            (sequence-gate-def-to-graph-node (rest gate-def))))
+  (:method ((gate-def null))
+    nil))
+
+(defun verify-no-loops (graph &optional (path NIL))
```
karlosz, line 59: Try to use `(cond ...)` here.

src/analysis/validate-seq-gate.lisp
```diff
@@ -0,0 +1,77 @@
...
+      (if graph
+          (if path
+              (map nil (lambda (x) (if (member x path)
+                                       (error (format nil "Defgate sequence dependencies contain loops: ~a " path))
+                                       (verify-no-loops graph (cons x path))))
+                   (neighboors-of graph (first path)))
+              (map nil (lambda (x) (verify-no-loops graph (list (first x)))) graph))
+          NIL)))
+
+(defun neighboors-of (graph key)
```
karlosz, line 72: `neighbors`?

Also, again `(cond ...)` seems more idiomatic here.

src/gates.lisp
```diff
@@ -190,6 +190,91 @@
...
+  ((parameters :initarg :parameters
+               :reader sequence-gate-parameters)
+   (arguments :initarg :arguments
+              :reader sequence-gate-arguments)
+   (sequence :initarg :sequence ;;<- I dont know about this one
+             :reader sequence-gate-sequence))
+  (:documentation "A gate specified by a sequence of other gates"))
+
+
+;;; we dont think this is neccesary because seq gate definition
```
karlosz, line 209: Grammar/spelling.

src/analysis/validate-seq-gate.lisp
```diff
@@ -0,0 +1,77 @@
...
+  (every #'gate-application-p sequence))
+
+(define-transform validate-defgate-loops (validate-defgate-loops)
+  "This transform traverses all defgate as sequence objects to verify there are no circular references.")
+
+(defun validate-defgate-loops (parsed-program)
+  (verify-no-loops (graph-of-seq-def parsed-program))
+  parsed-program)
+
+(defgeneric graph-of-seq-def (program)
```
karlosz, line 37: So the code here to build a graph and do cyclic dependency analysis probably needs at least a comment explaining the data structure used for the graph, and potentially a test to make sure this machinery works. If the graph is only being used for loop detection, there is probably a lighter weight way of doing this by just directly doing DFS, or if we'd like to have general "call" graphs, it might be better to have a more polished interface to that.

karlosz (changes requested):
My initial review. Good work.

src/analysis/validate-seq-gate.lisp
Spin1Half: Alternatively, one could opt for something like:
```
(cond ((and S1 S2) 
        (cond (S3 A) (t B)))
    (t Nil))
```

karlosz: I'd rewrite the condition like this:
```(when (and graph
                         (or (null (cdr path)))
                               (first path)))
        (if path
            ...
            ...))
```
actually. But its hard to tell whats going on because its unclear how the graph is being represented. I think you'll find with a better representation the conditionalization will simplify a lot.

`WHEN` is used specifically when you are just trying to do something for its side effect, as seems to be the case here.

src/gates.lisp
```diff
@@ -190,6 +190,91 @@
...
+                              (application-parameters seq-gate-application)
+                              (application-arguments seq-gate-application)))
+
+(defun instantiate-gate-application (gate-app lut-args lut-params)
+  (setf gate-app (copy-instance gate-app))
+  (setf (application-parameters gate-app) (instantiate-from-param-name (application-parameters gate-app) lut-params))
+  (setf (application-arguments gate-app) (instantiate-from-formal-name (application-arguments gate-app) lut-args))
+  gate-app)
+
+(defun instantiate-from-formal-name (list-names lut-names)
```
kilimanjaro, line 239: I think this could be `(mapcar (lambda (name) (gethash (formal-name name) lut-names)) lit-names)`, and similarly for `instantiate-from-param-name`. In general, I prefer using `mapcar` rather than direct list traversal, where sensible.

src/gates.lisp
```diff
@@ -190,6 +190,91 @@
...
+;;;
+
+(defun instantiate-sequence-gate (seq-gate parameters arguments)
+  "Expands an instantiation of a sequence gate into a list of gate applications"
+  (let* (
+         (lookup-arg (make-map-list-to-list (sequence-gate-definition-arguments seq-gate) arguments #'formal-name))
+         (lookup-param (make-map-list-to-list (sequence-gate-definition-parameters seq-gate) parameters #'param-name))
+         (instantiated-list nil))
+    (loop :for gate-app :in (reverse (sequence-gate-definition-sequence seq-gate))
+          :do (setf instantiated-list (cons (instantiate-gate-application gate-app lookup-arg lookup-param) instantiated-list)))
```
kilimanjaro, line 224: This is another place where `mapcar` is more convenient -- no need to reverse the list, just 

```
(mapcar (lambda (gate-app) (instantiate-gate-application gate-app lookup-arg lookup-param)) 
  (sequence-gate-definition-sequence seq-gate))
```

src/gates.lisp
```diff
@@ -190,6 +190,91 @@
...
+(defun instantiate-gate-sequence-application (seq-gate-application)
+  "Expands an instantiation of a sequence gate into a list of gate applications"
+  (instantiate-sequence-gate  (gate-application-resolution seq-gate-application)
+                              (application-parameters seq-gate-application)
+                              (application-arguments seq-gate-application)))
+
+(defun instantiate-gate-application (gate-app lut-args lut-params)
+  (setf gate-app (copy-instance gate-app))
+  (setf (application-parameters gate-app) (instantiate-from-param-name (application-parameters gate-app) lut-params))
+  (setf (application-arguments gate-app) (instantiate-from-formal-name (application-arguments gate-app) lut-args))
```
kilimanjaro, line 236: As a tiny convenience, you can do multiple assignments in a single `setf` form, e.g.
```
(setf a 1
        b 2)
```
They will be evaluated sequentially.

kilimanjaro (changes requested):

src/gates.lisp
```diff
@@ -190,6 +190,91 @@
 (defmethod gate-matrix ((gate parameterized-gate) &rest parameters)
   (apply (parameterized-gate-matrix-function gate) parameters))
 
+
+;;; SEQUENCE GATE
+;;;  
+;;;  Ordinarilly we would define a class like the following
+;;;
+#+ignore
```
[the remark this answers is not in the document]
stylewarning: I'm ok if it's `#+#:some-comment` since that can never be in `*features*`

src/gates.lisp
```diff
@@ -393,3 +478,135 @@ The Pauli sum is recorded as a list of PAULI-TERM objects, stored in the TERMS s
                    :terms terms
                    :dimension (expt 2 (length arguments))
                    :arity (length arguments))))
+
+(defmethod gate-definition-to-gate ((gate-def sequence-gate-definition))
+  ;;hack, should change as given new opimizations
+  gate-def
+  )
+
+;;;; some leftover stuff from standard-gates.lisp and elsewhere
```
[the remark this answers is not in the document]
Spin1Half: Might have gotten moved while I was working on this PR? I was confused why this showed up in the diff as well, meant to point it out.

src/gates.lisp
```diff
@@ -190,6 +190,91 @@
...
+
+(defun make-map-list-to-list (lista listb apply-to-key)
+  "evaluates apply-to-key on elements of lista and maps them to listb by index, result of apply-to-key must be string"
+  (let ((return-table (make-hash-table :test 'equal)))
+    (loop :for key :in lista
+          :for val :in listb
+          :do (setf (gethash (funcall apply-to-key key) return-table) val))
+    return-table))
+
+(defmethod gate-matrix ((gate sequence-gate-definition) &rest parameters)
```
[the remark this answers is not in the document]
Spin1Half: Definitely agree, I will include this comment in the "make defgate as sequence not shamelessly unoptimized" ticket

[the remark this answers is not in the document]
Spin1Half: https://github.com/quil-lang/quilc/issues/792

src/analysis/validate-seq-gate.lisp
Spin1Half: All done, looks much better now, way less code.

src/analysis/validate-seq-gate.lisp
```diff
@@ -34,14 +34,13 @@
...
-(defun verify-no-loops-in-gate-defs (gate-defs &optional (path NIL))
+(defun verify-no-loops-in-gate-defs (gate-defs &optional (ancestors NIL))
   "Takes list of gate defs, verifys there are no circular dependencies between sequence gate defs."
-  (mapcar (lambda (x) (when (typep x 'SEQUENCE-GATE-DEFINITION)
-                        (let ((gate-name (gate-definition-name x)))
-                          (if (member gate-name path)
-                              (error (format nil "Defgate sequence dependencies contains a loop: ~a " (reverse (cons gate-name path))))
-                              (verify-no-loops-in-gate-defs (neighbors-of-sequence-gate-def x) (cons gate-name path))))))
-          gate-defs))
+  (dolist (gate-def gate-defs) (when (typep gate-def 'SEQUENCE-GATE-DEFINITION)
```
karlosz, line 39: Idiomatic formatting for `dolist` is to have a newline before the body, i.e. before the `when`.

karlosz (approved):
Lgtm outside of one minor nit.

cl-quil.asd
```diff
@@ -164,7 +164,8 @@
                              (:file "expand-circuits")
                              (:file "rewrite-arithmetic")
                              (:file "fusion")
-                             (:file "simplify-arithmetic")))))
+                             (:file "simplify-arithmetic")
+                             (:file "validate-seq-gate")))))
```
stylewarning, line 168: can you spell out `sequence-gate`? and `git mv` the file to the new name

src/analysis/resolve-objects.lisp
```diff
@@ -95,4 +95,13 @@
...
+      (map nil (lambda (sd)
+                 (when (typep sd 'sequence-gate-definition)
+                       (let ((*in-definition-body* t))
```
stylewarning, line 101: indentation looks wrong on github; please verify this gets indented correctly.

src/analysis/validate-seq-gate.lisp
```diff
@@ -0,0 +1,47 @@
...
+(in-package #:cl-quil.frontend)
+
+(defun validate-resolved-seq-gate-definition (gate-def)
+  (declare (type sequence-gate-definition gate-def))
+  (let ((seq (sequence-gate-definition-sequence gate-def))
+        (args (sequence-gate-definition-arguments gate-def)))
+    (labels ((validate-argument (arg)
+               (cond
+                 ((qubit-p arg)
+                  (quil-parse-error "argument cannot be qubit"))
```
stylewarning, line 14: > argument of blah-blah cannot be qubit

src/analysis/validate-seq-gate.lisp
```diff
@@ -0,0 +1,47 @@
...
+                 ((not (is-formal arg))
+                  (quil-parse-error "argument is not formal"))
```
stylewarning, line 16: same here, improve error message

src/analysis/validate-seq-gate.lisp
```diff
@@ -0,0 +1,47 @@
...
+                 ((not (find arg args :test #'formal=))
+                  (quil-parse-error "argument ~a is not in gatedef arguments" arg))
+                 (T T))))
```
stylewarning, line 19: lowercase `t`'s

src/analysis/validate-seq-gate.lisp
```diff
@@ -0,0 +1,47 @@
...
+  (let ((seq (sequence-gate-definition-sequence gate-def))
+        (args (sequence-gate-definition-arguments gate-def)))
+    (labels ((validate-argument (arg)
+               (cond
+                 ((qubit-p arg)
+                  (quil-parse-error "argument cannot be qubit"))
+                 ((not (is-formal arg))
+                  (quil-parse-error "argument is not formal"))
+                 ((not (find arg args :test #'formal=))
+                  (quil-parse-error "argument ~a is not in gatedef arguments" arg))
```
stylewarning, line 18: spell out, use full english words

src/analysis/validate-seq-gate.lisp
```diff
@@ -0,0 +1,47 @@
...
+  "This transform traverses all defgate as sequence objects to verify there are no circular references.")
+
+(defun validate-defgate-loops (parsed-program)
+  (verify-no-loops-in-gate-defs (parsed-program-gate-definitions parsed-program))
+  parsed-program)
+
+(defun verify-no-loops-in-gate-defs (gate-defs &optional (ancestors NIL))
+  "Takes list of gate defs, verifys there are no circular dependencies between sequence gate defs."
+  (dolist (gate-def gate-defs)
+    (when (typep gate-def 'SEQUENCE-GATE-DEFINITION)
```
stylewarning, line 39: lowercase class name

src/analysis/validate-seq-gate.lisp
```diff
@@ -0,0 +1,47 @@
...
+      (let ((gate-name (gate-definition-name gate-def)))
+        (if (member gate-name ancestors)
+            (quil-parse-error (format nil "Defgate sequence dependencies contains a loop: ~a " (reverse (cons gate-name ancestors))))
```
stylewarning, line 42: `DEFGATE AS SEQUENCE`

src/gates.lisp
```diff
@@ -190,6 +190,87 @@
 (defmethod gate-matrix ((gate parameterized-gate) &rest parameters)
   (apply (parameterized-gate-matrix-function gate) parameters))
 
+
+;;; SEQUENCE GATE
```
stylewarning, line 194: I've changed my mind about this. I do think you should make something called `sequence-gate` which just has a single field containing the sequence gate definition, and nothing else.

In retrospect I think punning a definition as a gate will be too confusing to maintain and debug in the future.

src/gates.lisp
```diff
@@ -190,6 +190,87 @@
...
+;;; essentially contains all the data we need. As such, we will use seq
+;;; gate definitions to represent the actual gate objects.
+;;;
+;;; Perhaps we should multiply inheret from seq gate definition and
+;;; parameterized gate in order to construct a sequence gate.
+;;;
+
+(defun instantiate-sequence-gate (seq-gate parameters arguments)
+  "Expands a sequence gate definition, parameters, and arguments into a list of gate applications."
+  (let* (
```
stylewarning, line 219: no newline

src/parser.lisp
```diff
@@ -935,7 +935,17 @@ If ENSURE-VALID is T (default), then a memory reference such as 'foo[0]' will re
...
+                                                            :legal-parameters params)))
+              (:SEQUENCE
+               (let ((*shadowing-formals* args))
+                 (parse-gate-definition-body-into-sequence body-lines name args params :context op))))))))))
+
+(defun parse-gate-definition-body-into-sequence (body-lines name args params &key context)
+  (multiple-value-bind (parsed-body rest-lines) 
+    (parse-indented-body body-lines)
+    (values (make-instance 'sequence-gate-definition
+              :name name :parameters params :arguments args :sequence parsed-body :context context)
```
stylewarning, line 947: put keyword args on newlines

stylewarning (changes requested):
I think the biggest change is to do a backpedal and introduce a new gate class.

src/gates.lisp
Spin1Half: I've pushed some new changes up that try to flush out the sequence-gate class slightly more than just putting a sequence gate definition inside.  Let me know if that looks okay, or if doing it this way has problems, I'm happy to revert back to the original suggestion instead.

Spin1Half: I believe this has been implemented now, but you will want to check my most recent commit and make sure that I correctly interpreted what you were asking for.

src/analysis/validate-sequence-gate.lisp
```diff
@@ -0,0 +1,49 @@
...
+      (let ((gate-name (gate-definition-name gate-def)))
+        (if (member gate-name ancestors)
+            (quil-parse-error (format nil "DEFGATE AS SEQUENCE dependencies contains a loop: ~a " (reverse (cons gate-name ancestors))))
+            (verify-no-loops-in-gate-defs (neighbors-of-sequence-gate-def gate-def) (cons gate-name ancestors)))))))
+
+(defun neighbors-of-sequence-gate-def (gate-def)
+  "Returns all sequence gate defs that another sequence gate def depends on."
+  (mapcar (lambda (x) (when (typep x 'sequence-gate) (sequence-gate-gate-definition x) ))
+          (mapcar #'gate-application-gate
+                  (sequence-gate-definition-sequence gate-def))))
```
stylewarning, line 49:
```
(defun neighbors-of-sequence-gate-def (gate-def)
  "Returns all sequence gate defs that another sequence gate def depends on."
  (mapcan (lambda (x)
            (let ((g (gate-application-gate x)))
              (if (not (typep g 'sequence-gate))
                  nil
                  (list (sequence-gate-gate-definition g)))))
          (sequence-gate-definition-sequence gate-def)))
```

src/cl-quil.lisp
```diff
@@ -6,7 +6,7 @@
 (in-package #:cl-quil.frontend)
 
 (defvar *standard-post-process-transforms*
-  '(expand-circuits type-check simplify-arithmetic)
+  '(expand-circuits type-check simplify-arithmetic validate-defgate-loops)
```
stylewarning, line 9: put validation at the start

src/analysis/validate-sequence-gate.lisp
```diff
@@ -0,0 +1,49 @@
...
+      (unless (seq-contains-only-gate-definitions-p seq)
+        (quil-parse-error (format nil "Sequence gate definition body (~a) contains non gate-application" (gate-definition-name gate-def) )))
+      (loop :for isn :in seq
+            :for isn-args := (application-arguments isn)
+            :do (map nil #'validate-argument isn-args)))))
+
+(defun seq-contains-only-gate-definitions-p (sequence)
+  (every #'gate-application-p sequence))
+
+(define-transform validate-defgate-loops (validate-defgate-loops)
```
stylewarning, line 29: depend on process-includes

src/ast.lisp
```diff
@@ -1680,7 +1695,15 @@ For example,
...
                                 :prefix "    "))
-
+  
+  (:method ((gate sequence-gate-definition) (stream stream))
+    (format stream "DEFGATE ~A~@[(~{%~A~^, ~})~]~{ ~A~} AS SEQUENCE:~%"
+            (gate-definition-name gate)
+            (mapcar #'param-name (sequence-gate-definition-parameters gate))
+            (mapcar #'formal-name (sequence-gate-definition-arguments gate)))
+    (dolist (operation (sequence-gate-definition-sequence gate))
+      (format stream "    ~a~%" (print-instruction-to-string operation))))
```
stylewarning, line 1705:
```
(format stream "    ~/quil:instruction-fmt/~%" operation)
```

src/gates.lisp
```diff
@@ -190,6 +190,76 @@
...
+
+;;; SEQUENCE GATE
+;;;  
+(defclass sequence-gate (gate)
+    ((definition :initarg :definition
+                 :reader sequence-gate-gate-definition)))
+
+;;; Perhaps we should multiply inheret from seq gate definition and
+;;; parameterized gate in order to construct a sequence gate.
+;;;
```
stylewarning, line 204: delete extra ;;;

stylewarning: delete whole comment

src/gates.lisp
```diff
@@ -190,6 +190,76 @@
...
   (apply (parameterized-gate-matrix-function gate) parameters))
 
+
+
+
+;;; SEQUENCE GATE
+;;;  
+(defclass sequence-gate (gate)
+    ((definition :initarg :definition
+                 :reader sequence-gate-gate-definition)))
```
stylewarning, line 200: write a doc string, note that there's no practical difference between a sequence gate definition and a sequence gate itself

src/gates.lisp
```diff
@@ -190,6 +190,76 @@
...
+(defun make-map-list-to-list (lista listb apply-to-key)
+  "evaluates apply-to-key on elements of lista and maps them to listb by index, result of apply-to-key must be string"
+  (let ((return-table (make-hash-table :test 'equal)))
+    (loop :for key :in lista
+          :for val :in listb
+          :do (setf (gethash (funcall apply-to-key key) return-table) val))
+    return-table))
+
+(defmethod gate-matrix ((gate sequence-gate) &rest parameters)
+  (let ((n (length (sequence-gate-definition-arguments (sequence-gate-gate-definition gate)))))
```
stylewarning, line 243: gate-definition-qubits-needed

stylewarning (approved):

PR #789, opened 2022-01-28
Implement basic Quilc C API

### LINKED ISSUE #788: Expose basic Quilc functionality via a C API
# Problem
I want to be able to use Quilc as a shared library, e.g. `libquilc.[so|dll|dylib]`. This would serve as lighter alternative to the existing RPC server mode for users who want to use Quilc from other langauges like C or Python, especially when working on a single local machine.

# Scope
As a first pass, I am proposing that this shared library should expose Quilc functionality for:
- parsing and printing Quil programs
- creating basic chip specifications
- compiling Quil/ProtoQuil programs

This is a small but useful set of functionality to begin working off of.

# Solution
We can use [sbcl-librarian](https://github.com/quil-lang/sbcl-librarian) to expose Quilc functionality as a shared library.

kartik-s:
This PR resolves #788, providing a way to build a small subset of Quilc functionality into a shared library.

## Notes
- Tested on macOS, but not Linux.
- The Makefile for the C tests includes a condition to add the `-pagezero_size 0x100000` compilation option when compiling on macOS to address the issue at the bottom of the [sbcl-librarian README](https://github.com/quil-lang/sbcl-librarian#readme).
- The Python tests will not run on macOS unless either 1) you are using a custom build of SBCL that negates the aforementioned page zero size issue or 2) you are using a custom build of Python that was compiled with the `-pagezero_size 0x100000` option.
- The `test-build-lib` test that builds `libquilc.dylib` will fail unless you have `libsbcl.so` in your library path. Instructions:

1. Go to your SBCL source directory.
2. Run `sh make-shared-library.sh`.
3. Copy `src/runtime/libsbcl.so` into your library path, e.g. `/usr/local/lib/`.

### Parsing/printing
- [x] `quilc_parse_quil`
- [x] `quilc_print_program`

### Chip specifications
- [x] `quilc_build_nq_linear_chip`
- [x] `quilc_chip_spec_from_isa_descriptor`
- [x] `quilc_print_chip_spec`

### Compilation
- [x] `quilc_compile_quil`
- [x] `quilc_compile_protoquil`

## Testing
- [x] Write tests


kartik-s:
Need quil-lang/sbcl-librarian#13 to be resolved to fix up C function names.

lib/tests/c-tests.lisp
```diff
@@ -0,0 +1,54 @@
...
+           (processed-program (cl-quil:compiler-hook parsed-program chip-spec))
+           (expected-output (quilc::print-program processed-program nil)))
+      (multiple-value-bind (output error-output exit-code)
+          (uiop:run-program "tests/c/compile-quil" :input `(,input-source) :output :string)
+        (is (string= output expected-output))))))
+
+(deftest test-compile-protoquil ()
+  "Test compiling ProtoQuil from C."
+  (uiop:with-current-directory ("lib/")
+    (let* ((input-source "H 0")
```
kartik-s, line 36: Is there a better program that I can use as input that will actually exercise the ProtoQuil compilation flag?

lib/tests/python-tests.lisp
```diff
@@ -0,0 +1,40 @@
...
+           (processed-program (cl-quil:compiler-hook parsed-program chip-spec))
+           (expected-output (quilc::print-program processed-program nil)))
+      (multiple-value-bind (output error-output exit-code)
+          (uiop:run-program '("python3" "tests/python/compile_quil.py") :env '((:PYTHONPATH . ".")) :input `(,input-source) :output :string)
+        (is (string= output expected-output))))))
+
+(deftest test-python-compile-protoquil ()
+  "Test compiling ProtoQuil from Python."
+  (uiop:with-current-directory ("lib/")
+    (let* ((input-source "H 0")
```
kartik-s, line 22: Ditto

lib/src/libquilc.lisp
```diff
@@ -0,0 +1,37 @@
+(require '#:asdf)
+
+(asdf:load-system '#:quilc)
+(asdf:load-system '#:sbcl-librarian)
+
+(in-package #:quilc)
```
kilimanjaro, line 6: I noticed that you have a `libquilc` package defined elsewhere. On reflection, it would make sense to use that here (of course the package definition would need to be loaded) because it would prevent any accidental clobbering of symbols in the `quilc` package below. For example, while I'm pretty sure that `sbcl-librarian:define-handle-type` just installs methods on some generics, if it happened to do more than that we could end up accidentally redefining `quilc:quil-program` or `quilc:chip-specification` which might be meaningful elsewhere. It's more of an issue here than with smaller examples, since quilc has a lot of random symbols lurking around.

lib/tests/c-tests.lisp
```diff
@@ -0,0 +1,54 @@
+;;;; c-tests.lisp
```
kilimanjaro, line 1: these tests are a good idea!

lib/src/libquilc.lisp
```diff
@@ -0,0 +1,37 @@
...
+(sbcl-librarian:define-enum-type error-type "error_t"
+  ("ERROR_SUCCESS" 0)
+  ("ERROR_FAIL" 1))
+(sbcl-librarian:define-error-map error-map error-type 0
+  (t (condition) (declare (ignore condition)) 1))
+
+(defun compile-protoquil (parsed-program chip-specification)
+  (compiler-hook parsed-program chip-specification :protoquil t))
+
+(sbcl-librarian:define-library libquilc (:error-map error-map
```
kilimanjaro, line 20: One thing for us to think about is how we want `sbcl-librarian` to support bundling of libraries (which is necessarily a bit finicky if we are using this `save-lisp-and-die` functionality). What you have here is a very reasonable definition, but because the API spec here is part of the build script, we don't have a good way to get our hands on it for other purposes. The current closest mechanism in `sbcl-librarian` is the `define-api` macro, which results in a definition that can later get used. I'm not really convinced that it's right in the details, but it's food for thought if one is considering packaging both quilc & other things in a single shared library.

kilimanjaro (commented):
This looks like a great start to me!

lib/src/libquilc.lisp
kartik-s: Makes sense, will fix.

lib/tests/python-tests.lisp
```diff
@@ -0,0 +1,40 @@
...
+           (parsed-program (cl-quil:safely-parse-quil input-source))
+           (chip-spec (cl-quil::build-nq-linear-chip 8))
+           (processed-program (cl-quil:compiler-hook parsed-program chip-spec :protoquil t))
+           (expected-output (quilc::print-program processed-program nil)))
+      (multiple-value-bind (output error-output exit-code)
+          (uiop:run-program '("python3" "tests/python/compile_protoquil.py") :env '((:PYTHONPATH . ".")) :input `(,input-source) :output :string)
```
mhdavid-hrl, line 28: ditto my comment re line length

lib/tests/python-tests.lisp
```diff
@@ -0,0 +1,40 @@
...
+        (is (string= output expected-output))))))
+
+(deftest test-python-print-chip-spec ()
+  "Test printing chip specifications from Python."
+  (uiop:with-current-directory ("lib/")
+    (let ((chip-spec1 (cl-quil::build-nq-linear-chip 8))
+          (chip-spec2 (quilc::lookup-isa-descriptor-for-name "8Q")))
+      (multiple-value-bind (output error-output exit-code)
+          (uiop:run-program '("python3" "tests/python/print_chip_spec.py") :env '((:PYTHONPATH . ".")) :output :string)
```
mhdavid-hrl, line 37: ditto my comment re line length

lib/tests/suite.lisp
```diff
@@ -0,0 +1,29 @@
...
+
+(defun clean-up ()
+  (uiop:with-current-directory ("lib/")
+    (uiop:run-program '("make" "clean")))
+  (uiop:with-current-directory ("lib/tests/c/")
+    (uiop:run-program '("make" "clean"))))
+
+(defun run-libquilc-tests (&key (verbose nil) (headless nil))
+  "Run all libquilc tests. If VERBOSE is T, print out lots of test info. If HEADLESS is T, disable interactive debugging and quit on completion."
+  (setf fiasco::*test-run-standard-output* (make-broadcast-stream *standard-output*))
```
mhdavid-hrl, line 15: I'm curious about the use of `make-broadcast-stream` with a single arg here. Is this left over from debugging? Is it for expected future debugging? Perhaps a comment is worthwhile explaining it. Unless this is a well-known convention I've not known about until now.

mhdavid-hrl (commented):
Overall, really nice code, test suite, documentation, and terrific functionality. This will definitely be a great help for people wanting to glue quilc and c, etc.

- will be good to get this going on Linux, and also Windows, but OK not to hold up for those (but good to note their status)
- helpful to note the requirement to use the special version of SBCL (I think it requires branch `no-linker-flag`, right?)
- helpful to note the requirement to make the library and put it somewhere gcc can find it on macOS. Note that the issue https://github.com/quil-lang/sbcl-librarian/issues/8 is a help, but actually for some reason `LD_LIBRARY_PATH` environment variable isn't working on my macOS: my workaround was to explicitly do `cp -p /path/to/sbcl-repo/src/runtime/libsbcl.so /usr/local/lib/`
- see also a few comments sprinked over code, very minor style points

lib/tests/suite.lisp
kartik-s: I copied this from [the app test suite](https://github.com/quil-lang/quilc/blob/master/app/tests/suite.lisp#L9), so I'm not entirely sure what the justification is here; I'll add a TODO comment saying as much.

lib/tests/python-tests.lisp
```diff
@@ -0,0 +1,40 @@
...
+(deftest test-python-compile-quil ()
+  "Test compiling Quil from Python."
+  (uiop:with-current-directory ("lib/")
+    (let* ((input-source "H 0")
+           (parsed-program (cl-quil:safely-parse-quil input-source))
+           (chip-spec (cl-quil::build-nq-linear-chip 8))
+           (processed-program (cl-quil:compiler-hook parsed-program chip-spec))
+           (expected-output (quilc::print-program processed-program nil)))
+      (multiple-value-bind (output error-output exit-code)
+          (uiop:run-program '("python3" "tests/python/compile_quil.py") :env '((:PYTHONPATH . ".")) :input `(,input-source) :output :string)
```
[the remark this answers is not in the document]
kartik-s: Agreed, and fixed.

lib/tests/c-tests.lisp
```diff
@@ -0,0 +1,54 @@
...
+;;;; Author: Kartik Singh
+
+(in-package #:libquilc-tests)
+
+(deftest test-build-lib ()
+  "Test that libquilc builds."
+  (uiop:with-current-directory ("lib/")
+    (multiple-value-bind (output error-output exit-code)
+        (uiop:run-program "make")
+      (is (eql exit-code 0)))))
```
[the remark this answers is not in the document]
kartik-s: Fixed.

lib/tests/suite.lisp
mhdavid-hrl: Aha, maybe @stylewarning can share the lore

mhdavid-hrl (dismissed):
Looks all good to me.

lib/src/libquilc.lisp
kartik-s: Agreed that dumping the artifacts in the same place that we define the API is not conducive to reusing the API spec in other places. @karlosz and I have been discussing this, and the next release of `sbcl-librarian` will support decoupling these two concerns so that we can defer the artifact creation.

The details of that are still being worked out, so do you think it's OK if we keep this build script for now and then fix it up when `sbcl-librarian` gets updated?

lib/tests/c-tests.lisp
kilimanjaro: ProtoQuil is just a restricted subset of Quil with only definitions/declarations, pragmas, reset instructions, applications, and measurements with the caveat that resets come before applications and measurements come after (cf `check-protoquil-program`). So for a more interesting Protoquil program, something like `DECLARE ro BIT; RESET 0; H 0; MEASURE 0 ro`. Note that if this flag is set, then the following program would be erroneous: `DECLARE ro BIT; MEASURE 0 ro; H 0`, so you might as well test that you get a sensible error code from that.

kilimanjaro (approved):
lgtm

libquilc-tests.asd
```diff
@@ -0,0 +1,22 @@
+;;;; libquilc-tests.asd
+;;;;
+;;;; Author: Kartik Singh
+
+(asdf:defsystem #:libquilc-tests
+  :description "Regression tests for libquilc."
+  :author "Kartik Singh <kssingh@hrl.com"
```
Spin1Half, line 7: forgot a '>' at the end maybe?

Spin1Half (approved):

mhdavid-hrl (approved):
Looks good to me

stylewarning (approved):
Fantastic changes, and held to a high standard of quality!

PR #790, opened 2022-02-08
flush some unused code recently added in benchmarking code

mhdavid-hrl:
I just noticed that I'd left in some code which is useless by
mistake. It causes style warnings, which this commit eliminates.

The useless code came along with new modules introduced in my recent
commits for cl-quil-benchmarking system.  Tested by running on command
line: make benchmark-nq


stylewarning (approved):

stylewarning:
red PRs are the best

PR #795, opened 2022-02-16
Update to new SBCL librarian interface.

karlosz:
* Notably, we pull out handles into its own API so that other sbcl
librarian managed libraries can share the implementation. (And stuff
doesn't step on each other.)


stylewarning (approved):
lgtm

PR #796, opened 2022-02-16
Teach the "old gate definition model" about SWAP.

karlosz:
* Teach the chip reader how to deal with it.
* Adjust a test.

This just naively translates swap instructions directly when the
object supports it.

### DIFF
--- a/src/chip/chip-reader.lisp
+++ b/src/chip/chip-reader.lisp
@@ -244,6 +244,7 @@
                             ((null string) nil)
                             ((string= "CZ" string) ':cz)
                             ((string= "ISWAP" string) ':iswap)
+                            ((string= "SWAP" string) ':swap)
                             ((string= "CPHASE" string) ':cphase)
                             ((string= "PISWAP" string) ':piswap)
                             ((string= "XY" string) ':xy)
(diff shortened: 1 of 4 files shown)


karlosz:
I accidentally messed up the last pull request implementing this but I've adjusted tests and taught the reader how to deal with SWAP now.

tests/compiler-hook-tests.lisp
```diff
@@ -169,7 +169,7 @@ some test programs."
       (format t "~&    With *ENABLE-STATE-PREP-COMPRESSION* ~A~%" quil::*enable-state-prep-compression*)
       (dolist (file (uiop:directory-files *compiler-hook-test-file-directory* #P"*.quil"))
         (format t "      Testing file ~A:" (pathname-name file))
-        (dolist (architecture (list ':cz ':iswap ':cphase ':piswap ':cnot))
+        (dolist (architecture (list ':cz ':iswap ':swap ':cphase ':piswap ':cnot))
```
stylewarning, line 172: This can't be right. If SWAP is the only native 2Q gate of a gate set then it won't be universal. So the tests ought to fail.

karlosz: Right.

karlosz:
Updated to have more sensible tests.

stylewarning (approved):

PR #806, opened 2022-03-24
fix permutation gate synthesis

### LINKED ISSUE #805: DefPermutationGate compilation may be incorrect for some permutations
The code shown below indicates there is a problem compiling selected permutation gates (`DefPermutationGate`). The code uses four methods:

1. Wavefunction simulator
2. QVM, with the circuit compiled using `quilc`
3. QVM, without compilation
4. Manual calculation of the gate matrices using `numpy`,

to evaluate a circuit consisting of one `X` gate and one permutation gate.  The results indicate:

- If the permutation corresponds to a single `SWAP`, all four methods yield the same result.
- If the permutation corresponds to multiple `SWAP`s, or an arbitrary permutation, the result from method 2 (QVM with compilation) appears to differ from the other three methods.

```
from collections import Counter
import numpy as np
from pyquil.gates import X, MEASURE
from pyquil import Program, get_qc
from pyquil.quilbase import DefPermutationGate
from pyquil.api import WavefunctionSimulator

permutation = [0, 4, 2, 6, 1, 5, 3, 7]   # Corresponds to SWAP(0,2) - results from all methods agree
permutation = [0, 2, 1, 3, 4, 6, 5, 7]   # Corresponds to SWAP(1,2) - results from all methods agree
permutation = [0, 4, 1, 5, 2, 6, 3, 7]   # Corresponds to SWAP(1,2) + SWAP(0,2) - results from method 2 disagree
permutation = [5, 1, 2, 6, 7, 0, 4, 3]   # Arbitrary permutation - results from method 2 disagree

permutation_gate_def = DefPermutationGate("permutation_gate", permutation)
permutation_gate = permutation_gate_def.get_constructor()
program = Program()
program.declare("ro", "BIT", 3)
program += permutation_gate_def
program += X(0)
program += permutation_gate(0, 1, 2)
program += MEASURE(0, ("ro", 0))
program += MEASURE(1, ("ro", 1))
program += MEASURE(2, ("ro", 2))

# QVM wave function simulator
wf_sim = WavefunctionSimulator()
wf_results = wf_sim.wavefunction(program)
print(f"Wavefunction simulator result: {wf_results}")
print("\tNote: the wavefunction simulator reverses the bit order\n")

# QVM - with compilation
qc = get_qc('3q-qvm')
program.wrap_in_numshots_loop(100)
qvm_results = qc.run(qc.compile(program, to_native_gates=True, optimize=True)).readout_data.get("ro")
histogram = Counter([tuple(shot_result) for shot_result in qvm_results])
print(f"QVM result with compilation: {histogram}\n")

# QVM - without compilation
program.wrap_in_numshots_loop(100)
qvm_results = qc.run(program).readout_data.get("ro")
# Note: we could also disable compilation using:
#       qvm_results = qc.run(qc.compile(program, to_native_gates=False, optimize=False)).readout_data.get("ro")
histogram = Counter([tuple(shot_result) for shot_result in qvm_results])
print(f"QVM result without compilation: {histogram}\n")

# Compute the result manually
X_np = np.array([[0, 1], [1, 0]])
permutation_gate_np = np.array([[0] * col + [1] + [0] * (len(permutation) - 1 - col) for col in permutation])
I2 = np.array([[1, 0], [0, 1]])
XII = np.kron(X_np, np.kron(I2, I2))
psi_000 = np.array([1] + [0] * 7)
manual_result = permutation_gate_np @ XII @ psi_000
print(f"Manual result: {manual_result}")
```

Output:
```
Wavefunction simulator result: (1+0j)|011>
	Note: the wavefunction simulator reverses the bit order

QVM result with compilation: Counter({(1, 1, 1): 100})

QVM result without compilation: Counter({(1, 1, 0): 100})

Manual result: [0 0 0 0 0 0 1 0]
```
Note: the manual result `"0 0 0 0 0 0 1 0"` corresponds to state `|110⟩`, which I think is the correct result for the `[5, 1, 2, 6, 7, 0, 4, 3]` permutation.

The code was tested using `quilc 1.26.0`, `qvm 1.17.2` (as pulled from DockerHub) and `pyquil 3.1.0`.

The attachments contain the compiled instructions (_i.e._ the output from `qc.compile(program)`) for two cases:
- the arbitrary permutation, as shown above
- the permutation consisting of two `SWAP`s (`SWAP(1,2)` followed by `SWAP(0,2)`)
[compiled_program_arbitrary_permutation.txt](https://github.com/quil-lang/quilc/files/8339898/compiled_program_arbitrary_permutation.txt)
[compiled_program_two_SWAPs.txt](https://github.com/quil-lang/quilc/files/8339899/compiled_program_two_SWAPs.txt)

stylewarning:
Perm gates were being synthesized as inverses, leading to improper
compilations.

Added a battery of new tests for checking multi-qubit perm gate
synthesis.

Fixes #805.

### DIFF
--- a/src/compilers/permutation.lisp
+++ b/src/compilers/permutation.lisp
@@ -1,8 +1,12 @@
 ;;;; permutation.lisp
+;;;;
+;;;; Author: Charles Zhang
+
+(in-package #:cl-quil)
+
 ;;;; This file implements routines to decompose and synthesize
 ;;;; permutation matrices representing classical reversible circuits.
 
-(in-package #:cl-quil)
 
 ;;;; young group decomposition based algorithm for permutation gates
 
@@ -63,6 +67,7 @@
       (dotimes (i size)
         (assert (= (aref right (aref perm (aref left i)))
                    (aref old-perm i))))
+      ;; Return g1 and g2 in "mathematical" or "composition" order.
       (values left right))))
 
 (defstruct single-target-gate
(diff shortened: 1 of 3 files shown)


ecpeterson (commented):
Each of X, CNOT, and CCNOT are self-inverting, so the group inverse of a composite circuit is given by the same circuit with the reverse gate ordering. Maybe rather than invert the input permutation it would be preferable to reverse the order in which the gates are emitted? This may require more surgery, but to me it feels less “too wrongs make a right”.

stylewarning:
> Each of X, CNOT, and CCNOT are self-inverting, so the group inverse of a composite circuit is given by the same circuit with the reverse gate ordering. Maybe rather than invert the input permutation it would be preferable to reverse the order in which the gates are emitted? This may require more surgery, but to me it feels less “too wrongs make a right”.

I agree with you that inverting the inverse isn't very nice.

The only reason I decided against this was if in the future somebody modifies this code and the "just reverse the list" no longer implies an inverse. If we had a good "inverse sequence" function—which we do, but adds all sorts of `DAGGER`s–then maybe it'd be more palatable.

I think my preferred approach would be the one that I did, and post an issue to gently refactor it to just produce the right decomposition straight away.

ecpeterson:
I’m not sure I buy it — it seems like OP might have mixed up “left” and “right” conventions in the algorithm description vs circuit ordering vs Quil instruction ordering, so that we’re already looking at an accidentally inverted implementation and we’d be _eliminating_ DAGGER coincidences by untwisting some of the lefts and rights.

I’m also fine deferring further changes to an issue, but this fix as-is isn’t my long term preference.

karlosz:
Hello. I think I'm leaning towards agreeing with @ecpeterson here. What he said about mixing up the convention for left and right sounds exactly like what I did: When I implemented the algorithm I remember getting confused exactly how the different notations associated left and right with post and preapplication, and its very likely there is one too many reversals or not enough. It is standard to write permutation composition as applying in left to right order in some parts of grou theory and as the normal right to left order in others. Since this is looks like the original code bungled that it would be nice to fix it the right way by nailing down what the convention is supposed to be.

stylewarning:
Ok, I'll take a look and see if I can repair it / build it in the right direction.

stylewarning:
Even though I've simplified the computation, I still feel something isn't right about my approach. It seems the functions that are generating gates on qubits `0` to `n-1` should instead be doing so from `n-1` to `0`. I just don't know where exactly that should happen.

I really don't want the `(reverse qubits)` call there; it feels like the same issue as described in @ecpeterson and @karlosz's previous comments here.

ecpeterson:
That may not be avoidable because Quil argument ordering has descending significance. I think you’ll find examples of expressions like `(nth args (- (length args) index 1))` in the circuit builders which are necessary for this reason. Dunno tho.

Thanks for making the more complicated fix! 👍

P.S.: I wish diff tools were more intelligent about Lisp; removing an outer let should highlight its whole contents as being modified. Is there a way to suggest to git to use a different diff implementation on certain file types? Hm.

stylewarning:
> That may not be avoidable because Quil argument ordering has descending significance. I think you’ll find examples of expressions like `(nth args (- (length args) index 1))` in the circuit builders which are necessary for this reason. Dunno tho.

This part isn't irksome to me, it's that even if I pass qubits in "3 2 1" order, I have to reverse them to "1 2 3" so the circuit building code works. I suppose I'm suggesting that deep in the circuit-building code, it "natively" understands "3 2 1". 


> P.S.: I wish diff tools were more intelligent about Lisp; removing an outer let should highlight its whole contents as being modified. Is there a way to suggest to git to use a different diff implementation on certain file types? Hm.

There are a few [diff algorithms](https://luppeng.wordpress.com/2020/10/10/when-to-use-each-of-the-git-diff-algorithms/) built in to git but they're not visible in the GitHub interface. Agreed it's annoying.

stylewarning:
I'm no longer irked. I added documentation that I think justifies everything. I took out a `reverse` since it was really a red herring to it all. Things look good to me now.

kilimanjaro (approved):
LGTM. The comments & docstrings clear up any confusion I had.

karlosz (approved):
Looks great now!

PR #807, opened 2022-03-24
print warning when there are ambiguous gatedefs or circuitdefs by def…

Spin1Half:
change default behavior to print a warning when there are ambiguous gate definitions or circuit definitions

### DIFF
--- a/src/cl-quil.lisp
+++ b/src/cl-quil.lisp
@@ -9,6 +9,9 @@
   '(validate-defgate-loops expand-circuits type-check)
   "The standard transforms that are applied by PARSE-QUIL after parsing. (See also: *STANDARD-PRE-COMPILATION-TRANSFORMS*)")
 
+(defvar *default-ambiguous-definition-handler*
+  'warn-on-ambiguous-gate-or-circuit-definition)
+
 (define-condition ambiguous-definition-condition ()
   ((instruction :initarg :instruction
                 :reader ambiguous-definition-instruction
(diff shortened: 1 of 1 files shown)


stylewarning:
I would change to emitting a `style-warning` instead of a full warning. You could make a new condition class for this, or just use `alexandria:simple-style-warning`.

That should fix the tests.

src/cl-quil.lisp
```diff
@@ -44,6 +47,15 @@
 ;;; generally determined by a pattern matching system, and so we have to
 ;;; consider the arity and values of parameters and arguments).
 
+(defun warn-on-ambiguous-gate-or-circuit-definition (condition)
+  "Handler which prints a warning in the presence of a CONDITION indicating an ambiguous gate or circuit declaration."
+  (check-type condition ambiguous-definition-condition)
+  (let ((instr (ambiguous-definition-instruction condition)))
+    (typecase instr
+      (gate-definition (alexandria:simple-style-warning (format nil "Gate ~A has multiple definitions, this leads to ambiguous behavior.~%" (gate-definition-name instr))))
```
karlosz, line 55: You don't need to write `(format nil ...)` here. `warn` already can deal with a format string and format arguments by itself.

src/cl-quil.lisp
```diff
@@ -44,6 +47,15 @@
...
+      (circuit-definition (alexandria:simple-style-warning (format nil "Circuit ~A has multiple definitions, this leads to ambiguous behavior.~%" (circuit-definition-name instr)))))
```
karlosz, line 56: Ditto.

karlosz (changes requested):

src/cl-quil.lisp
Spin1Half: good point!

karlosz (dismissed):

src/cl-quil.lisp
```diff
@@ -44,6 +47,15 @@
 ;;; generally determined by a pattern matching system, and so we have to
 ;;; consider the arity and values of parameters and arguments).
 
+(defun warn-on-ambiguous-gate-or-circuit-definition (condition)
+  "Handler which prints a warning in the presence of a CONDITION indicating an ambiguous gate or circuit declaration."
+  (check-type condition ambiguous-definition-condition)
+  (let ((instr (ambiguous-definition-instruction condition)))
+    (typecase instr
```
stylewarning, line 54: either make this an `etypecase` or print a diagnostic in a `t` branch

stylewarning (changes requested):

stylewarning (approved):

PR #816, opened 2022-04-25
Scaffolding for CL-QUIL/DISCRETE

stylewarning:
This adds implementation and test **SCAFFOLDING** for "discrete compilation"—compilation of Quil programs to discrete gate sets—to QUILC. This does not actually contain an implementation, except for some minor defining functions.

The implementation will come later as a collaboration with other contributors. This PR is intended to help move that along by splitting the work up a bit.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -115,8 +115,8 @@ uninstall:
 # TEST
 ###############################################################################
 
-.PHONY: test test-cl-quil test-quilc test-quilt test-quilec test-libquilc
-test: test-cl-quil test-quilt test-quilec test-quilc test-libquilc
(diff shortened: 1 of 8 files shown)


notmgsk (approved):
Might be clearer to use naming specific to that set of gates rather than the general "discrete", but only a minor point

stylewarning:
> Might be clearer to use naming specific to that set of gates rather than the general "discrete", but only a minor point

I thought about that and I actually wanted to keep that abstracted away... Not sure if that's a good idea or not, but I'd like for the API to require the user to interrogate what that gate set is, since it may (?) change.

PR #820, opened 2022-05-20
Add tolerance check

aijony:
Description:
When using a tolerance pragma (e.g. `PRAGMA TOLERANCE "0.5"`), print global phase invariant distance ([arXiv:2106.07099](https://arxiv.org/abs/2106.07099)) instead of erroring when matrices are inevitably in the wrong projective class (e.g. `#Matrices have a phase invariant distance of 0.29746263355117152d0`).

Changes:
  - If tolerance pragma, prevent matrix scaling error
  - Make function for checking top pragmas

### DIFF
--- a/app/src/entry-point.lisp
+++ b/app/src/entry-point.lisp
@@ -404,10 +404,16 @@
                ;; can print the input and output unitary matrix as
                ;; comments at the end of the output.
                (when (and protoquil compute-matrix-reps)
-                 (let* ((processed-program-matrix (parsed-program-to-logical-matrix processed-program :compress-qubits t)))
-                   (print-matrix-comparision original-matrix
-                                             (quil::scale-out-matrix-phases processed-program-matrix
-                                                                            original-matrix)))))
+                 (let* ((processed-program-matrix (parsed-program-to-logical-matrix processed-program :compress-qubits t))
+                        ;; If a TOLERANCE pragma is found, allow non-zero phase invariant distance
+                        (tolerance (cl-quil::prog-find-top-pragma program "TOLERANCE")))
+                   (if tolerance
+                       (print-matrix-comparision
+                        original-matrix processed-program-matrix
+                        :tolerance t)
+                       (print-matrix-comparision
+                        original-matrix
+                        (quil::scale-out-matrix-phases processed-program-matrix original-matrix))))))
 
              ;; New and improved flow
              (when compile
--- a/app/src/printers.lisp
+++ b/app/src/printers.lisp
@@ -11,14 +11,23 @@
 (declaim (special *without-pretty-printing*
                   *human-readable-stream*))
 
-(defun print-matrix-comparision (m1 m2 &optional (stream *human-readable-stream*))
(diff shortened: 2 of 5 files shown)


src/utilities.lisp
```diff
@@ -93,3 +93,16 @@ appropriate method of comparison."
   (calculate-instructions-fidelity
    (coerce (parsed-program-executable-code program) 'list)
    chip))
+
+(defun prog-find-top-pragma (parsed-prog pragma)
```
stylewarning, line 97: This function needs to be defined later. (Maybe in ast.lisp or pragma.lisp?) The PARSED-PROGRAM struct and functions won't have been defined yet.

app/src/printers.lisp
```diff
@@ -11,14 +11,17 @@
 (declaim (special *without-pretty-printing*
                   *human-readable-stream*))
 
-(defun print-matrix-comparision (m1 m2 &optional (stream *human-readable-stream*))
+(defun print-matrix-comparision (m1 m2 &optional (stream *human-readable-stream*)
+                                         &key (distance nil))
```
stylewarning, line 15: add a little doc string that documents especially the DISTANCE argument

app/src/printers.lisp
```diff
@@ -11,14 +11,17 @@
 (declaim (special *without-pretty-printing*
                   *human-readable-stream*))
 
-(defun print-matrix-comparision (m1 m2 &optional (stream *human-readable-stream*))
+(defun print-matrix-comparision (m1 m2 &optional (stream *human-readable-stream*)
```
stylewarning, line 14: can we make both arguments &key? Having optional and key arguments are usually bad form.

Just do a quick search for any other uses of this function in the repo and make sure they're fixed up.

app/src/entry-point.lisp
```diff
@@ -404,10 +404,16 @@
...
-                   (print-matrix-comparision original-matrix
-                                             (quil::scale-out-matrix-phases processed-program-matrix
-                                                                            original-matrix)))))
+                 (let* ((processed-program-matrix (parsed-program-to-logical-matrix processed-program :compress-qubits t))
+                        ;; If a TOLERANCE pragma is found, allow non-zero phase invariant distance
+                        (tolerance (cl-quil::prog-find-top-pragma program "TOLERANCE")))
+                   (if tolerance
+                       (print-matrix-comparision
+                        original-matrix processed-program-matrix *human-readable-stream*
+                        :distance (cl-quil::phase-invariant-distance original-matrix processed-program-matrix))
```
stylewarning, line 413: I would make this a boolean argument, and calculate the distance in the printing function (since you're passing the matrices in anyway).

stylewarning (changes requested):
mostly small nits

src/utilities.lisp
aijony: It seems `prog-find-top-pragma` in is in `cl-quil` package which is dependent on `cl-quil.frontend` where PARSED-PROGRAM is defined (ast.lisp and pragma.lisp are also in `cl-quil.frontend`).

stylewarning (approved):

PR #827, opened 2022-06-08
`copy-instance` kwargs to overwrite slots

gefjon:
You can now do e.g. `(copy-instance SOME-BINARY-CLASSICAL-INSTRUCTION :left
NEW-LEFT-OPERAND)` to make a copy of `SOME-BINARY-CLASSICAL-INSTRUCTION` that has its
`left` slot overwritten with `NEW-LEFT-OPERAND`. I suspect there are plenty of places in
the QuilC code where this would enable a much nicer functional style by eliminating
patterns like `(let ((copy (copy-instance FOO))) (setf (FOO-BAR FOO) BAR))`.

### DIFF
--- a/src/ast.lisp
+++ b/src/ast.lisp
@@ -509,10 +509,18 @@ This replicates some of the behavior of CL-QUIL.CLIFFORD::PAULI, but it extends 
   prefactor
   arguments)
 
-(defmethod copy-instance ((term pauli-term))
-  (make-pauli-term :pauli-word (pauli-term-pauli-word term)
-                   :prefactor (pauli-term-prefactor term)
-                   :arguments (pauli-term-arguments term)))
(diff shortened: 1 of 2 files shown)


ecpeterson:
Neat! I tried writing something like this to chase this same pattern, and I was unhappy about having to treat struct and class instances separately, so it never made it in. Some years older and wiser, I think it’s a good enough idea that it should go in anyway.

gefjon:
I think the only way to do this for subclasses of `structure-object` is to define individual methods, though probably via a macro. But neither the CL spec nor the MOP defines any way to write a generic method.

Oh, speaking of which, the default method for `copy-instance` should really specialize on `standard-object`, not on `t`. I'll make that change tomorrow and see if it breaks anything.

PR #831, opened 2022-07-12
rename FOO.BAR packages to FOO/BAR

stylewarning:
This is a heavy renaming of all of our `foo.bar` package names to `foo/bar`.

The reason is that in old CMUCL and in Allegro CL, the `foo.bar` notation is for hierarchical packages, a non-standard Lisp feature that gives you nested namespaces. Moreover, the discrepancy between systems using `/` and packages using `.` had no good explanation.

This also removes all `rename-package` and conditionals on package-local-nicknames. We just assume you have it now.

This cleans up some existing nicknames due to issues/conflicts in the packages.lisp file

I'm a little worried I didn't replace *everything* in various scripts, dockerfiles, etc.

See also (required): https://github.com/quil-lang/qvm/pull/294


src/package.lisp
```diff
@@ -729,17 +722,23 @@
...
+  (:use #:cl)
+  (:local-nicknames (#:a      #:alexandria)
+                    (#:quilfe #:cl-quil/frontend))
+  (:import-from #:cl-quil/frontend #:tok #:token-type #:token-payload)
+
+  (:export
+   #:parse-qasm))
 
 (defpackage #:cl-quil
-  (:nicknames #:quil)
```
appleby, line 736: sobrave

src/package.lisp
```diff
@@ -1353,7 +1336,7 @@
 ;;; access "internal" symbols.  Having such symbols cataloged here is
 ;;; preferable to simply accessing any symbol whatsoever via full
 ;;; qualification, ala the use of package name "quil" with double
```
appleby, line 1338:
```suggestion
;;; qualification, ala the use of package name "cl-quil" with double
```
boom. gottem.

src/chip-library/package.lisp
```diff
@@ -2,7 +2,7 @@
 ;;;;
 ;;;; Author: A.J. Nyquist
 
-(defpackage #:cl-quil.chip-library
+(defpackage #:cl-quil/chip-library
   (:use #:cl)
   (:local-nicknames (:q :cl-quil)
```
appleby, line 7: `:q` -> `:quil` imho

appleby (approved):

karlosz (dismissed):

src/package.lisp
notmgsk: also `à la`

smh

notmgsk (dismissed):
Agree w the great Appleby

src/package.lisp
appleby: M-x ala-mode

stylewarning:
I'm not convinced this PR is a good idea anymore.

1. It does a bunch of different changes all in one PR.
2. It breaks `~/.../` format strings, which can't be fixed without hacks
3. It's not clear hierarchical packages are actually causing problems per se.

PR #837, opened 2022-07-29
Deduplicate code in logical-schedule

markasoftware:
Noticed that a couple functions were suspiciously similar. Adding commuting blocks to the schedule is not all that different than adding any other kind of instruction; just delay adding the new threads to the top and bottom of the lschedule until after all the threads have been added.

One question: Why are `COMMUTING_BLOCK`s that use the `+IMPOSSIBLY-FULL-RESOURCE+` disallowed? I didn't notice anything in `logical-scheduler.lisp` that would fail.

### DIFF
--- a/src/addresser/logical-schedule.lisp
+++ b/src/addresser/logical-schedule.lisp
@@ -261,25 +261,29 @@ as the size keyword arg to make-hash-table."
         (push instr traversed-instructions)))
     (setf (lscheduler-last-instrs lschedule) (nreverse new-bottommost))))
 
+(defun append-instructions-to-lschedule-parallel (lschedule instrs)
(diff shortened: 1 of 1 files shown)


ecpeterson (approved):
This looks OK to me.

I don't have an answer for why that `resource-all-p` was there. I've thought about it now for half an hour, and it's no clearer. I'm comfortable deleting it; [this](https://github.com/quil-lang/quilc/blob/master/src/addresser/logical-schedule.lisp#L160) is the set of instructions it's guarding against, and most — all? — of those can't appear in these blocks because they're absorbed into the CFG edges anyway. 🤷‍♀️

ecpeterson:
Oh I didn't see that you'd retracted the PR. this morning I'll ask why on Thursday.

markasoftware:
I thought I could make further simplifications related to commuting_blocks but they didn't work out. I'll undraftify the PR

PR #845, opened 2022-09-07
Work on removing FIND-DIAGONALIZER-IN-E-BASIS

### LINKED ISSUE #841: `find-diagonalizer-in-e-basis` can be simplified and use generalized Schur decomposition
I'm mostly interpreting some things @kilimanjaro told me (so credit goes to him), though errors I may have below are my own.

`find-diagonalizer-in-e-basis` aims to diagonalize a symmetric unitary matrix `gammag = u u^T` in terms of an orthogonal matrix of eigenvectors. Normally, all that the spectral theorem gives you (in this context) is that you can do this in terms of a unitary matrix of eigenvectors. In this case, however, the real and imaginary parts of `gammag` commute, so it's sufficient to simultaneously diagonalize them. They are real, symmetric matrices, their eigenvectors will be real and will give an orthogonal matrix

To simultaneously diagonalize a pair of commuting matrices, it's not quite enough to just compute eigenvectors and eigenvalues of one and then hope that this works for the other (consider that the identity matrix, which commutes with anything, but we need to write its eigenspaces as spanned by eigenvectors of some other matrix.) The current QUILC approach is to try to diagonalize a linear combination of the real and imaginary parts. Since there could be some relationships between them that we don't know, the current code tries to pick a random combination, and just repeats until we get something that works

This problem has a standard solution that linear algebra libraries implement, called the _generalized Schur decomposition_. LAPACK documents it [here](https://netlib.org/lapack/explore-html/d9/d8e/group__double_g_eeigen_ga8637d4b822e19d10327ddcb4235dc08e.html)

A Python implementation (which uses NumPy's `qz`) supplied by @kilimanjaro is here:

```
def orthogonal_decomposition(U):
    """
    Given unitary U, decompose UU^T into O @ np.diag(d) @ O.T where O is special orthogonal.
    """
    # Generalized Schur decomposition writes
    #    Re[U] = L D_r R^T,  Im[U] = L D_i R^T
    # In general, D_r, D_i are upper triangular, but for unitary U they end up diagonal
    diag_r, diag_i, left, right = scipy.linalg.qz(np.real(U), np.imag(U))
    diag = np.diagonal(diag_r) + np.diagonal(diag_i) * 1j
    if np.linalg.det(left) < 0:
        left[:,0] *= -1
    return left, diag*diag
```

If we want to implement this in QUILC, we'll need to get `qz` into MAGICL.

stylewarning:
This is an attempt to replace `F-D-I-E-B` with @kilimanjaro's idea of using a Schur decomposition.

This aims to fix #841 

CC @genos

### DIFF
--- a/src/compilers/approx.lisp
+++ b/src/compilers/approx.lisp
@@ -155,83 +155,62 @@
   :test #'magicl:=
   :documentation "This is a precomputed Hermitian transpose of +E-BASIS+.")
 
-(defun ensure-positive-determinant (m)
-  (let ((d (magicl:det m)))
-    (when *check-math*
-      (unless (double~ 0.0d0 (imagpart d))
-        (warn "Complex determinant found for a ~
-               matrix expected to be (real) orthogonal: ~
-                   det=~A" d)))
-    (if (double= -1d0 (realpart d))
-        (magicl:@ m (from-diag (list -1 1 1 1)))
-        m)))
+(defun real->complex (m)
+  "Convert a real matrix M to a complex one."
+  (let ((cm (magicl:zeros
+             (magicl:shape m)
+             :type `(complex ,(magicl:element-type m)))))
+    (magicl::map-to #'complex m cm)
+    cm))
 
-(defconstant +diagonalizer-max-attempts+ 16
-  "Maximum number of attempts DIAGONALIZER-IN-E-BASIS should make to diagonalize the input matrix using a random perturbation.")
+(defun special-orthogonal-p (m)
+  "Does M appear to be a special orthogonal matrix?"
+  (and (double~ 1.0d0 (magicl:det m))
+       (magicl:every
+        #'double~
+        (eye 4 :type 'double-float)
+        (magicl:@ (magicl:transpose m) m))))
 
-(define-condition diagonalizer-not-found (error)
-  ((matrix :initarg :matrix :reader diagonalizer-not-found-matrix)
-   (attempts :initarg :attempts :reader diagonalizer-not-found-attempts))
-  (:report (lambda (c s)
-             (format s "Could not find diagonalizer for matrix ~%~A~%after ~D attempt~:P."
-                     (diagonalizer-not-found-matrix c)
-                     (diagonalizer-not-found-attempts c))))
-  (:documentation "The diagonalizer for the given matrix was not found after a number of attempts."))
+(defun orthogonal-decomposition-of-UU^T (u)
+  "Given a unitary U, find a decomposition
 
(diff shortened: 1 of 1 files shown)


genos:
I think the original Python implementation from #841 fell prey to the note from [`scipy.linalg.qz`'s documentation](https://docs.scipy.org/doc/scipy/reference/generated/scipy.linalg.qz.html), namely

> Notes
> Q is transposed versus the equivalent function in Matlab.

Specifically, in the comment

```python
# Generalized Schur decomposition writes
#    Re[U] = L D_r R^T,  Im[U] = L D_i R^T
```

I think it should be `L^T` and `R`:

```python
In [1]: from scipy.linalg import norm, qz

In [2]: from scipy.stats import unitary_group

In [3]: rng = unitary_group(dim=2, seed=1729)

In [4]: print(u := rng.rvs())
[[-0.29585687+0.47303725j -0.76750556+0.31565756j]
 [ 0.71123856-0.42760284j -0.55613711-0.04479994j]]

In [5]: diag_r, diag_i, left, right = qz(u.real, u.imag)

In [6]: v = (left.T @ diag_r @ right) + 1j * (left.T @ diag_i @ right)

In [7]: norm(u - v)
Out[7]: 3.0436364816617974e-16
```

genos:
I take back my earlier comment; for unitary `U`, I think `orthogonal-decomposition-of-UU^T` does work correctly. Defining 
```lisp
(defun gen-gaussian (n)
  (magicl:from-list
    (loop :for i :from 0 :below (* n n) :collect (alexandria:gaussian-random))
    (list n n)))

(defun gen-unitary (n)
  (let* ((m (magicl:.complex (gen-gaussian n) (gen-gaussian n)))
         (z (magicl:./ m (sqrt 2))))
    (multiple-value-bind (q r) (magicl:qr z)
      (let* ((d (magicl:diag r))
             (normalized-d
               (magicl:from-list
                 (loop :for x :in d :collect (/ x (abs x)))
                 (list n 1)))
             (matrix-d (magicl:hstack (loop :for _ :from 0 :below n :collect normalized-d))))
        (magicl:.* q matrix-d)))))
```

and setting `*check-math*` to `T`, the following works without complaining:

```lisp
* (loop :for _ :from 0 :below 100 :do (orthogonal-decomposition-of-uu^t (gen-unitary 100)))
NIL
```

As such, I'm not sure what's up with the CI failures.

kilimanjaro:
FWIW there's a pretty serious flaw in my pitch to use gges for this. I detailed a bit more in https://github.com/quil-lang/quilc/issues/841

stylewarning:
going to close this since it needs re-thinking

PR #848, opened 2022-09-08
remove RANDOM from FIND-DIAGONALIZER-IN-E-BASIS

### LINKED ISSUE #847: remove randomization in `find-diagonalizer-in-e-basis`
I don't think the `F-D-I-E-B` is a "probabilistic algorithm" in the traditional sense (i.e., we don't need randomness to asymptotically converge to something), we just need numbers that break possible degenerate relationships between the Re and Im part. I propose we replace the RNG with something aperiodic but deterministic sequence like $r_k = \vert\sin \frac{k}{10}\vert$ for $k=1,2,\ldots$. This sequence will get arbitrarily close to every value in $[0,1]$, but that probably matters only to the theoretician.

stylewarning:
fixes #847

### DIFF
--- a/src/compilers/approx.lisp
+++ b/src/compilers/approx.lisp
@@ -178,6 +178,17 @@
                      (diagonalizer-not-found-attempts c))))
   (:documentation "The diagonalizer for the given matrix was not found after a number of attempts."))
 
+(defun diagonalizer-number-generator (k)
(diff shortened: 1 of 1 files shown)


kilimanjaro:
This works for me. The role of the RNG previously was just to break possible degeneracies, so making it deterministic (but still allowing the same number of attempts) is an improvement. We can still consider alternative strategies for doing the canonical decomposition in the future.

kilimanjaro (approved):

PR #854, opened 2022-09-19
Increase heap size when initializing runtime for libquilc

kartik-s:
## Notes
This PR uses the `initialize-lisp-args` option introduced by quil-lang/sbcl-librarian#36 to increase the default heap size to 8 GiB when initializing the runtime in the bindings for libquilc.

## Testing
Built libquilc against SBCL master with the bigger heap size on macOS, the `compile-quil` C test now runs without issue.


stylewarning (approved):

stylewarning:
we gotta get this heap size under control

PR #859, opened 2022-10-11
abstract out stdgates.quil path to *default-standard-gates-file*

stylewarning:
Instead of inlined calls to `asdf:system-relative-pathname`, instead we offer a new exported variable `*default-standard-gates-file*`


Spin1Half (approved):
LGTM

Issue #766, opened 2021-11-08
Implement SQiSW decomposition


ecpeterson:
Huang et al. recently produced an optimal compilation technique for XY(pi/4) = \sqrt{ISWAP}, which quilc still compiles using numerical search, despite this being a native interaction on certain Rigetti devices. Implementing this would be a significant improvement in compiler output on pertinent architectures.

https://arxiv.org/abs/2105.06074

PR #868, opened 2023-01-12
Add: sqisw-decompose compiler

macrologist:
This branch intends to add sqisw decomposition to quilc, as per #766 . 

What remains to be done: 
- [ ] separate `define-compiler` defs for specific known gates 
- [ ] tests passing
- [ ] amendments and alterations per review and critique 

What has been done so far: A `define-compiler` definition has been added that decomposes an arbitrary 2 qubit gate into at most three SQiSW gates interleaved by single qubit gates, and, for 80% of all 2 qubit gates, only requires two SQiSW gates. 

As this is my first attempt to contribute to the compiler, I am looking for critique on anything that is obviously wrong or obviously flouting conventions.

### DIFF
--- a/src/compilers/sqisw-decomposition.lisp
+++ b/src/compilers/sqisw-decomposition.lisp
@@ -35,7 +35,7 @@ Where L(x,y,z) is the canonical representative of the class with coordinates x,y
                          (sqrt (/ (* 4 cos-x-sq cos-z-sq sin-y-sq)
                                   (+ (* 4 cos-x-sq cos-z-sq sin-y-sq)
                                      (* cos2x cos2y cos2z))))))))
-    (values
+    (list
      (list (build-gate "RZ" (list gamma) q0)
            (build-gate "RX" (list alpha) q0 )
            (build-gate "RZ" (list gamma) q0))
@@ -51,7 +51,6 @@ Where L(x,y,z) is the canonical representative of the class with coordinates x,y
 
 (defun canonicalize-for-sqisw (x0 y0 z0 q0 q1)
   "x0 y0 and z0 are canonical parameters in the Weyl chamber satisfying pi/4 >= x >= |z|. q0 and q1 are qubit indices.
-
   
 Returns 9 values. x y z A0 A1 B0 B1 C0 C1 with the following interpretation.  
 If x0,y0,z0 are canonical paramters for an arbitrary 2 qubit gate, then that gate is equivalent to 
--- a/src/compilers/sqisw-decomposition.lisp
+++ b/src/compilers/sqisw-decomposition.lisp
@@ -1,0 +1,181 @@
+;;;; sqisw-decomposition.lisp
+;;;;
+;;;; Author: Colin O'Keefe Following the algorithm outlined in
+;;;; https://arxiv.org/pdf/2105.06074.pdf, Supplemental Material, I.B.
+
+(in-package #:cl-quil)
+
+(defun interleaving-1-qubit-gates (x y z q0 q1)
+  "x y and z are canonical parameters in the Weyl chamber satisfying pi/4 >= x >= y >= |z|. Furthermore, this function assumes that the class of gates indicated by x,y,z falls within the span of 2 SQISW gates, which is equivalent to the additional condtion that x >= y + |z|. 
+
+q0 and q1 are qubit indices
+
+Returns two values L0 and L1, both lists of single-qubit gate applications, acting on q0 and q1 respectively. 
+L0 is a list of three rotations RZ(γ) RX(α) RZ(θ), and L1 is a list containing one RX(β).
+
+L(x,y,z) ~ SQiSW [ RZ(γ)RX(α)RZ(γ) ⊗ RX(β)] SQiSW
+
+Where L(x,y,z) is the canonical representative of the class with coordinates x,y, and z."
+  (declare (type double-float x y z))
+  (assert (xyz-in-2-sqisw-span-p x y z) () "~a ~a ~a not in the span of 2 SQISW gates." x y z)
+  (let* ((c (* (sin (+ x y (- z)))
+               (sin (+ x (- y) z))
+               (sin (+ (- x) (- y) (- z)))
+               (sin (+ (- x) y z))))
+         (cos2x (cos (* 2 x)))
+         (cos2y (cos (* 2 y)))
+         (cos2z (cos (* 2 z)))
+         (cos-x-sq (* (cos x) (cos x)))
+         (cos-z-sq (* (cos z) (cos z)))
+         (sin-y-sq (* (sin y) (sin y)))
+         (2-sqrt-c (* 2 (sqrt c)))
+         (alpha (acos (+ cos2x (- cos2y) cos2z 2-sqrt-c)))
+         (beta (acos (+ cos2x (- cos2y) cos2z (- 2-sqrt-c))))
+         (gamma (acos (* (cond ((plusp z) 1d0) ((minusp z) -1d0) (t 0d0)) ; sign(z)
+                         (sqrt (/ (* 4 cos-x-sq cos-z-sq sin-y-sq)
+                                  (+ (* 4 cos-x-sq cos-z-sq sin-y-sq)
+                                     (* cos2x cos2y cos2z))))))))
+    (values
+     (list (build-gate "RZ" (list gamma) q0)
+           (build-gate "RX" (list alpha) q0 )
+           (build-gate "RZ" (list gamma) q0))
+     (list (build-gate "RX" (list beta) q1)))))
+
+
+(defun xyz-in-2-sqisw-span-p (x y z)
+  (and (in-weyl-chamber-p x y z)
+       (>= x (+ y (abs z)))))
+
+(defun in-weyl-chamber-p (x y z)
+  (>= #.(/ pi 4.0d0) x y (abs z)))
+
+(defun canonicalize-for-sqisw (x0 y0 z0 q0 q1)
+  "x0 y0 and z0 are canonical parameters in the Weyl chamber satisfying pi/4 >= x >= |z|. q0 and q1 are qubit indices.
+
+  
+Returns 9 values. x y z A0 A1 B0 B1 C0 C1 with the following interpretation.  
+If x0,y0,z0 are canonical paramters for an arbitrary 2 qubit gate, then that gate is equivalent to 
+
+     [A0 ⊗ A1] L(x,y,z) [C0 ⊗ C1] SQiSW [B0 ⊗ B1] 
+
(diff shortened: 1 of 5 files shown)


ecpeterson:
Nice! I’ll try to have a look over the weekend.

macrologist:
> Nice! I’ll try to have a look over the weekend.

@ecpeterson 

If you wait until next weekend, it might even work!  I've got the two sqisw case working but something is amiss in the three sqisw case. I think its that I'm mis-using the output of `canonical-decomposition` to extract the canonical coordinates and then halving them with fingers crossed. I have to step through and see that it actually makes sense. 

In any case, any feedback is appreciated!

src/compilers/sqisw-decomposition.lisp
```diff
@@ -0,0 +1,180 @@
...
+(define-compiler sqisw-decompose
+    ((instr (_ _ q1 q0)))
+  "This compiler will produce a sequence containing at most two SQISW
+applications whenever possible, and, when it cannot, is guaranteed to
+produce a sequence containing at most three SQiSW applications."
+  (destructuring-bind
+      (b0 b1 can1 a0 a1) (canonical-decomposition instr)
+    (destructuring-bind
+        ;; making an assumption that I can divide by two here.
+        ;; TODO: check that this isn't bunk.
```
ecpeterson, line 111: I myself haven't looked, but `canonical-decomposition` is where I'd recommend starting your bug search. The coordinates used in [PCS](https://arxiv.org/abs/1904.10541), which I think are what's used here, have a goofy normalization condition that causes its canonical gates to span a triangular prism rather than a tetrahedron. If that's indeed what's causing the problem, you could inject an extra step here to re-scissor its output into the tetrahedron you expect — but you would probably make everyone's life easier by changing it to use the same convention everyone else does. I don't think much (anything?) in quilc depends on that alternative convention, and it was definitely a flub on our part not to do that from the start.

src/compilers/sqisw-decomposition.lisp
```diff
@@ -0,0 +1,180 @@
...
+                            (list (build-gate "SQISW" () q1 q0)))))
+             (destructuring-bind
+                 (e0 e1 can2 d0 d1) (canonical-decomposition (anon-gate "V" (make-matrix-from-quil v)  q1 q0))
+               (declare (ignore can2))
+               (inst b0)
+               (inst (anon-gate "E0t" (magicl:conjugate-transpose (gate-matrix e0)) q0))
+               (inst b1)
+               (inst (anon-gate "E1t" (magicl:conjugate-transpose (gate-matrix e1)) q1))
+               (inst "SQISW" () q1 q0)
+               (mapc #'inst c0)
```
ecpeterson, line 129: I think there's an `inst*`. If there isn't, there should be!

src/compilers/sqisw-decomposition.lisp
```diff
@@ -0,0 +1,180 @@
...
+               (mapc #'inst c1)
+               (inst "SQISW" () q1 q0)
+               (inst (anon-gate "D0t" (magicl:conjugate-transpose (gate-matrix d0)) q0))
+               (inst a0)
+               (inst (anon-gate "D1t" (magicl:conjugate-transpose (gate-matrix d1)) q1))
+               (inst a1)))))
+
+        (t
+         (destructuring-bind
+             (cx cy cz f0 f1 g0 g1 h0 h1) (canonicalize-for-sqisw x y z q0 q1)
```
ecpeterson, line 139: Here again I'd recommend trampolining rather than duplicating the decomposition above. You can `define-compiler` version (a) of this routine for the two 2Q case — it operates only only `CAN` gates and has a `:where` clause which imposes that coordinate inequality, which the compiler manager will take to mean "this sometimes works and sometimes yields short output, so we'll try it first and not sweat a failure" — and then a second version (b) of this routine which handles the "general" case, which emits 1Q * SQiSW * 1Q * CAN * 1Q, where the compiler loop will re-walk this emitted sequence and automatically apply (a) to the emitted CAN.

ecpeterson (commented):
Even though you said this isn't finished, I took a look anyway — I'm more likely to have the time this holiday weekend than generally.

So far so good! Some comments. Feel free to ping again as you make progress.

src/compilers/sqisw-decomposition.lisp
macrologist: GTK. I may add one if there isn't.

src/compilers/sqisw-decomposition.lisp
macrologist: Yeah I was just thiking about this this morning: Two define compilers, one for the two-sqisw case using a "gaurd" to check the condition, and then utilizing that compiler in the body of the three-sqisw case's  compiler definition.

Thanks for your feedback! I really appreciated it.

src/compilers/sqisw-decomposition.lisp
```diff
@@ -0,0 +1,180 @@
...
+
+(defun interleaving-1-qubit-gates (x y z q0 q1)
+  "x y and z are canonical parameters in the Weyl chamber satisfying pi/4 >= x >= y >= |z|. Furthermore, this function assumes that the class of gates indicated by x,y,z falls within the span of 2 SQISW gates, which is equivalent to the additional condtion that x >= y + |z|. 
+
+q0 and q1 are qubit indices
+
+Returns two values L0 and L1, both lists of single-qubit gate applications, acting on q0 and q1 respectively. 
+L0 is a list of three rotations RZ(γ) RX(α) RZ(θ), and L1 is a list containing one RX(β).
+
+L(x,y,z) ~ SQiSW [ RZ(γ)RX(α)RZ(γ) ⊗ RX(β)] SQiSW
```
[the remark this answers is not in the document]
macrologist: yerp. typo.

src/compilers/sqisw-decomposition.lisp
```diff
@@ -0,0 +1,180 @@
...
+
+(in-package #:cl-quil)
+
+(defun interleaving-1-qubit-gates (x y z q0 q1)
+  "x y and z are canonical parameters in the Weyl chamber satisfying pi/4 >= x >= y >= |z|. Furthermore, this function assumes that the class of gates indicated by x,y,z falls within the span of 2 SQISW gates, which is equivalent to the additional condtion that x >= y + |z|. 
+
+q0 and q1 are qubit indices
+
+Returns two values L0 and L1, both lists of single-qubit gate applications, acting on q0 and q1 respectively. 
+L0 is a list of three rotations RZ(γ) RX(α) RZ(θ), and L1 is a list containing one RX(β).
```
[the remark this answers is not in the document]
macrologist: This is something that I really wasn't sure about: to perform the gate compositions myself and spit out anon gates, or to spit out the rotations with explicit angle parameters.  Thanks for the take! 

Generally, these kinds of "what is best for quilc" questions are what I'm most uncertain about right now.

src/compilers/sqisw-decomposition.lisp
```diff
@@ -0,0 +1,180 @@
...
+(defun in-weyl-chamber-p (x y z)
+  (>= #.(/ pi 4.0d0) x y (abs z)))
+
+(defun canonicalize-for-sqisw (x0 y0 z0 q0 q1)
+  "x0 y0 and z0 are canonical parameters in the Weyl chamber satisfying pi/4 >= x >= |z|. q0 and q1 are qubit indices.
+  
+Returns 9 values. x y z A0 A1 B0 B1 C0 C1 with the following interpretation.  
+If x0,y0,z0 are canonical paramters for an arbitrary 2 qubit gate, then that gate is equivalent to 
+
+     [A0 ⊗ A1] L(x,y,z) [C0 ⊗ C1] SQiSW [B0 ⊗ B1]
```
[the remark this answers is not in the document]
macrologist: They're lists of matrices. @stylewarning made the same suggestion. 

Also, that composition expression in the docstring is wrong - The C's and B's need to swap position. Just a typo. Sorry for any confusion while you were reviewing.   A corrected docstring will be pushed on Monday.

src/compilers/sqisw-decomposition.lisp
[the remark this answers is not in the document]
ecpeterson: This specific case can go either way, since the compressor is going to re-do the work I'm suggesting no matter what. But the general philosophy is that individual compiler routines should do as little work + express as few opinions as they can get away with, so that the compiler loop can creatively combine them. The "nativization" loop has more options for what to do with anonymous gates than it does with fixed gate sequences, hence the preference — but, again, the compressor's primary strategy is to re-anonymize sequences in an effort to force the nativizer to find unexpected cancellations, so it'll probably come out the same in the end.

src/compilers/sqisw-decomposition.lisp
ecpeterson: Similarly to my other reply-comment, leaning on the outer loop means (1) you don't even need to explicitly call out to the depth 2 case within the depth 3 case, since it'll do it for you; and (2) if someone ever implements an even clever-er inner decomposition, _not_ explicitly calling out to the depth 2 case in this code will leave the compiler loop the flexibility to select its preferred inner decomposition.

src/compilers/sqisw-decomposition.lisp
ecpeterson: More info on this scissors business in [Remark 2.7](https://arxiv.org/abs/2111.02535) and [all over this Python file](https://github.com/Qiskit-Extensions/monodromy/blob/main/monodromy/coordinates.py).

src/compilers/sqisw-decomposition.lisp
macrologist: :bulb:

src/compilers/sqisw-decomposition.lisp
stylewarning: ✂️➡️🔺

macrologist:
I've pushed again. the depth-3 case still is not working. I am going to narrow my search to `canonicalize-for-sqisw` (which, incidentally, I should rename) and check that each of the cases it addresses are conditioned correctly. This check entails stepping back through the proof and deriving each case condition from the cases treated in the proof (which are stated in superficially different, but presumably equivalent, terms).  I've pushed these changes in case inquiring minds wanted to know.

stylewarning:
adding this before i forget: I imagine that we should call the gate `SQRT-ISWAP` to fit the Quil vibes, but I'm not actually picky about this whatsoever. If not that, then `SQISWAP`.

macrologist:
This is not ready btw. Even though.. somewhat mysteriously, checks have passed. I should add tests specific to these compiler rules to  the suite.

macrologist:
I'm going to close this PR, flatten commits, and open a new one.

PR #870, opened 2023-01-24
fixes (for) boondoggle: quil -> cl-quil; workaround in print-program

macrologist:
I ran the boondoggle tests like so:

    cd <path-to>/qvm 
    make qvm 
    ./qvm -S 

    #elsewere ...
    sbcl 
    > (ql:quickload :boondoggle-tests) 
    > (asdf:test-system :boondoggle) 

and they passed. 

Some notes: 
- I changed a `dotimes` to a `loop repeat` because unused and/or shadowed counter variables make me nervous. 
- I altered `print-parsed-program-generic` in a way that, in a more perfect world, would not be necessary, but that shouldn't hurt anything.

### DIFF
--- a/boondoggle/src/producers.lisp
+++ b/boondoggle/src/producers.lisp
@@ -24,7 +24,7 @@ PARAMETER-BOUNDS is a list of maximum random values for the gate parameters."
   (error "This producer has not been implemented."))
 
 (defclass producer-random (producer)
-  ((chip-specification :initform (quil::build-8Q-chip)
+  ((chip-specification :initform (cl-quil::build-8Q-chip)
                        :initarg :chip-specification
                        :reader producer-random-chip-specification)
    (program-depth-limit :initform nil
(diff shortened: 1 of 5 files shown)


boondoggle/src/producers.lisp
```diff
@@ -81,16 +81,16 @@ PARAMETER-BOUNDS is a list of maximum random values for the gate parameters."
                          respect-topology     ; place multiqubit gates only on chip-specification links
                          )
   "Generates random Quil strings to the statistics specifications set out by the argument list."
-  (let ((parsed-program (make-instance 'quil::parsed-program))
+  (let ((parsed-program (make-instance 'cl-quil::parsed-program))
         (instruction-list nil)
         (random-gate-count 0))
-    (dotimes (j (random program-volume-limit))
+    (loop repeat (random program-volume-limit) do
```
stylewarning, line 87: style is to use colon keywords

stylewarning (approved):

boondoggle/src/producers.lisp
macrologist: ahh will do in the future.

PR #871, opened 2023-01-27
Add: define-compiler decomposing CAN gates to <= 3 SQISWAP gates

macrologist:
Defines two mutually exclusive compiler rules that match CAN gates. One rule decomposes a CAN into two SQISWAP gates interleaved by 1 qubit rotations; call this the case-2 rule. The other rule, call it case-3, decomposes a CAN into an SQISWAP gate and another CAN gate to which case-2 is applicable. 

There are some inline "tests" in comment blocks that I have left in for the time being as an aid to understanding. I can remove them if desired. 

This PR partially addresses #766

### DIFF
--- a/cl-quil-tests.asd
+++ b/cl-quil-tests.asd
@@ -51,4 +51,5 @@
                (:file "approximation-tests")
                (:file "addresser-tests")
                (:file "linear-reversible-circuit-tests")
-               (:file "permutation-tests")))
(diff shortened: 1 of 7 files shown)


stylewarning (approved):

stylewarning:
@macrologist Congratulations and welcome to being a QUILC contributor!

Issue #692, opened 2021-01-25
Confusion Between 'Schedule' and 'Scheduler'


ghost:
There seems to be confusion between 'schedule' and 'scheduler' throughout sources and doc. They seem to be used interchangeably, which seems confusing. Is a "schedule" the same as a "scheduleR"?

For example, consider the following code snippet in `src/compressor/compressor.lisp`:
```
  (let ((lschedule (make-lscheduler)))
    ;; load up the logical schedule
```
The variable `lschedule` and the comment refer to a "schedule", but `(make-lscheduler)` refers to a "scheduleR", and it creates an instance of `logical-scheduler`, i.e., scheduler with an R.

This sort of confusion seems pervasive throughout the system. A review and cleanup of the two usages may be worthwhile.

In case this helps, here are two grep commands I found useful to find 'scheduler' (with the R) as opposed to 'schedule' (without the R), from the root of the repo:
```
grep -r -i -E --include=\*.lisp "scheduler" .
```
```
grep -r -i -E --include=\*.lisp "schedule([^r]|$)" .
```

stylewarning:
This actually led to a bug (which has now been fixed) where, when `SAFETY` was sufficiently low, SBCL wasn't checking that a class slot type was a valid type. At some point somebody had written `logical-schedule` as a slot type, which doesn't exist, because it (the type) is called `logical-scheduler`.

Issue #662, opened 2020-07-24
ANON-GATE is risqué


stylewarning:
```
QUIL> (anon-gate "BUG" (random-special-unitary 2) nil 0)
#<BUG-22 nil 0>
```

It takes `nil` to be a `FORMAL` whose name is `nil`. I think this would be a common error and should be eliminated.

mhdavid-hrl:
I'm not sure I see why nil would be passed in in error. It does not seem to have any documented meaning.
I do not think it's great to rule out the symbol name "nil". 
I suppose you could have a special rule that if you want that, you have to use :nil (nil in keyword package).

PR #873, opened 2023-02-09
Consistency fixes

macrologist:
A few small changes, corrections,and bugfixes. 

Some notes:

I'm unsure about the decision I made in commit 15d1558 to prefer `lschedule` and `logical-schedule` over names that end in `scheduler`. It could go either way - I chose this language because "addding to a schedule" makes more intutive sense (to me) than "adding to a schedul**er**".   The commit is meant to address #692 

153e0c9 is meant to address #662. I think I may agree with the now ghostly commenter who suggests that it probably isn't much of an issue. However, disallowing `NIL` as the source of a formal's name while still alowing the name `"nil"` pulled out of a `:NIL` seems like a fine compromise between the two positions. Being wary of `NIL`s with unintended meanings seems like a fine suuperstition to adhere to in Common Lisp.


stylewarning:
typo in your commit message: "replacement"

src/ast.lisp
```diff
@@ -1329,7 +1329,7 @@ This definition does *not* incorporate the operator description (i.e., any opera
 If this slot is not supplied, then the gate is considered *anonymous*. If this is the case, then the GATE slot must be supplied.")
    ;; N.B. See the generic function GATE-APPLICATION-GATE as well.
    (gate :initarg :gate
-         :initform nil
+         ;:initform nil
```
stylewarning, line 1332: this looks suspect

stylewarning (changes requested):
can we move the LSCHED renaming to a separate PR? I did read it all, and it all looks good barring one `;`, but it'll be easier to read and ensure ok

src/ast.lisp
macrologist: oh fudge.

PR #877, opened 2023-02-16
Add: pseudo-gate-application class

macrologist:
Sometimes one has a need for a formal GATE-APPLICATION instruction that will never manifest as a gate object nor reify as a matrix.

This commit introduces the PSUEDO-GATE-APPLICATION class that makes these "purely formal" gate applications semantically distinct from "ordinary" GATE-APPLICATION instances.

A natural example is the STATE-PREP-APPLICATION class, which has been refactored to subclass PSEUDO-GATE-APPLICATION.

### DIFF
--- a/src/ast.lisp
+++ b/src/ast.lisp
@@ -1380,6 +1380,18 @@ N.B. This slot should not be accessed directly! Consider using GATE-APPLICATION-
     (t
      (assert (slot-boundp app 'gate) (app) "A gate application was made that doesn't have a definition. Please specify either its name resolution or a gate."))))
 
+(defclass pseudo-gate-application (gate-application)
+  ((gate :initform nil))
+  (:documentation "Represents a formal, \"stand-in\", gate-application instruction that will never be associated to an underlying gate."))
+
+(defmethod gate-application-gate ((app pseudo-gate-application))
+  (error "Attempted to return a GATE representation from a PSEUDO-GATE-APPLICATION instance: ~s" app))
+
+(defmethod gate-matrix ((app pseudo-gate-application) &rest params)
+  (declare (ignore params))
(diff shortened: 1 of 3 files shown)


macrologist:
@stylewarning A wee PR. We should either merge or scrap it. 

I do know of a few spots where it can be legitimately employed, trivial though the change is. Semantics matter imho.

stylewarning:
> @stylewarning A wee PR. We should either merge or scrap it. 
> 
> 
> 
> I do know of a few spots where it can be legitimately employed, trivial though the change is. Semantics matter imho. 

My only objection is that it doesn't seem to be a very meaningful description. When should I actually use a pseudo-gate?

The way it's being used now is to represent *some* non-specific unitary that can take us from one state to another. We *could* manufacture a matrix for this gate but it wouldn't be canonical, and it wouldn't be unique even up to phase.

So, I don't want to bikeshed, but it's not clear to me what other examples of specifically something called a "pseudo-gate-application" might look like.

macrologist:
I think I mostly agree. 

But to try to address the critique, I can imagine whole-program transformation that produces a sequence of instructions that act as "place holders" for yet another pass to come. I have seen this in practice. Using a gate-application for this you must remember to explicitly pass `nil` to the `:gate` initarg.  And even when you do - other users looking at your code down the line may not realize that this seemingly ordnary gate application is actually something that is never meant to have a representation you can apply the evolution of a quantum system. 

But, yes, such circumstances are perhaps a bit recondite.

It seems to me that if you really want to do  something like the above, you shouldn't abuse your `parsed-program` instance's executable code slot to hold  your intermediate states that way. And instead, should just use a custom structure fit for the purpose. 

So yeah. I'm cool with closing it. :shrug:

stylewarning:
Maybe could you convert this PR into an issue that describes some of the funny business going on?

PR #879, opened 2023-03-14
Fix: Coalton fixes for discrete-compilation

eliaslfox:
Update the discrete compilation code to match changes in the Coalton compiler and standard library.


colescott (dismissed):
Looks good. We should probably be running the discrete tests in CI so this doesn't bit-rot

stylewarning (approved):
seems reasonable to me

PR #880, opened 2023-03-15
Add Extern  Support

macrologist:
The EXTERN ~~Pragma~~ keyword allows for users to indicate that certain operators should be "virtual" from the perspective of the frontend parsing and compilation stages. By virtual, I mean that applications of these operators do not have a definition.

It is expected that users will provide definition for these externs during `backend-compile` for their own subclasses of `cl-quil:backend`.   

For example:

```
EXTERN PHONY

X 0
CNOT 1 0
PHONY 1
CNOT 0 1
```

The above code will parse and compile but result in an opaque instance of `extern-application` in the resulting code. Extern instructions may be rewired, but will otherwise remain untouched. Similar to the effect of `PRESERVE_BLOCK` pragmas.

Addtionally, the user can specify that an ordinarily "off the shelf" gate should be an EXTERN, meaning that, again, the user is expected to supply an implementation for that gate on the backend.

E.g. 

``` 
EXTERN H
EXTERN CNOT

X 0
CNOT 1 0
Y 0
H 1
CZ 1 0
```

Such a program will  allow the compiler to neither optimize nor nativize `CNOT` and `H` gate applications.


src/ast.lisp
```diff
@@ -1302,10 +1302,18 @@ consists of a CONTROLLED-OPERATOR acting on a NAMED-OPERATOR."
...
+
     * Application is an invalid application.
 
 Determining this requires the context of the surrounding program."))
 
+(defclass extern-application (application)
+  ()
+  (:documentation "Represents an externed operation. An extern is expected to receive its definition late in the compilation process, mostly likely for a specialization of cl-quil:backend-compile. 
+
+The intent is to allow users to supply definitions to these operations in a form that comports with their compilation target., but to still be able to refer to these operations in their Quil source code."))
```
stylewarning, line 1315: stray period

I would also reference PRAGMA PRESERVE_BLOCK, and to prefer to use that in the appropriate circumstance.

src/ast.lisp
```diff
@@ -1302,10 +1302,18 @@ consists of a CONTROLLED-OPERATOR acting on a NAMED-OPERATOR."
...
 
+    * Application is an extern application.
+
     * Application is an invalid application.
 
 Determining this requires the context of the surrounding program."))
 
+(defclass extern-application (application)
+  ()
+  (:documentation "Represents an externed operation. An extern is expected to receive its definition late in the compilation process, mostly likely for a specialization of cl-quil:backend-compile.
```
stylewarning, line 1313: Add to the docs that this, like UNRESOLVED-APPLICATION, is a tool for users to compile programs usually beyond the Quil level. These application classes don't in and of themselves carry meaningful semantics with respect to the quantum abstract machine (i.e., they can't be executed or simulated).

src/ast.lisp
```diff
@@ -1756,12 +1764,17 @@ For example,
    (executable-program :initarg :executable-code
                        :accessor parsed-program-executable-code
                        :type (vector instruction)
-                       :documentation "A vector of executable Quil instructions."))
+                       :documentation "A vector of executable Quil instructions.")
+   (externed-operations :initarg :externed-operations
```
stylewarning, line 1768: Do you think "extern-operations" is better for consistency?

src/cfg.lisp
```diff
@@ -175,6 +175,15 @@ Return the following values:
   (vector-push-extend instr (basic-block-code blk))
   (values blk nil nil))
 
+(defmethod process-instruction (cfg blk (instr extern-application))
+  (declare (ignore cfg))
+  (assert (not (null blk)) (blk))
+  ;; extern applications should be preserved from nativization and optimization
```
stylewarning, line 181: I only have hazy memory of the mechanics of this function, but is this going to inadvertently preserve all other instructions resident to this block too? Should extern instructions be pushed to their own block? Or maybe no preserve-blocks at all?

tests/extern-pragma-tests.lisp
```diff
@@ -0,0 +1,36 @@
+;;;; extern-pragma-tests.lisp
+;;;;
+;;;; Author: Colin O'Keefe
+
+(in-package #:cl-quil-tests)
+
+(deftest test-extern-pragmas ()
```
stylewarning, line 7: add a test that something like

    PRAGMA EXTERN H
    H 0
    Y 0

to see that H stays externed and doesn't compile further, but Y does.

stylewarning (changes requested):
Overall looking good. One thing I'm wondering is whether this should be a pragma or new keyword. Given it's newness, I'm fine with keeping it a pragma for now, but given the attitude I have toward programs (remove them to keep a valid program), they're a bit of an odd duck. I'm not intent on getting to bikesheddy about it now though. 

Make sure that a compiled program still maintains the EXTERN pragmas.

src/cfg.lisp
macrologist: I wasn't totally sure about this: I'd hoped compilation tests would reveal issues. I'll add more tests and report back.

macrologist:
@stylewarning I have taken your idea about the keyword approach to heart. As of 1481149 the pragma has been removed and replaced with a new keyword.

src/ast.lisp
```diff
@@ -421,6 +421,15 @@ If no exit rewiring is found, return NIL."
   (:documentation "A directive to include another file in a Quil file."))
 
 
+(defclass extern-operation ()
+  ((name :reader extern-operation-name
+         :initarg :name
+         :documentation "The Name of the operation being marked as an EXTERN"))
```
stylewarning, line 427: lower n

src/ast.lisp
```diff
@@ -421,6 +421,15 @@ If no exit rewiring is found, return NIL."
   (:documentation "A directive to include another file in a Quil file."))
 
 
+(defclass extern-operation ()
```
stylewarning, line 424: i would call it a directive (or just EXTERN): https://quil-lang.github.io/#3-3Main-Program-Elements

src/ast.lisp
```diff
@@ -1756,12 +1773,17 @@ For example,
...
+                      :accessor parsed-program-extern-operations
+                      :type hash-table
+                      :documentation "A hash table mapping string NAMEs to generalized booleans, indicating that an operation so named is an extern."))
   (:default-initargs
    :gate-definitions '()
    :circuit-definitions '()
    :memory-definitions '()
-   :executable-code #())
+   :executable-code #()
+   :extern-operations (make-hash-table :test #'equal))
```
stylewarning, line 1786: here, OPERATION is OK i think

src/package.lisp
```diff
@@ -401,6 +401,8 @@
 
    #:unresolved-application             ; CLASS
 
+   #:extern-application                 ; CLASS
```
stylewarning, line 404: export the AST node too

stylewarning (changes requested):
Add some parser tests to the right place too; you'll need to spelunk to see how they're done now

src/compiler-hook.lisp
```diff
@@ -314,13 +314,16 @@ Returns a value list: (processed-program, of type parsed-program
                  (remove-if-not #'global-pragma-instruction-p
                                 (parsed-program-executable-code parsed-program))
                  (parsed-program-executable-code processed-program)))
-          ;; retain the old circuit and gate definitions
+          ;; retain the old circuit and gate  definitions
```
stylewarning, line 317: double spacer

stylewarning (approved):
LGTM; we'll experiment with this feature. If it's ultimately blessed as useful over the next 6 months or so, we'll standardize it in Quil.

At some point we'll need to update the QVM too.

src/compiler-hook.lisp
macrologist: fudge 😓

PR #881, opened 2023-03-16
Refactor LIST-AVAILABLE-BACKENDS; include backend subclass trees

macrologist:


src/backends/common.lisp
```diff
@@ -53,9 +53,15 @@
 
 
 (defun list-available-backends ()
-  ;; This is simpleminded and assumes there isn't some crazy
-  ;; inheritance structure.
-  (mapcar #'class-name (c2mop:class-direct-subclasses (find-class 'backend))))
+  "Return a list of all subclasses of the class named BACKEND"
+  (let ((backends))
```
stylewarning, line 57: minor nit: since the value of backends is used prior to any SETF or other explicit initialization, we should supply an initial value here of NIL.

If OTOH backends were

    ;; complicated computation to derive foo
    (setf backends foo)

then I think

    (let (backends ...)

would be ok.

src/backends/common.lisp
```diff
@@ -53,9 +53,15 @@
...
+    (labels ((add-backends (class)
+               (let ((direct
+                       (c2mop:class-direct-subclasses class)))
+                 (when direct
```
stylewarning, line 61: minor nit: prefer COND here since

1. we depend on the return value 
2. the branches have side-effectful ops

stylewarning (changes requested):
just very minor stylistic patterns to be consistent elsewhere

src/backends/common.lisp
macrologist: I altered the implementation to not depend on the return value.

src/backends/common.lisp
```diff
@@ -53,9 +53,17 @@
...
+  (let ((backends nil))
+    (labels ((add-backends (class)
+               (let ((direct
+                       (c2mop:class-direct-subclasses class)))
+                 (when direct
+                   (setf backends (nconc backends direct))
+                   (dolist (cl direct)
+                     (add-backends cl))))))
+      (add-backends (find-class 'backend))
+      (remove-duplicates backends :test #'eq))))
```
stylewarning, line 66: all this NCONCing to not end with DELETE-DUPLICATES?

stylewarning (dismissed):
i can squash merge or you can rebase first, at your preference

stylewarning (approved):

PR #883, opened 2023-03-17
Hotfix: list-available-backends; nconc causes evil behavior

macrologist:
Two fixes. One: list-available-backends should return not classes but class names.

Two: the way it was written before, calling list-available-backends twice resulted in a hang leading to a heap overflow.

This change fixes quilc but does not address the issue, which may indicate a bug in sbcl.

### DIFF
--- a/src/backends/common.lisp
+++ b/src/backends/common.lisp
@@ -59,11 +59,11 @@
                (let ((direct
                        (c2mop:class-direct-subclasses class)))
                  (when direct
-                   (setf backends (nconc backends direct))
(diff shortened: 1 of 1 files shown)


macrologist:
(I shouldn't make PRs during my afternoon caffiene crash :-( )

macrologist:
In the interests of lispology: not a bug; just tricksy. The direct subclasses accessor is returning the actual list of those subclasses (duh, why wouldn't it?). NCONC is therefore mutating the list values on those class object slots.

stylewarning (approved):

PR #886, opened 2023-04-03
Call-Chip-Builder Function Accepts 'Linear' and 'Fully-Connected' Chips

Bluesquare99:
Added `linear` and `fully-connected` to the available chips that could be called via the `call-chip-builder` function within `chip-library.lisp`.  They accept one argument, `nQ`, which is the number of qubits for which to build the chip.  They return a chip object same as all other chip-building functions do.

### DIFF
--- a/src/chip-library/chip-table.lisp
+++ b/src/chip-library/chip-table.lisp
@@ -41,3 +41,5 @@ returns nil if not found."
 (install-chip-builder "16QMUX" 'q::build-16QMUX-chip)
 (install-chip-builder "bristlecone" 'q::build-bristlecone-chip)
 (install-chip-builder "ibmqx5" 'q::build-ibm-qx5)
+(install-chip-builder "linear" (lambda (nQ) (q::build-nq-linear-chip nQ)))
+(install-chip-builder "fully-connected" (lambda (nQ) (q::build-nq-fully-connected-chip nQ)))


stylewarning (changes requested):
this looks correct to me. in this case we can simplify a little bit:

    (lambda (x) (f x))

can be replaced with

    'f

or

    #'f

(the former is preferred for subtle technical reasons in this case)

fun fact: in the formal theory of lambda calculus, this is called *η-reduction* or *eta reduction*.

stylewarning (changes requested):
this looks correct to me. in this case we can simplify a little bit:

    (lambda (x) (f x))

can be replaced with

    'f

or

    #'f

(the former is preferred for subtle technical reasons in this case)

fun fact: in the formal theory of lambda calculus, this is called *η-reduction* or *eta reduction*.

stylewarning:
I don't know why it double-posted. That's just how it is sometimes.

stylewarning (approved):
actually, i'm deciding to accept this without the eta reduction. I think the lambda makes it clearer and more explicit that these require a parameter to be supplied.

Bluesquare99:
> actually, i'm deciding to accept this without the eta reduction. I think the lambda makes it clearer and more explicit that these require a parameter to be supplied.

okay great!  i'm not authorized to merge the pr so can't do so at this time.

PR #902, opened 2023-10-10
Fixing bugs to get discrete tests to load and run

Izaakwltn:
I ran into trouble getting sdiscrete tests to load and run:

1) I fixed some `.*` and `*.` to `s*` `*s` deprecated spots

2) There is an issue with a foreign function in the test `compile-rz`, so I commented the out (I think it's on the Magicl end). I can look into it, but wanted to get word out there first.

### DIFF
--- a/cl-quil.asd
+++ b/cl-quil.asd
@@ -113,7 +113,6 @@
                #:global-vars            ; Static globals
                #:salza2                 ; God table compression
                #:optima
-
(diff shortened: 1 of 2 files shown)


stylewarning:
I object to commenting out that test, since it is the most important test there is. Have you installed LAPACK and the latest MAGICL from source?

stylewarning (approved):

Issue #676, opened 2020-10-15
Inconsistent results compiling bell state program


mhodson-rigetti:
Compiling the following bell-state program via pyquil

```
    program = Program()
    ro = program.declare("ro", "BIT", 2)
    program.inst(H(1))
    program.inst(CNOT(1, 4))
    program.inst(MEASURE(1, ro[0]))
    program.inst(MEASURE(4, ro[1]))
```
... using the following quilc ISA patched to remove CZ gates (force use of XY)

```
[...]

PR #905, opened 2023-10-18
bugfix: copy-instance was not robust enough

macrologist:
This partially addresses one source of "nondeterminism" in quilc compilation (see #676) that was due to a violation of referential transparency.

Specifically, non destructive compilation was not non-destructive enough: compiling the same parsed-program instance twice produced two different outputs.  This was because copy-instance did not copy instances of measurement deeply, resulting in different outputs from compress-qubits, with downstream consequences.

This only addresses #676 in those cases where you are not parsing the same raw text-based source everytime you compile. Or equivalently, when you are not rebuilding your `Program` object from pyquil every time.


macrologist:
(oh: I commented out the clifford tests b/c they've been failing for a while now and I wanted to sanity check my change, we might consider this PR, and any other wide-range-effect PRs to be blocked until the clifford issue is addressed)

stylewarning:
what a great find

stylewarning:
re-enable clifford tests and we can merge

macrologist:
turns out the cliffort tests are failing b/c the fucntion `cl-permutation::iota-vector` does not produce values of type `cl-permutation::raw-perm`. the former is being used to to produce permutation representations passed in to `cl-permutation::%make-perm`, throwing a type error.  Sit tight.

macrologist:
should be good

PR #916, opened 2024-02-07
Cleanups having to do with fidelity of sets of instructions

macrologist:
Move defs near to use sites; refactor instrs fidelity

Fidelity calculations used by the compressor and by the fidelity addresser were needlessly constructing logical schedules.

Peeking at the logical schedule fidelity calculations, they too have been cleaned up and refatored for clarity and performance.


src/addresser/fidelity-addresser.lisp
```diff
@@ -26,6 +26,13 @@
 (defmethod cost-flatten ((cost fidelity-cost))
   (fidelity-cost-value cost))
 
+(defun calculate-instructions-log-fidelity (instructions chip-specification)
+  "Calculates the fidelity of a sequence of native INSTRUCTIONS on a chip with architecture governed by CHIP-SPECIFICATION (and with assumed perfect parallelization across resources)."
+  (reduce #'+ instructions
```
stylewarning, line 31: I prefer an initial value be provided.

Do we want to use + here? I forget if we ever use objects to represent fidelities or not. Probably not.

src/addresser/fidelity-addresser.lisp
```diff
@@ -26,6 +26,13 @@
...
+          :key (lambda (instr)
```
stylewarning, line 32: might be silly but maybe we should rewrite the reduce as a loop to avoid the closure, or you should FLET it (it would have a nice name) and declare it dynamic-extent

SBCL is maybe smart enough to do this already; I don't know

src/addresser/logical-schedule.lisp
```diff
@@ -658,55 +662,25 @@ Returns the reduction of all bumped values by COMBINE-VALUES, and a hash table m
...
-                 (declare (ignore val))
-                 (incf running-fidelity (get-fidelity instr)))
-               (lschedule-earlier-instrs lschedule))
-      (sqrt running-fidelity))))
+  "Computes sqrt( ∑ ln(f_i)^2 ) where f_i is a fidelity associated with
+an instruction i, i ranging over all instructions in the LSCHEDULE"
+  (let ((running-fidelity 0d0))
+    (map-lschedule-in-topological-order
+     lschedule
+     (lambda (instr)
```
stylewarning, line 672: consider flet + DX

src/addresser/logical-schedule.lisp
```diff
@@ -658,55 +662,25 @@ Returns the reduction of all bumped values by COMBINE-VALUES, and a hash table m
...
-
-  This relies on the fact that the function $\exp\{-\sqrt{\log(x)^2 + \log(y)^2}\}$ is approximately equal to $\min\{x, y\}$ for $x, y \in (0, 1]$."
-  (multiple-value-bind (max-value value-hash)
-      (lschedule-calculate-log-fidelity lschedule chip-spec)
-    (values (exp (- max-value)) value-hash)))
+  "Calculate fidelity as the minimum fidelity of the individual instructions."
+  (let ((min-fidelity 1.0d0)) 
+    (map-lschedule-in-topological-order
+     lschedule
+     (lambda (instr)
```
stylewarning, line 681: consider flet + DX

src/chip/chip-specification.lisp
```diff
@@ -809,3 +809,40 @@ Compilers are listed in descending precedence.")
...
+                                (hardware-object-gate-information qubit-obj))))
+       (if specs-obj
+           (gate-record-fidelity specs-obj)
+           (warn-and-return-perfect-fidelity instr))))
+
+    (application
+     (let (fidelity)
+       (a:when-let* ((obj        (lookup-hardware-object chip-spec instr))
+                     (specs-hash (hardware-object-gate-information obj))
+                     (binding    (and (< 0 (hash-table-count specs-hash))
```
stylewarning, line 837: plusp

src/compressor/compressor.lisp
```diff
@@ -87,17 +87,12 @@
...
-  (let ((lschedule (make-lschedule)))
-    ;; load up the logical schedule
-    (append-instructions-to-lschedule lschedule instructions)
-    ;; sift through it for fidelities
-    (lschedule-calculate-log-fidelity lschedule chip-specification)))
 
 (defun calculate-instructions-fidelity (instructions chip-specification)
   "Calculates the fidelity of a sequence of native INSTRUCTIONS on a chip with architecture governed by CHIP-SPECIFICATION (and with assumed perfect parallelization across resources)."
-  (exp (- (calculate-instructions-log-fidelity instructions chip-specification))))
+  (reduce #'min instructions
```
stylewarning, line 93: is this correct? why the minimum across instructions?

stylewarning (commented):
I did a once-over to check for obvious things. Looks pretty good.

In my initial review though I was a bit confused about using the minimum. It seemed the old code didn't actually calculate minima, but it mentioned it in a doc string?

I'd personally like some clarity on that.

Not approving or requesting changes until I understand it better.

src/compressor/compressor.lisp
macrologist: Yes so, the old function here depended upon a mathematical trick that approximated the minimum by passing a sum of log squared of individual fidelities to the removed exponential.  What you got was not the minimum but was "close". This cuts out the middle man, gives you the minimum, and doesn't have to calculate the sum of log squared terms. 

The function as it appears in compressor was more-or-less copied from its similarly named function in logical-schedule.  The trick was explained there in the (old) docstring to `lschedule-calculate-fidelity`. I.e. 

`min(x,y) ~  exp(-(sqrt (ln(x)^2 + ln(y)^2)))`. 

I assume this trick was used in order to recycle the `lschedule-calculate-log-fidelity` function. But that recycling isn't really worth it, in my opinion: it is much faster to just get the actual minimum.

src/addresser/fidelity-addresser.lisp
macrologist: This function has been moved out of `compressor` where it is no longer used (due to a refactor of its only caller there) and into `fidelity-addresser` where it is still called. 

This is function is equal to its old implementation, but avoids needlessly building a `logical-schedule` instance first.

src/addresser/fidelity-addresser.lisp
macrologist: I could use loop - I was just too lazy to determine the container type of `instructions`.

src/chip/chip-specification.lisp
```diff
@@ -809,3 +809,40 @@ Compilers are listed in descending precedence.")
       (adjoin-hardware-object (build-qubit j :type '(:RZ :X/2 :MEASURE)) chip-spec))
     (warm-hardware-objects chip-spec)
     chip-spec))
+
+(declaim (inline warn-and-return-perfect-fidelity))
+(defun warn-and-return-perfect-fidelity (instr)
+  (format-noise "Unknown fidelity for ~/cl-quil::instruction-fmt/. Skipping." instr)
```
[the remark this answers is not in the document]
macrologist: Formerly, this fidelity-lookup function was being used in the context of a summation of `(expt (log fidelity) 2)`. If `fidelity` could not be found, then rather than computer log fidelity squared we just added 0 to that sum.  But if the meaning of that 0 is a log fidelity squared, then that means that fidelity was 1.

Because I have factored out the "lets lookup fidelity on the device" into its own function, we should return 1 here, and we can now use this lookup in more functions. This is essential for getting rid of the unnecessary logical-schedule construction in a couple of other functions. (i.e. basically logical schedules were being built just so that all of the instructions in those schedules could be iterated over, just so that we could sum up their log fidelity squares, just so we could pass that sum to an `(exp (- sum))` to get the approximate minimum of the original fidelities. Lots of silly extra work).

src/addresser/logical-schedule.lisp
macrologist: what do you mean by `+ DX`? Are you asking that I not use `incf` ? I can't tell

src/compressor/compressor.lisp
macrologist: Oh, and I assume it wants the minimum b/c its doc string literally states that's what it is trying to do.

src/addresser/fidelity-addresser.lisp
macrologist: Opted for `flet` instead: citing above laziness.

src/addresser/logical-schedule.lisp
macrologist: I refactored to use `flet`.

src/compressor/compressor.lisp
ecpeterson: Huh. I think I'm originally responsible for this goofy expression, and I don't think I was looking for an approximate minimum — if x = y, for instance, you won't get min(x, y) = x but x^{-sqrt 2}. I don't think I'm responsible for that old docstring; there was a refactor in 2019–2020 where someone else did their best to read the tea leaves. That said, I don't think there's any true justification to what I wrote, as fidelities do not easily compose, neither horizontally nor vertically. If you want something with at least some thin mathematical justification, you could consider using `sum` everywhere, since _infinitesimal_ infidelities do compose OK and we're all paid to be very optimistic about long-term hardware performance.

src/addresser/logical-schedule.lisp
stylewarning: DX is SBCLian slang for "dynamic extent (declaration)"

src/compressor/compressor.lisp
macrologist: I suppose this is a way to use the composition-as-a-sum-of-log-squares and normalize it to [0,1] so that it looks like a fidelity?

ecpeterson: That is nice that it keeps to [0, 1]; I'm not sure if that's accidental or intentional. On the other hand, the minimum fidelity of an nQ gate is 1/(1 + 2^n), so maybe not that nice. And sums of infinitesimals won't escape [0, 1] either. I dunno.

macrologist:
Some of the changes I have made here have to do with removing needless logical schedule construction. The assessment of needlessness was based on looking at the code as it actually is and determining that extra work was being done, but not necessarily what it was intended to be. 

Specifically, all of these fidelity calculations first construct a logical schedule, but then use that schedule to iterate over every instruction, adding up the (log squared) fidelities of each, and then either using that sum or using the `(exp -sum)` in some kind of comparison.  In the calculations as they exist presently, the ordering of operations in a logical schedule doesn't seem to be relevant.

But! maybe they're meant to be. Maybe it was decided that the logical schedule is *the place* to calculate fidelities b/c there may at some point be a better model of whole-program fidelity calculation that takes into account the logical ordering of operations.

I now have some doubts about that optimisitic model based on what @ecpeterson mentioned above: that fidelities don't generally compose well. 

So, I don't know if this PR should remove the instantiation of logical schedules from the pipeline of instructions -> fidelity score because doing so may or may not be a violation of intended design.   @stylewarning

ecpeterson:
> In the calculations as they exist presently, the ordering of operations in a logical schedule doesn't seem to be relevant.
> ...

Something else to bear in mind is that the addresser has a cross-competing concern about instruction ordering: it _wants_ to make good decisions so that the total resulting schedule has good fidelity, but it also _needs_ to make a decision about what SWAPs to immediately insert, and so it can't get too wistful about instructions down the line. We dealt with this by adding a decay factor to logically-temporally distant instructions, which is fine for this scheduling application but less fine for calculating the total fidelity of a schedule.

I only bring this up to say (1) that there is a place where we jam in ordering info into these values + (2) something like that jamming is unavoidable in greedy-ish scheduling, so it's always going to distort the compiler's internal perspective on these values, and so maybe it isn't so important to come up with a really nice (/ computationally expensive) formula. (Users' external perspectives on these values are another matter! They probably do want meaningful values.)

macrologist:
Sorry I don't know if I made my concern clear. My question has to do with a very specific case: Several of these fidelity calculations start with a sequence of instructions, build a logical schedule out of them, calculate fidelity, then throw the recently built logical schedule away. In these situations, the logical schedule is only being built so that it can be passed to a fidelity calculation routine.  (This is the situation in some of the compressor code as well as a sub-step in the fidelity-addresser code. )

Maybe this is fine if what you want to do is to take into account the ordering of operations in your calculation of fidelity. Right now that is not happening - the actual computation doesn't take order into account. But perhaps one might in the future expect order to be significant - if that is the case, it makes sense to create disposable logical schedules. 

I just don't know if there is a meaningful way to think about fidelity on a logical schedule that doesn't also apply to simple sequences of instructions.

ecpeterson:
I myself think it is fine not to bother calculating these small logical schedules and to use a fairly dumb fidelity estimate which need not rely on logical ordering.

One last comment: Certain Places care foremost about coherence limitations and so only really care about the “longest chain” of instructions. I guess this is comparable to what we’re doing with the temporal strategy? It would be nice to have a similarly clear objective for the “fidelity strategy” if it’s not, in fact, computing fidelities.

macrologist:
I opted to replace most of the 

   (exp (- (sqrt (reduce #'+ FIDELITIES :key (lambda (f) (* (log f) (log f))))))) 

calculations with a simpler 

   (reduce #'min FIDELITIES) 

expression. Both will yield values in the interval `(0, 1]` when each member of `FIDELITIES` is also in `(0, 1]`. 

The first has the added benefit of getting worse as more "bad-but-equal" fidelities enter and so is perhaps more reflective of compounding bad-upon-bad. The downside is that more computation is required for empirically uncertain benefit.

Today I have pushed the most recent version of these changes as a kind of indication of life. I am considering what next to do. 

Some thoughts that have occurred to me: 

- re-refactor to make the (presently excised) construction of logical-schedules significant by considering, perhaps, fidelity on longest paths 
- generalize the selection of the fidelity-composition-routine into a PRAGMA for, e.g., compression strategies.
- OR instead of foisting this choice onto users, do the work to produce evidence on the affect of choices to compilation output, and based on those, select one over the other.

PR #926, opened 2024-09-25
 add a clifford-based optimizer module called foust to cl-quil

YarinHeffes:
This pull request adds a set of packages that build up an optimization routine, based on Intel's PCOAST, which takes advantage of the Clifford group as the normalizer of the Pauli group to perform global optimizations, including and especially in the context of measurements and preparations.

More details can be found in src/foust/README.md.

This commit does not integration to compiler-hook, which we expect in a follow-up commit and pull request.


stylewarning (approved):
Great work!

